Issue #188, opened 2014-12-02
iOS frameworks built with Carthage cannot be submitted to the App Store


jspahrsummers:
From @soffes in https://github.com/Carthage/Carthage/issues/174#issuecomment-65187702:

> I've been dragging the projects into my project and setting up dependencies in Xcode for building. You can't submit iOS frameworks with x86. Couldn't find a way to do a "release build" in carthage so that was my work around.

IOW, our fat binary is inadmissable to the App Store, apparently.

dreambt:
Not work.
Must execute above, why?

    lipo -remove i386 Carthage/Build/iOS/Base32.framework/Base32 -output Carthage/Build/iOS/Base32.framework/Base32
    lipo -remove x86_64 Carthage/Build/iOS/Base32.framework/Base32 -output Carthage/Build/iOS/Base32.framework/Base32

jspahrsummers:
@dreambt Please open a new issue, and detail _exactly what you've tried_, and what's not working about it. We can't help unless you provide more information.

ricardolpd:
hi guys, i am looking into using carthage for a set of frameworks i am building for my employee, is this issue still an issue?

ikesyo:
> is this issue still an issue?

I think so. At the time of Xcode 7 GM, it might change or not. :sweat_smile:

pietbrauer:
It is not an issue anymore (that's why this issue was closed), I am successfully submitting apps that contain carthage built frameworks to the AppStore.

ricardolpd:
@ikesyo so how do you guys work around it? build two framewoks a fat one and a non-fat one and switching with a script??

ikesyo:
@ricardolpd Did you check [the README section](https://github.com/Carthage/Carthage/blob/5fd867c4895b4f59d70181dec169a1644f4430e3/README.md#if-youre-building-for-ios)?

ikesyo:
> I think so.

This is about Xcode/App Store submission bug to which @jspahrsummers referred above: https://github.com/Carthage/Carthage/issues/188#issuecomment-66523227

ricardolpd:
ikesyo, sorry i never got around reading the section of the readme file about the script. thanks it makes sense now

ikesyo:
No worries!

adammendoza:
I'm using Carthage for AFNetwork and I also have to uncheck Include app symbols for your application to receive symbolicated crash logs from Apple.

izouxv:
[...]

PR #208, opened 2014-12-11
Add `copy-framework`

robb:
As mentioned in #188, this command can be used to

1) Copy a framework into an app bundle
2) Strip all unwanted architectures (think `x86_64` slices when building for ARM)
3) Re-sign the framework as needed

This changeset is now functional.

What's left to do:
- [x] Read architectures in the binary from `lipo`
- [x] Actually copy the frameworks that are passed in
- [x] Tests


dreambt:
I think not

PR #249, opened 2014-12-25
Add `carthage fetch` command

### LINKED ISSUE #116: `carthage fetch` for obtaining refs from an arbitrary remote
If I already have the repo in case on my hard-drive somewhere, it would be nice if Carthage could use that one, without me having to write a file URL into the `Cartfile` that may not be correct for other users.

Additionally, such a layer of indirection could make sure that old versions still check out cleanly even if the original URLs have long since rotted away. Decoupling the _what version of which depencency_ from _where to get it_ is pretty much my favorite thing about Cocoapods.

jspahrsummers:
```
$ carthage fetch https://github.com/ReactiveCocoa/ReactiveCocoa.git
*** Fetching ReactiveCocoa
```

```
$ carthage fetch https://github.com/jspahrsummers/Notifications
*** Cloning Notifications
```

Resolves #116.


robb:
:tada:

PR #257, opened 2014-12-26
Automatically use binary frameworks from GitHub Releases

### LINKED ISSUE #6: GitHub Releases support
We should support cloning and building arbitrary tags to start with, but maybe there's something we can do specifically with capital-R Releases that would be interesting.

For example, if a release has detailed notes, maybe we could present those somehow. Or if it has binaries attached, maybe we can use those instead of building our own from scratch (extra important given that every Xcode update breaks something).

jspahrsummers:
Resolves #6.

**To do:**

 - [x] Automatically fetch releases for GitHub repository dependencies
 - [x] Automatically paginate through all releases
 - [x] Download release assets with content type `application/zip`
 - [x] Unzip downloads into a temporary folder
 - [x] Drop binary frameworks into `Carthage.build/` based on platform
 - [x] Terminate asset/release fetching early when a matching binary is found
 - [x] Cache assets in `~/Library/Caches`
 - [x] Reuse existing downloaded assets
 - [x] Skip repository checkout when binaries are found
 - [x] Remove old repository checkout when binaries are found
 - [x] Only load Git credentials as needed
 - [x] Add documentation
 - [x] Add logging
 - [x] Split up signal chain
 - [x] Only download binaries when updating or bootstrapping with `--build`, or checking out with `--use-binaries`
 - [x] Only create `Build` folder symlink in dependency checkouts (not when binaries are used)
 - [x] Test unzipping


jspahrsummers:
I didn't want to go into too much detail in the README, partly because I want to see which bits need the most explanation, and partly because [Cartfile settings](https://github.com/Carthage/Carthage/issues/262) will change how the feature is commonly used.

Other than that, this is ready for review. :stars:

jspahrsummers:
@Carthage/carthage _bump_

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -154,3 +155,72 @@ extension NSScanner {
...
+	}
+}
+
+extension NSURL: JSONDecodable {
+	public class func decode(json: JSONValue) -> Self? {
+		if let URLString = String.decode(json) {
+			return self(string: URLString)
+		} else {
+			return nil
+		}
```
swizzlr: I think
```objc
switch j {
        case .JSONString(let string):
            return NSURL(string: string)
        default:
            return .None
}
```
is slightly more elegant but that's only because I wrote this exact code yesterday.

swizzlr:
Looks good to me!

Source/CarthageKit/FrameworkExtensions.swift
robrix: I’m partial to `Optional<T>.map` for simple cases like this one:

```swift
return String.decode(json).map { self(string: $0) }
```

Aside from that, should this be an optional initializer instead of a `class func`?

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -154,3 +155,72 @@ extension NSScanner {
...
+	}
+}
+
+extension NSFileManager {
+	/// Creates a directory enumerator at the given URL. Sends each URL
+	/// enumerated, along with the enumerator itself (so it can be introspected
+	/// and modified as enumeration progresses).
+	internal func rac_enumeratorAtURL(URL: NSURL, includingPropertiesForKeys keys: [String], options: NSDirectoryEnumerationOptions, catchErrors: Bool = false) -> ColdSignal<(NSDirectoryEnumerator, NSURL)> {
```
robrix: 1. If this always caught errors, then the signal would terminate, and resuming the enumeration from that point would be a pain, yes?
2. I really think this should be `carthage_…` instead of `rac_…`. I know you’re not likely to step on your own toes with this one, but it does sort of defeat the purpose of the prefix to use another project’s :smile:

Source/CarthageKit/GitHub.swift
```diff
@@ -55,3 +69,235 @@ extension GitHubRepository: Printable {
...
+			return self(ID: ID, name: name, contentType: contentType, downloadURL: downloadURL)
+		}
+
+		public static func decode(j: JSONValue) -> Asset? {
+			return self.create
+				<^> j <| "id"
+				<*> j <| "name"
+				<*> j <| "content_type"
+				<*> j <| "browser_download_url"
+		}
```
robrix: What do you think of [`check`](https://gist.github.com/andymatuschak/2b311461caf740f5726f#file-gistfile1-swift-L13-L16) as a more readable (and not very abstract) abstraction for handling this sort of thing?

Source/CarthageKit/Project.swift
```diff
@@ -292,6 +318,107 @@ public final class Project {
...
+							return .error(error)
+						}
+					}
+					.concatMap(unzipArchiveToTemporaryDirectory)
+					.concatMap { directoryURL in
+						return frameworksInDirectory(directoryURL)
+							.mergeMap(self.copyFrameworkToBuildFolder)
+							.on(completed: {
+								NSFileManager.defaultManager().trashItemAtURL(checkoutDirectoryURL, resultingItemURL: nil, error: nil)
+								return
```
robrix: I learned the other day that you can do:

```swift
_ = expression…
```

in order to conveniently return `()` instead of the type of `expression`.

README.md
```diff
@@ -116,6 +116,14 @@ Carthage determines which versions of your framework are available by searching
 
 Tags without any version number, or with any characters following the version number (e.g., `1.2-alpha-1`) are currently unsupported, and will be ignored.
 
+### Archive prebuilt frameworks into one zip file
+
+Carthage can automatically use prebuilt frameworks, instead of building from scratch, if they are attached to a [GitHub Release](https://help.github.com/articles/about-releases/) on your project’s repository.
+
+To offer prebuilt frameworks for a specific tag, the binaries for _all_ supported platforms should be zipped up together into _one_ archive, and that archive should be attached to a published Release corresponding to that tag. The attachment should include `.framework` in its name (e.g., `ReactiveCocoa.framework.zip`), to indicate to Carthage that it contains binaries.
```
robrix: How will this work if you give your iOS & Mac frameworks the same name?

Source/CarthageKit/FrameworkExtensions.swift
jspahrsummers: This is the same method we talked about before. `map` won't work because it would need the signature of `>>=`.

> Aside from that, should this be an optional initializer instead of a `class func`?

Source/CarthageKit/FrameworkExtensions.swift
jspahrsummers: > If this always caught errors, then the signal would terminate, and resuming the enumeration from that point would be a pain, yes?

You mean if it passed through errors? That's correct.

> I really think this should be `carthage_…` instead of `rac_…`. I know you’re not likely to step on your own toes with this one, but it does sort of defeat the purpose of the prefix to use another project’s :smile:

I've been using these as a testbed for methods that could maybe get lifted into RAC.

Source/CarthageKit/GitHub.swift
jspahrsummers: Since we're using Argo, I'd rather following Argo's conventions.

README.md
jspahrsummers: The zip can contain nested folders, and we'll search through them for frameworks.

Source/CarthageKit/FrameworkExtensions.swift
robrix: > This is the same method we talked about before. map won't work because it would need the signature of `>>=`.

Postfix `?` turns out to be monadic `join`, so:

```swift
return String.decode(json).map { self(string: $0) }?
```

should work. But I’m not bothered if you want to leave it as is.

> This is specifically adding conformance to an Argo protocol, so it has to match their signature.

Ah, of course.

Source/CarthageKit/FrameworkExtensions.swift
robrix: > You mean if it passed through errors? That's correct.

Cool, thanks.

> I've been using these as a testbed for methods that could maybe get lifted into RAC.

I figured. I’m still pretty uncomfortable with it, although I don’t know that it introduces any risk in practical terms.

Source/CarthageKit/GitHub.swift
robrix: :ok_hand:

README.md
robrix: :sparkles: Excellent.

robrix:
I’m :cool: to :shipit: with or without the above suggestions.

Source/CarthageKit/FrameworkExtensions.swift
jspahrsummers: I'll leave it as-is for now, but it's a good note for the future.

groue:
This feature has the unfortunate consequence that library users ask library developers to opt in Github release system and ship pre-built binaries, when they only want to speed up the way Carthage builds their application's framework.

PR #261, opened 2014-12-30
Check if codesigning is allowed

### LINKED ISSUE #239: Run script from embedded framework fails with "no identity found"
Every time carthage attempts to build a framework that uses carthage, I get an error:

```
The following build commands failed:
        PhaseScriptExecution Run\ Script /Users/justinmakaila/Library/Developer/Xcode/DerivedData/FacebookSDK-fdrhrpuxegaxkefhtkrmgrpzpmap/Build/Intermediates/FacebookSDK.build/Release-iphonesimulator/FacebookSDK.build/Script-7CECB7A51A48DA5400D38A4B.sh
```

[log here](https://gist.github.com/justinmakaila/a66b5fb28a047effbc15)

It appears to be a signing identity issue, but I'm not sure where to begin.

robb:
Do you think it's worth setting up full-blown integration tests for this?

Fixes #239

### DIFF
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -27,7 +27,7 @@ public struct CopyFrameworksCommand: CommandType {
 
 				return combineLatest(ColdSignal.fromResult(target), .fromResult(validArchitectures()))
 					.mergeMap { (target, validArchitectures) -> ColdSignal<()> in
-						return combineLatest(copyFramework(source, target), .fromResult(getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY")))
+						return combineLatest(copyFramework(source, target), codeSigningIdentity())
 							.mergeMap { (url, codesigningIdentity) -> ColdSignal<()> in
 								return stripFramework(target, keepingArchitectures: validArchitectures, codesigningIdentity: codesigningIdentity)
 							}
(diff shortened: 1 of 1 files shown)


akisute:
Guys I happened to run into an issue with this commit. This fix doesn't work when you build something for iPhone Simulator instead of real devices with the `/usr/local/bin/carthage copy-frameworks` charm.

The problem is,
1. `codeSigningAllowed()` always returns `false` on iPhone Simulator because `CODE_SIGNING_ALLOWED` is always set to `NO`.
2. Thus `codeSigningIdentity()` always returns a single nil `ColdSignal<String?>`.
3. This results in calling `stripFramework(keepingArchitectures:codesigningIdentity:)` with a nil String.
4. This makes Xcode cause error like `: no identity found`.

I suppose you should not do anything in `stripFramework(keepingArchitectures:codesigningIdentity:)` when nil codesigningIdentity: is passed.

jspahrsummers:
@akisute A `nil` codesigning identity means that `codesign` will not be invoked. Are you on Carthage 0.5.1? Version 0.5 does not include this fix.

akisute:
@jspahrsummers Ugh, I'm reading the latest code while I'm using Carthage 0.5.0 to build something. That's why :( Sorry about that!

jspahrsummers:
No problem! :smile:

PR #265, opened 2015-01-04
Switch to relative symlinks between build folders

### LINKED ISSUE #263: `carthage build` creates absolute symlink from dependency to main Carthage/Build directory
Running `carthage build` after checking out the dependencies results in a symlink with an absolute path from the dependencies’ `Carthage/Build` directory to the project’s `Carthage/Build` directory.

When the checkout command is ran without the `--use-submodules` option and one checks the `Carthage/Checkouts` directory into the git repo (as suggested by the documentation) the repo can’t be shared by multiple persons.

I’m running **Carthage version 0.5.1** (installed via Homebrew).

**Example without submodules**
Given a project at the path `/User/aron/Documents/SomeApp` and a Cartfile of:
```
github "Carthage/Commandant" "master"
```
Running `carthage bootstrap`, `carthage update` or `carthage checkout` _without_ the `--use-submodules` will result in the following on disk after the `carthage build` step has been performed:
```
Carthage/Build/Mac/Commandant.framework
Carthage/Checkouts/Commandant/Carthage/Build -> /Users/aron/Documents/SomeApp/Carthage/Build
```

That is an absolute path in the symlink which becomes problematic if the repo is shared.

----
I tried to resolve this by changing to using submodules instead by removing the checkouts and then running `carthage bootstrap --use-submodules`. This partially works. There’s no longer any need to check in the `Carthage/Build` symlink found in each dependency’s checkout directory. However each dependency’s submodule is marked as dirty by git.

**Example with submodules**
```shell
$ git clone https://github.com/Carthage/Carthage.git && cd Carthage
[…]
$ carthage bootstrap # This will check the dependencies out as submodules as they are in the .gitmodule file
*** Checking out Commandant at "2c0384bfa4d0c66ab0ebf5bc48739f9236ea2e0c"
[…]
*** Building scheme "Commandant" in Commandant.xcworkspace
[…]
$ git status -sb
## master...origin/master
 M Commandant
 M LlamaKit
 M Nimble
 M Quick
 M ReactiveCocoa
 M ReactiveTask
 M xcconfigs
$ git diff
diff --git a/Carthage/Checkouts/Commandant b/Carthage/Checkouts/Commandant
--- a/Carthage/Checkouts/Commandant
+++ b/Carthage/Checkouts/Commandant
@@ -1 +1 @@
-Subproject commit 2c0384bfa4d0c66ab0ebf5bc48739f9236ea2e0c
+Subproject commit 2c0384bfa4d0c66ab0ebf5bc48739f9236ea2e0c-dirty
[…]
$ cd Carthage/Checkouts/Commandant && git status -sb
## HEAD (no branch)
?? Carthage/Build
[…]
```

Dirty forever :crying_cat_face:

jspahrsummers:
Fixes #263.

/cc @rastersize

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -764,12 +764,11 @@ public typealias BuildSchemeSignal = ColdSignal<(ProjectLocator, String)>
 ///
 /// Returns signals in the same format as buildInDirectory().
 public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryURL: NSURL, withConfiguration configuration: String) -> (HotSignal<NSData>, ColdSignal<BuildSchemeSignal>) {
+	let rootBinariesURL = rootDirectoryURL.URLByAppendingPathComponent(CarthageBinariesFolderPath, isDirectory: true)
 	let dependencyURL = rootDirectoryURL.URLByAppendingPathComponent(dependency.relativePath, isDirectory: true)
 
 	let (buildOutput, schemeSignals) = buildInDirectory(dependencyURL, withConfiguration: configuration)
 	let copyProducts = ColdSignal<BuildSchemeSignal>.lazy {
-		let rootBinariesURL = rootDirectoryURL.URLByAppendingPathComponent(CarthageBinariesFolderPath, isDirectory: true)
-
 		var error: NSError?
 		if !NSFileManager.defaultManager().createDirectoryAtURL(rootBinariesURL, withIntermediateDirectories: true, attributes: nil, error: &error) {
 			return .error(error ?? CarthageError.WriteFailed(rootBinariesURL).error)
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -787,7 +786,17 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
 			}
 		}
 
-		if !NSFileManager.defaultManager().createSymbolicLinkAtURL(dependencyBinariesURL, withDestinationURL: rootBinariesURL, error: &error) {
+		// The relative path to this dependency's Carthage/Build folder, from
+		// the root.
+		let dependencyBinariesRelativePath = dependency.relativePath.stringByAppendingPathComponent(CarthageBinariesFolderPath)
+		let componentsForGettingTheHellOutOfThisRelativePath = Array(count: dependencyBinariesRelativePath.pathComponents.count - 1, repeatedValue: "..")
```
robrix: :grin:

robrix: This might also be the first and only case I’ve seen where `Repeat<T>` might apply. I don’t see any concrete benefits to using it over `Array` here, however, at least not relative to the cost of “what the hell is `Repeat`?”.

robrix:
Looks good to me :+1:

rastersize:
Yay, thanks! :smile:

Source/CarthageKit/Xcode.swift
jspahrsummers: I looked at `Repeat` before writing this code, and it didn't apply to my original conception. Of course, this final implementation would actually make perfect sense for it.

The benefit would be not needing to preallocate an array for the repeating components, but it's so small that it doesn't matter.

robrix: (Exactly my thinking.)

PR #269, opened 2015-01-09
Quick typo fix in readme for Run Script commands

patr1ck:
The default shell /bin/sh uses $VAR and not $(VAR) for variable substitution – the latter will actually error but Xcode won't notify you unless you go looking for it, so it could easily cause problems for people.

(Also added double quotes for good measure, though they aren't likely needed since hopefully no one has spaces in their framework names.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
   and add the paths to the frameworks you want to use under “Input Files”, e.g.:
 
   ```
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Yeah, it is appropriate. This fix isn't necessary.

patr1ck:
Oh wow, didn't even occur to me to use the input files section. Duh. Sorry for the noise!

jspahrsummers:
No problem! :sparkles:

PR #280, opened 2015-01-13
Fix copy-frameworks help bug

### LINKED ISSUE #277: `carthage help copy-frameworks` tries to execute the command
When I run `carthage help copy-frameworks` (with 0.5.1), I see the following result:
```
In a Run Script build phase, copies each framework specified by an SCRIPT_INPUT_FILE environment variable into the built app bundle

Environment variable not set: SCRIPT_INPUT_FILE_COUNT
```
It appears that when I request help with the `copy-frameworks` command, Carthage tries to run that command, but fails when it can't find the environment variable.

---
Looking closer, I've tracked down the cause of issue: `HelpCommand` [calls `command.run(.Usage)`](https://github.com/Carthage/Commandant/blob/master/Commandant/HelpCommand.swift#L38-40), but `CopyFrameworksCommand` completely [ignores the `mode` argument](https://github.com/Carthage/Carthage/blob/master/Source/carthage/CopyFrameworks.swift#L20-39) and tries to actually execute the command.

mattrubin:
Fixes #277:
A bug occurred when calling `carthage help copy-frameworks`. Previously, the CopyFrameworksCommand would ignore the mode argument passed to run() and would attempt to execute the command. This change fixes run() to only execute in Arguments mode.

### DIFF
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -18,23 +18,31 @@ public struct CopyFrameworksCommand: CommandType {
 	public let function = "In a Run Script build phase, copies each framework specified by an SCRIPT_INPUT_FILE environment variable into the built app bundle"
 
 	public func run(mode: CommandMode) -> Result<()> {
-		return inputFiles()
-			.map { frameworkPath -> ColdSignal<()> in
-				let frameworkName = frameworkPath.lastPathComponent
+		switch mode {
+		case .Arguments:
+			return inputFiles()
+				.map { frameworkPath -> ColdSignal<()> in
+					let frameworkName = frameworkPath.lastPathComponent
 
-				let source = NSURL(fileURLWithPath: frameworkPath, isDirectory: true)!
-				let target = frameworksFolder().map { $0.URLByAppendingPathComponent(frameworkName, isDirectory: true) }
+					let source = NSURL(fileURLWithPath: frameworkPath, isDirectory: true)!
+					let target = frameworksFolder().map { $0.URLByAppendingPathComponent(frameworkName, isDirectory: true) }
 
-				return combineLatest(ColdSignal.fromResult(target), .fromResult(validArchitectures()))
-					.mergeMap { (target, validArchitectures) -> ColdSignal<()> in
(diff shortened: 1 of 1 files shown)


Source/carthage/CopyFrameworks.swift
```diff
@@ -18,23 +18,31 @@ public struct CopyFrameworksCommand: CommandType {
...
+								.mergeMap { (url, codesigningIdentity) -> ColdSignal<()> in
+									return stripFramework(target, keepingArchitectures: validArchitectures, codesigningIdentity: codesigningIdentity)
+								}
+						}
+				}
+				.concat(identity)
+				.wait()
+
+		default:
+			break
```
jspahrsummers: Can you move the `return` statement below into here instead, replacing the `break`? That means that both conditions will return directly, which is perhaps less confusing.

Source/carthage/CopyFrameworks.swift
```diff
@@ -18,23 +18,31 @@ public struct CopyFrameworksCommand: CommandType {
...
+							return combineLatest(copyFramework(source, target), codeSigningIdentity())
+								.mergeMap { (url, codesigningIdentity) -> ColdSignal<()> in
+									return stripFramework(target, keepingArchitectures: validArchitectures, codesigningIdentity: codesigningIdentity)
+								}
+						}
+				}
+				.concat(identity)
+				.wait()
+
+		default:
```
jspahrsummers: Can you name `.Usage` explicitly here? We want any new enum values to cause an error, so that we have an opportunity to think about the correct behavior here.

jspahrsummers:
Thanks! Just a couple of notes. :notebook:

mattrubin:
Good feedback, thanks!

If you want to be explicit when switching on `CommandMode`, it might be worth making the same change to [Version](https://github.com/Carthage/Carthage/blob/56b6bdb39adcb1a296635b8b7974dc30fc72c02e/Source/carthage/Version.swift#L28-32) (though that is a much simpler command).

jspahrsummers:
Yeah, I agree it should change there too, although it's fairly low priority. I just happened to notice it here. :trollface: 

Thanks for the PR! :stars:

PR #284, opened 2015-01-18
Set DSTROOT to TEMPORARY_FOLDER in XCODEFLAGS

jpsim:
Without this, overriding `TEMPORARY_FOLDER` will only partially work since Xcode always uses `/tmp/Carthage.dst` as its intermediate built products location (see https://developer.apple.com/library/mac/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW22).

Supporting overriding `TEMPORARY_FOLDER` also means that a failed homebrew installation can still clean up temporary artifacts (see https://github.com/Homebrew/homebrew/pull/35926#discussion_r23104627).

I've confirmed this also works with `xctool`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
-XCODEFLAGS=-workspace 'Carthage.xcworkspace' -scheme 'carthage'
-
-TEMPORARY_FOLDER=/tmp/Carthage.dst
+TEMPORARY_FOLDER?=/tmp/Carthage.dst
 PREFIX?=/usr/local
 BUILD_TOOL?=xcodebuild
+
+XCODEFLAGS=-workspace 'Carthage.xcworkspace' -scheme 'carthage' DSTROOT=$(TEMPORARY_FOLDER)
 
 BUILT_BUNDLE=$(TEMPORARY_FOLDER)/Applications/carthage.app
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Just to add my :+1: here as I requested it as something that would help cleanup the Homebrew formula.

jspahrsummers:
Why wouldn't we use `-derivedDataPath` here?

jpsim:
> why wouldn't we use -derivedDataPath here?

Because I wasn't aware of that flag :wink:. If it has the same effect, why not? Probably safer, but setting `DSTROOT` has worked for me.

jspahrsummers:
I think we should try with `-derivedDataPath`, because assumably that flag was added to fix problems with people trying to override build settings.

jpsim:
Setting the `-derivedDataPath` still uses `/tmp/Carthage.dst` as a temporary path, so no dice.

jspahrsummers:
Okay, no worries. Thanks for trying it! :shipit:

mikemcquaid:
@jpsim Thanks again dude!
@jspahrsummers :zap: Just FYI if you're submitting the next Carthage PR: you can remove the `make clean` step now that this was merged. Thanks!

Issue #273, opened 2015-01-10
Include xcodebuild output inline with --verbose


jspahrsummers:
From https://github.com/Carthage/Carthage/issues/235#issuecomment-69409228:

> It may be helpful to have a flag that tail's the xcodebuild log during build, as Travis doesnt allow you to go in and examine the file system as far as I'm aware - makes it a bit tricky to debug this error.

/cc @sebbean

sebbean:
nice!

PR #291, opened 2015-01-28
Build Option: --verbose 

mattrubin:
Per #273, the `--verbose` option will print out the xcodebuild log as each build progresses.

 - This doesn't (yet) offer a `--verbose` option for `carthage bootstrap` or `carthage update`. I suspect we'll want that option, since a desired use case for verbose output is with `carthage bootstrap` on a non-interactive Travis CI session, but a verbose option on those commands might also print verbose output from the git clone/checkout stage of the process, so I've left it out for the moment.

 - My solution for tailing the build log was simply to print each line of data from `stdoutSignal` as it's written to the log file. This appears to correctly interleave the build logs with the non-verbose printing from `schemeSignals`. I have only a beginner's understanding of ReactiveCocoa, so if there's a more idiomatic approach I'd love to know what it might be.

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -31,6 +31,12 @@ public struct BuildCommand: CommandType {
 
 				let (stdoutSignal, schemeSignals) = self.buildProjectInDirectoryURL(directoryURL, options: options)
 				stdoutSignal.observe { data in
+					if options.verbose {
+						if let outString = NSString(data: data, encoding: NSUTF8StringEncoding) {
+							carthage.print(outString)
+						}
+					}
+
 					stdoutHandle.writeData(data)
 					stdoutHandle.synchronizeFile()
 				}
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -25,18 +25,12 @@ public struct BuildCommand: CommandType {
 
 	/// Builds a project with the given options.
 	public func buildWithOptions(options: BuildOptions) -> ColdSignal<()> {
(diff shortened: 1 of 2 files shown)


Source/carthage/Build.swift
```diff
@@ -31,6 +31,12 @@ public struct BuildCommand: CommandType {
 
 				let (stdoutSignal, schemeSignals) = self.buildProjectInDirectoryURL(directoryURL, options: options)
 				stdoutSignal.observe { data in
+					if options.verbose {
+						if let outString = NSString(data: data, encoding: NSUTF8StringEncoding) {
+							carthage.print(outString)
```
jspahrsummers: Instead of logging to two different places, maybe `--verbose` should just dump everything to the terminal. In that case, could we perhaps skip the temporary file, and use `stdoutHandle = NSFileHandle.fileHandleWithStandardOutput()`?

jspahrsummers:
Thanks! This looks great. :sparkles:

I am wondering about [continuing to log to a file](https://github.com/Carthage/Carthage/pull/291#discussion_r23794562), but we could always get this in and then tweak it more.

Source/carthage/Build.swift
mattrubin: I like this idea. Having a single destination is cleaner, and I can't think of a situation where I'd need both the inline log *and* the temporary log file.

mattrubin:
PR updated to write build logs to the terminal via `stdout`, in lieu of writing to a log file.

PR #292, opened 2015-01-28
PrettyColors — Take advantage of the terminal

jdhealy:
_Resolves #121_

Adds colors and formatting to most `carthage.println`s. Imports [jdhealy/PrettyColors](//github.com/jdhealy/PrettyColors).

Basically just the bare minimum of commits to Carthage to allow for flexibility based off your feedback.

I based most of the color and formatting choices off [homebrew](//github.com/homebrew/homebrew).

Output looks like this in iTerm 2 with custom pastel-ish colors:
screen shot 2015-01-12 at 1 00 04 pm

Output looks like this in Terminal.app using basic colors:
screen shot 2015-01-14 at 6 35 44 pm

## Feedback
- Color and formatting choices
  - Any preferences on colors or formatting?
  - I've styled the output so far with [widely-supported styles](https://github.com/jdhealy/PrettyColors/wiki/Terminal-Support) _(bold and underlined)_ and [nearly-universal colors](https://github.com/jdhealy/PrettyColors/blob/development/Source/Color/Named.swift). 

- Command line arguments and environment variables
  - Should we check any environment variables? `$TERM`?
  - `--no-colors` flag?

- PrettyColors as a framework of CarthageKit
  - I embedded PrettyColors inside of CarthageKit like all the other dependencies, but the library is actually only used in the `carthage` target. Should it be taken out of CarthageKit?

- Lastly, I'd really welcome feedback on the library. It's pinned in the Cartfile at [`edf8291df2`](https://github.com/jdhealy/PrettyColors/commit/edf8291df2f524ed62dcfa60227b988535ccf6eb), which is essentially `v1.0.0-rc1`. So, please raise any concerns or observations! :smiley:

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -10,7 +10,6 @@ import CarthageKit
 import Commandant
 import Foundation
 import LlamaKit
-import PrettyColors
 import ReactiveCocoa
 
 public struct BuildCommand: CommandType {
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -35,13 +35,15 @@ public struct BuildCommand: CommandType {
 					stdoutHandle.writeData(data)
 					stdoutHandle.synchronizeFile()
 				}
-
+				
+				let formatting = options.colorOptions.formatting
+				
 				return schemeSignals
 					.concat(identity)
 					.on(started: {
-						carthage.println(Formatting.bullets + "xcodebuild output can be found in " + Formatting.path.autowrap(temporaryURL.path!))
+						carthage.println(formatting.bullets + "xcodebuild output can be found in " + formatting.path(string: temporaryURL.path!))
 					}, next: { (project, scheme) in
-						carthage.println(Formatting.bullets + "Building scheme " + Formatting.quote(scheme) + " in " + Formatting.projectName.autowrap(project.description))
+						carthage.println(formatting.bullets + "Building scheme " + formatting.quote(scheme) + " in " + formatting.projectName(string: project.description))
 					})
 					.then(.empty())
 			}
@@ -69,7 +71,7 @@ public struct BuildCommand: CommandType {
 			}
 			.mergeMap { project in
 				return project
-					.migrateIfNecessary()
+					.migrateIfNecessary(options.colorOptions)
 					.on(next: carthage.println)
 					.then(.single(project))
 			}
@@ -117,17 +119,18 @@ public struct BuildCommand: CommandType {
 public struct BuildOptions: OptionsType {
 	public let configuration: String
 	public let skipCurrent: Bool
+	public let colorOptions: ColorOptions
 	public let directoryPath: String
 
-	public static func create(color: ColorOptions)(configuration: String)(skipCurrent: Bool)(directoryPath: String) -> BuildOptions {
-		return self(configuration: configuration, skipCurrent: skipCurrent, directoryPath: directoryPath)
+	public static func create(configuration: String)(skipCurrent: Bool)(colorOptions: ColorOptions)(directoryPath: String) -> BuildOptions {
+		return self(configuration: configuration, skipCurrent: skipCurrent, colorOptions: colorOptions, directoryPath: directoryPath)
 	}
 
 	public static func evaluate(m: CommandMode) -> Result<BuildOptions> {
 		return create
-			<*> ColorOptions.evaluate(m)
 			<*> m <| Option(key: "configuration", defaultValue: "Release", usage: "the Xcode configuration to build")
 			<*> m <| Option(key: "skip-current", defaultValue: true, usage: "don't skip building the Carthage project (in addition to its dependencies)")
+			<*> ColorOptions.evaluate(m)
 			<*> m <| Option(defaultValue: NSFileManager.defaultManager().currentDirectoryPath, usage: "the directory containing the Carthage project")
 	}
 }
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -39,9 +39,9 @@ public struct BuildCommand: CommandType {
 				return schemeSignals
 					.concat(identity)
 					.on(started: {
-						carthage.println(bullets + "xcodebuild output can be found in " + Color.Wrap(foreground: .Yellow).wrap(temporaryURL.path!))
+						carthage.println(Formatting.bullets + "xcodebuild output can be found in " + Formatting.path.autowrap(temporaryURL.path!))
 					}, next: { (project, scheme) in
-						carthage.println(bullets + "Building scheme " + Color.Wrap(foreground: .Green).wrap(quote(scheme)) + " in " + bold.wrap(project.description))
+						carthage.println(Formatting.bullets + "Building scheme " + Formatting.quote(scheme) + " in " + Formatting.projectName.autowrap(project.description))
 					})
 					.then(.empty())
 			}
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -119,12 +119,13 @@ public struct BuildOptions: OptionsType {
 	public let skipCurrent: Bool
 	public let directoryPath: String
 
-	public static func create(configuration: String)(skipCurrent: Bool)(directoryPath: String) -> BuildOptions {
+	public static func create(color: ColorOptions)(configuration: String)(skipCurrent: Bool)(directoryPath: String) -> BuildOptions {
 		return self(configuration: configuration, skipCurrent: skipCurrent, directoryPath: directoryPath)
 	}
 
 	public static func evaluate(m: CommandMode) -> Result<BuildOptions> {
 		return create
+			<*> ColorOptions.evaluate(m)
 			<*> m <| Option(key: "configuration", defaultValue: "Release", usage: "the Xcode configuration to build")
 			<*> m <| Option(key: "skip-current", defaultValue: true, usage: "don't skip building the Carthage project (in addition to its dependencies)")
 			<*> m <| Option(defaultValue: NSFileManager.defaultManager().currentDirectoryPath, usage: "the directory containing the Carthage project")
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -10,6 +10,7 @@ import CarthageKit
 import Commandant
 import Foundation
 import LlamaKit
+import PrettyColors
 import ReactiveCocoa
 
 public struct BuildCommand: CommandType {
@@ -38,9 +39,9 @@ public struct BuildCommand: CommandType {
 				return schemeSignals
 					.concat(identity)
 					.on(started: {
-						carthage.println("*** xcodebuild output can be found in \(temporaryURL.path!)")
+						carthage.println(bullets + "xcodebuild output can be found in " + Color.Wrap(foreground: .Yellow).wrap(temporaryURL.path!))
 					}, next: { (project, scheme) in
-						carthage.println("*** Building scheme \"\(scheme)\" in \(project)")
+						carthage.println(bullets + "Building scheme " + Color.Wrap(foreground: .Green).wrap(quote(scheme)) + " in " + bold.wrap(project.description))
 					})
 					.then(.empty())
 			}
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -95,7 +95,7 @@ private func inputFiles() -> ColdSignal<String> {
 		}
 }
 
-private func getEnvironmentVariable(variable: String) -> Result<String> {
+internal func getEnvironmentVariable(variable: String) -> Result<String> {
 	let environment = NSProcessInfo.processInfo().environment
 
 	if let value = environment[variable] as String? {
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -94,13 +94,3 @@ private func inputFiles() -> ColdSignal<String> {
 			return ColdSignal.fromValues(variables).concat(identity)
 		}
 }
-
-internal func getEnvironmentVariable(variable: String) -> Result<String> {
-	let environment = NSProcessInfo.processInfo().environment
-
-	if let value = environment[variable] as String? {
-		return success(value)
-	} else {
-		return failure(CarthageError.MissingEnvironmentVariable(variable: variable).error)
-	}
-}
--- a/Source/carthage/Environment.swift
+++ b/Source/carthage/Environment.swift
@@ -1,0 +1,39 @@
+//
+//  Environment.swift
+//  Carthage
+//
+//  Created by J.D. Healy on 2/6/15.
+//  Copyright (c) 2015 Carthage. All rights reserved.
+//
+
+import CarthageKit
+import Foundation
+import LlamaKit
+
+internal func getEnvironmentVariable(variable: String) -> Result<String> {
+	let environment = NSProcessInfo.processInfo().environment
+
+	if let value = environment[variable] as String? {
+		return success(value)
+	} else {
+		return failure(CarthageError.MissingEnvironmentVariable(variable: variable).error)
+	}
+}
+
+/// Information about the possible parent terminal.
+internal struct Terminal {
+	/// Terminal type retrieved from `TERM` environment variable.
+	static var terminalType: String? {
+		return getEnvironmentVariable("TERM").value()
+	}
+	
+	/// Whether terminal type is `dumb`.
+	static var isDumb: Bool {
+		return terminalType?.caseInsensitiveCompare("dumb") == NSComparisonResult.OrderedSame ?? false
+	}
+	
+	/// Whether STDOUT is a TTY.
+	static var isTTY: Bool {
+		return isatty(STDOUT_FILENO) != 0
+	}
+}
--- a/Source/carthage/Extensions.swift
+++ b/Source/carthage/Extensions.swift
@@ -13,6 +13,7 @@ import CarthageKit
 import Commandant
 import Foundation
 import LlamaKit
+import PrettyColors
 import ReactiveCocoa
 
 private let outputQueue = { () -> dispatch_queue_t in
@@ -55,21 +56,25 @@ extension GitURL: ArgumentType {
 	}
 }
 
+internal let bullets = Color.Wrap(foreground: .Blue, style: .Bold).wrap("***") + " "
+internal let bold = [StyleParameter.Bold] as Color.Wrap
+internal func quote(string: String) -> String { return "\u{0022}" + string + "\u{0022}" }
+
 /// Logs project events put into the sink.
 internal struct ProjectEventSink: SinkType {
 	mutating func put(event: ProjectEvent) {
 		switch event {
 		case let .Cloning(project):
-			carthage.println("*** Cloning \(project.name)")
+			carthage.println(bullets + "Cloning " + bold.wrap(project.name))
 
 		case let .Fetching(project):
-			carthage.println("*** Fetching \(project.name)")
+			carthage.println(bullets + "Fetching " + bold.wrap(project.name))
 
 		case let .CheckingOut(project, revision):
-			carthage.println("*** Checking out \(project.name) at \"\(revision)\"")
+			carthage.println(bullets + "Checking out " + bold.wrap(project.name) + " at " + Color.Wrap(foreground: .Green).wrap(quote(revision)))
 
 		case let .DownloadingBinaries(project, release):
-			carthage.println("*** Downloading \(project.name) release \"\(release)\"")
+			carthage.println(bullets + "Downloading " + bold.wrap(project.name) + " at " + Color.Wrap(foreground: .Green).wrap(quote(release)))
 		}
 	}
 }
@@ -90,7 +95,7 @@ extension Project {
 		let carthageBuild = "Carthage.build"
 		let carthageCheckout = "Carthage.checkout"
 
-		let migrationMessage = "*** MIGRATION WARNING ***\n\nThis project appears to be set up for an older (pre-0.4) version of Carthage. Unfortunately, the directory structure for Carthage projects has since changed, so this project will be migrated automatically.\n\nSpecifically, the following renames will occur:\n\n  \(cartfileLock) -> \(CarthageProjectResolvedCartfilePath)\n  \(carthageBuild) -> \(CarthageBinariesFolderPath)\n  \(carthageCheckout) -> \(CarthageProjectCheckoutsPath)\n\nFor more information, see https://github.com/Carthage/Carthage/pull/224.\n"
+		let migrationMessage = Color.Wrap(foreground: .Blue, style: .Bold).wrap("*** MIGRATION WARNING ***") + "\n\nThis project appears to be set up for an older (pre-0.4) version of Carthage. Unfortunately, the directory structure for Carthage projects has since changed, so this project will be migrated automatically.\n\nSpecifically, the following renames will occur:\n\n  \(cartfileLock) -> \(CarthageProjectResolvedCartfilePath)\n  \(carthageBuild) -> \(CarthageBinariesFolderPath)\n  \(carthageCheckout) -> \(CarthageProjectCheckoutsPath)\n\nFor more information, see " + ([StyleParameter.Underlined] as Color.Wrap).wrap("https://github.com/Carthage/Carthage/pull/224") + ".\n"
 		let signals = ColdSignal<ColdSignal<String>> { sink, disposable in
 			let checkFile: (String, String) -> () = { oldName, newName in
 				if fileManager.fileExistsAtPath(directoryPath.stringByAppendingPathComponent(oldName)) {
--- a/Source/carthage/Extensions.swift
+++ b/Source/carthage/Extensions.swift
@@ -56,25 +56,50 @@ extension GitURL: ArgumentType {
 	}
 }
 
-internal let bullets = Color.Wrap(foreground: .Blue, style: .Bold).wrap("***") + " "
-internal let bold = [StyleParameter.Bold] as Color.Wrap
-internal func quote(string: String) -> String { return "\u{0022}" + string + "\u{0022}" }
+extension Color.Wrap {
+	func autowrap(string: String) -> String {
+		return Formatting.color ? self.wrap(string) : string
+	}
+}
+
+internal struct Formatting {
+	static let color = Terminal.isTTY && !Terminal.isDumb
+	
+	static let bulletin = Color.Wrap(foreground: .Blue, style: .Bold)
+	static let bullets: String = {
+		return bulletin.autowrap("***") + " "
+	}()
+	
+	static let URL = [StyleParameter.Underlined] as Color.Wrap
+	static let projectName = [StyleParameter.Bold] as Color.Wrap
+	static let path = Color.Wrap(foreground: .Yellow)
+	
+	static func quote(string: String) -> String {
+		return Color.Wrap(foreground: .Green).autowrap("\u{0022}" + string + "\u{0022}")
+	}
+}
+
+internal struct Terminal {
+	static let term: String? = NSProcessInfo().environment["TERM"] as? String
+	static let isDumb: Bool = (Terminal.term? as NSString?)?.isEqualToString("dumb") ?? false
+	static let isTTY: Bool = isatty(1) == 1
+}
 
 /// Logs project events put into the sink.
 internal struct ProjectEventSink: SinkType {
 	mutating func put(event: ProjectEvent) {
 		switch event {
 		case let .Cloning(project):
-			carthage.println(bullets + "Cloning " + bold.wrap(project.name))
+			carthage.println(Formatting.bullets + "Cloning " + Formatting.projectName.autowrap(project.name))
 
 		case let .Fetching(project):
-			carthage.println(bullets + "Fetching " + bold.wrap(project.name))
+			carthage.println(Formatting.bullets + "Fetching " + Formatting.projectName.autowrap(project.name))
 
 		case let .CheckingOut(project, revision):
-			carthage.println(bullets + "Checking out " + bold.wrap(project.name) + " at " + Color.Wrap(foreground: .Green).wrap(quote(revision)))
+			carthage.println(Formatting.bullets + "Checking out " + Formatting.projectName.autowrap(project.name) + " at " + Formatting.quote(revision))
 
 		case let .DownloadingBinaries(project, release):
-			carthage.println(bullets + "Downloading " + bold.wrap(project.name) + " at " + Color.Wrap(foreground: .Green).wrap(quote(release)))
+			carthage.println(Formatting.bullets + "Downloading " + Formatting.projectName.autowrap(project.name) + " at " + Formatting.quote(release))
 		}
 	}
 }
@@ -95,7 +120,7 @@ extension Project {
 		let carthageBuild = "Carthage.build"
 		let carthageCheckout = "Carthage.checkout"
 
-		let migrationMessage = Color.Wrap(foreground: .Blue, style: .Bold).wrap("*** MIGRATION WARNING ***") + "\n\nThis project appears to be set up for an older (pre-0.4) version of Carthage. Unfortunately, the directory structure for Carthage projects has since changed, so this project will be migrated automatically.\n\nSpecifically, the following renames will occur:\n\n  \(cartfileLock) -> \(CarthageProjectResolvedCartfilePath)\n  \(carthageBuild) -> \(CarthageBinariesFolderPath)\n  \(carthageCheckout) -> \(CarthageProjectCheckoutsPath)\n\nFor more information, see " + ([StyleParameter.Underlined] as Color.Wrap).wrap("https://github.com/Carthage/Carthage/pull/224") + ".\n"
+		let migrationMessage = Formatting.bulletin.autowrap("*** MIGRATION WARNING ***") + "\n\nThis project appears to be set up for an older (pre-0.4) version of Carthage. Unfortunately, the directory structure for Carthage projects has since changed, so this project will be migrated automatically.\n\nSpecifically, the following renames will occur:\n\n  \(cartfileLock) -> \(CarthageProjectResolvedCartfilePath)\n  \(carthageBuild) -> \(CarthageBinariesFolderPath)\n  \(carthageCheckout) -> \(CarthageProjectCheckoutsPath)\n\nFor more information, see " + Formatting.URL.autowrap("https://github.com/Carthage/Carthage/pull/224") + ".\n"
 		let signals = ColdSignal<ColdSignal<String>> { sink, disposable in
 			let checkFile: (String, String) -> () = { oldName, newName in
 				if fileManager.fileExistsAtPath(directoryPath.stringByAppendingPathComponent(oldName)) {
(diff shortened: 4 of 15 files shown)


robrix:
:sparkles: looks great!

> Any preferences on colors or formatting?

If there’s any sort of reasonable way for us to have consistency with e.g. `grep`/`git`/other tools’ colours, then :+1:, but in general if it’s not eye-bleeding and will be ok on a white background with black text (the system default) or a black background with white text (which seems common), :+1:.

I know that doesn’t exactly narrow things down; sorry about that :confused:

> Should we check any environment variables? `$TERM`?

`$TERM` seems like a good place to start:

```
$ env
…
TERM=xterm-256color
```

> `--no-colors` flag?

`grep` and `git-diff` both use `--color=[when]` with `when` being `never`, `always`, or `auto` (the default). Can we do this?

> I embedded PrettyColors inside of CarthageKit like all the other dependencies, but the library is actually only used in the `carthage` target. Should it be taken out of CarthageKit?

This seems appropriate to me.

Source/carthage/Build.swift
```diff
@@ -38,9 +39,9 @@ public struct BuildCommand: CommandType {
 				return schemeSignals
 					.concat(identity)
 					.on(started: {
-						carthage.println("*** xcodebuild output can be found in \(temporaryURL.path!)")
+						carthage.println(bullets + "xcodebuild output can be found in " + Color.Wrap(foreground: .Yellow).wrap(temporaryURL.path!))
```
robrix: Only suggestion would be to abstract the specific colours such that we are describing e.g. _the colour for file paths_ or _the colour for errors_ rather than `.Yellow` or `.Green`.

jspahrsummers:
@robrix I went ahead and assigned this to you, based on your above feedback, but I'm happy to take it if you'd rather not own it. :sparkles:

Source/carthage/Extensions.swift
```diff
@@ -55,21 +56,50 @@ extension GitURL: ArgumentType {
...
+	static let bullets: String = {
+		return bulletin.autowrap("***") + " "
+	}()
+	
+	static let URL = [StyleParameter.Underlined] as Color.Wrap
+	static let projectName = [StyleParameter.Bold] as Color.Wrap
+	static let path = Color.Wrap(foreground: .Yellow)
+	
+	static func quote(string: String) -> String {
+		return Color.Wrap(foreground: .Green).autowrap("\u{0022}" + string + "\u{0022}")
```
robrix: Syntax highlighting is flagging this, which makes me wonder if these should be `\\u…` instead?

Source/carthage/Extensions.swift
```diff
@@ -55,21 +56,50 @@ extension GitURL: ArgumentType {
...
+	}
+}
+
+internal struct Terminal {
+	static let term: String? = NSProcessInfo().environment["TERM"] as? String
+	static let isDumb: Bool = (Terminal.term? as NSString?)?.isEqualToString("dumb") ?? false
```
robrix: That’s a thing? That’s hilarious :grin:

robrix:
@jdhealy It’ll probably be easier for me to review this if you don’t force push while it’s in development. We’re not too bothered about having history be pristine, so no worries on that front.

robrix:
Just the one note (on top of your own), but I’m removing my assignment for now since it’s `WIP` and all.

Source/carthage/Extensions.swift
jdhealy: That actually just interpolates to `"`. 

[⌘F for “arbitrary Unicode scalar” in these Swift docs](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/StringsAndCharacters.html).

jspahrsummers: So the intention is just to insert double quotes? Why not use `\"`?

jdhealy:
Sorry, about the force push. :disappointed: 

Accidentally pushed remotely when committing some junk in SourceTree. Not something I intend to make a habit of…

I'll just commit with Github for Mac from here on forward… :wink:

Source/carthage/Extensions.swift
robrix: (TIL!)

robrix:
_Recommended by GitHubbers!_ :grin:

Source/carthage/Extensions.swift
jdhealy: @jspahrsummers Mostly it's just the syntax highlighting in Xcode.

screen shot 2015-01-29 at 10 10 32 pm

Backslash escaped special characters don't have a differentiating highlight color. Seeing two-plus same-colored double-quotes in one valid string literal always makes me pause reflexively…

It definitely trades intuitiveness for scanability though, I'd be happy to change it back :dolphin:

jdhealy:
<b>Feedback Requested On</b>

- Data Dependency
	- `Color.Wrap.autowrap` can't be called before any `--color` flag is processed.
	- Calling `Color.Wrap.autowrap` before `ColorOptions.create` should hit the `fatalError`.
	- If you could recommend a more elegant ReactiveCocoa way there's probably a better (and more type and race safe) solution.

- Commandant
	- Currently, Commandant applies (`<*>`) `ColorOptions.evaluate(m)` in the `Build`, `Update`, `Fetch`, and `Update` files. Maybe there's a more universal way of doing this?

- Todo:
	- [ ] Tests

@robrix, please review? :smile:

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+	static let isDumb: Bool = (Terminal.term?.lowercaseString as NSString?)?.isEqualToString("dumb") ?? false
+	static let isTTY: Bool = isatty(STDOUT_FILENO) == 1
+}
+
+public enum ColorArgument: String, ArgumentType {
+	case Auto = "auto"
+	case Never = "never"
+	case Always = "always"
+	
+	public var isColorful: Bool {
```
robrix: I think we tend to prefer to omit `is` prefixes from API. That would also make it consistent with `ColorOptions.colorful`.

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+		switch self {
+		case .Always: return true
+		case .Never:  return false
+		case .Auto:   return Terminal.isTTY && !Terminal.isDumb
```
robrix: Style: bodies of `case` statements indented on following line:

```swift
switch self {
case .Always:
	return true
…
}
```

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+	
+	public static let name = "color"
+	
+	public static func fromString(string: String) -> ColorArgument? {
```
robrix: Should this be an initializer?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+	public let argument: ColorArgument
+	
+	static var colorful: Bool {
+		get {
+			if Static.colorful == nil { fatalError("`ColorOptions.create` might not have been called…") }
+			return Static.colorful!
+		}
+	}
+
+	private struct Static {
```
robrix: I’m not sure I follow the intent of this struct over `private static var`s?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+		static var colorful: Bool? = nil
+		static var token: dispatch_once_t = 0
+	}
+	
+	public static func create(argument: ColorArgument) -> ColorOptions {
+		dispatch_once(&Static.token, { Static.colorful = argument.isColorful })
```
robrix: Style: trailing closure syntax.

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+		return self(rawValue: string.lowercaseString)
+	}
+	
+}
+
+public struct ColorOptions: OptionsType {
+	public let argument: ColorArgument
+	
+	static var colorful: Bool {
+		get {
```
robrix: Style: omit `get {…}` and just do the computation directly.

Source/carthage/Formatting.swift
jdhealy: Nope, [conforming to ArgumentType](//github.com/Carthage/Commandant/blob/6d123ab08a7d7f0fa1181b3ddd1ead3166c23afc/Commandant/Option.swift#L99).

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+			if Static.colorful == nil { fatalError("`ColorOptions.create` might not have been called…") }
```
robrix: Maybe it’d be better to set up a sensible default instead? E.g. provide a `ColorOptions.color` property with a sensible default instead of this static `Bool`?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+	static let projectName = [StyleParameter.Bold] as Color.Wrap
+	static let path = Color.Wrap(foreground: .Yellow)
+	
+	static func quote(string: String, quotationMark: String = "\u{0022}" /* double quote */) -> String {
+		return Color.Wrap(foreground: .Green).autowrap(quotationMark + string + quotationMark)
+	}
+	
+}
+
+internal struct Terminal {
```
robrix: I think `Terminal` and `Formatting` would be better expressed as normal `struct`s with non-`static` members + an `internal let formatting = …` global.

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+	static let bulletin = Color.Wrap(foreground: .Blue, style: .Bold)
+	static let bullets: String = {
+		return bulletin.autowrap("***") + " "
+	}()
+	
+	static let URL = [StyleParameter.Underlined] as Color.Wrap
+	static let projectName = [StyleParameter.Bold] as Color.Wrap
+	static let path = Color.Wrap(foreground: .Yellow)
+	
+	static func quote(string: String, quotationMark: String = "\u{0022}" /* double quote */) -> String {
```
robrix: I would really prefer `"\""` instead of commenting the intent here. It just makes it harder to read & follow.

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,89 @@
...
+
+import Commandant
+import Foundation
+import LlamaKit
+import PrettyColors
+import ReactiveCocoa
+
+extension Color.Wrap {
+	func autowrap(string: String) -> String {
+		return ColorOptions.colorful ? self.wrap(string) : string
```
robrix: Style: omit `self.` wherever possible.

Source/carthage/Formatting.swift
robrix: Or if not a global for the instance, then a `static let defaultFormatting = …` in `Formatting`.

Source/carthage/Formatting.swift
robrix: Ah, right—thanks.

robrix:
Bunch of small notes :point_up:

Source/carthage/Formatting.swift
jdhealy: Just makes naming a little easier. Is there a performance disadvantage to the `struct`?

Would you like these instead?
```swift
private static var _colorful: Bool? = nil
private static var token: dispatch_once_t = 0
```

robrix:
I’m going to be otherwise occupied for the rest of the day at least, so if anyone else wants to review they should feel free :+1:

Source/carthage/Formatting.swift
jdhealy: I named the `isX` variables following the example of the `stdlib`’s `Array.isEmpty` and the global `isEmpty`.

Doing a bit more poking around, [`Float`](http://swiftdoc.org/type/Float/) and [`Double`](http://swiftdoc.org/type/Double/) have tons of `isX` variables. Even a few `hasX` variables (<a href='http://swiftdoc.org/type/String/' title='hasPrefix, hasSuffix'>hasPrefix, hasSuffix</a>, <a href='http://swiftdoc.org/type/StaticString/' title='hasPointerRepresentation'>hasPointerRepresentation</a>).

But, if the GitHub style is to avoid `isX`, then I'm happy to comply.

jdhealy:
**Pulling this out from @robrix's [comment](https://github.com/Carthage/Carthage/pull/292#discussion_r23852363) because it's important**

So, to clarify, the default — when running a command that applies `ColorOptions.evaluate(m)` for example _`carthage bootstrap --no-build`_ — is that [the `--color` flag has a default value of `.Auto`](//github.com/jdhealy/Carthage/commit/eec90adec02ff99b2fb6d3530af4c60bd278b3f2#diff-d2b778da6ed7a60a1048451fa06f6bceR87). 

When [`ColorOptions.create`](https://github.com/jdhealy/Carthage/commit/eec90adec02ff99b2fb6d3530af4c60bd278b3f2#diff-d2b778da6ed7a60a1048451fa06f6bceR80) gets called, the Singleton-esqe `static var colorful` of `ColorOptions.Static` [has a value set](https://github.com/jdhealy/Carthage/commit/eec90adec02ff99b2fb6d3530af4c60bd278b3f2#diff-d2b778da6ed7a60a1048451fa06f6bceR81).

In the case of `carthage bootstrap --no-build` with no `--color` flag in the arguments, the value `ColorArgument.Auto.isColorful` will overwrite the nil in `Static.colorful`.
- - -

But, we face a data dependency problem if we ever reach the scenario where `Color.Wrap.autowrap` gets called before the Commandant call-stack has `ColorOptions.create` overwrite the nil in `Static.colorful`.

- - -
The `fatalError` and the Singleton-esqe stuff are definitely inelegant and somewhat unsafe.

It seems to me that there are two fixes.

- Commandant — /cc @robb
  - A more universal way of checking the `--color` flag on every command to make sure `ColorOptions.create` gets called.

- ReactiveCocoa — /cc @jspahrsummers 
  - Replace the Singleton-esqe stuff and `dispatch_once` side-effects with ReactiveCocoa

I could definitely use some guidance on these… 

/cc @Carthage/carthage :heart:

Source/carthage/Formatting.swift
jdhealy: [reply is here](https://github.com/Carthage/Carthage/pull/292#issuecomment-72239348)

Source/carthage/Formatting.swift
robrix: Eh, don’t sweat it.

Source/carthage/Formatting.swift
robrix: > Just makes naming a little easier. Is there a performance disadvantage to the `struct`?

In Swift, who knows :wink: It’s more that it imposes unnecessary indirection on the reader.

Those vars look fine :+1:

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,111 @@
...
+	static let path = Color.Wrap(foreground: .Yellow)
+	
+	static func quote(string: String, quotationMark: String = "\"") -> String {
+		return Color.Wrap(foreground: .Green).autowrap(quotationMark + string + quotationMark)
+	}
+	
+}
+
+internal struct Terminal {
+	static let term: String? = getEnvironmentVariable("TERM").value()
```
jspahrsummers: Can this be a computed property instead? It's not immutable, after all.

Source/carthage/Formatting.swift
jspahrsummers: I agree that this use of structs to hold globals is pretty weird.

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,111 @@
...
+//
+
+import Commandant
+import Foundation
+import LlamaKit
+import PrettyColors
+import ReactiveCocoa
+
+extension Color.Wrap {
+	func autowrap(string: String) -> String {
```
jspahrsummers: Can you document this? (And, indeed, this whole file?)

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,111 @@
...
+		return ColorOptions.colorful ? wrap(string) : string
+	}
+}
+
+internal struct Formatting {
+
+	static let bulletin = Color.Wrap(foreground: .Blue, style: .Bold)
```
jspahrsummers: Where does this name come from?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,111 @@
...
+		case .Always:
+			return true
+		case .Never:
+			return false
+		case .Auto:
+			return Terminal.isTTY && !Terminal.isDumb
+		}
+	}
+	
+	public var description: String { return "‘auto’ || ‘always’ || ‘never’" }
```
jspahrsummers: I'm pretty sure the compiler will allow you to declare this as a constant (with `let`).

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,111 @@
...
+				// Most likely an illegal value for `--color`.
+				fatalError(error.description)
+			}
+
+		}
+	}
+
+	private struct Static {
+		struct Colorful {
+			static var singleton: Bool? = nil
```
jspahrsummers: Big :-1: on storing this information globally. It should instead be passed to whichever functions/methods need it.

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,111 @@
...
+	}
+	
+	public static func create(argument: ColorArgument) -> ColorOptions {
+		dispatch_once(&Static.Colorful.token) { Static.Colorful.singleton = argument.isColorful }
+		return self(argument: argument)
+	}
+	
+	public static func evaluate(m: CommandMode) -> Result<ColorOptions> {
+		return create
+			<*> m <| Option(key: "color", defaultValue: ColorArgument.Auto, usage: "Terminal coloring and styling")
```
jspahrsummers: For the `usage`, how about:

> whether to apply formatting and colors to terminal messages

Note the lowercase, too—this matches the other commands.

jspahrsummers:
Thanks! I've added several comments. :zap:

Source/carthage/Formatting.swift
jdhealy: It's used to color the bullets (`***`) and the `*** MIGRATION WARNING ***` which is kind of a bulletin-like notice…

Source/carthage/Formatting.swift
jdhealy: Just did a double-check: "Enums may not contain stored properties" says the compiler.

jspahrsummers: In that case, can you please split the body over multiple lines, to make its method-like nature apparent?

jdhealy:
@jspahrsummers ––– :eyes:

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,98 @@
...
+			return true
+		case .Never:
+			return false
+		case .Auto:
+			return Terminal.isTTY && !Terminal.isDumb
+		}
+	}
+	
+	public var description: String {
+		return "‘auto’ || ‘always’ || ‘never’"
```
mattrubin: Should this return `rawValue`? As it stands, it looks like `println(myColorArg)` will print a description of the `ColorArgument` type, rather than a description of a particular instance.

jdhealy: Yeah, changed this because in earlier commits `carthage help xxxxx` would have output like
`[--color (Enum value)]`


Looking at `carthage help checkout`'s output…
```
[--configuration Release]
the Xcode configuration to build (ignored if --no-build option is present)
```
… Commandant seems to print `[``--key` `defaultValue.description``]`-type output.

I guess I should move the `‘auto’ || ‘always’ || ‘never’` part back to `usage` [like it was here](https://github.com/jdhealy/Carthage/commit/eec90adec02ff99b2fb6d3530af4c60bd278b3f2) and change the description to the `rawValue`. Good catch :+1:

Source/carthage/Extensions.swift
```diff
@@ -57,19 +58,28 @@ extension GitURL: ArgumentType {
 
 /// Logs project events put into the sink.
 internal struct ProjectEventSink: SinkType {
+	let colorOptions: ColorOptions
```
jspahrsummers: Can you make this `private`? I don't think consumers of this sink type should need to read this property.

Source/carthage/Extensions.swift
```diff
@@ -57,19 +58,28 @@ extension GitURL: ArgumentType {
...
+	
+	init(_ colorOptions: ColorOptions) {
+		self.colorOptions = colorOptions
+	}
+	
 	mutating func put(event: ProjectEvent) {
+
```
jspahrsummers: Can you please remove this blank line?

Source/carthage/Extensions.swift
```diff
@@ -57,19 +58,28 @@ extension GitURL: ArgumentType {
 
 /// Logs project events put into the sink.
 internal struct ProjectEventSink: SinkType {
+	let colorOptions: ColorOptions
+	
+	init(_ colorOptions: ColorOptions) {
```
jspahrsummers: Normally, I don't give names to single-parameter initializers, but this has the potential to be confusing. For example, if you saw `ProjectEventSink(x)`, would you know what that meant?

Source/carthage/CopyFrameworks.swift
```diff
@@ -95,7 +95,7 @@ private func inputFiles() -> ColdSignal<String> {
 		}
 }
 
-private func getEnvironmentVariable(variable: String) -> Result<String> {
+internal func getEnvironmentVariable(variable: String) -> Result<String> {
```
jspahrsummers: This should be moved somewhere else if it needs to be used by other files.

Source/carthage/Extensions.swift
```diff
@@ -89,8 +99,10 @@ extension Project {
...
+		
+		let formatting = colorOptions.formatting
+		
+		let migrationMessage = formatting.bulletin(string: "*** MIGRATION WARNING ***") + "\n\nThis project appears to be set up for an older (pre-0.4) version of Carthage. Unfortunately, the directory structure for Carthage projects has since changed, so this project will be migrated automatically.\n\nSpecifically, the following renames will occur:\n\n  \(cartfileLock) -> \(CarthageProjectResolvedCartfilePath)\n  \(carthageBuild) -> \(CarthageBinariesFolderPath)\n  \(carthageCheckout) -> \(CarthageProjectCheckoutsPath)\n\nFor more information, see " + formatting.URL(string: "https://github.com/Carthage/Carthage/pull/224") + ".\n"
```
jspahrsummers: Can the formatter add these asterisks itself? It'd be nice to have that shared amongst all formatting of that type.

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+//
+
+import Commandant
+import Foundation
+import LlamaKit
+import PrettyColors
+import ReactiveCocoa
+
+/// Wraps or passes through a string.
+func wrap(colorful: Bool)(wrap: Color.Wrap)(string: String) -> String {
```
jspahrsummers: This doesn't appear to be used elsewhere. Can you please make it `private`?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+//
+//  Created by J.D. Healy on 1/29/15.
+//  Copyright (c) 2015 Carthage. All rights reserved.
+//
+
+import Commandant
+import Foundation
+import LlamaKit
+import PrettyColors
+import ReactiveCocoa
```
jspahrsummers: Where is RAC used here?

Source/carthage/Formatting.swift
jspahrsummers: Also, it looks like the first two arguments are always passed in together. Can those be part of the same argument list, with only the last one curried?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+
+/// Wraps or passes through a string.
+func wrap(colorful: Bool)(wrap: Color.Wrap)(string: String) -> String {
+	return colorful ? wrap.wrap(string) : string
+}
+
+/// Information about the possible parent terminal.
+internal struct Terminal {
```
jspahrsummers: This also doesn't appear to be used elsewhere. Can you please mark it as `private`?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+	static var term: String? {
```
jspahrsummers: Can we maybe call this `terminalType` or something instead?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+		return getEnvironmentVariable("TERM").value()
+	}
```
jspahrsummers: Can you please add blank lines between these implementations?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+	static var isDumb: Bool {
+		return (Terminal.term?.lowercaseString as NSString?)?.isEqualToString("dumb") ?? false
```
jspahrsummers: Why the use of `NSString` here? Could this just use `caseInsensitiveCompare` directly?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+	}
+	static var isTTY: Bool {
+		return isatty(STDOUT_FILENO) == 1
```
jspahrsummers: I don't see where this is guaranteed to be `1`. Can we just test `!= 0` instead?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+public enum ColorArgument: String, ArgumentType, Printable {
```
jspahrsummers: Can you please document this type and its members?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+//  Copyright (c) 2015 Carthage. All rights reserved.
+//
+
+import Commandant
+import Foundation
+import LlamaKit
+import PrettyColors
+import ReactiveCocoa
+
+/// Wraps or passes through a string.
```
jspahrsummers: Can you please clarify this documentation? What does it mean to “wrap” a string? (This is somewhat rhetorical, because I know the answer, but I'd love for the documentation to express it.)

Source/carthage/Formatting.swift
jspahrsummers: Can you please document the members of this type?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+		return self(rawValue: string.lowercaseString)
+	}
+	
+}
+
+public struct ColorOptions: OptionsType {
+	let argument: ColorArgument
+	let formatting: Formatting
+	
+	struct Formatting {
```
jspahrsummers: Why is this a nested type? It doesn't seem to depend upon anything from the outer type.

jspahrsummers: Can this either be `public` or `private`? I like having the explicit indicator of how this is supposed to be used.

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+		let colorful: Bool
+		
+		/// Wraps or passes through a string.
```
jspahrsummers: As [above](https://github.com/Carthage/Carthage/pull/292/files#r24192165), can you please clarify what this means?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+		typealias Wrap = (string: String) -> String
+		
+		init(_ colorful: Bool) {
+			self.colorful    = colorful
+			self.bulletin    = wrap(colorful)(wrap: Color.Wrap(foreground: .Blue, style: .Bold))
```
jspahrsummers: The use of `self` on most of these lines seems to be unnecessary.

Source/carthage/Formatting.swift
jspahrsummers: Can you please document this type and all its members?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+		
+		let URL: Wrap
+		let projectName: Wrap
+		let path: Wrap
+		
+		/// Wraps a string in quotation marks and formatting.
+		func quote(string: String, quotationMark: String = "\"") -> String {
+			return wrap(colorful)(wrap: Color.Wrap(foreground: .Green))(string: quotationMark + string + quotationMark)
+		}
+
```
jspahrsummers: Can you please :hocho: this extraneous blank line?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,103 @@
...
+		
+	}
+	
+	static func create(argument: ColorArgument) -> ColorOptions {
+		return self(argument: argument, formatting: Formatting(argument.isColorful))
+	}
+	
+	public static func evaluate(m: CommandMode) -> Result<ColorOptions> {
+		return create
+			<*> m <| Option(key: "color", defaultValue: ColorArgument.Auto, usage: "apply Terminal colors and formatting — ‘auto’ || ‘always’ || ‘never’")
```
jspahrsummers: Can you make this usage string a bit more human-readable? For example:

> whether to apply terminal colors and formatting (one of `auto`, `always`, or `never`)

Source/carthage/Update.swift
```diff
@@ -32,11 +32,11 @@ public struct UpdateCommand: CommandType {
 public struct UpdateOptions: OptionsType {
 	public let buildAfterUpdate: Bool
 	public let configuration: String
-	private let checkoutOptions: CheckoutOptions
+	internal let checkoutOptions: CheckoutOptions
```
jspahrsummers: Why this change?

jspahrsummers:
Sorry for all the notes! This is as rigorously as we review each other, too. :wink:

Source/carthage/Formatting.swift
jdhealy: Yeah, [not great documentation](https://developer.apple.com/library/ios/documentation/System/Conceptual/ManPages_iPhoneOS/man3/isatty.3.html). `!= 0` is probably safer.

Source/carthage/Update.swift
jdhealy: [Bootstrap evaluates UpdateOptions.](https://github.com/jdhealy/Carthage/blob/22d5892fa9a485aa2d0c68e1bf879ba439608260/Source/carthage/Bootstrap.swift#L23)

[We use the colorOptions here.](https://github.com/jdhealy/Carthage/blob/22d5892fa9a485aa2d0c68e1bf879ba439608260/Source/carthage/Bootstrap.swift#L31)

jspahrsummers: Then can this be `public` instead?

Source/carthage/Formatting.swift
jdhealy: The terseness is mostly a side effect of trying to keep the string around 70 characters, so that none of the values gets chopped in half by wrapping around a common 80-char Terminal width.

Preferably, we could be more wordy with the first part of the description and put the values after a new line.

I've opened a pull request (Carthage/Commandant#17) to indent every line of a multi-line string.

Your thoughts?

Source/carthage/Formatting.swift
jdhealy: This is probably the source, right?

http://www.opensource.apple.com/source/Libc/Libc-262/gen/isatty.c

robrix: For now, probably.

jspahrsummers: It doesn't matter what the source says—in C, 0 is false, and anything non-zero is true. We should honor that convention.

Source/carthage/Formatting.swift
jspahrsummers: > The terseness is mostly a side effect of trying to keep the string around 70 characters, so that none of the values gets chopped in half by wrapping around a common 80-char Terminal width.

I don't think we need to worry about that. Let's be as verbose as we want, and deal with wrapping separately/later.

Source/carthage/Formatting.swift
robrix: :100:% agreed. I primarily meant to indicate that the source should be considered `volatile` :smile:

jspahrsummers: @robrix Sorry for the confusion. My comment was meant for @jdhealy.

jdhealy: Yup, I'm in agreement. [Made the change](https://github.com/jdhealy/Carthage/commit/91efff1258841a69ee2b4122ccfef4e531fd34b6)

jdhealy:
Ready for review. Appreciate the notes and the rigor! :scroll: :clipboard: :ledger:

Source/carthage/Environment.swift
```diff
@@ -0,0 +1,39 @@
...
+
+	if let value = environment[variable] as String? {
+		return success(value)
+	} else {
+		return failure(CarthageError.MissingEnvironmentVariable(variable: variable).error)
+	}
+}
+
+/// Information about the possible parent terminal.
+internal struct Terminal {
```
robrix: This would be easier to test if it weren’t (effectively) a singleton. I.e. if its properties weren’t `static`, and if `ColorArgument` were parameterized (or at least parameterizable) by it.

I’m not sure how much we care about testing around this, though. @jspahrsummers, any thoughts?

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,95 @@
...
+	
+	/// Whether to color and format.
+	public var isColorful: Bool {
+		switch self {
+		case .Always:
+			return true
+		case .Never:
+			return false
+		case .Auto:
+			return Terminal.isTTY && !Terminal.isDumb
```
robrix: I was wondering if we’d want to check whether `TERM` contains the string `color`, but it’s probably not crucial for this PR.

There’s also apparently [`tput colors`](http://blog.sanctum.geek.nz/terminal-colour-tolerance/) if we ever wanted to get fancy and use more/fewer colours depending on the `TERM`’s support for them. That _definitely_ doesn’t need to be addressed in this PR, however.

Source/carthage/Formatting.swift
```diff
@@ -0,0 +1,95 @@
...
+		typealias Wrap = (string: String) -> String
+		
+		init(_ colorful: Bool) {
+			self.colorful = colorful
+			bulletin    = wrap(colorful)(wrap: Color.Wrap(foreground: .Blue, style: .Bold))
+			bullets     = self.bulletin(string: "***") + " "
+			URL         = wrap(colorful)(wrap: Color.Wrap(styles: .Underlined))
+			projectName = wrap(colorful)(wrap: Color.Wrap(styles: .Bold))
+			path        = wrap(colorful)(wrap: Color.Wrap(foreground: .Yellow))
+		}
```
robrix: This is kinda slick. I dig it.

Source/carthage/Environment.swift
jspahrsummers: I agree that it would be nice if this didn't simply contain static properties, but I think it's acceptable for the time being.

robrix: :+1:

jdhealy:
Team effort! Thanks for all the help, guys!

carlton dance

robrix:
:clap: nice one @jdhealy :smile:

Issue #288, opened 2015-01-24
Git clone output into log file


pietbrauer:
It would be nice to have an output for `git` as well (not only for `xcodebuild`). As in [some big projects](https://github.com/libgit2/objective-git) it can take multiple minutes (depending on the internet connection of course) to checkout a repository recursive.

PR #297, opened 2015-02-04
Add --verbose option to Update (and Bootstrap)

### LINKED ISSUE #296: carthage bootstrap --verbose not working
I'm trying to run carthage on CircleCI and the builds are failing. I can't view the log file so I'd like to see the verbose output but the `--verbose` flag isn't working for the `bootstrap` command. I'm not familiar enough with the way carthage and RAC 3 work to make a PR for this but if someone can point me in the right direction, I don't mind helping out here.

mattrubin:
Fixes #296.

When git output is written to the log file (#288), that should also be routed according to the `--verbose` option.


robrix:
:+1:

Issue #281, opened 2015-01-15
Code Sign error: No code signing identities found


refried:
Hi — first time Carthaging here, and having a bit of trouble.  It looks like it's related to code signing, but as far as I can tell, my signing profile is set up fine.
```
*** Fetching SwiftyJSON
*** Checking out SwiftyJSON at "2.1.3"
*** xcodebuild output can be found in /var/folders/kl/tdsq1jgx3_b33877x8vkmk480000gn/T/carthage-xcodebuild.MLdkEs.log
*** Building scheme "SwiftyJSON" in SwiftyJSON.xcworkspace
** BUILD FAILED **

The following build commands failed:
	Check dependencies
(1 failure)
```
The log file shows:
```
** BUILD SUCCEEDED **

Build settings from command line:
    SDKROOT = iphoneos8.1

=== BUILD TARGET SwiftyJSON OF PROJECT SwiftyJSON WITH CONFIGURATION Release ===

Check dependencies
Code Sign error: No code signing identities found: No valid signing identities (i.e. certificate and private key pair) 
matching the team ID “(null)” were found.
CodeSign error: code signing is required for product type 'Framework' in SDK 'iOS 8.1'
```

refried:
I guess a distribution profile is needed in addition to a development profile.  Might want to update the readme, if it isn't there already?

mattrubin:
Carthage uses a project's own release build settings when building frameworks. It looks like SwiftyJSON has [`CODE_SIGN_IDENTITY = "iPhone Distribution"` set in its Release configuration](https://github.com/SwiftyJSON/SwiftyJSON/blob/f5347e5a90dfe00c32fa1fe228472291cacb386a/SwiftyJSON.xcodeproj/project.pbxproj#L497).

Instead of setting up a distribution signing identity, you might be able to just change this build setting to "iOS Developer" in the SwiftyJSON project. I *think* that a custom framework embedded in an iOS app will be re-signed when it's copied into the app bundle, so signing for distribution on initial build may not be necessary.

refried:
Oh. Thanks!

jspahrsummers:
You could also pass `--configuration Debug` to Carthage, to select the Debug codesigning settings.

jspahrsummers:
I don't want to support simulator-only builds as a configuration option, but maybe that's something Carthage can do automatically if it fails to detect a code signing identity.

@Carthage/carthage Thoughts?

jspahrsummers:
[...]

Issue #235, opened 2014-12-22
Build with the null codesigning identity so a certificate is not required


pchatsu:
I don't join iOS developer program yet, and only use iOS simulator.
I used `carthage update` , but failed and the build log said,

```
Check dependencies
Code Sign error: No code signing identities found: No valid signing identities (i.e. certificate and private key pair) matching the team ID “(null)” were found.
CodeSign error: code signing is required for product type 'Framework' in SDK 'iOS 8.1'
```

So I tried to regist the code sign with Xcode's suggestion, but it said "You need to add an apple ID account that is enrolled in Developer's program.".

Can't I use carthage without joining iOS developer program for `xcodebuild`?

sebbean:
What about CI builds (Travis for instance)

I was hoping to use Carthage to pull in my private custom framework dependency. Works great locally, however breaks on Travis environment (presumably because of lack of code signing credentials)

Took me all day to track down this as the most likely culprit for this error:

```
carthage build --no-skip-current
*** xcodebuild output can be found in /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/carthage-xcodebuild.cnE6me.log
*** Building scheme "Gravy" in Gravy.xcodeproj
** BUILD FAILED **
```

Side note: It may be helpful to have a flag that tail's the xcodebuild log during build, as Travis doesnt allow you to go in and examine the file system as far as I'm aware - makes it a bit tricky to debug this error.

jspahrsummers:
@sebbean You'll have to ensure that Travis imports a code signing identity. You can see how Carthage itself does it in the [`cibuild`](https://github.com/Carthage/Carthage/blob/12fb7881876b4987584e49d00befcd2a8b6b354d/script/cibuild#L29-L51) script.

> Side note: It may be helpful to have a flag that tail's the xcodebuild log during build, as Travis doesnt allow you to go in and examine the file system as far as I'm aware - makes it a bit tricky to debug this error.

Filed as https://github.com/Carthage/Carthage/issues/273.

dwlnetnl:
I'd really like to option to build only for the simulator because I don't have a code signing identity yet. It keeps starting app developers from using Carthage.

jspahrsummers:
@dwlnetnl Thanks for the suggestion! Unfortunately, that would add a significant amount of complexity. Right now, it's very clear that when Carthage builds for iOS, you'll get a universal framework that you can use in both device and simulator builds.

[...]

PR #302, opened 2015-02-09
Add --platform flag to limit a build

### LINKED ISSUE #127: Argument to build only iOS or OS X targets
Building a large dependency graph can take some time. If you know that you'll only need the frameworks for a particular platform, the process could be expedited with a flag like `--platform Mac` or something like that.

jspahrsummers:
Resolves #127.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -95,14 +95,14 @@ public struct BuildArguments {
 	/// The configuration to use when building the project.
 	public var configuration: String?
 
-	/// The platform to build for.
-	public var platform: Platform?
+	/// The platform SDK to build for.
+	public var sdk: SDK?
 
-	public init(project: ProjectLocator, scheme: String? = nil, configuration: String? = nil, platform: Platform? = nil) {
+	public init(project: ProjectLocator, scheme: String? = nil, configuration: String? = nil, sdk: SDK? = nil) {
 		self.project = project
 		self.scheme = scheme
 		self.configuration = configuration
-		self.platform = platform
+		self.sdk = sdk
 	}
 
 	/// The `xcodebuild` invocation corresponding to the receiver.
@@ -117,8 +117,8 @@ public struct BuildArguments {
 			args += [ "-configuration", configuration ]
 		}
 
-		if let platform = platform {
-			args += platform.arguments
+		if let sdk = sdk {
+			args += sdk.arguments
 		}
 
 		return args
@@ -238,8 +238,8 @@ public func schemesInProject(project: ProjectLocator) -> ColdSignal<String> {
 		.map { (line: String) -> String in line.stringByTrimmingCharactersInSet(NSCharacterSet.whitespaceCharacterSet()) }
 }
 
-/// Represents a platform or SDK buildable by Xcode.
-public enum Platform {
+/// Represents an SDK buildable by Xcode.
+public enum SDK {
 	/// Mac OS X.
 	case MacOSX
 
@@ -249,9 +249,8 @@ public enum Platform {
 	/// iOS, for the simulator.
 	case iPhoneSimulator
 
-	/// Attempts to parse a platform name from a string returned from
-	/// `xcodebuild`.
-	public static func fromString(string: String) -> Result<Platform> {
+	/// Attempts to parse an SDK name from a string returned from `xcodebuild`.
+	public static func fromString(string: String) -> Result<SDK> {
 		switch string {
 		case "macosx":
 			return success(.MacOSX)
@@ -263,11 +262,11 @@ public enum Platform {
 			return success(.iPhoneSimulator)
 
 		default:
-			return failure(CarthageError.ParseError(description: "unexpected platform key \"(string)\"").error)
+			return failure(CarthageError.ParseError(description: "unexpected SDK key \"(string)\"").error)
 		}
 	}
 
-	/// Whether this platform targets iOS.
+	/// Whether this SDK targets iOS.
 	public var targetsiOS: Bool {
 		switch self {
 		case .iPhoneOS:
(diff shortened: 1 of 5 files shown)


robrix:
:sparkles:

Source/CarthageKit/Xcode.swift
```diff
@@ -238,8 +238,65 @@ public func schemesInProject(project: ProjectLocator) -> ColdSignal<String> {
...
+		}
+	}
+}
+
+public func == (lhs: Platform, rhs: Platform) -> Bool {
+	switch (lhs, rhs) {
+	case (.Mac, .Mac):
+		return true
+
+	case (.iOS, .iOS):
```
robrix: Quibble: these `case`s can be combined with `,`.

Source/CarthageKit/Xcode.swift
```diff
@@ -306,6 +350,37 @@ public enum Platform {
...
 
+public func == (lhs: SDK, rhs: SDK) -> Bool {
+	switch (lhs, rhs) {
+	case (.MacOSX, .MacOSX):
+		return true
+
+	case (.iPhoneSimulator, .iPhoneSimulator):
+		return true
+
+	case (.iPhoneOS, .iPhoneOS):
```
robrix: Same quibble.

Source/CarthageKit/Xcode.swift
```diff
@@ -689,28 +776,38 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
...
 			let folderURL = workingDirectoryURL.URLByAppendingPathComponent(platform.relativePath, isDirectory: true)
 
-			switch platform {
-			case .iPhoneSimulator, .iPhoneOS:
-				return settingsByTarget(buildPlatform(.iPhoneSimulator))
-					.map { simulatorSettingsByTarget -> ColdSignal<(BuildSettings, BuildSettings)> in
-						return settingsByTarget(buildPlatform(.iPhoneOS))
-							.map { deviceSettingsByTarget -> ColdSignal<(BuildSettings, BuildSettings)> in
-								assert(simulatorSettingsByTarget.count == deviceSettingsByTarget.count, "Number of targets built for iOS Simulator (\(simulatorSettingsByTarget.count)) does not match number of targets built for iOS Device (\(deviceSettingsByTarget.count))")
+			// TODO: Generalize this further?
```
robrix: If it would clarify things, sure; otherwise I wouldn’t sweat it.

robrix: What’s the SDK for Apple Watch called? :trolleybus:

robrix:
Couple of quibbles above which you may or may not care about; :sparkles: overall!

:cherries:

jspahrsummers:
:cherry_blossom:

jspahrsummers:
I also addressed your point from https://github.com/Carthage/Carthage/pull/303#discussion_r24330943.

Source/carthage/Build.swift
```diff
@@ -126,21 +126,100 @@ public struct BuildCommand: CommandType {
...
+		}
+	}
+}
+
+extension BuildPlatform: ArgumentType {
+	public static let name = "platform"
+
+	private static let acceptedStrings: [String: BuildPlatform] = [
+		"Mac": .Mac, "macosx": .Mac,
+		"iOS": .iOS, "iphoneos": .iOS, "iphonesimulator": .iOS,
```
robrix: So if it’s `iphonesimulator` we’ll build `iphoneos` as well. That seems fine for now, but maybe something that we’d want to tackle later on.

robrix:
:+1:

Source/carthage/Build.swift
jspahrsummers: This is specifically just for `$PLATFORM_NAME` support in the Run Script phase, which is why these variants aren't documented.

robrix: Yep, we’re on the same page there. Just musing on whether it’d be valuable to build for `iphonesimulator`-but-not-`iphoneos` (or vice versa).

jspahrsummers: Ah, I don't think so, for the same reasons expressed in https://github.com/Carthage/Carthage/issues/235#issuecomment-73570007.

robrix: That is an excellent point :+1:

rromanchuk: This flag should be added to the docs

robrix: As @jspahrsummers noted above, this isn’t usage we want to encourage for general use.

That said, if we do document it, we should document _that_ as well, since if Apple changes these, presumably `carthage` will track those changes.

rromanchuk: @robrix maybe i'm missing something fundamental with carthage, but I usually can never successfully run a `carthage update` because it explodes on trying to sign the code with a mac developer cert, or whatever the developer set.  I thought everything was signed on copy, shouldn't it be best practice to use 'DO NOT SIGN'  in build settings?

robrix: I’m sorry, I don’t think I follow the connection between `--platform`, the (undocumented) values I was commenting on in this thread, and signing or not signing. Could you open an issue to describe what you’re running into please?

rromanchuk: @robrix  `carthage update` almost never completes "successfully" because of code signing issues.  --platform cuts the chance of "failed builds" in half

https://www.google.com/search?q=carthage+code+signing&oq=carthage+code+signing&aqs=chrome..69i57j69i60l2.4609j0j1&sourceid=chrome&es_sm=91&ie=UTF-8

https://github.com/Carthage/Carthage/issues/281

Imagine a user who downloads xcode, creates a `Cartfile` adds `SwiftyJSON` and runs `carthage update`.... it's not a good first experience.

robrix: Okay, thank you. Could you please file an issue describing the documentation you would like to see?

PR #310, opened 2015-02-11
Add a few lines about transitive dependencies

### LINKED ISSUE #270: Note in the README that Carthage supports transitive dependencies
This seems to be a recurring point of confusion. The absence of any “spec” file leads some to believe that we don't look at dependencies' dependencies.

Carthage will automatically include dependencies' `Cartfile`s in the dependency graph, and only select versions that satisfy the constraints of _all_ projects in the graph.

Palleas:
Should fix #270

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,18 @@ By default, Carthage will directly [check out][Carthage/Checkouts] dependencies�
 
 When run this way, Carthage will write to your repository’s `.gitmodules` and `.git/config` files, and automatically update the submodules when the dependencies’ versions change.
 
+### Transitive dependencies
+
+If the framework you want to add to your project has dependencies, Carthage will automatically fetch and build them for you. Let's take a real life example with the command-line arguments parsing framework [Commandant](https://github.com/Carthage/Commandant) (also used in Carthage).
+
+Commandant currently depends on [the following frameworks](https://github.com/Carthage/Commandant/blob/master/Cartfile):
+
+  * [Quick](http://github.com/Quick/Quick)
+  * [Nimble](http://github.com/Quick/Nimble)
+  * Carthage's [own fork](https://github.com/Carthage/LlamaKit) of [LlamaKit](https://github.com/LlamaKit/LlamaKit)
+
+Once you've added [Commandant](https://github.com/Carthage/Commandant) as a dependency and ran `carthage update`, Carthage will look for a Cartfile, fetch those repositories for you and build the associated frameworks. You do not need the framework to have any spec file whatsoever.
+
 ## Supporting Carthage for your framework
 
 **Carthage only officially supports dynamic frameworks**. Dynamic frameworks can be used on any version of OS X, but only on **iOS 8 or later**.
--- a/README.md
+++ b/README.md
@@ -84,23 +84,15 @@ In rare cases, you may want to also copy each dependency into the build product 
 
 If you’ve modified your [Cartfile][], or you want to update to the newest versions of each framework (subject to the requirements you’ve specified), simply run the `carthage update` command again.
 
+### Transitive dependencies
+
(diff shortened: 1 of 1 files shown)


swizzlr:
It would be cool if this could mention the fact that you also need to bring in the dependency into your project – Carthage does not embed the dependency.

gfontenot:
I also just opened a PR to move Quick and Nimble into Cartfile.private for Commandant. This PR made me realize that hadn't ever happened. Might be good to make a note of the fact that Quick and Nimble are only development dependencies and won't get built alongside actual dependencies.

jspahrsummers:
Thanks for submitting this! :sparkles:

I think we can simplify this section a great deal, though. The key thing I want to convey is:

> If a framework you add has a `Cartfile` of its own, Carthage will automatically bring in the dependencies specified there as well.

@swizzlr's point is valuable too. In any case, I don't think we need to dive into an example—it leads to verbosity that I'm not sure people will read through.

Can we also switch this section and the section about submodules? Submodules are IMO the least important thing to talk about here.

Palleas:
@jspahrsummers Something like that?

README.md
```diff
@@ -84,6 +84,10 @@ In rare cases, you may want to also copy each dependency into the build product
 
 If you’ve modified your [Cartfile][], or you want to update to the newest versions of each framework (subject to the requirements you’ve specified), simply run the `carthage update` command again.
 
+### Transitive dependencies
+
+If the framework you want to add to your project has dependencies, Carthage will automatically fetch and build them for you. Once you've added a dependency and ran `carthage update`, Carthage will look for a Cartfile, fetch those repositories and build the associated frameworks that **you will have to drag yourself into your project**. You do not need the framework to have any spec file whatsoever.
```
jspahrsummers: Can we clarify that this only applies to frameworks _that have a `Cartfile`_. It may have “dependencies,” but Carthage won't know about them without the `Cartfile`.

jspahrsummers: Also, I'd like to be opaque about the building part, because of the existence of the `--no-build` flag. We can just say that the dependencies-of-dependencies are “brought in” or “added” as well.

jspahrsummers: Finally, I think we should remove the last sentence, because the `Cartfile` could be interpreted as a kind of spec. It's true that we can just read the project file for everything else, but Carthage does need _some_ information to make transitivity work.

jspahrsummers:
Sorry, just a few more notes. :smile:

README.md
```diff
@@ -84,6 +84,10 @@ In rare cases, you may want to also copy each dependency into the build product
 
 If you’ve modified your [Cartfile][], or you want to update to the newest versions of each framework (subject to the requirements you’ve specified), simply run the `carthage update` command again.
 
+### Transitive dependencies
```
robrix: Can we find clearer wording than “transitive dependencies”? I think this will be needlessly opaque and off-putting; it makes it sound more complicated than it is.

Palleas: Trying to think of a better title but I'm stuck on "dependencies' dependencies" right now... maybe "Framework's dependencies" since we are in the "Adding frameworks to an application" section?

robrix: Nested dependencies?

Palleas: Works for me :)

README.md
Palleas: > Can we clarify that this only applies to frameworks that have a Cartfile. It may have “dependencies,” but Carthage won't know about them without the Cartfile.

Done.

> Also, I'd like to be opaque about the building part, because of the existence of the --no-build flag. We can just say that the dependencies-of-dependencies are “brought in” or “added” as well.

I went for retrieved. :grimacing: 

@jspahrsummers thanks for the feedback, let me know how it is now!

jspahrsummers:
:shipit: Thanks @Palleas!

Palleas:
Wouhou! \o/

PR #332, opened 2015-02-15
Remove xcconfigs from public Cartfile.

paulyoung:
This already exists in `Cartfile.private`, which is the correct place for it to live according to @jspahrsummers.

See https://twitter.com/jspahrsummers/status/566837562736508928 for a brief explanation.


jspahrsummers:
Thanks! :sparkles:

PR #344, opened 2015-02-18
Include symlinks as-is in `carthage archive`

### LINKED ISSUE #343: `carthage archive` does not include symlinks properly
We need to pass `--symlinks` to `zip` in order for symlinks to be encoded as such. Right now, they're just archived as copies of the original file.

jspahrsummers:
Resolves #343.


jspahrsummers:
Since this works as intended, and people are beginning to use Carthage's binary frameworks support, I'm going to self-merge this and push a new release.

PR #349, opened 2015-02-20
Don't pass in -sdk macosx

### LINKED ISSUE #347: --platform still sometimes builds platforms not specified
This appears to be an `xcodebuild` issue, because sometimes implicit dependency lookup in a workspace results in the wrong platform being chosen for a dependency, which then causes `carthage build` to fail or do unnecessary work.

For example, `ReactiveCocoa.xcodeproj` depends on `LlamaKit.xcodeproj`, but they're siblings in a workspace together. Sometimes, building ReactiveCocoa's Mac target will result in LlamaKit's iOS target being picked up as a dependency. :worried:

jspahrsummers:
Hopefully fixes #347. :pray:

**To do:**

- [x] Real world testing

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -339,7 +339,13 @@ public enum SDK: Equatable {
 	private var arguments: [String] {
 		switch self {
 		case .MacOSX:
-			return [ "-sdk", "macosx" ]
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Seems to work for me.

robrix:
LGTM

PR #355, opened 2015-02-23
Don't synchronize every `stdoutHandle`.

### LINKED ISSUE #350: synchronizeFile() is not supported on piped stdout
If you pipe `carthage build` into something else (or, say, run it from Xcode), the `--verbose` flag will cause the following crash:

> *** Terminating app due to uncaught exception 'NSFileHandleOperationException', reason: '*** -[_NSStdIOFileHandle synchronizeFile]: Operation not supported'

We shouldn't invoke [invoke `synchronizeFile()`](https://github.com/Carthage/Carthage/blob/26ea162f001f5333bcd5efcc73f7574705a5f46a/Source/carthage/Build.swift#L35) on an `stdout` handle.

jdhealy:
Fixes #350.

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -31,16 +31,8 @@ public struct BuildCommand: CommandType {
 
 				let (stdoutSignal, schemeSignals) = self.buildProjectInDirectoryURL(directoryURL, options: options)
 				stdoutSignal.observe { data in
-
 					stdoutHandle.writeData(data)
-
-					switch stdoutHandle {
-					case NSFileHandle.fileHandleWithStandardOutput(): fflush(stdout)
-					case NSFileHandle.fileHandleWithStandardError(): fflush(stderr)
-					case NSFileHandle.fileHandleWithStandardInput(): break
-					default: stdoutHandle.synchronizeFile()
-					}
-
+					stdoutHandle.synchronizeFile()
 				}
 
 				let formatting = options.colorOptions.formatting
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -31,8 +31,16 @@ public struct BuildCommand: CommandType {
 
 				let (stdoutSignal, schemeSignals) = self.buildProjectInDirectoryURL(directoryURL, options: options)
 				stdoutSignal.observe { data in
+
 					stdoutHandle.writeData(data)
-					stdoutHandle.synchronizeFile()
+
+					switch stdoutHandle {
+					case NSFileHandle.fileHandleWithStandardOutput(): fflush(stdout)
+					case NSFileHandle.fileHandleWithStandardError(): fflush(stderr)
+					case NSFileHandle.fileHandleWithStandardInput(): break
+					default: stdoutHandle.synchronizeFile()
+					}
+
 				}
 
 				let formatting = options.colorOptions.formatting
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -31,8 +31,8 @@ public struct BuildCommand: CommandType {
 
 				let (stdoutSignal, schemeSignals) = self.buildProjectInDirectoryURL(directoryURL, options: options)
 				stdoutSignal.observe { data in
-					stdoutHandle.writeData(data)
-					stdoutHandle.synchronizeFile()
+					stdoutHandle.write(data)
+					stdoutHandle.flush()
 				}
 
 				let formatting = options.colorOptions.formatting
@@ -95,9 +95,9 @@ public struct BuildCommand: CommandType {
 
 	/// Opens a temporary file for logging, returning the handle and the URL to
 	/// the file on disk.
-	private func openTemporaryLogFile(options: BuildOptions) -> ColdSignal<(NSFileHandle, NSURL?)> {
+	private func openTemporaryLogFile(options: BuildOptions) -> ColdSignal<(FileHandle, NSURL?)> {
 		if options.verbose {
-			let out: (NSFileHandle, NSURL?) = (NSFileHandle.fileHandleWithStandardOutput(), nil)
+			let out: (FileHandle, NSURL?) = (FileHandle(fileDescriptor: STDOUT_FILENO), nil)
 			return .single(out)
 		}
 
@@ -115,10 +115,10 @@ public struct BuildCommand: CommandType {
 				return String.fromCString(ptr.baseAddress)!
 			}
 
-			let stdoutHandle = NSFileHandle(fileDescriptor: logFD, closeOnDealloc: true)
+			let stdoutHandle = FileHandle(fileDescriptor: logFD)
 			let temporaryURL = NSURL.fileURLWithPath(temporaryPath, isDirectory: false)!
 
-			let out: (NSFileHandle, NSURL?) = (stdoutHandle, temporaryURL)
+			let out: (FileHandle, NSURL?) = (stdoutHandle, temporaryURL)
 			return .single(out)
 		}
 	}
--- a/Source/carthage/Extensions.swift
+++ b/Source/carthage/Extensions.swift
@@ -47,7 +47,7 @@ internal func print<T>(object: T) {
 	}
 }
 
-internal class FileHandle {
+internal final class FileHandle {
 	let fileDescriptor: Int32
 	let pointer: UnsafeMutablePointer<FILE>
 
--- a/Source/carthage/Extensions.swift
+++ b/Source/carthage/Extensions.swift
@@ -56,12 +56,22 @@ internal final class FileHandle {
 		pointer = fdopen(fd, "r+")
 	}
 
-	func write(data: NSData) -> Result<()> {
-		return Darwin.write(fileDescriptor, data.bytes, UInt(data.length)) != 0 ? success() : failure()
+	func write(data: NSData) -> Result<Int> {
+		switch Darwin.write(fileDescriptor, data.bytes, UInt(data.length)) {
+		case (let value) where value != 0:
+			return success(value)
+		default:
+			return failure()
+		}
 	}
 
-	func flush() -> Result<()> {
-		return fflush(pointer) != 0 ? success() : failure()
+	func flush() -> Result<Int32> {
+		switch fflush(pointer) {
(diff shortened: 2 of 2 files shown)


Source/carthage/Build.swift
```diff
@@ -31,8 +31,16 @@ public struct BuildCommand: CommandType {
 
 				let (stdoutSignal, schemeSignals) = self.buildProjectInDirectoryURL(directoryURL, options: options)
 				stdoutSignal.observe { data in
+
 					stdoutHandle.writeData(data)
-					stdoutHandle.synchronizeFile()
+
+					switch stdoutHandle {
```
jspahrsummers: Hm, I don't like having to introspect this value.

Could `openTemporaryLogFile` give us a `FILE *` instead? Then we can treat them all the same.

jspahrsummers:
Thanks for the pull request!

jdhealy:
Sorry, this took so long to push up…

I've never gone this low level before (all the way down to Darwin), but I've been testing it every which way with #369 for over a week and it seems solid.

A few people with more low-level experience might want to take a close :eyes: and make sure there's not anything subtle I overlooked.

Source/carthage/Extensions.swift
```diff
@@ -47,6 +47,28 @@ internal func print<T>(object: T) {
...
+	let fileDescriptor: Int32
+	let pointer: UnsafeMutablePointer<FILE>
+
+	init(fileDescriptor fd: Int32) {
+		fileDescriptor = fd
+		pointer = fdopen(fd, "r+")
+	}
+
+	func write(data: NSData) -> Result<()> {
+		return Darwin.write(fileDescriptor, data.bytes, UInt(data.length)) != 0 ? success() : failure()
```
robrix: Can `failure` capture the `errno`?

Source/carthage/Extensions.swift
```diff
@@ -47,6 +47,28 @@ internal func print<T>(object: T) {
...
+	}
+
+	func flush() -> Result<()> {
+		return fflush(pointer) != 0 ? success() : failure()
```
robrix: Ditto here: capturing the `errno` would be good.

Source/carthage/Extensions.swift
```diff
@@ -47,6 +47,28 @@ internal func print<T>(object: T) {
...
+	}
+
+	deinit {
+		fclose(pointer)
```
robrix: Noe that `fclose` can also error.

Source/carthage/Extensions.swift
```diff
@@ -47,6 +47,28 @@ internal func print<T>(object: T) {
 	}
 }
 
+internal class FileHandle {
+	let fileDescriptor: Int32
+	let pointer: UnsafeMutablePointer<FILE>
+
+	init(fileDescriptor fd: Int32) {
+		fileDescriptor = fd
+		pointer = fdopen(fd, "r+")
```
robrix: Note that `fdopen` can also error.

Source/carthage/Build.swift
```diff
@@ -31,8 +31,8 @@ public struct BuildCommand: CommandType {
 
 				let (stdoutSignal, schemeSignals) = self.buildProjectInDirectoryURL(directoryURL, options: options)
 				stdoutSignal.observe { data in
-					stdoutHandle.writeData(data)
-					stdoutHandle.synchronizeFile()
+					stdoutHandle.write(data)
+					stdoutHandle.flush()
```
robrix: I’m not sure we should be ignoring errors, but we were doing so before, too, so I guess it’s orthogonal to this PR.

Source/carthage/Extensions.swift
```diff
@@ -47,6 +47,28 @@ internal func print<T>(object: T) {
 	}
 }
 
+internal class FileHandle {
```
robrix: Can this be `final`?

Source/carthage/Extensions.swift
jdhealy: As this occurs in the `deinit` how would we best handle the error?

Attempt `fclose` a few times and `fatalError` if unsuccessful?

Source/carthage/Extensions.swift
jdhealy: Should this be a failable initializer then?

What should the client do if a FileHandle is repeatedly unable to be initialized?

Source/carthage/Extensions.swift
robrix: I think we should simply avoid RAII altogether, preferring explicit open/close API.

robrix: Rationale: we have no way to guarantee minimal lifetime, and therefore resource cleanup is (potentially) nondeterministic (thanks to the intrusion of autorelease pools, etc.). That makes close-on-`deinit` pretty treacherous.

Source/carthage/Extensions.swift
robrix: > Should this be a failable initializer then?

I think we should do an explicit `open` API. I’d like to hear @jspahrsummers’ thoughts on this, tho; this may be considered sufficient for Carthage’s needs, and he’s in a better place to estimate that than I am.

> What should the client do if a FileHandle is repeatedly unable to be initialized?

That’s pretty much up to the client.

jspahrsummers: I think this should be a class function that returns a `Result`. The error can be from `NSPOSIXErrorDomain` if something goes wrong here.

Source/carthage/Extensions.swift
```diff
@@ -47,6 +47,38 @@ internal func print<T>(object: T) {
...
+	let fileDescriptor: Int32
+	let pointer: UnsafeMutablePointer<FILE>
+
+	init(fileDescriptor fd: Int32) {
+		fileDescriptor = fd
+		pointer = fdopen(fd, "r+")
+	}
+
+	func write(data: NSData) -> Result<Int> {
+		switch Darwin.write(fileDescriptor, data.bytes, UInt(data.length)) {
```
jspahrsummers: This should use `fwrite`.

Source/carthage/Extensions.swift
```diff
@@ -47,6 +47,38 @@ internal func print<T>(object: T) {
 	}
 }
 
+internal final class FileHandle {
+	let fileDescriptor: Int32
+	let pointer: UnsafeMutablePointer<FILE>
+
+	init(fileDescriptor fd: Int32) {
+		fileDescriptor = fd
```
jspahrsummers: We shouldn't need to hold onto this for anything.

Source/carthage/Extensions.swift
```diff
@@ -47,6 +47,38 @@ internal func print<T>(object: T) {
 	}
 }
 
+internal final class FileHandle {
+	let fileDescriptor: Int32
+	let pointer: UnsafeMutablePointer<FILE>
```
jspahrsummers: This should probably be `private`.

jspahrsummers:
Did you see @robrix' comment about [avoiding RAII](https://github.com/Carthage/Carthage/pull/355#discussion_r25706578)?

Source/carthage/Extensions.swift
```diff
@@ -47,6 +47,38 @@ internal func print<T>(object: T) {
...
+
+	init(fileDescriptor fd: Int32) {
+		fileDescriptor = fd
+		pointer = fdopen(fd, "r+")
+	}
+
+	func write(data: NSData) -> Result<Int> {
+		switch Darwin.write(fileDescriptor, data.bytes, UInt(data.length)) {
+		case (let value) where value != 0:
```
jspahrsummers: This seems unnecessarily confusing. Why can't we assign to a temporary and use `if`? I think that would express our intention much more clearly.

Source/carthage/Extensions.swift
jdhealy: I started off using `fwrite`… Frustratingly, I can't remember why I had to switch though :weary:

Source/carthage/Extensions.swift
jdhealy: Eh, no technical reason why it's not an `if`. Potato, _potato_.

I can change it.

Source/carthage/Extensions.swift
jdhealy: Flashbacks of `EXC_BAD_ACCESS` :boom: :confused:

jspahrsummers: That would seem to indicate a memory management error. :wink:

jdhealy: Yeah, I had tried debugging it for a while.

Any particular reason why Darwin's `write` is insufficient?

jspahrsummers: It skips buffering that's implemented in stdio (at the `FILE *` level).

jdhealy:
Given the rathole we're in, does [the initial commit](https://github.com/jdhealy/Carthage/commit/9526584541d33cc3ccdb081fe4d7adecfc24a076) look any more satisfactory now?

Source/carthage/Extensions.swift
robrix: Oof, good catch.

jspahrsummers:
I don't think comprehensive review notes and discussion mean that we should abandon reviewing. :wink:

It's just part of the process. We want the simplest, clearest, and most expressive implementation possible. Sometimes that requires iterating.

Like I mentioned to @bavarious, we [do this to each other](https://github.com/Carthage/Carthage/pulls?q=is%3Apr+is%3Aclosed+sort%3Acomments-desc) all the time too, to great effect. :smile:

jdhealy:
> abandon reviewing
Whoa, didn't mean to imply this at all — sorry, if that came across as snide…

> It's just part of the process. We want the simplest, clearest, and most expressive implementation possible. Sometimes that requires iterating.

We just seem to be diving into a complex rewrite of NSFileHandle — reengineering it to avoid RAII, not close on `deinit`. 

Just suggesting that maybe the introspection on NSFileHandle that we're trying to avoid might be simplest and clearest, at least for fixing this small bug.

jspahrsummers:
I've opened an alternative implementation in #376. Can we move any discussion there? :pray:

@jdhealy Thanks for getting this started! :sparkles:

PR #360, opened 2015-02-24
Issue #234: suggest GitHub issue when no shared schemes in GitHub repository

### LINKED ISSUE #234: Cathage should suggest filing issues in the offending repo if a project is not configured correctly
e.g.:

```
** Could not find a shared Framework schema in foo/baz

If you believe this to be an error, please file an issue with the maintainers at https://github.com/foo/baz/issues/new
```

bavarious:
Fixes #234.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -39,8 +39,8 @@ public struct GitHubRepository: Equatable {
 	}
 
 	/// The URL for filing a new GitHub issue for this repository.
-	public var newIssueURL: NSURL? {
-		return NSURL(string: "https://github.com/\(owner)/\(name)/issues/new")
+	public var newIssueURL: NSURL {
+		return NSURL(string: "https://github.com/\(owner)/\(name)/issues/new")!
 	}
 
 	public init(owner: String, name: String) {
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -38,6 +38,11 @@ public struct GitHubRepository: Equatable {
 		return GitURL("ssh://git@github.com/\(owner)/\(name).git")
 	}
 
+	/// The URL for filing a new GitHub issue for this repository.
+	public var newIssueURL: NSURL? {
+		return NSURL(string: "https://github.com/\(owner)/\(name)/issues/new")
+	}
+
 	public init(owner: String, name: String) {
 		self.owner = owner
 		self.name = name
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -879,23 +879,28 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
 		}
 
 		return schemeSignals
-			// For no shared scheme errors in GitHub repositories, suggest
-			// filing a GitHub issue.
 			.catch { error in
 				switch (dependency, error.code) {
 				case (.GitHub(let repo), CarthageErrorCode.NoSharedSchemes.rawValue):
-					let msg = error.localizedDescription + "\n\nIf you believe this to be an error, please file an issue with the maintainers at \(repo.newIssueURL.absoluteString!)"
-					let fileAnIssueError = CarthageErrorCode.NoSharedSchemes.error([
-						NSLocalizedDescriptionKey: msg
-						])
-					return .error(fileAnIssueError)
+					return .error(addIssueFilingSuggestionToError(error, repo))
 				default:
 					return .error(error)
 				}
-		}
+			}
 	}
 
 	return (buildOutput, copyProducts)
+}
+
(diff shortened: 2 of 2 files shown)


Source/CarthageKit/GitHub.swift
```diff
@@ -38,6 +38,11 @@ public struct GitHubRepository: Equatable {
 		return GitURL("ssh://git@github.com/\(owner)/\(name).git")
 	}
 
+	/// The URL for filing a new GitHub issue for this repository.
+	public var newIssueURL: NSURL? {
```
jspahrsummers: I think we could make this non-optional, since the URL should be valid if this really refers to a GitHub repository (if it doesn't, that's a problem for initialization time).

Source/CarthageKit/Xcode.swift
```diff
@@ -878,6 +878,23 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
 		}
 
 		return schemeSignals
+			// For no shared scheme errors in GitHub repositories, suggest
+			// filing a GitHub issue.
```
jspahrsummers: I think this is reasonably self-documenting to where we might not need this comment, but if we wanted to make it even clearer, maybe the innards of the `case` statement could be pulled out into a function—something like `addIssueFilingSuggestionToError`?

Source/CarthageKit/Xcode.swift
```diff
@@ -878,6 +878,23 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+			.catch { error in
+				switch (dependency, error.code) {
+				case (.GitHub(let repo), CarthageErrorCode.NoSharedSchemes.rawValue):
+					var msg = error.localizedDescription + "\n\nIf you believe this to be an error, please file an issue with the maintainers"
+					if let newIssueURL = repo.newIssueURL?.absoluteString {
+						msg += " at \(newIssueURL)"
+					}
+					let fileAnIssueError = CarthageErrorCode.NoSharedSchemes.error([
+						NSLocalizedDescriptionKey: msg
+						])
```
jspahrsummers: Could you please outdent this to match the declaration?

jspahrsummers:
Just a few comments, but this looks great! Thank you for the pull request. :sparkles:

Source/CarthageKit/Xcode.swift
bavarious: Any preference on where this new function should be placed?

jspahrsummers:
Is this ready for another review? (We usually drop a note or an emoji to indicate that, since commits don't generate notifications on their own.)

bavarious:
Sorry! It is, though, as mentioned elsewhere, not quite sure about that indentation comment.

Source/CarthageKit/Xcode.swift
```diff
@@ -222,6 +222,7 @@ public func schemesInProject(project: ProjectLocator) -> ColdSignal<String> {
 			//
 			// '    This project contains no schemes.'
 			// 'There are no schemes in workspace "Carthage".'
+			return .error(CarthageError.NoSharedSchemes(project).error)
```
jspahrsummers: Is this from debugging? :smile:

Source/CarthageKit/Xcode.swift
```diff
@@ -878,11 +879,30 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
 
 	return (buildOutput, copyProducts)
 }
 
+/// Augments a no shared schemes error description to include a suggestion
+/// to file a GitHub issue in GitHub repositories.
+///
+/// Returns a no shared schemes error with an issue filing suggestion.
+private func addIssueFilingSuggestionToError(error: NSError, repo: GitHubRepository) -> NSError {
+	let msg = error.localizedDescription + "\n\nIf you believe this to be an error, please file an issue with the maintainers at \(repo.newIssueURL.absoluteString!)"
```
jspahrsummers: What does the URL look like printed? (I would expect it'd be the same as `absoluteString`, but maybe not.)

jspahrsummers:
> not quite sure about that indentation comment.

Don't worry about it. I found a couple of other minor things, though.

After that, this seems ready to go! :zap:

Source/CarthageKit/Xcode.swift
bavarious: Ooops! Let’s all just pretend this has never happened. 😳

Source/CarthageKit/Xcode.swift
bavarious: It looks the same, yeah. In theory, we have no guarantee that `.description` is something that can be copied and pasted, but that likely doesn’t matter in practice.

jspahrsummers: Okay, no worries either way.

bavarious:
Hopefully your last review of this PR.

PR #364, opened 2015-02-27
Detect duplicate Cartfile.private dependencies

### LINKED ISSUE #333: Error if a dependency is in both Cartfile and Cartfile.private
This would have prevented #332, Carthage/Commandant#20, and others.

bavarious:
This PR should fix #333.

At the moment, it only lists duplicate dependencies. I would like to discuss the items below; feedback is greatly appreciated!

1. ~~Should Carthage always error out when it finds duplicate dependencies or can it be a warning only? This would mean another command-line option.~~ Always error out.
2. ~~How do I error out in this case? That closure returns a `Cartfile`.~~
3. ~~Should I make that match O(n log n)? Note: 8 seconds with 1,000 non-matching entries in either Cartfile.~~ Use Swift 1.2 `Set`.
4. ~~That `println()` should be directed to `stderr`, and the fix probably depends on 1 and 2 above.~~


jspahrsummers:
Thanks @bavarious! I'll respond to the open items, but wait on the code review until you're ready:

> Should Carthage always error out when it finds duplicate dependencies or can it be a warning only? This would mean another command-line option.

I think erroring out is fine. There's no reason why duplicate dependencies should exist, and it's easy enough to correct.

Furthermore, what would happen if they specified different version constraints? I don't even know the answer, and it certainly shouldn't be relied upon, so it's better to let the user resolve it.

> Should I make that match O(n log n)? Note: 8 seconds with 1,000 non-matching entries in either Cartfile.

I'm _generally_ unconcerned with algorithmic complexity in Carthage. Constant factors seem to count for much more, given the short length of most `Cartfile`s and even `Cartfile.resolved`.

So, if it's easy enough to reduce the running time, that's awesome—but no worries if not.

bavarious:
> So, if it's easy enough to reduce the running time, that's awesome—but no worries if not.
I’d rather wait for Swift 1.2 support and use native sets. For 100 non-matching entries in either file, the current implementation is instantaneous.

bavarious:
> So, if it's easy enough to reduce the running time, that's awesome—but no worries if not.

I’d rather wait for Swift 1.2 support and use native sets. For 100 non-matching entries in either file, the current implementation is instantaneous.

PR is ready for review. 💨

Source/CarthageKit/Errors.swift
```diff
@@ -25,6 +25,7 @@ public enum CarthageErrorCode: Int {
 	case InvalidArchitectures
 	case MissingEnvironmentVariable
 	case NoSharedSchemes
+	case DuplicateDependenciesInPrivateCartfile
```
jspahrsummers: How about just `DuplicateDependencies`? We could use this error to refer to any incidence of duplicate listings.

Source/CarthageKit/Errors.swift
```diff
@@ -71,6 +72,10 @@ public enum CarthageError {
 	/// them.
 	case NoSharedSchemes(ProjectLocator)
 
+	/// The private cartfile contains dependencies that are already present in
+	/// the main cartfile.
+	case DuplicateDependenciesInPrivateCartfile([Dependency<VersionSpecifier>])
```
jspahrsummers: Since it's not clear which side the version specifier represents, maybe we should just use a list of `ProjectIdentifier`s here.

Source/CarthageKit/Project.swift
```diff
@@ -187,11 +187,19 @@ public final class Project {
 		}
 
 		return cartfile.zipWith(privateCartfile)
-			.map { (var cartfile, privateCartfile) -> Cartfile in
-				cartfile.appendCartfile(privateCartfile)
+			.map { (var cartfile, privateCartfile) -> ColdSignal<Cartfile> in
+				// TODO: use Set when Carthage supports Swift 1.2
+				let duplicateDeps = filter(privateCartfile.dependencies) { privateDep in
+					contains(cartfile.dependencies, {$0.project == privateDep.project})
+				}
```
jspahrsummers: As [hinted above](https://github.com/Carthage/Carthage/pull/364/files#r25580874), could we generalize this check to find any project that occurs more than once (whether the occurrences are split across `Cartfile` and `Cartfile.private`, or only in one)?

Source/CarthageKit/Project.swift
```diff
@@ -187,11 +187,19 @@ public final class Project {
 		}
 
 		return cartfile.zipWith(privateCartfile)
-			.map { (var cartfile, privateCartfile) -> Cartfile in
-				cartfile.appendCartfile(privateCartfile)
+			.map { (var cartfile, privateCartfile) -> ColdSignal<Cartfile> in
```
jspahrsummers: This can be `tryMap` instead.

Source/CarthageKit/Project.swift
```diff
@@ -187,11 +187,19 @@ public final class Project {
...
+				}
+
+				if duplicateDeps.count == 0 {
+					return .single(cartfile)
+				}
 
-				return cartfile
+				return .error(CarthageError.DuplicateDependenciesInPrivateCartfile(duplicateDeps).error)
 			}
+			.first().value()!
```
jspahrsummers: If the above is [changed to `tryMap`](https://github.com/Carthage/Carthage/pull/364/files#r25580922), this can go away.

Even if it remained `map` though, it'd be better to use something like `merge` here. `first` is a blocking call that should be avoided as much as possible.

Source/CarthageKit/Errors.swift
```diff
@@ -133,6 +138,12 @@ public enum CarthageError {
 			return CarthageErrorCode.NoSharedSchemes.error([
 				NSLocalizedDescriptionKey: "Project \"\(project)\" has no shared schemes"
 			])
+
+		case let .DuplicateDependenciesInPrivateCartfile(duplicateDeps):
+			let deps = duplicateDeps.reduce("", combine: {(acc, dep) in "\(acc), \(dep.project)"})
```
jspahrsummers: The combining function can be specified as a trailing literal:

```swift
duplicateDeps.reduce("") { (acc, dep) in … }
```

jspahrsummers: Could we list one project per line, tab-indented?

jspahrsummers:
If you have the time to do so, could you please add a test that verifies this behavior? :pray:

bavarious:
More issues than code! 😅

Source/CarthageKit/Project.swift
bavarious: I’ll write two functions: one that lists dupes in an array and another that lists dupes in two different arrays. What’s the best place to put them in the code base: as generic code in a new file with extensions to the standard library, or as methods in struct Cartfile?

I’ve thought of using Foundation for this, but that would mean converting struct ProjectIdentifier to an objc class.

Source/CarthageKit/Project.swift
bavarious: Worked like a charm.

Source/CarthageKit/Project.swift
bavarious: Tentative output:
```
The following dependencies are duplicates:
	github "selfdup1/selfdup1"	Cartfile
	github "selfdup2/selfdup2"	Cartfile
	github "selfdup2/selfdup2"	Cartfile
	github "dup1/dup1"	Cartfile and Cartfile.private
	github "dup2/dup2"	Cartfile and Cartfile.private
	github "dup3/dup3"	Cartfile and Cartfile.private
	github "dup4/dup4"	Cartfile and Cartfile.private
```

jspahrsummers: Those functions could go into [`FrameworkExtensions.swift`](https://github.com/Carthage/Carthage/blob/1ee40ca0dc982abbf88b79f35cefd5e67fa62198/Source/CarthageKit/FrameworkExtensions.swift).

> Tentative output:

Could we make it clearer that the `Cartfile` part of the line is the source? Maybe something like:

```
github "dup1/dup1" (found in Cartfile and Cartfile.private)
```

jspahrsummers:
> More issues than code! 😅

Sorry about that! It's not uncommon on our OSS projects, no matter who has opened the PR. :wink:

See https://github.com/Carthage/Carthage/pull/208 for some picking on poor @robb.

bavarious:
Could you review 4a788b9 before I start writing test cases? I’ve moved detection of duplicate dependencies to `Cartfile`, which will allow me to separately write test cases for dupes in a single file and across two files. Let me know if that makes sense.

Source/CarthageKit/Cartfile.swift
```diff
@@ -92,6 +92,35 @@ extension Cartfile: Printable {
...
+extension Cartfile {
+	/// Returns an array containing projects that are listed as duplicate
+	/// dependencies.
+	public func duplicateProjects() -> [ProjectIdentifier] {
+		return repeatedElements(self.sortedProjects())
+	}
+
+	/// Returns an array containing projects that are listed as dependencies
+	/// in both the receiver and the argument.
+	public func duplicateProjects(cartfile: Cartfile) -> [ProjectIdentifier] {
```
jspahrsummers: This feels more like a static method or free function to me (since the receiver and the argument conceptually have equal weight)—perhaps `duplicateProjectsInCartfiles` or something.

Source/CarthageKit/Cartfile.swift
```diff
@@ -92,6 +92,35 @@ extension Cartfile: Printable {
...
+			otherProjects.binarySearch($0, isOrderedBefore: { $0.description < $1.description })
+		}
+	}
+
+	/// Returns an array containing the underlying projects of all dependencies
+	/// in the receiver, sorted by project description.
+	private func sortedProjects() -> [ProjectIdentifier] {
+		return self.dependencies
+			.map { $0.project }
+			.sorted { $0.description < $1.description }
```
jspahrsummers: `ProjectIdentifier` is hashable, so we could use a “fake” set with `[ProjectIdentifier: ()]` and get O(1) lookups.

I'm not really concerned about the time/algorithmic complexity, but the sorting and binary searching stuff seems more likely to contain a bug.

Source/CarthageKit/Errors.swift
```diff
@@ -71,6 +72,10 @@ public enum CarthageError {
 	/// them.
 	case NoSharedSchemes(ProjectLocator)
 
+	/// The private cartfile contains dependencies that are already present in
+	/// the main cartfile.
+	case DuplicateDependencies([(ProjectIdentifier, /*label*/ String)])
```
jspahrsummers: Can we encapsulate this tuple's information in a custom `struct` type instead, to make it clearer?

jspahrsummers: We could also implement `Printable` that way, making the string formatting below easier.

jspahrsummers:
Sorry for the churn. Hope those comments are helpful!

Source/CarthageKit/Cartfile.swift
bavarious: Indeed. Now, `[ProjectIdentifier: ()]` or `[ProjectIdentifier: Int]` (set vs counted set)? In the unlikely case that a dependency is listed more than twice, the latter would make it trivial for us not to list that dupe more than once.

bavarious:
Review away! 🛃

I’ll work on test cases.

bavarious:
Feel free to review those test cases as well! 😜

Source/CarthageKit/Errors.swift
```diff
@@ -133,6 +138,76 @@ public enum CarthageError {
...
+				NSLocalizedDescriptionKey: "The following dependencies are duplicates:\(deps)"
+			])
+		}
+	}
+}
+
+/// A duplicate dependency, used in CarthageError.DuplicateDependencies.
+public struct DuplicateDependency {
+	/// The duplicate dependency as a project.
+	let project: ProjectIdentifier
```
jspahrsummers: Can you please make these `public` as well?

Source/CarthageKit/Errors.swift
```diff
@@ -133,6 +138,76 @@ public enum CarthageError {
...
+		self.locations = locations.sorted(<)
+	}
+}
+
+extension DuplicateDependency: Printable {
+	public var description: String {
+		return "\(project) \(printableLocations())"
+	}
+
+	private func printableLocations() -> String {
```
jspahrsummers: I'd probably make this a computed property (`var`), but it doesn't matter much. If it's convenient to fix along the way, sure—if not, don't worry about it.

Source/CarthageKit/Errors.swift
```diff
@@ -133,6 +138,76 @@ public enum CarthageError {
...
+		if locations.count == 0 {
+			return ""
 		}
+
+		return " (found in \(locations[0])"
+			+ locations[1..<locations.count].reduce("") { (acc, location) in
```
jspahrsummers: You can use `join` for this. :smile:

Source/CarthageKit/Errors.swift
```diff
@@ -133,6 +138,76 @@ public enum CarthageError {
...
+	}
+
+	if lhs.locations.count < rhs.locations.count {
+		return true
 	}
+	else if lhs.locations.count > rhs.locations.count {
+		return false
+	}
+
+	for (index, lhsLocation) in enumerate(lhs.locations) {
```
jspahrsummers: You can use this instead:

```swift
for (lhsLocation, rhsLocation) in Zip2(lhs.locations, rhs.locations) { … }
```

Source/CarthageKit/Cartfile.swift
```diff
@@ -92,6 +92,27 @@ extension Cartfile: Printable {
...
+		let projectSet = buildCountedSet(self.dependencies.map { $0.project })
+		return filter(projectSet) { $0.1 > 1}
+			.map { $0.0 }
+	}
+}
+
+/// Returns an array containing projects that are listed as dependencies
+/// in both arguments.
+public func duplicateProjectsInCartfiles(cartfile1: Cartfile, cartfile2: Cartfile) -> [ProjectIdentifier] {
+	let projectSet1 = buildCountedSet(cartfile1.dependencies.map { $0.project }).keys
```
jspahrsummers: By immediately converting from a dictionary to an ordered collection, we lose O(1) lookups. Can this wait until after the `filter`?

Source/CarthageKit/Cartfile.swift
```diff
@@ -92,6 +92,27 @@ extension Cartfile: Printable {
 	}
 }
 
+// Duplicate dependencies
+extension Cartfile {
+	/// Returns an array containing projects that are listed as duplicate
+	/// dependencies.
+	public func duplicateProjects() -> [ProjectIdentifier] {
+		let projectSet = buildCountedSet(self.dependencies.map { $0.project })
```
jspahrsummers: This bit is duplicated below. Could it be factored out into a `private` function and reused by both?

jspahrsummers:
:soon:

Source/CarthageKit/Errors.swift
bavarious: I thought I’d tried it before and got an error like ‘extensions cannot have computed properties’, but I’ve just tried it again and it worked. *shrug*

Source/CarthageKit/Errors.swift
bavarious: TIL₁.

Source/CarthageKit/Errors.swift
bavarious: TIL₂

Source/CarthageKit/Cartfile.swift
bavarious: That Homer moment…

bavarious:
Yet another round. Fight! 👊

Issue #219, opened 2014-12-18
`carthage checkout` hangs when accessing git repo that requires authentication if credentials aren't already set


gfontenot:
If you specify a private repo in your `Cartfile` and your credentials aren't set up for https access, `carthage checkout` will get to the point where it needs authentication and hang. The user will be prompted for their username, but after entering it they are never prompted for their password.

The workaround for this is to make sure your credentials will be saved (with `osxkeychain`, for example) and then force authentication by cloning a private repo through https elsewhere on your system. After that, `carthage checkout` works as expected.

I'm assuming this has to do with some asynchronous behavior around the checking out of repos, but I haven't looked into it yet.

angerman:
Just got bitten by this as well. Why doesn't the `launchGitTask` time out?

jspahrsummers:
@angerman Because it hasn't been written. :blush: 

If you'd like to submit a pull request, we'd happily review it!

ctava:
If you've been using ssh with 2FA to clone repos and are using carthage to clone private repos (which is using https) here's the workaround for you:

setup an access token for command line use:
https://help.github.com/articles/creating-an-access-token-for-command-line-use/

when prompted enter your github.com username and for your github.com password, enter the access token that you have just generated. caching of username/password is handled.

jarsen:
Ran into this issue this morning. One thing the original poster alluded to but I had never done before and so caused me some more confusion was be sure to enable git to cache your credentials. Instructions are here https://help.github.com/articles/caching-your-github-password-in-git/

pizthewiz:
Just ran into this after upgrading from 0.7.0 to 0.7.1 for a Cartfile of only public repos…

mdiep:
0.7.2 was released, which should fix this issue for anyone using Xcode 6.3+.

hansott:
+1

``$ carthage checkout``
``Username for 'https://github.com': hansott``

``$ carthage version``
``0.7.4``

``$ git version``
``git version 2.4.3``

mdiep:
@hansott What's the output of `git config --get credential.helper`?

hansott:
@mdiep  `osxkeychain`

mdiep:
@hansott **Don't post the output**, but what happens if you run `GIT_TERMINAL_PROMPT=0 echo "url=https://github.com"|git credential fill`? Does it prompt you for credentials?

hansott:
@mdiep Yes, it does. :-)

[...]

PR #369, opened 2015-03-03
[WIP] Test the CLI

jdhealy:
# CarthageTests

_Hopefully, tests and catches issues like #350_

- [x] Remove last vestiges of original CarthageTests (which tested arguments parsing)
- [x] Add CarthageTests target.
- [x] Link and copy frameworks.
- [x] Resolve path of Carthage CLI executable.
	- [x] Check user-set `CARTHAGE_PATH` environment variable.
	- [x] Fallback to searching adjacent to the `xctest` bundle.
- [x] `CLI` enum for paths and launchers.
	- [x] Launcher modifiers
		- [x] Ensure TTY
		- [x] Ensure No TTY
		- [x] Run in ZSH Shell
			- [x] Allow for testing piping into other processes.
				- [ ] Write tests for piping into other processes.
- [x] Zipped fixtures
	- [x] `Fixture	` enum
	- [x] Temporary Directory [using Archive.swift](https://github.com/Carthage/Carthage/blob/1ee40ca0dc982abbf88b79f35cefd5e67fa62198/Source/CarthageKit/Archive.swift)
		- [x] Move `temporaryDirectory` of fixture to Trash on completion of signal.
- [x] `Error` enum.
	- [x] NSError function for cases.
	- [x] LlamaKit.failure function for cases.
	- [x] NSError static function with code parameter.
	- [x] LlamaKit.failure static function for generic errors.
- [ ] Remove CLISpec :question:
- [ ] Change Pragma Mark formatting :question:
- [x] :lemon::lemon::lemon: Lemons :lemon::lemon::lemon:

## Tests Carthage commands:
- [x] version
- [x] bootstrap
- [ ] build
- [ ] checkout
- [ ] archive
- [ ] copy-frameworks
- [ ] fetch
- [ ] update

## Tests currently work in:
- [x] `Xcode.app`
- [x] `xcodebuild`
- [x] `xctest`
- [ ] `xctool`


Tests/BootstrapSpec.swift
```diff
@@ -0,0 +1,53 @@
...
+						arguments: ["bootstrap", "--platform", "🍋"],
+						workingDirectoryPath: Fixture.DependsOnPrelude
+					).single().error()
+				).notTo(beNil())
+			}
+
+			for ttyEnabled in [true, false] {
+				let suffix = " with" + (ttyEnabled ? "" : "out") + " TTY"
+
+				let with·or·without = "xxx"
```
jdhealy: :sleepy: Didn't mean to commit this — oversight.

Tests/CLISpec.swift
```diff
@@ -0,0 +1,40 @@
...
+						modify: [echo, .ZSH()]
+					)
+					.try( Assertion.output(contains: "🍋") )
+					.single().error()
+				).to(beNil())
+			}
+			
+		}
+	}
+}
```
jdhealy: :sleepy: Missed the closing new line here.

jdhealy:
Not sure if the `[WIP]` normally precludes any comments, but got any comments, initial reviews?

jspahrsummers:
My initial reaction is that this is a _lot_ of work in order to fix a relatively rare class of problem. Are there better, more effective ways that we can be catching issues like that?

I definitely don't think we should be trying to shell out _to Carthage itself_ in our test suite. That strikes me as super bizarre.

jdhealy:
The analogy I see with these tests is automated UI testing. Unit testing tests that a function's input produces the correct output, then UI testing tests that a user's input is correctly hooked up to a function's input.

With the command line we also have many issues with “invisible” input like teletypes, environment variables, piping, output, default arguments — many of which can cause crashes.

These tests launch Carthage as an executable command line interface in the same fashion that iOS UI tests would launch the `IPA` in the Simulator.

Carthage is a command line app, we should test it in relation to the command line.

jspahrsummers:
> The analogy I see with these tests is automated UI testing.

It's funny you should mention that, because I generally advocate _against_ automated UI tests. :grin: 

If the UI (or, in this case, CLI) layer needs tests, that implies that it isn't just a “dumb wrapper” around the important business logic. If that's the case, then it seems like there are opportunities to push more logic into the directly testable components instead.

If our CLI's direct dependencies (Commandant, PrettyColors, etc.) are well-tested, if CarthageKit is well-tested, and if the majority of our logic is in CarthageKit, it seems like any further gains from automated CLI tests would be very marginal—and, therefore, more of a maintenance burden and liability than a help.

@Carthage/carthage Anyone else have thoughts?

robrix:
I’m sympathetic to black box acceptance testing, since there’s never a guarantee that the interface _is_ just a dumb wrapper (ditto those other “if”s in @jspahrsummers’ comment). However, we mitigate that risk quite a bit by code review.

On the balance, I’m similarly doubtful of the effort/reward ratio, both for building this and for maintaining it, especially under significant change to the project—new commands, new options, etc., deprecations, etc.

jdhealy:
> However, we mitigate that risk quite a bit by code review.

I agree, thorough code review catches a lot of the discrepancies between CLI code and framework code.

However, I'd argue that there are certain classes of bug that are particularly difficult to spot when scanning code in review.

- stdio problems — #167, #350 
- hangs — #219 
- bugs that don't touch CarthageKit — #277
- weird issues with our command-line dependancies — #347 

Additionally, this would be an appropriate place to tests for regressions in all of the aforementioned.

>  I’m similarly doubtful of the effort/reward ratio, both for building this and for maintaining it, especially under significant change to the project—new commands, new options, etc., deprecations, etc.

Yes, this definitely adds a bit of overhead, but consider these changes also require updates to `CarthageKitTests`.

jspahrsummers:
> However, I'd argue that there are certain classes of bug that are particularly difficult to spot when scanning code in review.

I think most of the cases you've listed wouldn't have been caught by CLI tests, or else could've been caught by more thorough CarthageKit modularization and testing.

#350 is probably the best candidate for a CLI test, but we can just fix that now that we know it's an issue.

> Yes, this definitely adds a bit of overhead, but consider these changes also require updates to `CarthageKitTests`.

That doesn't sound like a benefit to me. :confused: 

----

Since it sounds like @robrix and I are in agreement, I don't think we want to incorporate this right now. We may change our minds and spend time investigating this in the future, but the cost/benefit analysis doesn't seem to work in our favor today.

Sorry about that! I do sincerely appreciate the work you've put into the proposal, and demonstrating some code instead of discussing it in the abstract. :sparkles:

jdhealy:
Well, thanks for considering it. :sunny:

PR #375, opened 2015-03-06
[WIP] Show all projects that require a dependency

### LINKED ISSUE #244: Show which cloned repositories are dependencies in `carthage update` output.
With following contents of Cartfile:

```
github "ReactiveCocoa/ReactiveCocoa" >= 2.3.1
github "AFNetworking/AFNetworking" >= 2.5
```

The output of running `carthage update` shows flat list of 4 extra repositories (which are dependencies of those defined in Cartfile):

```
*** Cloning ReactiveCocoa
*** Cloning AFNetworking
*** Cloning xcconfigs
*** Cloning Quick
*** Cloning Nimble
*** Cloning LlamaKit
```

This seems to be confusing if a user does not know about dependencies of the repositories in the Cartfile. Adding indentation to clonal dependencies is better way to distinguish between requested frameworks and their dependencies.

bavarious:
This pull request should fix #244.

The initial commit runs BFS on the dependency graph, annotating each node with its root ancestors.

We need to decide:

 1. Whether BFS should be run separately (as done in the first commit) or it should be part of some pre-existing traversal.
 2. How to propagate this information to `cloneOrFetchProject()` surroundings, which outputs cloning information. At that point, full `Dependency` values, which contain root ancestors, have already been mapped to `ProjectIdentifier` values, which do not contain that information.

### DIFF
--- a/Carthage.xcodeproj/project.pbxproj
+++ b/Carthage.xcodeproj/project.pbxproj
@@ -41,6 +41,8 @@
 		D0AAAB5019FB0960007B24B3 /* CarthageKit.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = D0D1216D19E87B05005E4BAA /* CarthageKit.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		D0AAAB5519FB1062007B24B3 /* TestCartfile in Resources */ = {isa = PBXBuildFile; fileRef = D0AAAB5419FB1062007B24B3 /* TestCartfile */; };
 		D0AAC2401A14933100060F2E /* Update.swift in Sources */ = {isa = PBXBuildFile; fileRef = D0AAC23F1A14933100060F2E /* Update.swift */; };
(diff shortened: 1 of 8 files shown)


bavarious:
It seems that we cannot know all requirements that lead to a dependency being cloned/fetched at the moment it is cloned/fetched. Is it okay to list only one (the first) of those requirements? E.g., in

A depends on Z
B depends on C depends on Z

we’d print either `Cloning Z (dependency of A)` or `Cloning Z (dependency of B)`. This would tell the user why a particular dependency is being cloned/fetched, but not the whole story.

Thoughts? @jspahrsummers

jspahrsummers:
> It seems that we cannot know all requirements that lead to a dependency being cloned/fetched at the moment it is cloned/fetched.

Ah, that's true. Sorry for not thinking of that sooner. :flushed: 

> we’d print either `Cloning Z (dependency of A)` or `Cloning Z (dependency of B)`

How about something like `Cloning Z (because of A)`? That doesn't imply that A is the only reason it might be cloned, but it at least tells you why Carthage is finding that project.

mdiep:
Closing this because it's been >1 year and the branch doesn't contain any changes.

PR #368, opened 2015-03-01
Upgrade dependencies

jspahrsummers:
Updates to the latest version of RAC and Swift.

Resolves #421.

**To do:**

 - [x] Fix binaries incorrectly being downloaded
 - [x] Fix segmentation fault
 - [x] ~~Make output signals finite?~~
 - [x] Tag ReactiveTask and pin to the semantic version

PR #376, opened 2015-03-07
Encapsulate log streaming in a FileSink type

### LINKED ISSUE #350: synchronizeFile() is not supported on piped stdout

jspahrsummers:
Resolves #350.

**To do:**

 - [x] Unit tests for `FileSink`

### DIFF
--- a/Source/CarthageKit/FileSink.swift
+++ b/Source/CarthageKit/FileSink.swift
@@ -23,7 +23,7 @@ public final class FileSink: SinkType {
 	}
 
 	/// Creates a sink that will take over the given file descriptor.
-	public class func sinkWithDescriptor(fileDescriptor: Int32, closeWhenDone: Bool) -> ColdSignal<FileSink> {
+	private class func sinkWithDescriptor(fileDescriptor: Int32, closeWhenDone: Bool) -> ColdSignal<FileSink> {
 		return ColdSignal.lazy {
 			let pointer = fdopen(fileDescriptor, "a")
 			
--- a/Source/CarthageKit/FileSink.swift
+++ b/Source/CarthageKit/FileSink.swift
@@ -1,0 +1,100 @@
+//
+//  FileSink.swift
+//  Carthage
+//
+//  Created by Justin Spahr-Summers on 2015-03-06.
+//  Copyright (c) 2015 Carthage. All rights reserved.
+//
+
+import Foundation
+import ReactiveCocoa
+
+/// A sink which streams its events to a file handle.
+public final class FileSink: SinkType {
+	private var filePointer: UnsafeMutablePointer<FILE>?
+	private let closeWhenDone: Bool
+
+	/// Creates a sink that will take over the given FILE * pointer.
+	private init(filePointer: UnsafeMutablePointer<FILE>, closeWhenDone: Bool) {
+		setlinebuf(filePointer)
+
+		self.filePointer = filePointer
+		self.closeWhenDone = closeWhenDone
+	}
+
+	/// Creates a sink that will take over the given file descriptor.
+	public class func sinkWithDescriptor(fileDescriptor: Int32, closeWhenDone: Bool) -> ColdSignal<FileSink> {
+		return ColdSignal.lazy {
+			let pointer = fdopen(fileDescriptor, "a")
+			
+			if pointer == nil {
+				return .error(NSError(domain: NSPOSIXErrorDomain, code: Int(errno), userInfo: nil))
+			} else {
+				return .single(self(filePointer: pointer, closeWhenDone: closeWhenDone))
+			}
+		}
+	}
+
+	/// Creates a sink that will open and write to a temporary file.
+	/// 
+	/// Sends the created sink and the URL to the temporary file.
+	public class func openTemporaryFile() -> ColdSignal<(FileSink, NSURL)> {
+		return ColdSignal.lazy {
+			var temporaryDirectoryTemplate: ContiguousArray<CChar> = NSTemporaryDirectory().stringByAppendingPathComponent("carthage-xcodebuild.XXXXXX.log").nulTerminatedUTF8.map { CChar($0) }
+			let logFD = temporaryDirectoryTemplate.withUnsafeMutableBufferPointer { (inout template: UnsafeMutableBufferPointer<CChar>) -> Int32 in
+				return mkstemps(template.baseAddress, 4)
+			}
+
+			if logFD < 0 {
+				return .error(NSError(domain: NSPOSIXErrorDomain, code: Int(errno), userInfo: nil))
+			}
+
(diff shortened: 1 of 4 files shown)


Source/carthage/Build.swift
```diff
@@ -25,15 +25,14 @@ public struct BuildCommand: CommandType {
...
+			.map { (stdoutSink, temporaryURL) -> ColdSignal<()> in
 				let directoryURL = NSURL.fileURLWithPath(options.directoryPath, isDirectory: true)!
 
 				let (stdoutSignal, schemeSignals) = self.buildProjectInDirectoryURL(directoryURL, options: options)
-				stdoutSignal.observe { data in
-					stdoutHandle.writeData(data)
-					stdoutHandle.synchronizeFile()
-				}
+				stdoutSignal
+					.map { .Next(Box($0)) }
```
jspahrsummers: This is a temporary hack. Once we're using the latest version of RAC, everything will be `Event`s, so we can hook up the standard output signal directly to the sink.

jspahrsummers:
Ready for review.

Source/CarthageKit/FileSink.swift
```diff
@@ -0,0 +1,110 @@
...
+	/// Upon a terminating event, the file will be flushed, and closed if
+	/// appropriate.
+	public func put(event: Event<NSData>) {
+		if let filePointer = filePointer {
+			switch event {
+			case let .Next(data):
+				fwrite(data.unbox.bytes, UInt(data.unbox.length), 1, filePointer)
+
+			case let .Error(error):
+				fputs(error.description + "\n", filePointer)
```
robrix: I’m a little surprised that this is writing the errors out as well; I had expected it would resurface them in another signal so you could e.g. maintain the stdout/stderr divide. Do you think that would be appropriate for the type?

Source/carthage/Build.swift
```diff
@@ -93,33 +92,15 @@ public struct BuildCommand: CommandType {
...
-			}
-
-			let temporaryPath = temporaryDirectoryTemplate.withUnsafeBufferPointer { (ptr: UnsafeBufferPointer<CChar>) -> String in
-				return String.fromCString(ptr.baseAddress)!
-			}
-
-			let stdoutHandle = NSFileHandle(fileDescriptor: logFD, closeOnDealloc: true)
-			let temporaryURL = NSURL.fileURLWithPath(temporaryPath, isDirectory: false)!
-
-			let out: (NSFileHandle, NSURL?) = (stdoutHandle, temporaryURL)
```
robrix: This is pretty nice.

Source/CarthageKit/FileSink.swift
```diff
@@ -0,0 +1,110 @@
...
+	/// Sends the created sink and the URL to the temporary file.
+	public class func openTemporaryFile() -> ColdSignal<(FileSink, NSURL)> {
+		return ColdSignal.lazy {
+			var temporaryDirectoryTemplate: ContiguousArray<CChar> = NSTemporaryDirectory().stringByAppendingPathComponent("carthage-xcodebuild.XXXXXX.log").nulTerminatedUTF8.map { CChar($0) }
+			let logFD = temporaryDirectoryTemplate.withUnsafeMutableBufferPointer { (inout template: UnsafeMutableBufferPointer<CChar>) -> Int32 in
+				return mkstemps(template.baseAddress, 4)
+			}
+
+			if logFD < 0 {
+				return .error(NSError(domain: NSPOSIXErrorDomain, code: Int(errno), userInfo: nil))
```
robrix: Would it be useful to capture the path we attempted to use in `userInfo`?

Source/CarthageKit/FileSink.swift
```diff
@@ -0,0 +1,110 @@
...
+			}
+
+			let temporaryPath = temporaryDirectoryTemplate.withUnsafeBufferPointer { (ptr: UnsafeBufferPointer<CChar>) -> String in
+				return String.fromCString(ptr.baseAddress)!
```
robrix: Can this be reduced to `temporaryDirectoryTemplate.withUnsafeBufferPointer(String.fromCString)!`?

Source/CarthageKitTests/FileSinkSpec.swift
```diff
@@ -0,0 +1,52 @@
...
+			sink?.put(.Next(Box(foobarData)))
+			expect(NSString(contentsOfURL: URL, encoding: NSUTF8StringEncoding, error: nil)).to(equal("foobar\n"))
+
+			// Verify line buffering.
+			sink?.put(.Next(Box(fuzzbuzzData)))
+			expect(NSString(contentsOfURL: URL, encoding: NSUTF8StringEncoding, error: nil)).to(equal("foobar\n"))
+
+			// Verify output flushing.
+			sink?.put(.Completed)
+			expect(NSString(contentsOfURL: URL, encoding: NSUTF8StringEncoding, error: nil)).to(equal("foobar\nfuzzbuzz"))
```
robrix: :+1: These are great tests.

robrix:
:+1:

Source/CarthageKit/FileSink.swift
jspahrsummers: I don't think so, since it's just a template.

Source/CarthageKit/FileSink.swift
jspahrsummers: Nope:

```
error: 'UnsafeBufferPointer<CChar>' is not a subtype of 'UnsafePointer<CChar>'
```

Source/CarthageKitTests/FileSinkSpec.swift
jspahrsummers: Thanks!

jspahrsummers:
@robrix I refactored `FileSink` a bit based on your feedback, primarily to remove the confusion about the error stuff.

In #368, I can restore the handling of `Event`, but explicitly mark it as `NoError`. Then, it can only be hooked up to pure logging streams that cannot error.

Source/CarthageKit/FileSink.swift
robrix: :ok_hand:

Source/CarthageKit/FileSink.swift
robrix: Thanks for checking :bow:

robrix:
@jspahrsummers Okay. Did you want this to depend on #368? If not I’m :+1: to :shipit:

jspahrsummers:
Nope, I'll fix it up within that PR. :ship:

PR #379, opened 2015-03-07
Documented --use-ssh flag in README

### LINKED ISSUE #378: Submodules are set up with HTTP instead of SSH
Is there a reason for this? SSH would be useful to be able to contribute to dependencies inside of `Carthage/Checkouts` more easily :)

NachoSoto:
Fixes #378.


jspahrsummers:
:-1: I don't think the Markdown files should be in the business of documenting every possible setup option.

It's your responsibility to use `carthage help` and figure out what flags you want. :wink:

NachoSoto:
Err my bad, I was only trying `carthage update --help`, now I see I can do `carthage help update` :fearful:

Issue #299, opened 2015-02-07
Nondeterministic “project has no shared schemes” error


truppelito:
I wanted to use Carthage on a new app I'm developing so I created a dummy Xcode project just to try it out. Inside the dummy project folder (where the xcodeproj file is) I added a Cartfile with this in it:

```
github "robb/Cartography"
github "kishikawakatsumi/KeychainAccess"
github "Alamofire/Alamofire"
github "SwiftyJSON/SwiftyJSON"
github "hubertr/Swell"
```
(all these projects say they are Carthage compatible, btw)

Then I ran on Terminal:

    cd /path/to/folder/containing/dummy/xcodeproj
    carthage update

Sure enough a `Carthage.resolved` file is created which contains this:

```
github "Alamofire/Alamofire" "1.1.4"
github "robb/Cartography" "0.2.0"
github "kishikawakatsumi/KeychainAccess" "v1.1.1"
github "hubertr/Swell" "1.0.1"
github "SwiftyJSON/SwiftyJSON" "2.1.3"
```

And a Carthage folder is created. Inside are the Build and Checkouts folders. The Checkouts folder contains 5 subfolders (one for each project), as expected. My problem is that the Build folder is totally and completely empty!

Here is the output produced by `carthage update`:
```
*** Fetching Cartography
*** Fetching KeychainAccess
*** Fetching Alamofire
*** Fetching SwiftyJSON
*** Fetching Swell
*** Checking out Alamofire at "1.1.4"
*** Checking out Cartography at "0.2.0"
*** Checking out KeychainAccess at "v1.1.1"
*** Checking out Swell at "1.0.1"
*** Checking out SwiftyJSON at "2.1.3"
*** xcodebuild output can be found in /var/folders/7p/r8sw1dw15ks2_pcb0tbm5lkw0000gn/T/carthage-xcodebuild.bBK0SN.log
Project "Alamofire.xcodeproj" has no shared schemes
```

The xcodebuild output file is completely empty, and the error (`Project "Alamofire.xcodeproj" has no shared schemes`) is not correct. If I go to the Alamofire folder inside the Checkouts folder and open that xcodeproj file and check the schemes it contains I can see that it has two shared schemes (one for iOS and one for Mac).

Am I doing something wrong? Thanks.

mdiep:
Can you check which version of carthage you're running with `carthage version`?

truppelito:
Yes, of course, `carthage version` outputs: 0.5.2.

[...]

PR #423, opened 2015-04-12
When enumerating schemes, increase timeout for `xcodebuild -list` from 4s to 8s

bavarious:
Also adds CarthageError.XcodebuildListTimeout. See issue #299.

### DIFF
--- a/Source/CarthageKit/Errors.swift
+++ b/Source/CarthageKit/Errors.swift
@@ -145,7 +145,7 @@ public enum CarthageError {
 
 		case let .XcodebuildListTimeout(project):
 			return CarthageErrorCode.XcodebuildListTimeout.error([
-				NSLocalizedDescriptionKey: "Timeout whilst listing project \"\(project)\" schemes. This could mean that no shared schemes could be found or xcodebuild is just taking too long to execute"
+				NSLocalizedDescriptionKey: "Failed to discover shared schemes in project \"(project)\"—either the project does not have any shared schemes, or xcodebuild never returned"
 			])
 			
 		case let .DuplicateDependencies(duplicateDeps):
--- a/Source/CarthageKit/Errors.swift
+++ b/Source/CarthageKit/Errors.swift
@@ -25,6 +25,7 @@ public enum CarthageErrorCode: Int {
 	case InvalidArchitectures
 	case MissingEnvironmentVariable
 	case NoSharedSchemes
+	case XcodebuildListTimeout
 	case DuplicateDependencies
 
 	func error(userInfo: [NSObject: AnyObject]?) -> NSError {
@@ -71,6 +72,9 @@ public enum CarthageError {
 	/// The project is not sharing any schemes, so Carthage cannot discover
 	/// them.
 	case NoSharedSchemes(ProjectLocator)
+
+	/// Timeout whilst running `xcodebuild list` to enumerate shared schemes.
+	case XcodebuildListTimeout(ProjectLocator)
 
 	/// A cartfile contains duplicate dependencies, either in itself or across
 	/// other cartfiles.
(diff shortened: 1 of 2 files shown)


bavarious:
Please review.

Source/CarthageKit/Errors.swift
```diff
@@ -139,6 +143,11 @@ public enum CarthageError {
 				NSLocalizedDescriptionKey: "Project \"\(project)\" has no shared schemes"
 			])
 
+		case let .XcodebuildListTimeout(project):
+			return CarthageErrorCode.XcodebuildListTimeout.error([
+				NSLocalizedDescriptionKey: "Timeout whilst listing project \"\(project)\" schemes. This could mean that no shared schemes could be found or xcodebuild is just taking too long to execute"
```
jspahrsummers: How about something like:

> Failed to discover shared schemes in project \"\(project)\"—either the project does not have any shared schemes, or xcodebuild never returned

Source/CarthageKit/Xcode.swift
```diff
@@ -234,7 +234,9 @@ public func schemesInProject(project: ProjectLocator) -> ColdSignal<String> {
 		// xcodebuild has a bug where xcodebuild -list can sometimes hang
 		// indefinitely on projects that don't share any schemes, so
 		// automatically bail out if it looks like that's happening.
-		.timeoutWithError(CarthageError.NoSharedSchemes(project).error, afterInterval: 4, onScheduler: QueueScheduler())
+		// xcodebuild may also take a while to execute;
+		// see https://github.com/carthage/carthage/issues/299
```
jspahrsummers: I think it's understood that we could create false negatives here. I don't think it needs a comment.

jspahrsummers:
Thanks for the PR! Can you please make sure we call `addIssueFilingSuggestionToError` for this new error case too?

bavarious:
@jspahrsummers Do you think we should reword that file-an-issue error? Currently it’s

>If you believe this to be an error, please file an issue with the maintainers at

And now ‘this’, besides an actual no shared schemes error, can also be ‘Failed to discover shared schemes in project X—either the project does not have any shared schemes, or xcodebuild never returned’

jspahrsummers:
If we can clarify that “an error” means “a project configuration error,” the rest of the message should be fine, I think. What are your thoughts?

bavarious:
I was thinking of something like ‘If you believe project X has no shared schemes’ or something to that effect. ‘Configuration error’ also works.

Source/CarthageKit/Xcode.swift
```diff
@@ -891,15 +896,16 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+/// GitHub issue for repositories with potentially no shared schemes.
 ///
-/// Returns a no shared schemes error with an issue filing suggestion.
+/// Returns an error with an issue filing suggestion.
 private func addIssueFilingSuggestionToError(error: NSError, repo: GitHubRepository) -> NSError {
-	let msg = error.localizedDescription + "\n\nIf you believe this to be an error, please file an issue with the maintainers at \(repo.newIssueURL.absoluteString!)"
-	return CarthageErrorCode.NoSharedSchemes.error([
-		    NSLocalizedDescriptionKey: msg
-		])
+	var userInfo = error.userInfo!
```
bavarious: Here I’m assuming that this is a Carthage error which a localised description. Is it better to fail when this is not the case or to work around in case there is no localised description?

jspahrsummers: It's technically possible for an error to have a `localizedDescription` without having the corresponding `userInfo` key (even though I don't think third party libraries can do that).

But the easiest solution would just be to add `?? NSMutableDictionary()` here.

bavarious:
🙇🏻

jspahrsummers:
:shipit::sparkles:

mdiep:
:clap: Good work, @bavarious!

PR #434, opened 2015-04-24
Fix Authenticated GitHub API Usage

mdiep:
`carthage` wasn't downloading binaries for me this week because my IP was rate-limited for the anonymous GitHub API. The API errors were getting swallowed, so the code to try an authenticated request never executed.

This PR rectifies the situation by:

1. Creating an API error that will pass through errors like rate-limiting and invalid credentials
2. Switches the logic to try an authenticated request before making an unauthenticated one
2. Fixes the logic so these errors result in the request being tried a second time (anonymously)

Here's what it looks like if it can't authenticate and you've been rate-limited:

```
$ carthage checkout
*** Checking out your-project at "4ad2b0a1ad07fbbeaefaaf82474c650c9b8c677b"
GitHub API request failed: API rate limit exceeded for 173.194.43.97. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```


mdiep:
AFAICT, the build failure is just Travis being stupid.

Source/CarthageKit/GitHub.swift
```diff
@@ -211,6 +211,9 @@ internal struct GitHubCredentials {
 
 				return nil
 			}
+			.catch { error in
+				return .single(nil)
```
jspahrsummers: Is this change part of the fix?

Source/CarthageKit/GitHub.swift
```diff
@@ -265,6 +268,28 @@ private func fetchAllPages(URL: NSURL, credentials: GitHubCredentials?) -> ColdS
 			let thisData = ColdSignal.single(data)
 
 			if let HTTPResponse = response as? NSHTTPURLResponse {
+				if HTTPResponse.statusCode != 200 && HTTPResponse.statusCode != 404 {
```
jspahrsummers: Maybe this should just check for codes in the 4xx and 5xx blocks? 2xx and 3xx aren't really failures AFAIK.

Source/CarthageKit/GitHub.swift
```diff
@@ -265,6 +268,28 @@ private func fetchAllPages(URL: NSURL, credentials: GitHubCredentials?) -> ColdS
...
+					return thisData
+						.tryMap { data, error -> AnyObject? in
+							return NSJSONSerialization.JSONObjectWithData(data, options: nil, error: error)
+						}
+						.map { dictionary -> String in
+							let message = JSONValue.parse(dictionary)["message"]
```
jspahrsummers: Can we create a real model structure for this? It would clean up this parsing code, clarify intention, and might come in handy if we want to use fields other than `message` in the future.

Source/CarthageKit/GitHub.swift
```diff
@@ -265,6 +268,28 @@ private func fetchAllPages(URL: NSURL, credentials: GitHubCredentials?) -> ColdS
...
+							switch message {
+							case let .Some(.JSONString(message)):
+								return message
+							default:
+								return NSString(data: data, encoding: NSUTF8StringEncoding) ?? "Unknown error requesting \(URL)"
```
jspahrsummers: We could use `+[NSHTTPURLResponse localizedStringForStatusCode:]` for the unknown case.

jspahrsummers:
:+1: Thanks for fixing this! I imagine this also fixes releases on private repositories, which were probably broken before.

Source/CarthageKit/GitHub.swift
robrix: Also do we need to handle 3xx manually?

jspahrsummers: Nah, we shouldn't.

robrix: :+1:

Source/CarthageKit/GitHub.swift
mdiep: Yes. I wanted to ensure that we tried an unauthenticated request if we failed to load GitHub credentials.

You could definitely argue that this is the callers responsibility. I opted to put it here because it already returns an optional and we don't really care what the error is. IMO, it's hard to say what the best approach is because this method is only used in one place.

mdiep:
:dango:

Source/CarthageKit/GitHub.swift
```diff
@@ -265,6 +300,27 @@ private func fetchAllPages(URL: NSURL, credentials: GitHubCredentials?) -> ColdS
 			let thisData = ColdSignal.single(data)
 
 			if let HTTPResponse = response as? NSHTTPURLResponse {
+				let statusCode = HTTPResponse.statusCode
+				if statusCode > 400 && statusCode < 600 && statusCode != 404 {
```
jspahrsummers: Why do we want to ignore 404s here? Seems like they should also cause an error with the HTTP message.

Maybe we just need to handle potentially `nil` data?

jspahrsummers:
Sorry, 1 :one: (one) more note.

Source/CarthageKit/GitHub.swift
mdiep: Fair question.

Right now, we try to get the releases for whatever we're pinned to—even if it's a SHA. But if it's a SHA, then we'll request something like `https://api.github.com/repos/Quick/Nimble/releases/tags/56c3b79d5a32686fe5529d7aa305000b489ebdaa`, which 404s.

This is an error in the HTTP sense, but is not an unrecoverable error, which is really what's being modeled here. (If you were to remove the 404 check, then trying to checkout when pinned to something other than a tag would cause Carthage to error `GitHub API request failed: Not Found`.)

This method should probably send an HTTP-specific error and let the caller decide how to interpret them. But:

1. This method is _only_ used for releases currently
2. The version of RAC we're on doesn't have custom error types for signals, so we'd have to return an `NSError`, which is a huge pain.

I'd be inclined to leave this for further cleanup.

mdiep: I meant for cleanup in a later PR, in case that wasn't clear.

mdiep:
I merged in the Swift/RAC changes from master. This is ready for another review.

jspahrsummers:
Sorry for the delay. I think this looks good. :shipit:

PR #440, opened 2015-05-05
Fix building in symlinked checkouts

jspahrsummers:
This allows folders in Carthage/Checkouts to be symlinks, if users want to overwrite or otherwise manipulate those folders after `carthage checkout` completes but before `carthage build` runs.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -855,7 +855,7 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
 		// Link this dependency's Carthage/Build folder to that of the root
 		// project, so it can see all products built already, and so we can
 		// automatically drop this dependency's product in the right place.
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -854,7 +855,7 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
 		// Link this dependency's Carthage/Build folder to that of the root
 		// project, so it can see all products built already, and so we can
 		// automatically drop this dependency's product in the right place.
-		let dependencyBinariesURL = dependencyURL.URLByAppendingPathComponent(CarthageBinariesFolderPath, isDirectory: true)
+		let dependencyBinariesURL = dependencyURL.URLByResolvingSymlinksInPath!.URLByAppendingPathComponent(CarthageBinariesFolderPath, isDirectory: true)
```
robb: Aren't the symlinks already resolved in `let dependencyURL = rawDependencyURL.URLByResolvingSymlinksInPath!`?

jspahrsummers: Oops, yes.

jspahrsummers:
@robb :sparkles:

PR #453, opened 2015-05-11
Fix #452

### LINKED ISSUE #452: Document usage of local repository in Cartfile
During the development of a Carthage module, that would actually make sense to be able to reference a local repository. 
Instead of remotely checking a repo out, Carthage could simply copy the local repo in the Checkout folder or Symlink the repo in the checkouts folder

flovilmart:
Updates docs. with local repository example


mdiep:
I doesn't seem right that this changed all the submodules.

flovilmart:
That's better!

robrix:
:+1:

robrix:
Thanks very much @flovilmart!

PR #461, opened 2015-05-15
permuteWith notifies its observer outside of the lock.

robrix:
Contributes towards fixing the deadlocks, but does not fix all of the deadlocks.

:tophat: @mdiep

### DIFF
--- a/Source/CarthageKit/FrameworkExtensions.swift
+++ b/Source/CarthageKit/FrameworkExtensions.swift
@@ -55,22 +55,28 @@ internal func permuteWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, 
 			lock.lock()
 
 			signalValues.append(value)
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Hmm, is this valid? Seems like this could jumble permutations if the `otherSignal` happens to send something at the same time (hypothetically, its observer callbacks could forward events faster than those of `signal`).

robrix:
Hm, that’s a point. To be quite honest I’m not sure what or why this function is, which is making it really hard to reason on :cry:

jspahrsummers:
We use this operator to (lazily) create every possible dependency graph, with filled-in versions. The provided signals fire in preferential order, so it's more likely that earlier events result in a valid graph than later ones—which is why the ordering of output events is important.

The structure here is very similar to `combineLatestWith`, it simply combines all values received instead of just the latest.

PR #470, opened 2015-05-16
Tighten version constraints for graph permutations

jspahrsummers:
Just an optimization I noticed while investigating other issues.

Before this change, we were creating graph permutations for the _original_ nodes passed in, even though our `graph.addNode` calls might result in a tightening of those nodes' constraints. The effect was that we could end up creating graphs that we already know to be impossible.

This uses the nodes _after_ their constraints have been tightened, so we only explore paths that may result in something valid.


robrix:
:+1:

Issue #448, opened 2015-05-10
Duplicate dependencies fetched for Stream


robrix:
Using 0.7.1.

1. Make a new dir & `cd` into it.
2. Cartfile:

    ```
    github "antitypical/Stream" ~> 0.0.3
    ```

3. Bootstrap:

    ```
    rob@Resonance ~/D/Blah> carthage bootstrap
    *** No Cartfile.resolved found, updating dependencies
    *** Fetching Stream
    *** Fetching Memo
    *** Fetching Either
    *** Fetching Box
    *** Fetching Prelude
    *** Fetching Box
    ```

**Expected:** There. Is. One. Box.

**Actual:** Two boxes for the price of one!

robrix:
Cartfile.resolved:

```
github "robrix/Box" "1.2.2"
github "robrix/Prelude" "1.5.0"
github "robrix/Memo" "1.1.4"
github "robrix/Either" "1.2.2"
github "antitypical/Stream" "0.0.3"
```

BObereder:
I have the same thing, and #449 does not resolve that problem.

mojotalantikite:
I'm actually still hitting the deadlock from #449 on the 0.7.3 release as well. This is from both a fresh install using the Homebrew bottle and from building from source with a cleaned out cache.

Happy to provide any details that may help.

PR #471, opened 2015-05-17
Fix multiple fetches for projects without tags

jspahrsummers:
Since `self.cachedVersions[project]` was always `nil` when there are no tags, we would always try re-fetching the affected project.

Confusingly, this does _not_ fix https://github.com/Carthage/Carthage/issues/448, since the affected project in that example does indeed have tags.


robrix:
:+1:

PR #483, opened 2015-05-19
Add `DebugSinkOf`

norio-nomura:
Xcode debugger can't step into `SinkType.put()`.
Using `DebugSinkOf` instead of `SinkOf` allow step into `SinkType.put()`

https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2021


norio-nomura:
`DataSink` has gone by def65c9

Issue #445, opened 2015-05-08
Issues with 0.7.1 version


zocario:
I am having multiple issues since I have updated to latest version (0.7.1).
First I have the same issue as #444 but even if I delete Realm from my Cartfile I am still having strange behavior as Fetching or Checking out that never ends, or permissions denied errors.

My Cartfile is very simple:

    github "Alamofire/Alamofire"
    github "Hearst-DD/ObjectMapper"

Here is an example of permission error I had:

![capture d ecran 2015-05-08 a 13 15 27](https://cloud.githubusercontent.com/assets/800617/7535375/a9683260-f584-11e4-929f-05faf6213ce3.png)

After deleting Carthage folder and Cartfile.resolved file, I am stuck on ```Fetching``` step, but before it was the case for the ```Checking out``` step:

![capture d ecran 2015-05-08 a 13 24 11](https://cloud.githubusercontent.com/assets/800617/7535456/8e30a760-f585-11e4-8707-d41d7b8e0876.png)

Is there any known issues about this? 
What can I do to give you more informations to solve this problem?

Since this morning I am not able to retrieve dependencies for my new project.

MichaelMcGuire:
I'm having the same problem and was about to open another issue.  I'd like to try to diagnose this myself, but I'm having a hard time figuring out how to run a debug version of Carthage.  My Cartfile:

````
github "ReactiveCocoa/ReactiveCocoa" "swift-development"
github "thoughtbot/Argo" "master"
github "AliSoftware/OHHTTPStubs"
github "danielgindi/ios-charts" "master"

github "Quick/Quick"
github "Quick/Nimble"
````

and the output is stuck on Result running `carthage update`
````
$ carthage update
*** Fetching Nimble
*** Fetching Quick
*** Fetching ios-charts
*** Fetching OHHTTPStubs
*** Fetching Argo
*** Fetching ReactiveCocoa
*** Fetching Box
*** Fetching Runes
*** Fetching Result
````
If I delete the cache in ` /Users/user/Library/Caches/org.carthage.CarthageKit` and try the command again, it gets stuck on cloning Result:
````
$ carthage update
*** Cloning Nimble
*** Cloning Quick
*** Cloning ios-charts
*** Cloning OHHTTPStubs
*** Cloning Argo
*** Cloning ReactiveCocoa
*** Cloning Box
*** Cloning Runes
*** Cloning Result
````
Started happening after upgrading Carthage yesterday via homebrew.

[...]

PR #492, opened 2015-05-26
Add ContentsOfFileCache to contentsOfFileInRepository

norio-nomura:
This will reduce launching `git show` operations.
But this does not fix high cpu usage and segmentation fault by using Cartfile of https://github.com/Carthage/Carthage/issues/445#issuecomment-100301785
Maybe higher level caching will be appropriate.


norio-nomura:
This currently works as negative cache for fails of retrieving Cartfile using specific tags.
Some bugs of Dependency resolver or other cause launching `git show` operations repeatedly for them.

norio-nomura:
It seems this PR is not needed any more.

jspahrsummers:
Thank you anyways!

Issue #469, opened 2015-05-16
Oddities when parsing versions like v1.0.0-rc.1


jspahrsummers:
Given [Nimble](https://github.com/Quick/Nimble)'s current [tags](https://github.com/Quick/Nimble/tags):

![screen shot 2015-05-16 at 13 52 31](https://cloud.githubusercontent.com/assets/432536/7667919/d1c6ce98-fbd2-11e4-96b2-6d30a699c650.png)

a Cartfile with:

```
github "Quick/Nimble"
```

will result in version 0.4.2 being used, while

```
github "Quick/Nimble" ~> 1.0  # Or ==
```

will result in version 1.0 being used.

Something is wrong in our version parsing, because `v1.0.0-rc.1` should either be rejected by the parser, or else considered as greater than `v0.4.2`.

jspahrsummers:
While investigating an unrelated issue, I've also noticed this happening for RAC's recent [tags](https://github.com/ReactiveCocoa/ReactiveCocoa/tags). As an example, an invalid `DependencyNode`:

> github "ReactiveCocoa/ReactiveCocoa" @ "v3.0-beta.4" (restricted to ~> 2.4.3)

pietbrauer:
How would one tackle this?

I tried adding an `appendix` to `SemanticVersion` so that you can create one by using `SemanticVersion(major: 1, minor: 0, patch: 0, appendix: "alpha")` from `1.0.0-alpha`. But then the `Equatable` has to be updated too and one has to consider all the permutations, e.g. `"beta" < "pre"` (pseudocode).

My current idea is to introduce an enum: `SemanticVersion(major: 1, minor: 0, patch: 0, appendix: .Alpha)` this makes comparing easier. What other challenges do you see? Do you find that an okish approach?

aschuch:
I have a similar issue. Given the following tags in my repo:

* `1.0.1`
* `swift-1.1`

`swift-1.1` always "wins", although it is not the newest version.

jspahrsummers:
@aschuch That behavior is correct. After all, how is Carthage to know that those aren't part of the same release series?

I'd recommend renaming the tags, or else using explicit pins instead of open-ended constraints.

aschuch:
Ok, I didn't know that this was the intended behaviour. Thanks for the heads up, I'll rename the tags.

jdhealy:
[...]

PR #499, opened 2015-05-27
Add a first novice take on supporting -pre

pietbrauer:
This is related to https://github.com/Carthage/Carthage/issues/469 and
more than incomplete but I thought I might get some initial feedback to go along the correct path.

### DIFF
--- a/Source/CarthageKit/Version.swift
+++ b/Source/CarthageKit/Version.swift
@@ -13,6 +13,51 @@ import ReactiveCocoa
 /// An abstract type representing a way to specify versions.
 public protocol VersionType: Scannable, Equatable {}
 
+/// A development stage on a version
+public enum Stage: Int, Equatable {
+	/// No stage specified
+	///
+	/// This is only here to specify the default
+	case None
+
+	/// The alpha stage
+	///
+	/// This is usually represented as "-Alpha"
+	case Alpha
+
+	/// The beta stage
+	///
+	/// This is usually represented as "-Beta"
+	case Beta
+
+	/// The Pre stage
+	///
+	/// This is usually represented as "-pre"
+	case PreRelease
+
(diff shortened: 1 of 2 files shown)


jspahrsummers:

While it would be nice to support this, all that's necessary to fix https://github.com/Carthage/Carthage/issues/469 itself is just to make sure we ignore versions with trailing text (which is how the code is _supposed_ to behave right now, though it apparently broke).

If you do want to continue with the enhancement, we'll need to decide how prerelease software fits into semantic versioning constraints. My gut says that it kinda doesn't, because any prerelease version can have breaking changes, which defeats the purpose of Cartfile operators like `~>` and `>=`.

Is there a particular use case where you would _want_ to obtain a prerelease version, and yet don't want to refer to a tag by name?

pietbrauer:
For me everything fits into Semantic Versioning:

>  Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format. - [semver.org](http://semver.org)

 A pre release version is a not yet fully released version with the new version number (so breaking changes are actually reflected).

The possible order would be `1.0.0 > 1.0.0-pre > 1.0.0-beta > 1.0.0-alpha`.

For me it is necessary for Carthage to support these else Semantic Versioning support is incompatible.

Does this make sense?

jspahrsummers:
Sorry, yes, prerelease versions _are_ part of SemVer. What I meant is that it makes semantic version _constraints_ (`>=`, `~>`) confusing and potentially problematic (because many users will not want to receive prereleases).

Maybe we should only support `==` for version specifiers like this? But then what advantage would it have over specifying `"1.0.0-alpha"` as a tag name?

pietbrauer:
Sounds legit. Thanks for clarifying.

jspahrsummers:
No problem. Sorry it took so long to reply!

Issue #428, opened 2015-04-17
Stricter git tag to semantic version mapping


neilpa:
Currently carthage allows any prefix string in tags when attempting to parse out semantic versions. This works as intended for tags like `v1.2.3` or `version-4.5.6`. However, the RAC repo has a few tags like `swift-on-carthage-3.0.2` which are problematic since they trump `v3.0-beta.1` if you use something like `~> 3.0` in your Cartfile.

I explored a couple fixes in the resolver and version sorting that would de-prioritize non `v[ersion][-]` tag prefixes but still allow others. However, that got complicated quickly and could lead to unexpected results. Alternatively the tag to version mapping could be stricter in the allowed prefixes. That would always lead to consistent results but puts an extra burden on projects that don't follow the prescribed prefix scheme.

Thoughts?

jspahrsummers:
I think the current implementation is fine. In fact, those RAC tags were specifically created to test Carthage anyways, so arguably there's not a real problem here (besides my obnoxious use of those tags :trollface:).

If we're okay being Bad Citizens™, I think we could just delete those tags and be done with it. I've seen users intentionally pick them too, so they're doing active harm. :confused:

neilpa:
I inadvertently picked it up which is why I opened the issue. I agree that being stricter is probably the wrong approach since I've seen other projects use various other prefixes for version tags.

PR #508, opened 2015-06-01
[Version] Only strip leading "v" or "version-" when parsing SemVer

segiddins:
This means you can have arbitrary, non-version tags that _happen_ to end in numbers, and those trailing numbers will no longer be interpreted as a semantic version.

### DIFF
--- a/Source/CarthageKitTests/VersionSpec.swift
+++ b/Source/CarthageKitTests/VersionSpec.swift
@@ -31,8 +31,11 @@ class SemanticVersionSpec: QuickSpec {
 
 		it("should parse semantic versions") {
 			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("1.4")).value).to(equal(SemanticVersion(major: 1, minor: 4, patch: 0)))
-			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("v1")).value).to(beNil())
 			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("v2.8.9")).value).to(equal(SemanticVersion(major: 2, minor: 8, patch: 9)))
+		}
+
+		it("should fail on invalid semantic versions") {
+			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("v1")).value).to(beNil())
 			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("v2.8-alpha")).value).to(beNil())
 			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("null-string-beta-2")).value).to(beNil())
 		}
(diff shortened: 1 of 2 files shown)


neilpa:
This is probably too strict a requirement (c.f. #428). Can this instead require a `digit . digit` prefix? That should fix most of the false positives that match a single number

jspahrsummers:
Thanks for the pull request, @segiddins! :sparkles:

I agree with @neilpa's feedback, and think the requirement for two version components makes a lot of sense.

segiddins:
Should that check be done in `SemanticVersion.fromScanner` or here in `fromPinnedVersion` ?

segiddins:
(Also, reading #428, it seems like that issue was requesting something more like this?)

neilpa:
`fromScanner` should require a minor version component. But that also requires `fromPinnedVersion` to keep scanning if in case of tags like `RAC2-2.5.1`

neilpa:
RE #428 - More-or-less, but I didn't have strong feelings for any specific solution when I filed that.

segiddins:
I'm happy to implement that, but it seems like then you're just moving the restriction from `foo-\d` to `foo-\d.\d`?

neilpa:
Yea, but in practice that should solve most of the false identification.

segiddins:
Making the change to require a minor version in `fromScanner` makes an existing test fail -- should I just remove that test? (it's for `v1`)

neilpa:
Or you could make it expect to fail like with the other test you added.

segiddins:
Amended.

Source/CarthageKitTests/VersionSpec.swift
```diff
@@ -31,9 +31,10 @@ class SemanticVersionSpec: QuickSpec {
 
 		it("should parse semantic versions") {
 			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("1.4")).value).to(equal(SemanticVersion(major: 1, minor: 4, patch: 0)))
-			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("v1")).value).to(equal(SemanticVersion(major: 1, minor: 0, patch: 0)))
+			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("v1")).value).to(beNil())
 			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("v2.8.9")).value).to(equal(SemanticVersion(major: 2, minor: 8, patch: 9)))
 			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("v2.8-alpha")).value).to(beNil())
+			expect(SemanticVersion.fromPinnedVersion(PinnedVersion("null-string-beta-2")).value).to(beNil())
```
neilpa: Can you move these to a new spec for testing failed parsings

segiddins: done!

neilpa:
:+1: 

Note that this will still fail on `good2go-1.2.3`. Although, I'm okay with that for now since this would've failed in the same way before this change.

jdhealy:
Hmm, I almost just tagged something with `swiftlang-700.0.38.1`, which gets interpreted as a version.

Would disallowing tags with more than three dot-joined digit sets be too strict? :confused:

jspahrsummers:
@jdhealy No, that seems appropriate, but let's please discuss it in a new issue or PR.

PR #512, opened 2015-06-01
Refactored `userAgent` closure

alskipp:
Use `flatMap` and nil coalescing operator to remove a `var` declaration and `if clause` checking for nil.


alskipp:
Note to self: never issue a pull request just before bed (I missed the `kCFBundleVersionKey` as an alternative value – now fixed).

mdiep:
:+1: Looks legit, thanks!

PR #516, opened 2015-06-02
Copy dSYM files alongside built frameworks

### LINKED ISSUE #507: Preserve debug symbols when building
This was touched on in the discussion of https://github.com/Carthage/Carthage/issues/169.

It would be great if Carthage placed dSYM files in the `Carthage/Build/` directory after building. (Or anywhere where it's available to the project, really.)

sharplet:
Fixes #507.

This is basically working at this point. I'm open to any feedback you have on the approach I've taken. Currently I just look for `*.framework.dSYM` and copy that into Carthage's build directory alongside the framework.

The main thing missing here is creating a universal dSYM for both iOS platforms. I'm not entirely sure whether this is necessary, as you usually won't need to symbolicate crash reports for simulator builds—but maybe there's a use case I'm not thinking of?

This could probably do with a test that the iOS dSYM includes the arm architectures rather than simulator.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -866,6 +866,25 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 				fatalError("SDK count \(platform.SDKs.count) for platform \(platform) is not supported")
 			}
 		}
+		|> flatMapTaskEvents(.Concat) { builtProductURL in
+			return createDebugInformation(builtProductURL)
+				|> then(SignalProducer(value: builtProductURL))
+		}
+}
+
+public func createDebugInformation(builtProductURL: NSURL) -> SignalProducer<TaskEvent<NSURL>, CarthageError> {
+	let dSYMURL = builtProductURL.URLByAppendingPathExtension("dSYM")
+
+	if let builtProduct = builtProductURL.path, dSYM = dSYMURL.path {
+		let executable = builtProduct.stringByAppendingPathComponent(builtProduct.lastPathComponent.stringByDeletingPathExtension)
+		let dsymutilTask = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: ["dsymutil", executable, "-o", dSYM])
+
+		return launchTask(dsymutilTask)
+			|> mapError { .TaskError($0) }
+			|> flatMapTaskEvents(.Concat) { _ in SignalProducer(value: dSYMURL) }
+	} else {
+		return .empty
+	}
 }
 
 /// A producer representing a scheme to be built.
--- a/Source/CarthageKitTests/XcodeSpec.swift
+++ b/Source/CarthageKitTests/XcodeSpec.swift
@@ -67,13 +67,21 @@ class XcodeSpec: QuickSpec {
 
 			for dependency in projectNames {
 				let macPath = buildFolderURL.URLByAppendingPathComponent("Mac/\(dependency).framework").path!
+				let macdSYMPath = macPath.stringByAppendingPathExtension("dSYM")!
 				let iOSPath = buildFolderURL.URLByAppendingPathComponent("iOS/\(dependency).framework").path!
+				let iOSdSYMPath = iOSPath.stringByAppendingPathExtension("dSYM")!
 
 				var isDirectory: ObjCBool = false
 				expect(NSFileManager.defaultManager().fileExistsAtPath(macPath, isDirectory: &isDirectory)).to(beTruthy())
 				expect(isDirectory).to(beTruthy())
 
+				expect(NSFileManager.defaultManager().fileExistsAtPath(macdSYMPath, isDirectory: &isDirectory)).to(beTruthy())
+				expect(isDirectory).to(beTruthy())
+
 				expect(NSFileManager.defaultManager().fileExistsAtPath(iOSPath, isDirectory: &isDirectory)).to(beTruthy())
+				expect(isDirectory).to(beTruthy())
+
+				expect(NSFileManager.defaultManager().fileExistsAtPath(iOSdSYMPath, isDirectory: &isDirectory)).to(beTruthy())
 				expect(isDirectory).to(beTruthy())
 			}
 			let frameworkFolderURL = buildFolderURL.URLByAppendingPathComponent("iOS/ReactiveCocoaLayout.framework")


Source/CarthageKit/Xcode.swift
```diff
@@ -1032,6 +1064,9 @@ public func copyFramework(from: NSURL, to: NSURL) -> SignalProducer<NSURL, Carth
 	}
 }
 
+/// Alias of `copyFile` for backwards compatiblity.
+public let copyFramework = copyFile
```
sharplet: I included this because `copyFramework` is marked `public`. Let me know if you think this is unnecessary.

sharplet:
Re. universal dSYMs, this could be useful information: http://lists.apple.com/archives/xcode-users/2009/Apr/msg00034.html.

jspahrsummers:
Thanks for the pull request!

How do these dSYMs ultimately get used (i.e., what's the process for using them)?

sharplet:
- Copy these dSYMs into the built products folder
- Package them, along with the dSYMs for my app and its frameworks, into a single .dSYM.zip bundle: `zip -r App.dSYM.zip *.dSYM`

I use HockeyApp for crash reporting, where you can upload the .dSYM.zip file and it will do the rest. (I imagine it's the same or similar for other services.) Or alternatively you can just archive it for manual symbolication.

I think it would also be worth testing whether copying the dSYMs into the built products folder is enough for them to be picked up by Xcode when you Build & Archive: https://developer.apple.com/library/ios/technotes/tn2151/_index.html

sharplet:
Ok, so I went ahead and did a bit of investigation around universal dSYMs, and it's actually very simple. In fact, directly after running `carthage build`, you can just do something like this:

```
$ cd Carthage/Build/iOS
$ dsymutil Result.framework/Result -o Result.framework.dSYM
$ file Result.framework.dSYM/Contents/Resources/DWARF/Result
Result.framework.dSYM/Contents/Resources/DWARF/Result: Mach-O universal binary with 4 architectures
Result.framework.dSYM/Contents/Resources/DWARF/Result (for architecture i386):	Mach-O dSYM companion file i386
Result.framework.dSYM/Contents/Resources/DWARF/Result (for architecture x86_64):	Mach-O 64-bit dSYM companion file x86_64
Result.framework.dSYM/Contents/Resources/DWARF/Result (for architecture armv7):	Mach-O dSYM companion file arm
Result.framework.dSYM/Contents/Resources/DWARF/Result (for architecture arm64):	Mach-O 64-bit dSYM companion file
```

This should just work provided that the original object files haven't been cleaned up between building the framework and running `dsymutil`.

I think I'll rewrite this to just call `dsymutil` instead of copying `.dSYM` files around.

sharplet:
Ok! So I've rewritten this to use `dsymutil` to create the dSYMs. It's much simpler, and as a bonus they are also now universal.

Let me know what you think!

jspahrsummers:
@Carthage/carthage I commented on this, but I don't think I'll have time to review it in detail. Anyone else interested? :pray:

sharplet:
Just been thinking on this further and I definitely think this feature would be best if we can verify that Carthage dSYMs can be included in apps submitted to TestFlight or the App Store. In my app I'm copying them into my app's built products directory in a build phase, but I haven't tried a TestFlight deployment yet so I'm not sure if that's enough.

In any case, if all that's needed is to copy them somewhere during a build, perhaps a new `carthage copy-dsyms` command could work, or even an addition to the `copy-frameworks` command?

Source/CarthageKit/Xcode.swift
```diff
@@ -866,6 +866,29 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 				fatalError("SDK count \(platform.SDKs.count) for platform \(platform) is not supported")
 			}
 		}
+
+	return builtProduct
```
mdiep: Can you remove this variable and just add the `|> …` to the end of the chain of operations?

Source/CarthageKit/Xcode.swift
```diff
@@ -866,6 +866,29 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
...
+		|> flatMapTaskEvents(.Concat) { builtProductURL in
+			return createDebugInformation(builtProductURL)
+				|> then(SignalProducer(value: builtProductURL))
+			}
```
mdiep: This `}` looks over-indented. It should be on the same line as the `|>`.

Source/CarthageKit/Xcode.swift
```diff
@@ -866,6 +866,29 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
...
+
+	if let builtProduct = builtProductURL.path, dSYM = dSYMURL.path {
+		let executable = builtProduct.stringByAppendingPathComponent(builtProduct.lastPathComponent.stringByDeletingPathExtension)
+		let dsymutilTask = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: ["dsymutil", executable, "-o", dSYM])
+
+		return launchTask(dsymutilTask)
+			|> catch { error in SignalProducer(error: .TaskError(error)) }
+			|> map { taskEvent in
+				taskEvent.map { _ in dSYMURL }
+		}
```
mdiep: This `}` is under-indented.

Source/CarthageKit/Xcode.swift
```diff
@@ -866,6 +866,29 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
...
+
+public func createDebugInformation(builtProductURL: NSURL) -> SignalProducer<TaskEvent<NSURL>, CarthageError> {
+	let dSYMURL = builtProductURL.URLByAppendingPathExtension("dSYM")
+
+	if let builtProduct = builtProductURL.path, dSYM = dSYMURL.path {
+		let executable = builtProduct.stringByAppendingPathComponent(builtProduct.lastPathComponent.stringByDeletingPathExtension)
+		let dsymutilTask = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: ["dsymutil", executable, "-o", dSYM])
+
+		return launchTask(dsymutilTask)
+			|> catch { error in SignalProducer(error: .TaskError(error)) }
```
mdiep: I think this can just be `|> mapError { .TaskError($0) }`.

Source/CarthageKit/Xcode.swift
mdiep: Can this use `flatMapTaskEvents`?

mdiep:
Sorry for the long delay in reviewing! I made a few notes.

> I definitely think this feature would be best if we can verify that Carthage dSYMs can be included in apps submitted to TestFlight or the App Store. In my app I'm copying them into my app's built products directory in a build phase, but I haven't tried a TestFlight deployment yet so I'm not sure if that's enough.

So you've successfully used this for HockepApp but haven't verified with TestFlight or the App Store?

sharplet:
> Sorry for the long delay in reviewing! I made a few notes.

No problem! I've addressed your feedback, let me know if there's anything I've missed.

> So you've successfully used this for HockepApp but haven't verified with TestFlight or the App Store?

Yep, but I just successfully verified that if the dSYMs are copied into `BUILT_PRODUCTS_DIR` during the build, Xcode will copy them into your .xcarchive bundle when archiving.

mdiep:
Looks great—thanks!

sharplet:
Thank you!

jscubida:
Just to verify, does this pull request make this part of the README unnecessary? 

> Copying debug symbols for debugging and crash reporting
> 
> 1. On your application target’s “Build Phases” settings tab, click the “+” icon and choose “New Copy Files Phase”.
> 2. Click the “Destination” drop-down menu and select “Products Directory”.
> 3. For each framework you’re using, drag and drop its corresponding dSYM file.

ianbytchek:
@sharplet can you comment on the above? I'm coming from [here](https://github.com/fastlane/fastlane/issues/8163#issuecomment-278117940), trying to figure out the best approach.

sharplet:
@ianbytchek I haven't tested this for some time, but I just checked with a recent app and the dSYMs are definitely in the .xcarchive.

Are you using the `carthage copy-frameworks` command? Are you downloading binaries or building them locally?

ianbytchek:
@sharplet I'm not, I'm working on macOS project and linking / embedding them directly in build phases. Would that be the reason why?

sharplet:
I think so. If you're not using `copy-frameworks`, you'll need to copy the dSYMs manually.

PR #496, opened 2015-05-27
Call TaskDescription.waitForAllTaskTermination() on waitOnCommand()

norio-nomura:
Fix segmentation fault on termination.
https://github.com/Carthage/Carthage/pull/474#issuecomment-105467060

PR #474, opened 2015-05-18
Fix hanging during `carthage update`

jspahrsummers:
_Depends on https://github.com/Carthage/ReactiveTask/pull/23._

:pray: Fixes #455. Fixes #445. :pray:

**To do:**

 - [ ] Merge dependencies and bump Cartfile
 - [x] Fix fetching after checkout
 - [x] Fix nondeterministic `NoTaggedVersion` and `RequiredVersionNotFound` errors
 - [x] Fix crashes

PR #519, opened 2015-06-03
TaskDescription.waitForAllTaskTermination() on waitOnCommand()

norio-nomura:
*This is rebased version of #496.*

Fix segmentation fault on termination of Carthage.

Segmentation fault on termination of Carthage 0.7.4 beta-2 happens following scenario:
1. launchTask's thread call `sendCompleted()` and kick main thread
2a. main thread wake up and call `exit()`
2b. launchTask's thread continue disposing
3. main thread cause segmentation fault.

Waiting `completed` from launchTask's thread is not enough for fixing the segmentation fault.
It needs waiting actual termination of launchTask's thread

This PR depends on https://github.com/Carthage/ReactiveTask/pull/24

https://github.com/Carthage/Carthage/pull/474#issuecomment-107949958


Source/carthage/Extensions.swift
```diff
@@ -108,6 +109,7 @@ internal func waitOnCommand<T>(producer: SignalProducer<T, CommandError>) -> Res
 		|> then(SignalProducer<(), CommandError>.empty)
 		|> wait
 	
+	TaskDescription.waitForAllTaskTermination()
```
robrix: Should this be finer-grained, e.g. waiting per-task when they’re executed?

norio-nomura: I can't determined where I should place `dispatch_group_wait()` on another place.
RAC's `completed` event timing is not fit for place it.

norio-nomura: The segmentation fault happens while disposing after `completed` event from launch task's thread and main thread calling `exit()` concurrently.
Waiting per-task will need waiting some RAC event. That will be same what just current implementation is doing.
I think fixing the segmentation fault needs doing outside of RAC event, and waiting per-task without RAC event may be too complicated unnecessarily.

robrix: Okay, thanks for looking into this!

robrix:
I’ve merged [the dependency](https://github.com/Carthage/ReactiveTask/pull/24) and have [a beta tag](https://github.com/Carthage/ReactiveTask/releases/tag/0.7-beta1) of ReactiveTask to include it.

Could you please update the `Cartfile` & submodules to pin to that tag?

robrix:
Once that’s done I think we can merge and release a new beta of Carthage. I may not be aesthetically content with waiting for all child processes like this, but I can’t argue with its results.

norio-nomura:
I've pushed the commit updating the `Cartfile` & submodules to pin to `0.7-beta1`

robrix:
Thanks very much!

norio-nomura:
You're welcome. :bow:

PR #521, opened 2015-06-03
Clarify how to build before creating archive.

kareman:
It took a bit of research to find out that you need to use «carthage
build --no-skip-current» first. I think this error message is a bit more helpful.


mdiep:
Seems like a good idea to me, but I'll leave it open in case anyone else in @Carthage/carthage has a different opinion.

neilpa:
:+1: 

I always appreciate "how to fix it" messages in warnings and errors.

PR #524, opened 2015-06-04
Allow github releases with null names

erichoracek:
JSON deserialization of a GitHub release fails when the name field is null. Currently, the implementation of `GitHubRelease` only accepts empty strings as values of name. This ammends releases to allow nil strings as well.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -118,11 +118,14 @@ public struct GitHubRelease: Equatable {
 	public let ID: Int
 
 	/// The name of this release.
-	public let name: String
+	public let name: String?
 
-	/// The name of this release, with fallback to its tag when the name is an empty string.
+	/// The name of this release, with fallback to its tag when the name is an empty string or nil.
 	public var nameWithFallback: String {
-		return name.isEmpty ? tag : name
+		if let name = name {
+			return name.isEmpty ? tag : name
+		}
+		return tag
 	}
 
 	/// The name of the tag upon which this release is based.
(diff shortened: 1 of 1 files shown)


mdiep:
AFAIK, the release name cannot be null. Were you seeing a null release name somewhere?

erichoracek:
@mdiep We're uploading a release via the GitHub API as part of CI without the `name` parameter (just `tag`). See the [create release method](https://octokit.github.io/octokit.rb/Octokit/Client/Releases.html#create_release-instance_method) in octokit.rb, in which `name` is an optional argument.

Source/CarthageKit/GitHub.swift
```diff
@@ -118,11 +118,14 @@ public struct GitHubRelease: Equatable {
...
+	public let name: String?
 
-	/// The name of this release, with fallback to its tag when the name is an empty string.
+	/// The name of this release, with fallback to its tag when the name is an empty string or nil.
 	public var nameWithFallback: String {
-		return name.isEmpty ? tag : name
+		if let name = name {
+			return name.isEmpty ? tag : name
+		}
+		return tag
```
mdiep: I don't like the duplication of `tag` here. Could we use a signal conditional here so that it doesn't have to be repeated? :pray:

You should be able to use [Swift 1.2's improved optional binding](http://nshipster.com/swift-1.2/).

mdiep:
Thanks for opening a pull request! Just the one note, then this is good to go!

jspahrsummers:
:+1: but I'll let @mdiep have the final say.

mdiep:
:metal:

PR #525, opened 2015-06-04
Use case insensitive comparison for GitHub repository names

### LINKED ISSUE #503: GitHub repository names are currently considered case-sensitive
This can lead to the same repository being treated as multiple projects, if two other projects include it as a dependency with different cases.

For example:

```
github "Carthage/CarthageKit"
github "carthage/carthagekit"
```

Besides it being annoying to see duplicate entries in `Cartfile.resolved`, this can cause real problems like version incompatibility (they'd each resolve separately) and trying to modify the same Git repository on disk multiple times concurrently.

GitHub treats repository names case-insensitively—we should too.

eliperkins:
Fixes #503

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -97,7 +97,7 @@ public struct GitHubRepository: Equatable {
 }
 
 public func ==(lhs: GitHubRepository, rhs: GitHubRepository) -> Bool {
-	return lhs.owner == rhs.owner && lhs.name == rhs.name
+	return lhs.owner.lowercaseString == rhs.owner.lowercaseString && lhs.name.lowercaseString == rhs.name.lowercaseString
 }
 
 extension GitHubRepository: Hashable {
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -97,7 +97,7 @@ public struct GitHubRepository: Equatable {
 }
 
 public func ==(lhs: GitHubRepository, rhs: GitHubRepository) -> Bool {
-	return lhs.owner.caseInsensitiveCompare(rhs.owner) == NSComparisonResult.OrderedSame && lhs.name.caseInsensitiveCompare(rhs.name) == NSComparisonResult.OrderedSame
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/GitHub.swift
```diff
@@ -97,7 +97,7 @@ public struct GitHubRepository: Equatable {
 }
 
 public func ==(lhs: GitHubRepository, rhs: GitHubRepository) -> Bool {
-	return lhs.owner == rhs.owner && lhs.name == rhs.name
+	return lhs.owner.lowercaseString == rhs.owner.lowercaseString && lhs.name.lowercaseString == rhs.name.lowercaseString
```
mdiep: It's usually preferable to use a case-insensitive comparison operation (like `String.caseInsensitiveCompare`). But since the strings here are limited to ASCII, I'm not sure if it matters.

Anyone else in @Carthage/carthage have an opinion?

mdiep:
Thanks for opening a pull request! :sparkles:

Source/CarthageKit/GitHub.swift
robrix: `String.caseInsensitiveCompare` expresses the intent much more clearly. And who knows if we’ll allow ÜTF8 usernames someday.

mdiep: @eliperkins Would you mind making that change?

eliperkins: Done!

eliperkins:
:nail_care:

Source/CarthageKit/GitHub.swift
```diff
@@ -97,7 +97,7 @@ public struct GitHubRepository: Equatable {
 }
 
 public func ==(lhs: GitHubRepository, rhs: GitHubRepository) -> Bool {
-	return lhs.owner == rhs.owner && lhs.name == rhs.name
+	return lhs.owner.caseInsensitiveCompare(rhs.owner) == NSComparisonResult.OrderedSame && lhs.name.caseInsensitiveCompare(rhs.name) == NSComparisonResult.OrderedSame
```
mdiep: Could you remove the `NSComparisonResult` here so it's just `== .OrderedSame`? :pray:

eliperkins: Heh. Good call. :+1:

mdiep:
Looks great! Thanks!

PR #529, opened 2015-06-05
Fix release downloads by using GitHub API rather than browser URL

erichoracek:
Currently, Carthage uses the `browser_url` field on a GitHub release asset to download precompiled binaries. However, when downloading a private binary, this URL only succeeds if the user is logged into their GitHub account on Safari. This is because the cookies are shared between Safari and Carthage, and thus the browser URL request is successful because the user is already authenticated. As such, If the user is **not** logged into their GitHub account in Safari, the download fails with an opaque error stating that the release zip file is invalid. When you inspect the contents of the downloaded file in the `~/Libarary/Caches/org.carthage.CarthageKit/binaries/` directory, the contents of the file are the string "Not Found", rather than the zipped `.framework` that you'd expect.

To rememdy this, we use the GitHub release asset API with a content type of `application/octet-stream` to download the asset, as described [here](https://developer.github.com/v3/repos/releases/#get-a-single-release-asset). This ensures that private release assets are downloaded successfully regardless of whether the user is logged into Safari.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -28,6 +28,10 @@ private let userAgent: String = {
 /// The type of content to request from the GitHub API.
 private let APIContentType = "application/vnd.github.v3+json"
 
+/// The type of content to request from the GitHub API when downloading assets
+/// from releases.
+private let APIAssetDownloadContentType = "application/octet-stream"
+
 /// Represents an error returned from the GitHub API.
 public struct GitHubError: Equatable {
 	public let message: String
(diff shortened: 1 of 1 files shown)


robrix:
Hah. Good fix :+1:

robrix:
Also: good diagnosis! That sounds like it mustn’t have been fun :grimacing:

robrix:
Short & sweet PR :+1:

mdiep:
Thank you for opening a PR for this! :sparkles:

PR #531, opened 2015-06-06
Don't add an @rpath that's already set

jspahrsummers:
Resolves https://github.com/Homebrew/homebrew/pull/40450#issuecomment-109534187.


robrix:
:+1: noticed this yesterday while building, but wasn’t sure I wasn’t holding it wrong.

PR #552, opened 2015-06-13
Support for watchOS builds

JaviSoto:
Tested with `Result`:

screen shot 2015-06-12 at 6 09 06 pm

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -258,8 +258,11 @@ public enum Platform: Equatable {
 	/// iOS for device and simulator.
 	case iOS
 
+	/// Apple Watch device and simulator.
+	case watchOS
+
 	/// All supported build platforms.
-	public static let supportedPlatforms: [Platform] = [ .Mac, .iOS ]
+	public static let supportedPlatforms: [Platform] = [ .Mac, .iOS, .watchOS ]
 
 	/// The relative path at which binaries corresponding to this platform will
 	/// be stored.
@@ -270,6 +273,9 @@ public enum Platform: Equatable {
 
 		case .iOS:
 			return CarthageBinariesFolderPath.stringByAppendingPathComponent("iOS")
+
+		case .watchOS:
+			return CarthageBinariesFolderPath.stringByAppendingPathComponent("watchOS")
 		}
 	}
 
@@ -281,6 +287,9 @@ public enum Platform: Equatable {
 
 		case .iOS:
 			return [ .iPhoneSimulator, .iPhoneOS ]
+
+		case .watchOS:
+			return [ ]
 		}
 	}
 }
@@ -303,6 +312,9 @@ extension Platform: Printable {
 
 		case .iOS:
 			return "iOS"
+
+		case .watchOS:
+			return "watchOS"
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


Source/carthage/Build.swift
```diff
@@ -258,6 +270,7 @@ extension BuildPlatform: ArgumentType {
 	private static let acceptedStrings: [String: BuildPlatform] = [
 		"Mac": .Mac, "macosx": .Mac,
 		"iOS": .iOS, "iphoneos": .iOS, "iphonesimulator": .iOS,
+		"watchOS": .watchOS, "watchos": .watchOS, "whatchsimulator": .iOS,
```
ikesyo: `"watchsimulator": .watchOS,`?

JaviSoto: Yup! Nice catch.

Source/carthage/Build.swift
```diff
@@ -258,6 +270,7 @@ extension BuildPlatform: ArgumentType {
 	private static let acceptedStrings: [String: BuildPlatform] = [
 		"Mac": .Mac, "macosx": .Mac,
 		"iOS": .iOS, "iphoneos": .iOS, "iphonesimulator": .iOS,
+		"watchOS": .watchOS, "watchos": .watchOS, "whatchsimulator": .watchOS,
```
nanoxd: `whatchsimulator` -> `watchsimulator`

JaviSoto: oh wow I'm definitely tired.
:+1:

JaviSoto: Fixed

nanoxd: Happens to us all! :beer:

Source/CarthageKit/Xcode.swift
```diff
@@ -330,6 +352,12 @@ public enum SDK: Equatable {
 		case "iphonesimulator":
 			return .success(.iPhoneSimulator)
 
+		case "watchos":
+			return .success(.watchOS)
+
+		case "watchsimulator":
+			return .success(watchSimulator)
```
nubbel: Shouldn't that be `.success(.watchSimulator)`?

robrix: Both work (this is a `static func` so `watchSimulator` doesn’t need to be resolved with `.` or `self.`), but it’d be nice to be consistent with the other uses.

robrix:
Short and sweet :clap: Nice one, @JaviSoto. Just the one note :point_up:

Source/CarthageKit/Xcode.swift
JaviSoto: Oh, I didn't even know you could omit the dot in this case, so that was definitely a (lucky) typo on my part. Fixing it!

JaviSoto:
OK ready!

robrix:
:tada:

Source/carthage/Build.swift
```diff
@@ -258,6 +270,7 @@ extension BuildPlatform: ArgumentType {
 	private static let acceptedStrings: [String: BuildPlatform] = [
 		"Mac": .Mac, "macosx": .Mac,
 		"iOS": .iOS, "iphoneos": .iOS, "iphonesimulator": .iOS,
+		"watchOS": .watchOS, "watchos": .watchOS, "watchsimulator": .watchOS,
```
jspahrsummers: These are compared case-insensitively, so the two variants of `watchos` aren't needed.

soffes:
:heart:

PR #554, opened 2015-06-14
Remove redundant watchOS definition from BuildPlatform.acceptedStrings

ikesyo:
Resolves https://github.com/Carthage/Carthage/pull/552#discussion_r32374292.


robrix:
Thanks!

PR #556, opened 2015-06-15
Remove Equatable implementations for enums without associated values

ikesyo:
This is the same as ReactiveCocoa/ReactiveCocoa#2049.


neilpa:
Do you still need to declare the enums as `Equatable` or can you get that "for free" too?

ikesyo:
It compiles successfully without the declarations. :confetti_ball:

neilpa:
:+1:

PR #565, opened 2015-06-17
Use Himotoki as a JSON decoder

ikesyo:
__Disclosure:__ I am the author of [that library](https://github.com/ikesyo/Himotoki). :wink: 

This replaces Argo (and Runes) with Himotoki, aming to simplify JSON decoding implementations and reduce dependencies.


jspahrsummers:
Although I'm definitely a fan of Himotoki, this doesn't really seem to result in that much of a simplification (from the perspective of the Carthage codebase).

Ideally, we could switch to something like [JSONCodable](https://github.com/matthewcheok/JSONCodable), but that's blocked on Swift 2 support. Since we can't do that yet, let's keep the JSON stuff the way it is right now.

ikesyo:
Thanks for the response. :bow:

PR #570, opened 2015-06-18
Skip projects without shared dynamic framework schemes

ikesyo:
Resolves https://github.com/Carthage/Carthage/issues/476#issuecomment-104721552.


jspahrsummers:

Unfortunately, the bug doesn't seem to be fixed. For example, this Cartfile:

```
github "SwiftyJSON/SwiftyJSON" == 2.2
```

doesn't result in anything being built. Any ideas?

ikesyo:
That's a bad news... :dizzy_face: It works at my local machine and the test is passing in Travis. I have no idea why it doesn't at your machine at the moment.

ikesyo:
I've made some updates to simplify the implementation.

jspahrsummers:
@Carthage/carthage Can someone else take this for a spin and see if it works?

Source/CarthageKit/Xcode.swift
```diff
@@ -971,31 +984,52 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
...
+						|> map { _ in SharedScheme(schemeName: scheme, project: project) }
+				}
+				|> collect
+				|> flatMap(.Concat) { schemes in
+					if schemes.isEmpty {
+						return .empty
+					} else {
+						return SignalProducer(value: schemes)
+					}
+				}
```
mdiep: I think you could simplify this a bit by replacing the `flatMap` with a `filter`:

```swift
|> filter { !$0.isEmpty }
```
```

File: Source/CarthageKit/Xcode.swift [hunk 2]
```diff
@@ -971,31 +984,52 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
...
+			let projectToUse: ProjectLocator
+			switch project {
+			case .Workspace:
+				projectToUse = project
 
-					return BuildSchemeProducer(value: .Success(Box(initialValue)))
-						|> concat(buildProgress)
+			case .ProjectFile:
+				projectToUse = scheme.project
+			}
```
mdiep: This doesn't seem quite right to me. It seems like we should either (a) choose the first workspace or (b) choose the first workspace that includes that scheme, failing back in either case to building from the project where the scheme was defined.

Also… I know you didn't write it, but I think it'd be clearer replace the `combineLatestWith` with a `flatMap` and move the this logic into that block.

Source/CarthageKit/Xcode.swift
```diff
@@ -177,6 +177,18 @@ private func <(lhs: ProjectEnumerationMatch, rhs: ProjectEnumerationMatch) -> Bo
 	return lhs.locator < rhs.locator
 }
 
+/// A shared scheme defined in a project or workspace
+private struct SharedScheme {
+	let schemeName: String
+	let project: ProjectLocator
+}
```
mdiep: I don't really like this struct. Since Xcode projects contain schemes, it doesn't seem right for `SharedScheme` to have a `ProjectLocator`. Maybe instead we could use a named tuple?

mdiep:
Thanks for working on this! :sparkles: Sorry it took me so long to review.

This worked for me, but I made a few notes about some changes I'd like to see.

Source/CarthageKit/Xcode.swift
ikesyo: Yep, that's true. :sweat_smile:

ikesyo:
Thank you for the review and taking time for this! :sparkling_heart: I've refined what you pointed out. :boom:

Source/CarthageKit/Xcode.swift
```diff
@@ -971,31 +975,52 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
...
-						// use our initial value instead of waiting for
-						// completion.
-						|> map { taskEvent in
-							return taskEvent.map { _ in initialValue }
-						}
-						|> filter { taskEvent in taskEvent.value == nil }
+		|> take(1)
+		|> flatMap(.Merge) { schemes in SignalProducer(values: schemes) }
+		|> flatMap(.Merge) { scheme, project -> SignalProducer<(String, ProjectLocator), CarthageError> in
+			return locatorProducer
```
ikesyo: Should we avoid enumerating the given directory per scheme? That would be inefficient.

mdiep: I think we should, yes.

ikesyo: Done by 0aa6a0f. :wink:

ikesyo:
Diff without whitespace: https://github.com/Carthage/Carthage/pull/570/files?w=1

Source/CarthageKit/Xcode.swift
```diff
@@ -952,51 +952,89 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+					|> filter { project in
+						switch project {
+						case .Workspace:
+							return true
+
+						default:
+							return false
 						}
-						|> filter { taskEvent in taskEvent.value == nil }
+					}
```
mdiep: This should probably check that the workspace can actually build the scheme?

mdiep:
This is looking really good. :sparkles:

Source/CarthageKit/Xcode.swift
ikesyo: I've addressed that. :bomb:

mdiep:
Looks great! Thank you for all the work you put in on this! :sparkling_heart:

ikesyo:
:tada:

ikesyo:
A follow-up: #632.

PR #572, opened 2015-06-18
Put collect() operator in its place

ikesyo:
Now is the time. See https://github.com/Carthage/Carthage/pull/368#discussion_r29791444.


jspahrsummers:
Thanks! :fireworks:

PR #580, opened 2015-06-23
Set ONLY_ACTIVE_ARCH=NO when building

### LINKED ISSUE #504: Force ONLY_ACTIVE_ARCH=NO when building
Most Debug builds set `ONLY_ACTIVE_ARCH=YES`, which is currently honored when you run `carthage build --configuration Debug`.

Unfortunately, this can cause problems for iOS targets, because it means that you may not be able to actually run your build on the device you have connected to your machine (for example, if the active arch is `arm64` but you're trying to run on a 32-bit device).

Since part of the appeal of Carthage is that it gives you frameworks which mostly “just work,” I think we should build for all valid architectures regardless of the chosen build configuration.

ikesyo:
Resolves #504.

~~This currently sets the flag for all `xcodebuild` tasks other than `xcodebuild build`.~~

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -207,7 +207,7 @@ public func locateProjectsInDirectory(directoryURL: NSURL) -> SignalProducer<Pro
 /// Creates a task description for executing `xcodebuild` with the given
 /// arguments.
 public func xcodebuildTask(task: String, buildArguments: BuildArguments) -> TaskDescription {
-	return TaskDescription(launchPath: "/usr/bin/xcrun", arguments: buildArguments.arguments + [ task ])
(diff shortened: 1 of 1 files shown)


jspahrsummers:
> This currently sets the flag for all `xcodebuild` tasks other than `xcodebuild build`.

This seems backwards. Shouldn't it _only_ set the flag for `xcodebuild build`?

ikesyo:
I limited setting the flag only when building (when `sdk` is specified). Or should we add another property to `BuildArguments`?

jspahrsummers:
It does seem like another property in `BuildArguments` would make sense. We don't want to pass `ONLY_ACTIVE_ARCH` when we're trying to read a project's build settings, for example.

ikesyo:
@jspahrsummers Thanks for the feedback! I've addressed the points. :boom:

mdiep:
This looks fine to me.

@jspahrsummers, do you want to take another look?

jspahrsummers:
:+1: Sorry for the delay!

Issue #576, opened 2015-06-18
No way to avoid code signing?


abbeycode:
I'm trying to validate my project's Carthage compatibility with a script on Travis, and I can't get it to succeed there or locally for iOS. With the default code signing ("iOS Developer"), it complains:

> Code Sign error: No code signing identities found: No valid signing identities (i.e. certificate and private key pair) matching the team ID “(null)” were found.
> CodeSign error: code signing is required for product type 'Framework' in SDK 'iOS 8.3'

When I set it to "Don't Code Sign", then I just get the second "code signing is required" error. Is there any way around this? This is the way I'm calling Carthage:

    carthage bootstrap --configuration Debug --verbose

The project is [UnzipKit](https://github.com/abbeycode/UnzipKit), with Carthage output [on Travis here](https://travis-ci.org/abbeycode/UnzipKit/jobs/67438706#L1109). It looks like this is related to issue #281, but comments are closed on that one with no update.

mdiep:
I believe this is an Xcode limitation that Carthage can't anything about.

abbeycode:
The problem is that Carthage forces building for the device in addition to building for the simulator. For some purposes, though, the simulator is enough. Not specifying a signing identity could be taken as a signal that no device build is required. As long as a warning is logged, that shouldn't be a problem, right?

abbeycode:
@mdiep @jspahrsummers Any thoughts?

robrix:
@abbeycode As we’re all pretty busy with our day-jobs and Carthage is a spare time thing, it can take a little while for us to get around to responding. It’s not a sign of disinterest, just that time are short and obligations are many. I promise they’ll get to this issue when they have a chance :smile:

abbeycode:
@robrix Of course, apologies

jspahrsummers:
@abbeycode See #235 and specifically my conclusions in https://github.com/Carthage/Carthage/issues/235#issuecomment-68114529.

abbeycode:
Is it a lot of effort to support simulator-only builds, though? Point me in the right direction and I can submit a pull request. It's worth it to me, because otherwise my CI setup won't be able to tell me if my iOS build using Carthage is broken.

abbeycode:
[...]

PR #583, opened 2015-06-24
Allow builds to skip iOS and watchOS device SDKs when the codesigning identity is not installed

abbeycode:
This addresses issues #576, #235, and #281, by adding a new `--simulator-only` flag, and also properly treating the `iphonesimulator` and `watchsimulator` values for the `--platform` flag.

I added associated value for `Platform` and `BuildPlatform` attached to the `.iOS` and `.watchOS` values, in addition to passing a `simulatorOnly` parameter around between functions. After adding the associated values, I also had to implement the `==` overload to keep the enums `Equatable` and allow them to be used in `switch` statements.


abbeycode:
Should I make any changes to this guy?

mdiep:
Most of the core contributors have been pretty busy recently. I'll try to take a look at this soon. :relaxed:

abbeycode:
Anything I can do to help with this review?

mdiep:
I started reviewing all the open PRs over the weekend. I hope to get to this sometime this week. Sorry for the delay!

mdiep:
Thank you for opening a pull request! And sorry it took me so long to look at this. It's been a busy month. :smile:

However, as discussed in #281, we're not interested in adding a command-line flag for this. Carthage should build for device if it's able and automatically build only for the simulator in the case where code signing identities haven't been set up.

abbeycode:
Ok, and what about specifying "iphonesimulator" or "watchsimulator"? In master, those have no semantic difference from "iphoneos" or "watchos" – should I keep my change to treat those as simulator-only?

mdiep:
No, I don't think that's something that Carthage should be concerned with.

abbeycode:
@mdiep I updated (and simplified) the pull request to check with the "security" command's output, and only build for devices if an iOS signing identity is present. Can you please take a look? Also, this is my first time writing ReactiveCocoa code, so any constructive feedback is greatly appreciated.

mdiep:
That's great! Thanks for tackling this! I'll add this to my todo list and try to review it in the next few days.

Source/CarthageKit/Xcode.swift
```diff
@@ -913,6 +917,47 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+		|> mapError { CarthageError.TaskError($0) }
+		|> map { (data: NSData) -> String in
+			return NSString(data: data, encoding: NSStringEncoding(NSUTF8StringEncoding))! as String
+		}
+		|> flatMap(.Merge) { (string: String) -> SignalProducer<String, CarthageError> in
+			return string.linesProducer |> promoteErrors(CarthageError.self)
+		}
+		|> map { (identityLine: String) -> String? in
+			var error: NSError? = nil
+			let regex = NSRegularExpression(pattern: "\".+\"", options: nil, error: &error)
```
mdiep: I'd really like to see this split out into a private static variable, like `targetSettingsRegex` above.

It'd also be nice for it to be a bit more specific. I think something like this would work:

```
^\s*\d\)\s+\w+\s+"(.+)"$
```

Note the use of a capturing group for the actual name of the identity. Bonus points if you use `AllowCommentsAndWhitespace` to separate the comments to make it a little clearer. :smile:

Source/CarthageKit/Xcode.swift
```diff
@@ -913,6 +917,47 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+			if let match = regex?.firstMatchInString(identityLine, options: nil, range: NSRange(location: 0, length: count(identityLine))) as NSTextCheckingResult? {
```
mdiep: It'd be nice if the `NSRange(...)` were broken out into a separate variable to clean things up a bit.

Source/CarthageKit/Xcode.swift
```diff
@@ -913,6 +917,47 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+				let quotedIdentityName = (identityLine as NSString).substringWithRange(match.range)
+				let identityName = quotedIdentityName.stringByTrimmingCharactersInSet(NSCharacterSet(charactersInString: "\""))
```
mdiep: If you use the capturing group in the regex like I suggested above, you can get rid of most of this and use `let identityName = (identityLine as NSString).substringWithRange(match.rangeAtIndex(1))`.

Source/CarthageKit/Xcode.swift
```diff
@@ -913,6 +917,47 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+		}
+		|> ignoreNil
+}
+
+/// Returns true if the current user has any iOS signing identities configured
+public func iOSSigningIdentitiesConfigured(identities: SignalProducer<String, CarthageError> = signingIdentities()) -> Bool {
+	let iOSIdentities = identities
+		|> filter { (identity) in
+			let id = identity as NSString
+			return id.containsString("iOS") || id.containsString("iPhone")
```
mdiep: I don't think this is the correct approach. Instead of checking that the name of the identity contains `iOS` or `iPhone`, we should cross-reference the installed signing identities with the build settings of the scheme (Carthage already has code to get the build settings).

I'm not positive which setting we need to look at, but #281 suggests that we should use `EXPANDED_CODE_SIGN_IDENTITY`.

mdiep:
Thanks again for tackling this! :sparkles: I left some higher level notes above, but I think you're on the right track here. :+1:

The other thing that I'd like to see is for Carthage to print a warning if it falls back to a simulator only build.

abbeycode:
Alright, I've taken care of cleaning up the regex stuff, but now I'm left to the hard part and can use some guidance. The only way I see to get any settings from the project in Carthage is through `xcodebuild`. When running `BuildSettings.loadWithArguments()`, `EXPANDED_CODE_SIGN_IDENTITY` is empty. The only setting returned from there that seemed to indicate _something_ useful was the `CODE_SIGNING_ALLOWED` key. This returned "NO" for every iOS target in the unit tests, and "YES" for every Mac target, but that seems like shaky ground to stand on. We'd ideally also be able to determine whether the build wants a developer or distribution certificate.

I think we need to get the `CODE_SIGN_IDENTITY` value out of the `project.pbxproj` value. Do you know of any way to leverage Carthage's existing code to accomplish that? CocoaPods's [`Xcodeproj`](https://github.com/CocoaPods/Xcodeproj) does this.

mdiep:
With `BuildSettings.loadWithArguments()`, I can get a `CODE_SIGN_IDENTITY` for a project that has a set identity. Strangely, Xcode itself is showing `iOS Developer` when the `.pbxproj` and `xcodebuild` show `iPhone Developer`. I don't know if that's important.

Will that do what we need?

abbeycode:
Thanks for telling me it worked for you – allowed me to find my bug and now I'm seeing `iPhone Developer` also. For the `iOS`/`iPhone` divide, it looks like they started showing `iOS` in the UI at some point, but kept on using `iPhone` behind the scenes consistently in the keychain and project, I suppose for backwards compatibility. I should have a new solution ready soon.

abbeycode:
Alright, it's most of the way there - I was having trouble with printing out the warning line, though. Could you lend a hand with that part? I left a `TODO` comment online line 808 where I'm creating a warning string with some useful info, but I wasn't sure how best to propagate that up the signal chain (perhaps it will require some refactoring of my code).

coffeechris:
Hi @abbeycode 
This is cool. I am definitely looking forward to this being added to Carthage. I checked out your branch and built it from scratch but unfortunately I am encountering the following error:

*** Building scheme "Alamofire OSX" in Alamofire.xcodeproj
*** Building scheme "Alamofire iOS" in Alamofire.xcodeproj
fatal error: SDK count 0 for platform iOS is not supported: file
/XXX/Carthage/Source/CarthageKit/Xcode.swift, line 900

Any ideas as to what is going on?

abbeycode:
@coffeechris Is that error occurring when you build Carthage, or when you use Carthage to build your own project. Also, are you using Xcode 6.4 or 7?

abbeycode:
@coffeechris Can you share a link to the project you're trying to build? Also, are you using Xcode 6.4 or 7?

That error means that Carthage filtered out all of the SDKs for that project based on not finding matching signing identities, which shouldn't happen, since it should at least have a Simulator build available.

abbeycode:
@coffeechris I was able to build Alamofire successfully with my `simulator-only` branch using this Cartfile, on a system with an iOS signing identity:

    github "Alamofire/Alamofire"

But, on my system without an iOS signing identity installed, I get the same error as you, followed by many repeats of "Caught signal triggered by the Swift runtime!". Are you getting this with any other libraries?

copperdrake:
@abbeycode, the error you describe is the same one that @coffeechris and I are getting (we're collaborating on a project). I'm using the most recent (non-beta) version of xcode (don't have the version number handy), with no signing identity configured. Alamofire is the library throwing the error for me. After Alamofire fails, it stops trying to build other libraries, so that is the only library I have seen throw this error.

abbeycode:
@coffeechris @copperdrake I see what's going on. Alamofire is configured to require codesigning for device _and simulator_ builds, so my new filtering removes both builds (since they would fail anyway). This should, correctly, be a fatal error. The only part of this pull request that isn't done yet is printing a warning message when a target SDK is skipped over, which would have helped you to diagnose this.

The Swift runtime signal is a different story, though. That definitely shouldn't happen, and I'm not sure what's causing it. It's being triggered by the `fatalError()` call, which is in the Swift standard library. It happens with and without string interpolation in the argument for it.

I made a change to give an even clearer message before that happens, though.

coffeechris:
Well that answers that. Thanks @abbeycode

abbeycode:
@mdiep The trouble I'm having is in using `carthage.println()` from `Xcode.swift`. It looks like nothing in CarthageKit writes to the output directly, but I'm not sure what the best way would be to pass on some output to the Build command for printing. Can you lend any guidance?

abbeycode:
Ok, this guy is feature-complete, pending code review. I tackled printing by passing a nullable block into the building functions, that the warning message gets passed to. The Build command calls with a block that prints the string out.

@mdiep did you see the discussion above about the fatal errors causing Swift signals? This happens now if the device and simulator builds require a missing identity. Have you seen that before?

mdiep:
I'm trying to get back to this PR. I promise. :smile: I've been busy moving this past week, but I'll hoping find time in the next week.

abbeycode:
@mdiep No problem, I was just updating as I made progress.

mdiep:
> I tackled printing by passing an optional callback block into the building functions, that the warning message gets passed to. The Build command calls with a block that prints the string out.

This is a sign to me that the composition is wrong. That's not your fault, but it makes this quite nasty.

I think the best way to clean this up would be to do the following:

1. Make `buildScheme(withConfiguration:inProject:workingDirectoryURL:)` take an SDK argument
2. Move the logic that decides which SDKs to build from CarthageKit into carthage
3. Place the warning that you've added into carthage

What do you think of that approach?

mdiep:
> did you see the discussion above about the fatal errors causing Swift signals? This happens now if the device and simulator builds require a missing identity. Have you seen that before?

I think erroring out is okay in this case. I haven't seen it before, but I mostly do OS X development, so I wouldn't know how common it is.

abbeycode:
> What do you think of that approach?

I think that makes sense, but won't that break any consumers of CarthageKit using `buildScheme(...)`, who will then need to figure out the SDKs to build? Let's make the `SDKs` argument optional, and default to all of the platform's SDKs if none is passed in?

> I think erroring out is okay in this case

The issue isn't erroring out – I agree that's correct (if you're trying to build for only SDKs the output of which you can't sign, there's no correct way to continue). The problem is that when the fatal error is thrown, it produces mountains of useless output (this message, repeated countless times):

> Caught signal triggered by the Swift runtime!
Illegal instruction: 4

> Unfortunately, this is probably a bug in Swift and not Carthage. If

> Please also consider filing a radar with Apple, containing the version

mdiep:
> I think that makes sense, but won't that break any consumers of CarthageKit using buildScheme(...), who will then need to figure out the SDKs to build? Let's make the SDKs argument optional, and default to all of the platform's SDKs if none is passed in?

That sounds good. :+1: 

> The issue isn't erroring out – I agree that's correct (if you're trying to build for only SDKs the output of which you can't sign, there's no correct way to continue). The problem is that when the fatal error is thrown, it produces mountains of useless output (this message, repeated between 2 and 1000 times):

Weird. :confused: I think we can tackle that separately if you'd be kind enough to open an issue. :pray:

abbeycode:
Will do, on both counts

abbeycode:
As I started implementing this, I'm seeing that it's difficult to move the whole SDK list up the call chain, as information is missing. How about using a callback block inside `buildScheme()` to filter the SDK list, like so:

    shouldBuildSDK: (sdk: SDK, scheme: String, configuration: String, project: ProjectLocator) -> Bool

The default implementation would return `true`, but the function passed in from `carthage` would use the existing logic, and be able to print.

mdiep:
> As I started implementing this, I'm seeing that it's difficult to move the whole SDK list up the call chain, as information is missing

Can you elaborate on this?

abbeycode:
The list of SDKs isn't easily known from the `build` command. It's calculated based on a ProjectLocator and scheme, both of which are created/discovered in the `buildInDirectory()` method. All the `build` command knows up front is the directory of the project. The list of SDKs is calculated based on multiple command line calls to Xcode, which doesn't seem like it should be happening upstream of the `Xcode` module.

That's why I'm suggesting a callback function, which would allow the caller to optionally filter the list of SDKs, rather than prescribing a predetermined list. I could also refactor those up the call chain also, so that the `build` command has a list of schemes and a `ProjectLocator` instance, but that logic feels at home in the `Xcode` module.

mdiep:
I'm :+1: with the callback approach. I can take a look at refactoring once we have the functionality in if I'm feeling ambitious. :smile:

abbeycode:
Alright, I pushed the change with the callback, ans synced in upstream changes. Let me know how it looks. I tested locally with signing identities present, and not, and it worked in both cases (though when they were missing I get the 74,000 lines of output regarding the Swift runtime crash that I will report once this pull request is merged (since I don't know of any other way to easily reproduce it). If you want to avoid that (to actually see the output), you can update `Xcode.swift` starting at line 887 like so:

``` Swift
			switch sdksToBuild.count {
			case 0:
				return SignalProducer(error: .MissingBuildSetting("no_valid_sdks"))
```

Carthage/Checkouts/Commandant
```diff
@@ -1 +1 @@
-Subproject commit 6c2426a43a6159a7cbe7fcee8e4b5ce16c056b1e
+Subproject commit 63eabdc1023e962dbd63352b15e2b86468764442
```
mdiep: It looks like you accidentally set Commandant back to an older version.

Source/CarthageKit/Xcode.swift
```diff
@@ -789,14 +789,19 @@ private func mergeBuildProductsIntoDirectory(firstProductSettings: BuildSettings
...
+
+/// A callback function used to determine whether or not an SDK should be built
+public typealias SDKFilterCallback = (sdks: [SDK], scheme: String, configuration: String, project: ProjectLocator) -> [SDK];
+
 /// Builds one scheme of the given project, for all supported SDKs.
 ///
 /// Returns a signal of all standard output from `xcodebuild`, and a signal
 /// which will send the URL to each product successfully built.
-public func buildScheme(scheme: String, withConfiguration configuration: String, inProject project: ProjectLocator, #workingDirectoryURL: NSURL) -> SignalProducer<TaskEvent<NSURL>, CarthageError> {
+public func buildScheme(scheme: String, withConfiguration configuration: String, inProject project: ProjectLocator, canBuildSDK: SDKFilterCallback? = nil, #workingDirectoryURL: NSURL, sendWarning: (String -> Void)? = nil) -> SignalProducer<TaskEvent<NSURL>, CarthageError> {
```
mdiep: Rather than specifying `nil` as the default value for `canBuildSDK`, you could set it to `{ $0 }`. That would let you eliminate the optional and simplify some logic below.

Source/CarthageKit/Xcode.swift
```diff
@@ -1025,12 +1042,109 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+}
+
+extension CodeSigningIdentity: NilLiteralConvertible {
+	public init(nilLiteral: ()) {
+		IDHash = ""
+		IdentityType = ""
+		DeveloperName = ""
+		DeveloperID = ""
+	}
+}
```
mdiep: What's this for?

Source/CarthageKit/Xcode.swift
```diff
@@ -1025,12 +1042,109 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+					)
+			}
+			
+			return nil
+		}
+		|> ignoreNil
+}
+
+/// Returns true if the current user has any iOS signing identities configured
+public func iOSSigningIdentitiesConfigured(identities: SignalProducer<CodeSigningIdentity, CarthageError> = parseSecuritySigningIdentities()) -> Bool {
```
mdiep: It looks like this is only used in unit tests, so it could be removed.

Source/CarthageKit/Xcode.swift
```diff
@@ -1025,12 +1042,109 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+		|> mapError { CarthageError.TaskError($0) }
+		|> map { (data: NSData) -> String in
+			return NSString(data: data, encoding: NSStringEncoding(NSUTF8StringEncoding))! as String
+		}
+		|> flatMap(.Merge) { (string: String) -> SignalProducer<String, CarthageError> in
+			return string.linesProducer |> promoteErrors(CarthageError.self)
+		}
+}
+
+public struct CodeSigningIdentity {
```
mdiep: This type seems like overkill because AFAICT we only every use the UUID. Can we just return that instead of capturing all the properties?

Source/CarthageKit/Xcode.swift
abbeycode: I had started out that way, but then the default gets repeated in the signature of all the methods in the call chain, whereas this way allows me to specify the default behavior one time, closest to where it's used. Do you know of a way around repeating the ` = { $0 }` in each of the other methods (or is that an acceptable tradeoff to you)?

As a side note, ` = { $0 }` doesn't compile, though this does: `= { (sdks, _, _, _) in sdks }`. I've [asked about that](http://stackoverflow.com/q/32216909/105717) on Stack Overflow, as it makes no sense to me.

Source/CarthageKit/Xcode.swift
abbeycode: Sure, I can move that into the definition of the unit test.

Source/CarthageKit/Xcode.swift
ikesyo: I answered in SO, `{ $0.0 }` should be OK.

Source/CarthageKit/Xcode.swift
abbeycode: Well no, actually only the `IdentityType` is used (e.g. "iPhone Developer"), since that's what's specified in the project file. The rest aren't used, but does it hurt anything? I could see it being useful/helpful in the future, for presenting more detailed messages to the user.

abbeycode:
I pushed a bunch of commits, addressing all points but the last.

Source/CarthageKit/Xcode.swift
mdiep: Oh, that's right: We're checking the name.

> The rest aren't used, but does it hurt anything? I could see it being useful/helpful in the future, for presenting more detailed messages to the user.

It just adds unneeded complexity. I'd prefer to leave it out until it's needed.

Source/carthage/Build.swift
```diff
@@ -187,6 +192,64 @@ public struct BuildCommand: CommandType {
...
+			
+			BuildSettings.loadWithArguments(identityCheckArgs)
+				|> on(next: { settings in
+					let identityResult = settings["CODE_SIGN_IDENTITY"]
+					
+					var signingAllowed = true
+					
+					switch identityResult {
+					case .Success(let configuredSigningIdentity):
+						let matchingidentity = parseSecuritySigningIdentities()
```
mdiep: Can we get the installed signing identities once instead of once per SDK?

mdiep:
I'm very happy with where this is heading! :smile:

Source/carthage/Build.swift
```diff
@@ -187,6 +192,66 @@ public struct BuildCommand: CommandType {
...
+	var identityCheckArgs = BuildArguments(project: project, scheme: scheme, configuration: configuration)
+	
+	var availableIdentities = Set<CodeSigningIdentity>()
+	var result = Set<SDK>()
+	
+	parseSecuritySigningIdentities()
+		|> on(next: {identity in
+			availableIdentities.insert(identity)
+		})
+		|> wait
```
abbeycode: @mdiep Is there a cleaner way to do this? A ReactiveCocoa function, perhaps?

mdiep: You can use `collect` and `first` to return the identities as an array:

```swift
let identities: Result<[CodeSigningIdentity], CarthageError>? = parseCodeSigningIdentities
    |> collect
    |> first
```

It'd probably be better for `buildableSDKs` to take and return a `SignalProducer`, but it should at least only have **1** `wait`. You can do this by combining the identities and SDKs with `combineLatest` and returning the result of `BuildSettings.loadWithArguments` using `flatMap`.

abbeycode:
I made a bunch of improvements to `buildableSDKs`, but it does still return an array (but no `wait` anymore!. Since its results get used directly as the `sdkFilter` callback, I'd have to convert the signal to an array, either in its body, or when it's called. Do you prefer doing the conversion on line 109, where its result is read in?

mdiep:
Yeah, let's do it at line 109 where `buildableSDKs` is called.

Those nested `first`s are still a code smell! You should be able to use `combineLatest` and `flatMap` to get rid of the inner. Let me know if you need help.

mdiep:
Okay, I haven't tested this, but it compiles. :smile:

```swift
public func buildableSDKs(sdks: [SDK], scheme: String, configuration: String, project: ProjectLocator, formatting: ColorOptions.Formatting) -> SignalProducer<[SDK], CarthageError> {
	var identityCheckArgs = BuildArguments(project: project, scheme: scheme, configuration: configuration)
	
	// combine the SDKs and identities into tuples of (SDK, [CodeSigningIdentity])
	return combineLatest(SignalProducer(values: sdks), parseSecuritySigningIdentities() |> collect)
		// check that each SDK is buildable
		|> flatMap(.Concat) { (sdk: SDK, signingIdentities: [CodeSigningIdentity]) -> SignalProducer<SDK, CarthageError> in
			let availableIdentities = Set(signingIdentities)
			
			identityCheckArgs.sdk = sdk
			
			return BuildSettings.loadWithArguments(identityCheckArgs)
                           	// Ignore values if the identity isn't on the machine
				|> filter { settings -> Bool in
					if let configuredSigningIdentity = settings["CODE_SIGN_IDENTITY"].value where !availableIdentities.contains(configuredSigningIdentity) {
						let quotedSDK = formatting.quote(sdk.rawValue)
						let quotedIdentity = formatting.quote(configuredSigningIdentity)
						let message = "Skipping build for \(quotedSDK) SDK, because the necessary signing identity \(quotedIdentity) is not installed"
						carthage.println("\(formatting.bullets)WARNING: \(message)")
						return false
					}
					
					return true
				}
				// Replace the settings with the actual SDK, since that's what we want
				|> map { _ in sdk }
		}
		// Combine all the buildable SDKs into an array
		|> collect
}
```

The comments don't need to be checked in—they're just there to help explain to you what I did. Let me know if you have questions.

abbeycode:
Thanks, that's awesome, and clearly better. It looks like it's not working totally correctly, though. When I print out `sdks` before starting `combineLatest()`, it gives what I'd expect:

> *** *** ***  SDKs: [iOS Simulator, iOS Device]

But then, inside the `flatMap()`, I print out the individual `sdk`, and get this:

> *** *** *** ***  SDK: iOS Device
> *** *** *** ***  SDK: iOS Device

Why would it be passing on the same `sdk` twice and skipping over the first one?

mdiep:
Oops, you're right: `combineLatest` won't do what I want here in RAC3.

This looks like it works though:

```swift
public func buildableSDKs(sdks: [SDK], scheme: String, configuration: String, project: ProjectLocator, formatting: ColorOptions.Formatting) -> SignalProducer<[SDK], CarthageError> {
	var identityCheckArgs = BuildArguments(project: project, scheme: scheme, configuration: configuration)
	
	carthage.println("\(formatting.bullets): SDKS: \(sdks)")
	return parseSecuritySigningIdentities()
		|> collect
		|> flatMap(.Concat) { identities in
			return SignalProducer(values: sdks) |> map { ($0, identities) }
		}
		|> flatMap(.Concat) { (sdk: SDK, signingIdentities: [CodeSigningIdentity]) -> SignalProducer<SDK, CarthageError> in
			let availableIdentities = Set(signingIdentities)
			
			identityCheckArgs.sdk = sdk
			
			return BuildSettings.loadWithArguments(identityCheckArgs)
				|> filter { settings -> Bool in
					if let configuredSigningIdentity = settings["CODE_SIGN_IDENTITY"].value where !availableIdentities.contains(configuredSigningIdentity) {
						let quotedSDK = formatting.quote(sdk.rawValue)
						let quotedIdentity = formatting.quote(configuredSigningIdentity)
						let message = "Skipping build for \(quotedSDK) SDK, because the necessary signing identity \(quotedIdentity) is not installed"
						carthage.println("\(formatting.bullets)WARNING: \(message)")
						return false
					}
					
					return true
				}
				|> map { _ in sdk }
				|> take(1)
		}
		|> collect
}
```

abbeycode:
Alright, I have a good feeling about this one :smile: 

One question, though: why include the `take(1)`? Just to future-proof? `loadWithArguments` only returns a single `BuildSettings` signal.

mdiep:
> One question, though: why include the take(1)? Just to future-proof? loadWithArguments only returns a single BuildSettings signal.

I wasn't sure how many values it sent. :laughing:

Source/carthage/Build.swift
```diff
@@ -187,6 +195,50 @@ public struct BuildCommand: CommandType {
...
+			
+			identityCheckArgs.sdk = sdk
+			
+			return BuildSettings.loadWithArguments(identityCheckArgs)
+				|> filter { settings -> Bool in
+					if let configuredSigningIdentity = settings["CODE_SIGN_IDENTITY"].value
+						where !availableIdentities.contains(configuredSigningIdentity) {
+							let quotedSDK = formatting.quote(sdk.rawValue)
+							let quotedIdentity = formatting.quote(configuredSigningIdentity)
+							let message = "Skipping build for \(quotedSDK) SDK, because the necessary signing identity \(quotedIdentity) is not installed"
```
mdiep: Can you remove the `,` here? :pray:

mdiep:
This looks and works really good! There's just one small nit and then we can merge it! :zap: :zap: :zap:

Source/CarthageKit/Xcode.swift
```diff
@@ -1017,12 +1034,62 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+		"\"(.+):\\s"         + // Identity type (e.g. Mac Developer, iPhone Developer)
+		"(.+)\\s\\("         + // Developer Name
+		"([A-Z0-9]+)\\)\"\\s*" // Developer ID (e.g. DUCNFCN445)
+	),
+ options: nil, error: nil)!
+
+public func parseSecuritySigningIdentities(securityIdentities: SignalProducer<String, CarthageError> = getSecuritySigningIdentities()) -> SignalProducer<CodeSigningIdentity, CarthageError> {
+	return securityIdentities
+		|> map { (identityLine: String) -> CodeSigningIdentity? in
+			var error: NSError? = nil
```
ikesyo: This `error` is not used.

Source/CarthageKit/Xcode.swift
```diff
@@ -1017,12 +1034,62 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+			let fullRange = NSMakeRange(0, count(identityLine))
+			
+			if let matches = signingIdentitiesRegex.matchesInString(identityLine, options: nil, range: fullRange) as? [NSTextCheckingResult],
+				let match = matches.first {
```
ikesyo: These two lines can be written as follows:

```swift
if let match = signingIdentitiesRegex.matchesInString(identityLine, options: nil, range: fullRange).first as? NSTextCheckingResult {
```

I think that is cleaner way.

Source/CarthageKit/Xcode.swift
```diff
@@ -781,14 +781,19 @@ private func mergeBuildProductsIntoDirectory(firstProductSettings: BuildSettings
 		}
 }
 
+
+/// A callback function used to determine whether or not an SDK should be built
+public typealias SDKFilterCallback = (sdks: [SDK], scheme: String, configuration: String, project: ProjectLocator) -> [SDK];
```
ikesyo: Trailing `;` is not needed.

Source/CarthageKit/Xcode.swift
```diff
@@ -858,21 +863,33 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 	}
 
 	return BuildSettings.SDKsForScheme(scheme, inProject: project)
-		|> map { $0.platform }
-		|> flatMap(.Concat) { (platform: Platform) in
+		|> collect
+		|> flatMap(.Concat) { (schemeSDKList: [SDK]) in
+			let platform: Platform! = schemeSDKList.first!.platform
```
ikesyo: If `schemeSDKList` is empty, unwrapping `first` would crash, so the following `fatalError()` message would not be printed, I think. Using IUO is OK, so we should avoid unwrapping:

```swift
let platform: Platform! = schemeSDKList.first?.platform
```

Source/CarthageKit/Xcode.swift
```diff
@@ -858,21 +863,33 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
...
 			let folderURL = workingDirectoryURL.URLByAppendingPathComponent(platform.relativePath, isDirectory: true).URLByResolvingSymlinksInPath!
-
+			
+			let sdksToBuild = sdkFilter(sdks: schemeSDKList, scheme: scheme, configuration: configuration, project: project)
+			
 			// TODO: Generalize this further?
-			switch platform.SDKs.count {
+			switch sdksToBuild.count {
+			case 0:
+				let isMissingSigningIdentities = schemeSDKList.count > 0
```
ikesyo: This should be `!schemeSDKList.isEmpty`?

Source/CarthageKit/Xcode.swift
```diff
@@ -1017,12 +1034,62 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
 		}
 }
 
+
+public func getSecuritySigningIdentities() -> SignalProducer<String, CarthageError> {
+	let securityTask = TaskDescription(launchPath: "/usr/bin/security", arguments: ["find-identity", "-v", "-p", "codesigning"])
```
ikesyo: Could you please add spaces as follows: `[ "find-identity", "-v", "-p", "codesigning" ]`

Source/CarthageKit/Xcode.swift
```diff
@@ -1017,12 +1034,62 @@ public func buildDependencyProject(dependency: ProjectIdentifier, rootDirectoryU
...
+	
+	return launchTask(securityTask)
+		|> ignoreTaskData
+		|> mapError { CarthageError.TaskError($0) }
```
ikesyo: `CarthageError` could be omitted.

Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -268,5 +275,85 @@ class XcodeSpec: QuickSpec {
...
+public func iOSSigningIdentitiesConfigured(identities: SignalProducer<CodeSigningIdentity, CarthageError> = parseSecuritySigningIdentities()) -> Bool {
+	let iOSIdentities = identities
+		|> filter { identity in
+			let id = identity as NSString
+			return id.containsString("iPhone") || id.containsString("iOS")
+		}
+		|> last
+	
+	return iOSIdentities != nil
+}
```
ikesyo: Could you add a blank line at the end?

Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -268,5 +275,85 @@ class XcodeSpec: QuickSpec {
 			let result = locateProjectsInDirectory(directoryURL.URLByAppendingPathComponent("ReactiveCocoaLayout")) |> first
 			expect(result).to(beNil())
 		}
+		
+		it("should parse signing identities correctly") {
+
```
ikesyo: Could you remove these empty lines after `it()`?

Source/carthage/Build.swift
```diff
@@ -104,6 +104,14 @@ public struct BuildCommand: CommandType {
 	/// Returns a producer of producers, representing each scheme being built.
 	private func buildProjectInDirectoryURL(directoryURL: NSURL, options: BuildOptions) -> SignalProducer<BuildSchemeProducer, CarthageError> {
 		let project = Project(directoryURL: directoryURL)
+		let formatting = options.colorOptions.formatting
+		let sdkFilter: SDKFilterCallback = {(sdks, scheme, configuration, project) in
```
ikesyo: The parentheses are not needed: `{ sdks, scheme, configuration, project in`

Source/carthage/Build.swift
```diff
@@ -187,6 +195,50 @@ public struct BuildCommand: CommandType {
...
+:param: sdk           The list of SDKs to filter
+:param: scheme        The scheme name to be built
+:param: configuration The Xcode configuration to be built
+:param: project       The project being built
+:param: formatting    An instance of formatting options
+
+:returns: A list of SDKs that can be built with the configured signing identities
+*/
+public func buildableSDKs(sdks: [SDK], scheme: String, configuration: String, project: ProjectLocator, formatting: ColorOptions.Formatting) -> SignalProducer<[SDK], CarthageError> {
+	var identityCheckArgs = BuildArguments(project: project, scheme: scheme, configuration: configuration)
```
ikesyo: This line should be in the second `flatMap()` closure?

Source/carthage/Build.swift
```diff
@@ -187,6 +195,50 @@ public struct BuildCommand: CommandType {
...
+			identityCheckArgs.sdk = sdk
+			
+			return BuildSettings.loadWithArguments(identityCheckArgs)
+				|> filter { settings -> Bool in
+					if let configuredSigningIdentity = settings["CODE_SIGN_IDENTITY"].value
+						where !availableIdentities.contains(configuredSigningIdentity) {
+							let quotedSDK = formatting.quote(sdk.rawValue)
+							let quotedIdentity = formatting.quote(configuredSigningIdentity)
+							let message = "Skipping build for \(quotedSDK) SDK, because the necessary signing identity \(quotedIdentity) is not installed"
+							carthage.println("\(formatting.bullets)WARNING: \(message)")
```
ikesyo: `formatting.bullets + "WARNING: \(message)"` might be preferable for consistency, or using `formatting.bulletinTitle()` is OK as [this line](https://github.com/Carthage/Carthage/blob/fafa472e77811be3967b71e8bf67bebfe0b811f9/Source/carthage/Extensions.swift#L174) does.

ikesyo:
I'm sorry for the nitpickings! :bow:

abbeycode:
@ikesyo Don't worry about it, those are all helpful tips. Let me know how it all looks! @mdiep I got yours too (first commit above)

mdiep:
@ikesyo I'll let you give this another look. You can merge when you're satisfied with it! Thank you for reviewing. :heart:

ikesyo:
Looks good now! I'm happy to merge the great work of @abbeycode and @mdiep :sparkling_heart:

mdiep:
Thanks for the pull request, @abbeycode! And thank you for being patient through the review process!!

abbeycode:
No problem, it was a great learning experience. When do you expect the next release to be cut?

mdiep:
I plan to do a release sometime this week.

abbeycode:
:+1:

guidomb:
@mdiep Could you update this thread when a new release is cut so we can make a new homebrew release?

mdiep:
I just released [0.9 Sad Effects](https://github.com/Carthage/Carthage/releases/tag/0.9). :tada:

DenHeadless:
@mdiep Can I ask you to release 0.9.0 on Homebrew? Been looking forward to this release for several months.

guidomb:
I have made the PR for Carthage 0.9 in Homebrew https://github.com/Homebrew/homebrew/pull/43884

mdiep:
@guidomb Just in time for me to release [0.9.1](https://github.com/Carthage/Carthage/releases/tag/0.9.1). :smile:

guidomb:
@abbeycode Could you take a look at https://github.com/Carthage/Carthage/issues/776. I think that that bug was introduced by this patch.

abbeycode:
It seems like Homebrew/homebrew#43884 has stalled a bit (but it's on the home stretch). What's holding it up at this point?

AAverin:
Was this feature released?
Anything I need to configure to use it?

abbeycode:
@AAverin It was released, and you don't need to do anything to use it. It skips device SDKs (only building for the simulator) when there are no signing identities present.

PR #584, opened 2015-06-24
Inherit the environment of the parent

mdiep:
This can be important, as I’ve found out. Specifically, our CI depends on having the environment set. Without it, our credentials can't be looked up, leading to anonymous API usage which can be rate limited.


robrix:
:dancer:

mdiep:
:green_heart: :green_heart: :green_heart: :green_heart: :green_heart: :green_heart:

keithduncan:
Should `launchGitTask` merge the given `environment` into the current process’ environment by default?

mdiep:
> Should `launchGitTask` merge the given `environment` into the current process’ environment by default?

Probably. I think we're now merging the current process's environment in all cases where we're specifying variables.

PR #598, opened 2015-07-03
Add an 'outdated' command

### LINKED ISSUE #510: Add an `outdated` command
Homebrew and cocoa pods both have an `outdated` command that is useful when considering what's available and what to update. It should probably split explicit vs transient dependencies and only show the latter in a `--verbose` scenario.

mattprowse:
Fixes #510.

I've tried to keep it as self-contained as possible. My approach was to combine `Project`'s `loadResolvedCartfile()` and `updatedResolvedCartfile()`, looking for pairs of `Dependencies` where `project` is equal and `version` is not equal.

The requested `--verbose` flag has been implemented. Unless `--verbose` is specified, the list of dependencies is filtered to only include those that are present in the output of `Project`'s `loadCombinedCartfile()`.

This is my first attempt at contributing to project on GitHub, and indeed my first ever pull request on GitHub, so I'm very much open to feedback and willing to work to implement it.

Some suggested improvements:

* The `OutdatedVersion` type allows the outdated command to reuse the existing `Dependency` type rather than creating an entirely new type, but it feels a bit wrong maybe.

* The list of outdated dependencies currently includes dependencies that will be downgraded to an older version. Perhaps these should be excluded, or the fact that a downgrade and not an upgrade is happening should be made explicit.

* For dependencies that specify a git object as their version, the current and proposed versions are just the SHAs of the appropriate commits. It might make sense to run the SHAs through git describe to make things a little friendlier.

* It would be nice to be able to output the latest available version of a dependency in addition to the version that will be updated to.


teddynewell:
Is there anything still in progress here, or is this ready for review @mattprowse?

mattprowse:
Hey, sorry about the radio silence. I was planning to work on some extra stuff for this PR but I'm not sure when I'll get the time, so it's probably best to review it as is. I can submit the additions as a new PR when they're ready.

Source/CarthageKit/Project.swift
```diff
@@ -354,6 +354,45 @@ public final class Project {
...
+		let currentDependenciesProducer = loadResolvedCartfile()
+			.map { $0.dependencies }
+		let updatedDependenciesProducer = updatedResolvedCartfile()
+			.map { $0.dependencies }
+		let outdatedDependenciesProducer = combineLatest(currentDependenciesProducer, updatedDependenciesProducer)
+			.map { (currentDependencies, updatedDependencies) -> [Dependency<OutdatedVersion>] in
+				var outdatedDependencies: [Dependency<OutdatedVersion>] = []
+
+				for updatedDependency in updatedDependencies {
+					if let resolvedDependency = currentDependencies.filter({ $0.project == updatedDependency.project }).first where resolvedDependency.version != updatedDependency.version {
```
mdiep: Rather than filtering the dependencies on each iteration, could you build a `[Project: Dependency<PinnedVersion>]` upfront and then check `let resolvedDependency = currentDependencies[updatedDependency.project]`?

Source/CarthageKit/Project.swift
```diff
@@ -354,6 +354,45 @@ public final class Project {
 			.collect()
 			.map(ResolvedCartfile.init)
 	}
+	
+	/// Attempts to determine which of the project's Carthage
+	/// dependencies are out of date.
+	///
+	/// This will fetch dependency repositories as necessary, but will not check
+	/// them out into the project's working directory.
+	public func outdatedDependencies(includeTransientDependencies: Bool) -> SignalProducer<[Dependency<OutdatedVersion>], CarthageError> {
```
mdiep: Rather than introducing the new `OutdatedVersion` type, how about return `SignalProducer<[(Dependency<PinnedVersion>, Dependency<PinnedVersion>)], CarthageError>`?

It's a bit of a mouthful, but it removes the need for an additional type, and I'm not keen on add more version types.

mdiep: Also, we usually call these _nested_ dependencies. We should use that terminology here.

Source/CarthageKit/Project.swift
```diff
@@ -354,6 +354,45 @@ public final class Project {
...
+			}
+
+		if !includeTransientDependencies {
+			let explicitDependencyIdentifiersProducer = loadCombinedCartfile()
+				.map { $0.dependencies.map { $0.project } }
+
+			return outdatedDependenciesProducer
+				.combineLatestWith(explicitDependencyIdentifiersProducer)
+				.map { (oudatedDependencies: [Dependency<OutdatedVersion>], explicitDependencyIdentifiers: [ProjectIdentifier]) -> [Dependency<OutdatedVersion>] in
+					return oudatedDependencies.filter{ explicitDependencyIdentifiers.contains($0.project) }
```
mdiep: There should be a space here between `filter` and `{`.

Source/carthage/Outdated.swift
```diff
@@ -0,0 +1,70 @@
...
+			return SignalProducer(value: project)
+		}
+	}
+	
+	public let verb = "outdated"
+	public let function = "Check for updates to the project's dependencies"
+
+	public func run(options: Options) -> Result<(), CarthageError> {
+		return options.loadProject()
+		.flatMap(.Merge) { $0.outdatedDependencies(options.verbose) }
```
mdiep: This line and the following lines should be indented another level. (1 level past the `return`.)

Source/carthage/Outdated.swift
```diff
@@ -0,0 +1,70 @@
...
+		public static func create(useSSH: Bool) -> Bool -> ColorOptions -> String -> Options {
+			return { verbose in { colorOptions in { directoryPath in
+				return self.init(useSSH: useSSH, verbose: verbose, colorOptions: colorOptions, directoryPath: directoryPath)
+				} } }
+		}
+		
+		public static func evaluate(m: CommandMode) -> Result<Options, CommandantError<CarthageError>> {
+			return create
+				<*> m <| Option(key: "use-ssh", defaultValue: false, usage: "use SSH for downloading GitHub repositories")
+				<*> m <| Option(key: "verbose", defaultValue: false, usage: "include transient dependencies")
```
mdiep: This should say nested dependencies too.

Source/carthage/Outdated.swift
```diff
@@ -0,0 +1,70 @@
...
+			
+			var eventSink = ProjectEventSink(colorOptions: colorOptions)
+			project.projectEvents.observeNext { eventSink.put($0) }
+			
+			return SignalProducer(value: project)
+		}
+	}
+	
+	public let verb = "outdated"
+	public let function = "Check for updates to the project's dependencies"
```
mdiep: Maybe say `compatible updates` since we're only showing updates that are compatible with the `Cartfile` requirements.

Source/carthage/Outdated.swift
```diff
@@ -0,0 +1,70 @@
...
+		.on(next: { outdatedDependencies in
+			let formatting = options.colorOptions.formatting
+
+			if outdatedDependencies.count > 0 {
+				carthage.println(formatting.bullets + formatting.path(string: "The following dependencies are outdated:"))
+				for dependency in outdatedDependencies {
+					carthage.println(formatting.bullets + formatting.projectName(string: dependency.project.name) + " \(dependency.version)")
+				}
+			} else {
+				carthage.println(formatting.bullets + "All dependencies are up to date.")
```
mdiep: I think we should not prefix the lines here with bullets since they're the point of the command. Bullets indicate progress updates—things that are incidental to the command itself.

mdiep:
Very nice! There are some things that I think we should fix up, but none of them should be very difficult.

I'm excited to see this added!

mattprowse:
Thanks for the detailed feedback! I'll make the requested changes ASAP.

mattprowse:
I think I've covered all of the requested changes. Let me know if there's anything I missed or if there anything else you notice that needs attention.

Carthage/Checkouts/xcconfigs
```diff
@@ -1 +1 @@
-Subproject commit d78854b22b9567e7f30d748bdd4966c86bcc93a5
+Subproject commit ec5753493605deed7358dec5f9260f503d3ed650
```
mdiep: Unfortunately, you mistakenly changed the submodules when you merged `master` in. That's always really confusing when you get a merge conflict. 😩

Running `carthage checkout --use-submodules` should fix it up.

mdiep:
Apart from the submodules, this looks good! Thanks for keeping it up-to-date. Once that's fixed, I expect the tests will pass and I'll merge. 👍

mattprowse:
Sorry about that! I quickly pulled in the recent Swift 2.2 changes this morning and I didn't even notice the changed submodules when I committed.

mdiep:
✨

mattprowse:
🎉

PR #601, opened 2015-07-06
Fix license badge

kaishin:
The license badge was no longer able to parse the license type since the CC license was appended to it.


jspahrsummers:
:+1: Thanks!

PR #605, opened 2015-07-07
Adds support to configure the GitHub access token to be used for GitHub API calls

### LINKED ISSUE #603: Configure Github API token to avoid rate limit
I am experiencing the following [issue](https://github.com/travis-ci/travis-ci/issues/4195) in Travis. Is there a way to tell Carthage to use a specific token to access the Github API? If not could you point me in the right direction to add this feature? Or is there a better way to avoid rate limit issues when running carthage in a shared environment with the same public IP?

guidomb:
fixes #603 

To avoid Github rate limit issues now it is possible
to provide a Github API access token.

In shared environment where several virtual machines
are using the same public ip address (like travis ci),
carthage user could hit a Github API rate limit. By
providing a Github API access token, carthage can get
a higher rate limit.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -11,7 +11,6 @@ import Foundation
 import Result
 import ReactiveCocoa
 import Runes
-import Box
 
 /// The User-Agent to use for GitHub requests.
 private let userAgent: String = {
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -11,6 +11,7 @@ import Foundation
 import Result
 import ReactiveCocoa
 import Runes
+import Box
 
 /// The User-Agent to use for GitHub requests.
 private let userAgent: String = {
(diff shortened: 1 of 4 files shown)


guidomb:
You can check https://github.com/Wolox/ReactiveArray to see a running version of this patch in TravisCI

mdiep:
I have mixed feels about this. @Carthage/carthage?

robrix:
Seems like maybe this shouldn’t be a command line argument, but maybe an environment variable?

robb:
The need is real, but I agree that an environment variable is probably the way to go here.

mdiep:
:+1: to an environment variable.

guidomb:
I will remove the command line parameter and add support for the `GITHUB_ACCESS_TOKEN` env variable.

I wasn't sure about adding a the `githubAccessToken` parameter to [this](https://github.com/guidomb/Carthage/blob/github-access-token-option/Source/CarthageKit/Project.swift#L286) and [this](https://github.com/guidomb/Carthage/blob/github-access-token-option/Source/CarthageKit/Project.swift#L444) method. It just doesn't feel right but by adding the env variable all the logic needed to decided the authentication method can be encapuslated in the `Github.swift` file.

Should I amend the previous commit or add a new one?

mdiep:
New commits are preferred to rewriting history. :smile:

guidomb:
Just replaced the command line argument for the env variable. Let me know if you want to change something else.

Source/CarthageKit/GitHub.swift
```diff
@@ -266,6 +284,14 @@ internal struct GitHubCredentials {
 	}
 }
 
+internal func loadGitHubCredentials(accessToken: GitHubAccessToken?) -> SignalProducer<GitHubCredentials?, CarthageError> {
+	if let accessToken = accessToken {
+		return SignalProducer(value: AccessTokenGitHubCredentials(accessToken: accessToken))
```
mdiep: Can we read `GITHUB_ACCESS_TOKEN` here instead of passing it in? I think it's best to read it from CarthageKit.

Source/CarthageKit/GitHub.swift
```diff
@@ -266,6 +284,14 @@ internal struct GitHubCredentials {
 	}
 }
 
+internal func loadGitHubCredentials(accessToken: GitHubAccessToken?) -> SignalProducer<GitHubCredentials?, CarthageError> {
```
mdiep: Would you be up for changing this method to something like this?

```swift
internal func loadGitHubAuthorization() -> SignalProducer<String?, CarthageError>
```

Then we could get rid of `GitHubCredentials`, `AccessTokenGitHubCredentials`, and `BasicGitHubCredentials`. All we really care about is the authorization header value. We don't need to pass around the actual credentials; it's just adding complexity.

mdiep:
I really like where this is headed—thanks for working on it! :sparkles: And sorry for the delayed review.

Source/CarthageKit/GitHub.swift
guidomb: Yeah we can definetly do that. My retaionale behind this was that for me the command line app is the one responsabile for obtaining the input (user or env) and feed that input to the CarthageKit lib.

Source/CarthageKit/GitHub.swift
guidomb: What do you mean by adding complexity? I know that thre are more types but I think it makes the code easier to read and the intent more clear. 

Maybe I am missing something but I think that we would be putting too much responsability in the `loadGitHubAuthorization` method. 

Also we would have to change every method that receives as `GitHubCredentials` to receive a `String` with the value of the authorization header. Which I don't think is good but I don't have a strong argument just that I prefer having a `GitHubCredential` that knows how to build the value for the authorization header. Maybe we can do some refactoring in https://github.com/guidomb/Carthage/blob/github-access-token-option/Source/CarthageKit/Project.swift#L311 and put all the logic of getting the github credentials inside the `GitHub.swift` file.

What do you think?

mdiep: > Also we would have to change every method that receives as GitHubCredentials to receive a String with the value of the authorization header.

There's only a few methods in the project that takes a `GitHubCredentials`. They all pass them through to `createGitHubRequest`, which only uses them to get the `authorizationHeaderValue`.

If you think having the type is preferable, then how about we make `GitHubCredentials` an enum?

```swift
internal enum GitHubCredentials {
    case .Basic(username: String, password: String)
    case .Token(String)

    var authorizationHeaderValue: String { ... }
}
```

guidomb: Yeah I like that! Thanks for your feedback!

guidomb: @mdiep I tried to implement the `GitHubCredentials` enum but after doing the refactor I got a compiler seg fault. The problem is with `internal func loadGitHubCredentials() -> SignalProducer<GitHubCredentials?, CarthageError>`. When I try to invoke this method in https://github.com/guidomb/Carthage/blob/github-access-token-option/Source/CarthageKit/Project.swift#L310 I get a seg fault from the compiler. Apparently the compiler cannot handle a generic type paremeter of an optional enum.

So we have two options: I either change the method signature to `internal func loadGitHubCredentials() -> SignalProducer<Box<GitHubCredentials>?, CarthageError>` or I do what you originally suggested (maybe you can see the future :-P) and delete anything related to `GitHubCredentials`, change the method signature to `internal func loadGitHubAuthorization() -> SignalProducer<String?, CarthageError>` and change the method that received a `GitHubCredential` to receive the value of the authorization header (which I still think is not a good option)

Another option is to change `downloadAsset` and `releaseForTag` not to receive the credentials and load them inside those method. The only problem is that when somebody uses this two methods the credentials are going to be loded more than once. This actually happens inside `installBinariesForProject` in `Project.swift`.

What do you think is best?

mdiep: Oh, Swift. Maybe that will work in 2.0.

Let's go with my original suggestion of returning the authorization header value as a `String`.

Thanks for trying the `enum`!

guidomb:
@mdiep I've just commited the fixes we were discussing. Let me know if there is anything else. Should I amend all the commits before merge?

Source/CarthageKit/GitHub.swift
```diff
@@ -11,6 +11,7 @@ import Foundation
 import Result
 import ReactiveCocoa
 import Runes
+import Box
```
ikesyo: Is this import needed?

Source/CarthageKit/Project.swift
```diff
@@ -348,8 +348,8 @@ public final class Project {
 	///
 	/// Sends the URL to each downloaded zip, after it has been moved to a
 	/// less temporary location.
-	private func downloadMatchingBinariesForProject(project: ProjectIdentifier, atRevision revision: String, fromRepository repository: GitHubRepository, withCredentials credentials: GitHubCredentials?) -> SignalProducer<NSURL, CarthageError> {
-		return releaseForTag(revision, repository, credentials)
+	private func downloadMatchingBinariesForProject(project: ProjectIdentifier, atRevision revision: String, fromRepository repository: GitHubRepository, withAuthorizationHeaderValue authorizationHeaderValue: String? = .None) -> SignalProducer<NSURL, CarthageError> {
```
ikesyo: Is there any benefit of the default argument for `withAuthorizationHeaderValue`? This method is called only in two places above, so explicitly passing nil is better I think.

mdiep:
> Should I amend all the commits before merge?

No, you don't need to amend the commits. We prefer history to be unaltered.

Source/CarthageKit/GitHub.swift
guidomb: Sorry for that I forgot to delete it. I was trying a different approch that didn't work and I left that. I am adding a new commit to fix it.

Source/CarthageKit/Project.swift
guidomb: OK I'll revert that change in another commit.

mdiep:
Thanks for all the work you put into this. It looks great! :sparkles:

ikesyo:
:tada:

robb:
@mdiep would you mind building a new release including this feature?

mdiep:
@robb I'll try to do so in the next few days. I've been busy with a move, but would really like to get all the new bits out in the wild.

guidomb:
@mdiep I would be nice (if possible) to have #583 in the next release too. Thanks!

mdiep:
I just released 0.8: Llama Calculus. #583 didn't make it in, but I'll cut another release once the current batch of PRs is merged.

I haven't opened a PR for homebrew since I don't use it, but it'd be :sparkles: if someone wanted to do so.

guidomb:
@mdiep To open a PR for homebrew do we need to collaborators or anybody can do it? I've never released a brew formula before.

mdiep:
I think anybody, but I've never done it either. You should be able to search for Carthage in their PRs to find an example.

guidomb:
Here it is https://github.com/Homebrew/homebrew/pull/42632/files

Blackjacx:
Hey any support for .env files to store the `GITHUB_ACCESS_TOKEN`? It feels unnecessary to do this each time we need to update our dependencies. It's better to store it in the .env file and let the carthage read this file like fastlane and many shell commands do it.

mdiep:
PRs welcome :)

ikesyo:
Using [direnv](https://direnv.net/) will be a possible solution.

Blackjacx:
Oh thx, that’s a great tip 👍

Issue #630, opened 2015-07-20
Should the --verbose flag affect Git output


abbeycode:
I'm troubleshooting a Carthage issue, and noticed that the `--verbose` flag passed into it doesn't make the Git commands it runs any more verbose. Are there any issues I should be aware of before making a pull request for this?

mdiep:
That sounds like a bug to me. :+1:

danielphillips:
Any news on this one?

abhiramkrishnan:
Is there an update on this one? We have random failures while carthage clones public github repos and this feature would be a great help.

mdiep:
I don't think anyone has worked on this. PRs and sponsorships are welcome. ☺️

PR #631, opened 2015-07-21
Extended use of "--verbose" flag to invocations of Git

abbeycode:
1. Every time a Git command is called, it knows whether the user passed "--verbose" - this involved adding a verbose flag to a couple of commands that didn't have one before (`checkout`, `fetch`)
2. If `verbose`, prints output like the following before calling Git:

    ```none
    Calling Git
      arguments: config --blob HEAD:.gitmodules -z --get-regexp submodule\..*\.path --verbose
      working dir: /Users/Username/Source/ProjectName
    ```
3. When calling Git, if verbose, remove any `--quiet` flags passed in, then add a `--verbose` flag

This change touched a lot of files, adding `verbose: Bool` arguments to many functions, allowing the value specified in the various Options classes to get passed on through.

Addresses Issue #630


mdiep:
I feel bad, given that you've already spent time on this, but I think the correct approach here is to remove the use of `--quiet` and use `TaskEvent` to capture the output.

I would look at how the `xcodebuild` output is captured and copy that. I think we could output the git output to a file if `--verbose` isn't specified as well—but it's probably best to use a different log than the Xcode log.

What do you think?

abbeycode:
That approach makes sense, and no worries regarding time spent - I want to get it done right, also. I have some comments and questions, though.

I looked at how the Build command redirects output: if `--verbose`, it sends it to `stdout`, otherwise sending it to a temp file. What I can see will make this approach tricky for `git` output, is that a bunch of commands call `git`, and call it multiple times, whereas only the Build command calls `xcodebuild`. That will make it trickier to redirect each of their outputs from the command-level like Build does.

While investigating, I found it also looks like Carthage doesn't actually pass a `--verbose` flag to `xcodebuild`, either. Should I add that as part of this pull request?

It also occurs to me that now, understanding what the `--verbose` flag does, maybe it should be renamed, or something. It seems like there should be one flag that controls verbosity of Carthage and its associated tools, and another that controls whether output is sent to a file or `stdout`. Do you know why those are grouped together? It would seem useful to have separate control over each.

mdiep:
For now, let's stick with the current meaning of `--verbose`:

1. Don't worry about passing the `--verbose` flag to `git` or `xcodebuild`
2. When `--verbose` is passed to `carthage`, print the `git` and `xcodebuild` output to `stdout`

Afterwards, we can consider if wether we should also have the ability to pass `--verbose` to `git` and `xcodebuild`.

I'm not terribly familiar with the `TaskEvent` APIs, but I'd imagine that you can `concat` them, which should let you combine the output from multiple commands.

abbeycode:
@mdiep This should be all set now - let me know what you think

abbeycode:
I'd love to close out this PR too @mdiep when you have a chance to take a look.

mdiep:
I thought you were going to use the `TaskEvent` APIs to return the output from the git commands? (Like how `buildScheme(withConfiguration:inProject:workingDirectoryURL:)` returns a signal that sends `TaskEvent`s.)

abbeycode:
I looked at doing it that way, but it didn't look like that was a good fit, since the output wouldn't always propagate up to the command level. I can take another look, if you'd like.

abbeycode:
One problem I hit into, was how to insert the other output I added (printing the Git args and working directory). How would I insert those into the top of the `TaskEvent` chain?

mdiep:
> I can take another look, if you'd like.

Yeah, I really think that's the best approach. It'd be :sparkles: if you could try.

> One problem I hit into, was how to insert the other output I added (printing the Git args and working directory). How would I insert those into the top of the TaskEvent chain?

You should be able to do something like this:

```swift
	let header = "..."
	let output = TaskEvent<NSData>.StandardOutput((header as NSString).dataUsingEncoding(NSASCIIStringEncoding)!)
	let taskDescription = TaskDescription(launchPath: "/usr/bin/env", arguments: [ "git" ] + arguments, workingDirectoryPath: repositoryFileURL?.path, environment: updatedEnvironment, standardInput: standardInput)
	
	return SignalProducer<TaskEvent<NSData>, ReactiveTaskError>(value: output)
		|> concat(launchTask(taskDescription))
		|> mapError { CarthageError.TaskError($0) }
		|> map { taskEvent in
			return taskEvent.map { data in
				return NSString(data: data, encoding: NSUTF8StringEncoding)! as String
			}
		}
```

(That compiles, but I didn't test that it works.)

abbeycode:
Thanks, that's what I was looking for. I agree it compiles, and will report back once I can run it. However, I've hit what I think will be the last roadblock, which is connecting the "final mile", so to speak. As an example, look at `Fetch.fetchWithOptions()` in this branch. I started to modify its return statement like so:

``` swift
return cloneOrFetchProject(project, preferHTTPS: true)
    |> on(started: {
        if let temporaryURL = temporaryURL {
            carthage.println(formatting.bullets + "git output can be found in " + formatting.path(string: temporaryURL.path!))
        }
        }, next: { event, _ in
            switch event {
            case let .StandardOutput(data):
                fileHandle.writeData(data)
                
            case let .StandardError(data):
                fileHandle.writeData(data)
                stderrHandle.writeData(data)
            }

            eventSink.put(event)
    })
    |> then(.empty)
```

But... `cloneOrFetchProject()` returns `SignalProducer<(ProjectEvent, NSURL), CarthageError>`, and because its return value is used in a `Project.swift` call, I can't just change it to return a `TaskEvent<NSData>`, as I've updated all the Git-related methods I can. What would you suggest there?

mdiep:
> But... `cloneOrFetchProject()` returns `SignalProducer<(ProjectEvent, NSURL), CarthageError>`, and because its return value is used in a `Project.swift` call, I can't just change it to return a `TaskEvent<NSData>`, as I've updated all the Git-related methods I can. What would you suggest there?

Hmm… I'm struggling to understand what you're asking here. (Sorry!) I can't tell if you're worried about changing the public API, trying to figure out how to use an updated `cloneOrFetchProject` in `cloneOrFetchDependency`, or something else.

I _think_ `cloneOrFetchProject()` should return `SignalProducer<TaskEvent<(ProjectEvent, NSURL), CarthageError>`. The `TaskEvent`s will need to continue bubbling up the API chain until they're part of the public CarthageKit API.

If I've completely misunderstood the question, let me know.

abbeycode:
No, you did misunderstand, and your comments helped me move forward a little bit. I'm struggling to get this wrapped up, though. I've got some pieces that I just can't get to compile. If I commit what I have so far, could you take a look? I've increased my understanding of what's happening with the TaskEvent API and ReactiveCocoa a lot, but I feel like something is still missing.

mdiep:
If you commit what you have, I'll try to take a look this weekend. :smile:

abbeycode:
Great, thanks. I've been banging my head against the wall.

mdiep:
It looks like you're on the right track, but just have a little more to do.

I did a little of versionsForProject:

```swift
		let fetchVersions = cloneOrFetchDependency(project)
			|> flatMap(FlattenStrategy.Merge) { taskEvent -> SignalProducer<TaskEvent<String>, CarthageError> in
				switch taskEvent {
				case let .StandardOutput(data):
					return SignalProducer(value: TaskEvent.StandardOutput(data))
					
				case let .StandardError(data):
					return SignalProducer(value: TaskEvent.StandardError(data))
					
				case let .Success(repositoryURL):
					return listTags(repositoryURL.value)
				}
			}
			|> flatMapTaskEvents(FlattenStrategy.Concat) { SignalProducer(value: PinnedVersion($0)) }
			|> collect
			|> on(next: { newVersions in
				self.cachedVersions[project] = newVersions
			})
			|> flatMap(FlattenStrategy.Concat) { versions in SignalProducer(values: versions) }
```

But the last `on` and `flatMap` still need to be updated to account for the `TaskEvent`s.

The trick is to use a local variable when you get an error about finding an overload for `|>`. In this case, I pulled things out like this:

```swift
		let fetchVersions1 = cloneOrFetchDependency(project)
			|> flatMap(FlattenStrategy.Merge) { taskEvent -> SignalProducer<TaskEvent<String>, CarthageError> in
				switch taskEvent {
				case let .StandardOutput(data):
					return SignalProducer(value: TaskEvent.StandardOutput(data))
					
				case let .StandardError(data):
					return SignalProducer(value: TaskEvent.StandardError(data))
					
				case let .Success(repositoryURL):
					return listTags(repositoryURL.value)
				}
			}
		
		let fetchVersions2 = fetchVersions1
			|> flatMapTaskEvents(FlattenStrategy.Concat) { SignalProducer(value: PinnedVersion($0)) }
		
		let fetchVersions = fetchVersions2
			|> collect
			|> on(next: { newVersions in
				self.cachedVersions[project] = newVersions
			})
			|> flatMap(FlattenStrategy.Concat) { versions in SignalProducer(values: versions) }
```

That lets you get much better errors from the type inferencer. I'll move the blocks up until the problematic one is on a variable by itself.

The error in `updatedResolvedCartfile` also seems to be because it hasn't been updated for `TaskEvent`s yet. The `Resolver` doesn't take them, and it doesn't return one.

I think you'll be able to get the rest, but let me know if you need more help.

AndrewSB:
@abbeycode @mdiep how's this feature looking?

abbeycode:
@AndrewSB I got partway there and gave up.

danielphillips:
@abbeycode @mdiep what's the state of this then? I'm having rate limit issues with GitHub.com, I also have some dependancies on our enterprise GitHub account, I've got access tokens setup but it's failing with BAD CREDENTIALS for the GitHub one, I'm quite confused and I can't see any git commands being performed. 

This would be really helpful to have.

Issue #532, opened 2015-06-06
SwiftyJSON not building, no error, no Xcode build output


petereddy:
This is in version 0.7.4. I created a Cartfile with the following contents:

github "SwiftyJSON/SwiftyJSON"

$ carthage update
*** Fetching SwiftyJSON
*** Checking out SwiftyJSON at "2.2.0"
*** xcodebuild output can be found in /var/folders/qm/6l_xyqjs35b5zbhhnf4fyvy00000gn/T/carthage-xcodebuild.hiSkw1.log
$ cat /var/folders/qm/6l_xyqjs35b5zbhhnf4fyvy00000gn/T/carthage-xcodebuild.hiSkw1.log
$ 

In other words, no errors and the log is empty. If I add other projects to the Cartfile it builds them normally.

jspahrsummers:
This is because Carthage is finding the `Example.xcodeproj` of SwiftyJSON, instead of the project file containing the framework itself. See https://github.com/Carthage/Carthage/issues/476 for a related discussion.

In the meantime, possible workarounds would be to rename `Example.xcodeproj` so it's sorted after the main project, or move it into a folder so it’s more nested.

Thomvis:
I've created a PR for SwiftyJSON that 'fixes' this issue by moving the example project to a subdirectory: https://github.com/SwiftyJSON/SwiftyJSON/pull/244. I'll be using my own fork until it is merged: https://github.com/Thomvis/SwiftyJSON. I tagged it as 2.2.1.

petereddy:
Thanks, Thomvis.

By the way, I created this issue because I thought it was bad to fail silently, not that I thought that carthage should be able to handle any odd project structure. 

It'd be nice if there were some message to the effect that the project's structure wasn't supported or something. The current behavior make it look like a bug in carthage.

jspahrsummers:
See https://github.com/Carthage/Carthage/issues/476 and https://github.com/Carthage/Carthage/issues/108, which kinda cover this.

hartbit:
> See #476 and #108, which kinda cover this.

I'd like to get this issue re-opened. I agree with @petereddy that it seems wrong to skip a dependency silently - so entirely a reporting issue for me. The issues you linked do provide more information, but nothing concerning this silent reporting.

ikesyo:
@hartbit I'm working on it (#476) now. :full_moon_with_face:

kortina:
agree w @hartbit

[...]

PR #632, opened 2015-07-23
Add .NoSharedFrameworkSchemes error related to #570

ikesyo:
Addresses https://github.com/Carthage/Carthage/issues/532#issuecomment-112984683, https://github.com/Carthage/Carthage/issues/532#issuecomment-121513872 and other comments.

If `.NoSharedSchemes` error happens, failover to other projects would not work well before the fix.


Source/CarthageKit/Xcode.swift
```diff
@@ -992,6 +995,15 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
 							|> map { _ in scheme }
 					}
 					|> collect
+					|> catch { error in
+						switch error {
+						case .NoSharedSchemes:
```
ikesyo: Should the `.XcodebuildListTimeout` error be handled also?

Source/CarthageKit/Xcode.swift
```diff
@@ -992,6 +995,15 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
...
+							return .empty
+
+						default:
+							return SignalProducer(error: error)
+						}
+					}
```
mdiep: What's the purpose of this `catch`?

ikesyo: > If `.NoSharedSchemes` error happens, failover to other projects would not work well before the fix.

mdiep: Do you mean that this lets Carthage continue to run after finding a project with no shared dependencies?

ikesyo: In other words, #570 works well if some projects does not have a shared framework scheme, but have some other shared schemes. If those projects have no any shared schemes, `buildInDirectory()` errors out at the first project locator of them.

Source/CarthageKit/Xcode.swift
mdiep: No, I think it's best not to catch those.

mdiep:
Looks good. :+1: Thanks!

ikesyo:
:smile:

ikesyo:
I'm sorry, this pull request breaks the compatibility with some repositories which have no projects/workspaces (e.g. https://github.com/jspahrsummers/xcconfigs). We might consider not to send `.NoSharedFrameworkSchemes` error if a dependency have no projects/workspaces.

ikesyo:
A follow-up: #637. :bow:

PR #634, opened 2015-07-23
Upgrade ReactiveTask and ReactiveCocoa

ikesyo:
ReactiveTask [0.7-beta.2](https://github.com/Carthage/ReactiveTask/releases/tag/0.7-beta.2) specified the `swift-development` branch that is removed now already in its Cartfile. So we should avoid that now.


mdiep:
Seems legit. :+1:

PR #649, opened 2015-07-29
Use the tag name instead of its SHA for Cartfile.resolved when pinned to an exact tag

### LINKED ISSUE #431: When pinned to an exact tag, the SHA is written to Cartfile.resolved instead of the name
For example, this `Cartfile` line:

```
github "ReactiveCocoa/ReactiveCocoa" "v3.0-beta.1"
```

will become this in `Cartfile.resolved`:

```
github "ReactiveCocoa/ReactiveCocoa" "233a3c01e32b44610fdcd648d6a8603d7befa626"
```

Unlike a branch or short commit SHA, we can (and should) copy the name as-is, to be consistent with the behavior of version specifiers like `~> 3.0`, which will record the chosen tag name.

ikesyo:
Fixes #431.


mdiep:
This is great! Thanks! :sparkles:

Issue #417, opened 2015-04-07
Carthage build fails with iOS 7.1 simulator installed


erichoracek:
Steps to reproduce:

- Install the iOS 7.1 Simulator in Xcode 6.2
- Create a new Cocoa Touch Framework project in Xcode
- Run `carthage build --no-skip-current` in the directory that the Framework project was created in

You will receive the following build error:

```
CompileC ...
invalid architecture 'arm64' for deployment target '-mios-simulator-version-min=8.2'
Command /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang failed with exit code 1
```

This is due to the fact that Carthage (as far as I can tell) is running the following command during `carthage build --no-skip-current`:

```
xcodebuild build -scheme Framework -sdk iphonesimulator
```

If you run the above command on your new framework project, you will see the same build error that is produced when you run `carthage build`.

As far as I can tell, the reason for this failure is that when no destination is specified in a build command, `xcodebuild` implicitly selects a destination simulator to build for. Since we have the iOS 7.1 Simulator installed, it is (for some reason) picked as the default destination, causing the above error to occur. However, if you select the latest simulator as the build destination via the command line:

```bash
xcodebuild build -scheme Framework -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 4s,OS=latest'
```

There is no build failure. The solution for this (as far as I can tell) would be to have Carthage allow for a `destination` option to be specified in addition to the existing `platform` and `configuration` options, or simply to not support Xcode installs that have pre-iOS 8 simulators installed.

jspahrsummers:
Good catch!

Looking at the manpage for `xcodebuild`, it looks like we could maybe pass `-destination 'generic/platform=iOS Simulator'`, or perhaps fit `OS=latest` in there somehow. The flag seems poorly documented, so I'm not exactly sure.

pizthewiz:
I've run into this with SwiftHTTP daltoniam/SwiftHTTP#83 and notably the iOS 7 Simulators are *not* installed.

pizthewiz:
And strangely after upgrading to Xcode 6.4 Beta 2 and activating its command line tools, SwiftHTTP builds just fine…

Ben-G:
Has anyone found a workaround for this issue? Besides building the library without carthage 😉

sharpfive:
[...]

PR #653, opened 2015-08-04
Add -destination flag to build arguments when building for the simulator

jernejstrasner:
This fixes problems when the project deployment version is lower than the target's deployment version. In that case `xcodebuild` chooses the first destination on the list, which is iPad 2 (7.1) as of Xcode 6.4. Some projects that have targets for both a static library and dynamic framework have such a configuration (for example [HTMLReader](https://github.com/nolanw/HTMLReader)), as they support a lower deployment version for the static library than the framework.

The destination name is hardcoded to iPhone 6 because I'm currently not aware of a way to list all valid destinations.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -102,7 +102,7 @@ public struct BuildArguments {
 	/// The run destination to try building for.
 	public var destination: String?
 
-	/// The amount of time xcodebuild spends looking for the destination (in seconds).
+	/// The amount of time xcodebuild spends searching for the destination (in seconds).
 	public var destinationTimeout: UInt?
 
 	/// The build setting whether the product includes only object code for
@@ -804,29 +804,34 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 		var argsForBuilding = argsForLoading
 		argsForBuilding.onlyActiveArchitecture = .No
 
-		let fetchDestination = { (var arguments: BuildArguments) -> SignalProducer<BuildArguments, ReactiveTaskError> in
+		func fetchDestination() -> SignalProducer<String?, ReactiveTaskError> {
 			if sdk == .iPhoneSimulator {
-				let destinationLookup = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: ["simctl", "list", "devices"])
+				let destinationLookup = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: [ "simctl", "list", "devices" ])
 				return launchTask(destinationLookup)
 					|> ignoreTaskData
-					|> map { (data: NSData) -> BuildArguments in
+					|> map { data in
 						let string = NSString(data: data, encoding: NSStringEncoding(NSUTF8StringEncoding))!
 						let regex = NSRegularExpression(pattern: "-- iOS [0-9.]+ --\\n.*?\\(([0-9A-Z]{8}-([0-9A-Z]{4}-){3}[0-9A-Z]{12})\\)", options: nil, error: nil)!
-						if let lastDevice = regex.matchesInString(string as String, options: nil, range: NSRange(location: 0, length: (string as NSString).length)).last as? NSTextCheckingResult {
-							let deviceID = string.substringWithRange(lastDevice.rangeAtIndex(1))
-							arguments.destination = "platform=iOS Simulator,id=\(deviceID)"
+						let lastDeviceResult = regex.matchesInString(string as String, options: nil, range: NSRange(location: 0, length: (string as NSString).length)).last as? NSTextCheckingResult
+						return lastDeviceResult.map { result in
+							let deviceID = string.substringWithRange(result.rangeAtIndex(1))
+							return "platform=iOS Simulator,id=\(deviceID)"
 						}
-						arguments.destinationTimeout = 3
-						return arguments
 				}
 			}
-			return SignalProducer(value: arguments)
+			return SignalProducer(value: nil)
 		}
 
-		return fetchDestination(argsForBuilding)
-			|> flatMap(.Concat) { (buildArguments: BuildArguments) -> SignalProducer<TaskEvent<NSData>, ReactiveTaskError> in
-				var buildScheme = xcodebuildTask("build", buildArguments)
+		return fetchDestination()
+			|> flatMap(.Concat) { destination -> SignalProducer<TaskEvent<NSData>, ReactiveTaskError> in
+				if let destination = destination {
+					argsForBuilding.destination = destination
+					argsForBuilding.destinationTimeout = 3
+				}
+
+				var buildScheme = xcodebuildTask("build", argsForBuilding)
 				buildScheme.workingDirectoryPath = workingDirectoryURL.path!
+
 				return launchTask(buildScheme)
 			}
 			|> mapError { .TaskError($0) }
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -168,7 +168,7 @@ public struct BuildArguments {
 		}
 
 		if let destination = destination {
-			args += [ "-destination", destination.description]
+			args += [ "-destination", destination.description ]
 		}
 
 		args += onlyActiveArchitecture.arguments
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -85,33 +85,6 @@ extension ProjectLocator: Printable {
 	}
 }
 
-/// Configures a build's destination.
-public enum BuildDestination {
-	/// OSX with optional architecture parameter. Can be x86_64 (the default) or i386.
-	case OSX(arch: String?)
-	/// iOS with a device identifier as shown in the Devices tab of the Xcode Organizer.
-	case iOS(id: String)
-	/// iOS Simulator with the full name of the device to simulate (as presented in Xcode) and OS version (eg. '8.1') or 'latest' (the default).
-	case iOSSimulator(name: String, OS: String?)
-}
-
-extension BuildDestination: Printable {
-	public var description: String {
-		switch self {
-		case .OSX(arch: .Some(let arch)):
-			return "platform=OSX,arch=\(arch)"
-		case .OSX(arch: _):
-			return "platform=OSX"
-		case .iOS(id: let id):
-			return "platform=iOS,id=\(id)"
-		case .iOSSimulator(name: let name, OS: .Some(let os)):
-			return "platform=iOS Simulator,name=\(name),OS=\(os)"
-		case .iOSSimulator(name: let name, OS: _):
-			return "platform=iOS Simulator,name=\(name)"
-		}
-	}
-}
-
 /// Configures a build with Xcode.
 public struct BuildArguments {
 	/// The project to build.
@@ -127,7 +100,7 @@ public struct BuildArguments {
 	public var sdk: SDK?
 
 	/// The run destination to try building for.
-	public var destination: BuildDestination?
+	public var destination: String?
 
 	/// The amount of time xcodebuild spends looking for the destination (in seconds).
 	public var destinationTimeout: UInt?
@@ -160,7 +133,7 @@ public struct BuildArguments {
 		}
 
 		if let destination = destination {
-			args += [ "-destination", destination.description ]
+			args += [ "-destination", destination ]
 		}
 
 		if let destinationTimeout = destinationTimeout {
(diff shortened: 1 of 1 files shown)


ikesyo:
ref: #417

jernejstrasner:
One way to avoid hardcoding the destination name would be to get the deployment iOS version from the target being built, then parse `~/Library/Developer/CoreSimulator/Devices/device_set.plist` and finally pick the first device from the appropriate list.
But I'm not sure if it's worth it, considering anyone who will be building dynamic frameworks will definitely have an iPhone 6 simulator.

Ben-G:
@jernejstrasner [Here's an article](http://mgrebenets.github.io/xcode/2015/02/18/whats-your-destination/) that outlines different ways to retrieve the installed simulators. Of the suggested commands, `xcrun simctl list devices` seems to run the fastest. But it might not be worth the effort actually parsing this list.

jernejstrasner:
@Ben-G yeah, that looks pretty much the same as `~/Library/Developer/CoreSimulator/Devices/device_set.plist`, minus the plist format, which should be easier to parse.
But imho, not worth it at the moment.

ikesyo:
Because the simulator names can be changed in “Devices” window of Xcode, “iPhone 6” name may not be legal in some circumstances. However, those might be rare cases, just passing hardcoded “iPhone 6” seems reasonable at the moment.

Source/CarthageKit/Xcode.swift
```diff
@@ -97,7 +124,21 @@ public struct BuildArguments {
...
+			// Example: Target is at 8.0, project at 7.0, xcodebuild chooses the first
+			// simulator on the list, iPad 2 7.1, which is invalid for the target.
+			if sdk == .iPhoneSimulator {
+				self.destination = BuildDestination.iOSSimulator(name: "iPhone 6", OS: "latest")
+			}
+		}
+	}
+
+	/// The run destination to try building for.
+	private var destination: BuildDestination?
```
ikesyo: If this is a `private` property, `BuildDestination` should also be `private`?

Source/CarthageKit/Xcode.swift
```diff
@@ -97,7 +124,21 @@ public struct BuildArguments {
...
-	public var sdk: SDK?
+	public var sdk: SDK? {
+		didSet {
+			// If SDK is the iOS simulator, then also set the destination.
+			// This fixes problems when the project deployment version is lower than the target's
+			// and includes simulators unsupported by the target.
+			// Example: Target is at 8.0, project at 7.0, xcodebuild chooses the first
+			// simulator on the list, iPad 2 7.1, which is invalid for the target.
+			if sdk == .iPhoneSimulator {
+				self.destination = BuildDestination.iOSSimulator(name: "iPhone 6", OS: "latest")
```
ikesyo: We should set this property only when building a project as well as `onlyActiveArchitecture` does (#580), not in `sdk`'s `didSet`, I think.

Source/CarthageKit/Xcode.swift
ikesyo: Or because we would just pass a hardcoded `platform=iOS Simulator,name=iPhone 6,OS=latest` to the `-destination` setting, we might not need the enum at all.

ikesyo:
> Because the simulator names can be changed in “Devices” window of Xcode, “iPhone 6” name may not be legal in some circumstances. However, those might be rare cases, just passing hardcoded “iPhone 6” seems reasonable at the moment.

Any thoughts? @Carthage/carthage

Source/CarthageKit/Xcode.swift
```diff
@@ -126,6 +167,10 @@ public struct BuildArguments {
 			args += sdk.arguments
 		}
 
+		if let destination = destination {
+			args += [ "-destination", destination.description]
```
ikesyo: Style: Need an extra space before `]`.

ikesyo:
@jernejstrasner Thank you for tackling this! :sparkling_heart: Could you please check the some notes above?

jdhealy:
@jernejstrasner @Ben-G Looks like Xcode 7 (at least Beta 4 — build 7A165t) supports a `--json` flag for simctl, which could eliminate the need for a custom parser:

```
$ strings /Applications/Xcode-7A165t.app/Contents/Developer/usr/bin/simctl | grep -i 'json'
dataWithJSONObject:options:error:
--json
[-j | --json] [devices|devicetypes|runtimes|pairs]
```

redacted and abbreviated output:

```
$ /Applications/Xcode-7A165t.app/Contents/Developer/usr/bin/simctl list --json devices
{
  "devices" : {
    "iOS 9.0" : [
      {
        "state" : "Shutdown",
        "availability" : "(available)",
        "name" : "iPhone 4s",
        "udid" : "AAAAAAAA-AAAA-AAAA-AAAA-AAAAAAAAAAAA"
      },
…
    ],
    "com.apple.CoreSimulator.SimRuntime.iOS-8-4" : [
      {
        "state" : "Shutdown",
        "availability" : " (unavailable, runtime profile not found)",
        "name" : "iPhone 4s",
        "udid" : "AAAAAAAA-AAAA-AAAA-AAAA-AAAAAAAAAAAA"
      },
…
    ],
    "watchOS 2.0" : [
      {
        "state" : "Shutdown",
        "availability" : "(available)",
        "name" : "Apple Watch - 38mm",
        "udid" : "AAAAAAAA-AAAA-AAAA-AAAA-AAAAAAAAAAAA"
      },
…
    ]
  }
}
```

**Unfortunately**, the flag doesn't appear to be available in Xcode 6.4…
```
$ strings /Applications/Xcode.app/Contents/Developer/usr/bin/simctl | grep -i 'json'

$ /Applications/Xcode.app/Contents/Developer/usr/bin/simctl list --json devices
Usage: simctl list [devices|devicetypes|runtimes]
```

Source/CarthageKit/Xcode.swift
jernejstrasner: Yeah, I think you're right. I was not sure here (first time digging trough Carthage) so I put it in the setter, but will move it :+1:

Source/CarthageKit/Xcode.swift
jernejstrasner: Good point about renaming devices in organizer. I will play around with parsing the list today a bit, so I might keep the enum if that goes well. Will report back!

mdiep:
> Because the simulator names can be changed in “Devices” window of Xcode, “iPhone 6” name may not be legal in some circumstances. However, those might be rare cases, just passing hardcoded “iPhone 6” seems reasonable at the moment.

My main concern would be to future proof this somehow. This is only needed because people still deploy to iOS 7, but the fix is only valid as long as iPhone 6 exists in the simulator. I'd like to balance those two factors if possible. Is there a way to detect this case and only pass a destination when needed?

Note that you can set the project's deployment target to 8.0 and the static library's target to 7.0 and still hit the bug.

jernejstrasner:
@mdiep this will also be needed when people will have projects with mixed iOS 8 and 9 targets. If building an iOS 9 target then `xcodebuild` will pick an iOS 8 destination, which is not valid.

One way would be to check if the target being built has a higher deployment target version than any other targets and the project, and append the destination flag only if it does. But I'd rather figure out a way to properly choose a destination and have the flag always when building for the simulator.

mdiep:
But the real reason to have targets for two different OS versions is to build a static library for iOS 7 since it doesn't support a dynamic framework. Otherwise you could have one target.

If we feel that multiple targets with multiple deployment targets will still be common after iOS 7 says its last goodbye, then we should solve this in a more robust way. Using the JSON output from Xcode 7+ would be a fine approach—falling back to the `iPhone 6` value if that's not available.

jernejstrasner:
@mdiep I'm actually half done with parsing the text output of `simclt` in Xcode 6.4.

Although the case should be rare, some projects might keep a static library target for performance reasons (https://github.com/artsy/eigen/issues/586).

Anyway, I'll report back when I have more on reliably parsing `simctl` output.

mdiep:
I think it would be okay to use the JSON output from Xcode 7, but either way I'm :+1: on using the output. Thanks for tackling this! :metal:

jernejstrasner:
Just pushed working parsing logic for `simctl` output.

I'm new to RAC so the code might not be as polished as it should. Please check if it could be done in a nicer way.

Source/CarthageKit/Xcode.swift
```diff
@@ -790,10 +804,31 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 		var argsForBuilding = argsForLoading
 		argsForBuilding.onlyActiveArchitecture = .No
 
-		var buildScheme = xcodebuildTask("build", argsForBuilding)
-		buildScheme.workingDirectoryPath = workingDirectoryURL.path!
+		let fetchDestination = { (var arguments: BuildArguments) -> SignalProducer<BuildArguments, ReactiveTaskError> in
```
ikesyo: Passing `argsForBuilding` for just mutating it seems not good. How about the following?

```swift
func fetchDestination() -> SignalProducer<String?, ReactiveTaskError> {
	if sdk == .iPhoneSimulator {
		let destinationLookup = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: [ "simctl", "list", "devices" ])
		return launchTask(destinationLookup)
			|> ignoreTaskData
			|> map { data in
				let string = NSString(data: data, encoding: NSStringEncoding(NSUTF8StringEncoding))!
				let regex = NSRegularExpression(pattern: "-- iOS [0-9.]+ --\\n.*?\\(([0-9A-Z]{8}-([0-9A-Z]{4}-){3}[0-9A-Z]{12})\\)", options: nil, error: nil)!

				let lastDeviceResult = regex.matchesInString(string as String, options: nil, range: NSRange(location: 0, length: (string as NSString).length)).last as? NSTextCheckingResult
				return lastDeviceResult.map { result in
					let deviceID = string.substringWithRange(result.rangeAtIndex(1))
					return "platform=iOS Simulator,id=\(deviceID)"
				}
		}
	}

	return SignalProducer(value: nil)
}

return fetchDestination()
	|> flatMap(.Concat) { destination -> SignalProducer<TaskEvent<NSData>, ReactiveTaskError> in
		if let destination = destination {
			argsForBuilding.destination = destination
			argsForBuilding.destinationTimeout = 3
		}

		var buildScheme = xcodebuildTask("build", argsForBuilding)
		buildScheme.workingDirectoryPath = workingDirectoryURL.path!

		return launchTask(buildScheme)
	}
```

And also we can use a local function here instead of a closure.

Source/CarthageKit/Xcode.swift
```diff
@@ -790,10 +804,31 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
...
+			if sdk == .iPhoneSimulator {
+				let destinationLookup = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: ["simctl", "list", "devices"])
+				return launchTask(destinationLookup)
+					|> ignoreTaskData
+					|> map { (data: NSData) -> BuildArguments in
+						let string = NSString(data: data, encoding: NSStringEncoding(NSUTF8StringEncoding))!
+						let regex = NSRegularExpression(pattern: "-- iOS [0-9.]+ --\\n.*?\\(([0-9A-Z]{8}-([0-9A-Z]{4}-){3}[0-9A-Z]{12})\\)", options: nil, error: nil)!
```
ikesyo: :+1:

Source/CarthageKit/Xcode.swift
```diff
@@ -99,6 +99,12 @@ public struct BuildArguments {
 	/// The platform SDK to build for.
 	public var sdk: SDK?
 
+	/// The run destination to try building for.
+	public var destination: String?
+
+	/// The amount of time xcodebuild spends looking for the destination (in seconds).
```
ikesyo: `searching` instead of `looking` in accordance with `man xcodebuild`?

jernejstrasner: :+1:

Source/CarthageKit/Xcode.swift
jernejstrasner: much nicer :+1:

Source/CarthageKit/Xcode.swift
```diff
@@ -790,10 +804,36 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
...
-		buildScheme.workingDirectoryPath = workingDirectoryURL.path!
+		func fetchDestination() -> SignalProducer<String?, ReactiveTaskError> {
+			if sdk == .iPhoneSimulator {
+				let destinationLookup = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: [ "simctl", "list", "devices" ])
+				return launchTask(destinationLookup)
+					|> ignoreTaskData
+					|> map { data in
+						let string = NSString(data: data, encoding: NSStringEncoding(NSUTF8StringEncoding))!
+						let regex = NSRegularExpression(pattern: "-- iOS [0-9.]+ --\\n.*?\\(([0-9A-Z]{8}-([0-9A-Z]{4}-){3}[0-9A-Z]{12})\\)", options: nil, error: nil)!
+						let lastDeviceResult = regex.matchesInString(string as String, options: nil, range: NSRange(location: 0, length: (string as NSString).length)).last as? NSTextCheckingResult
```
ikesyo: Because `string` is already `NSString`, the `(string as NSString)` cast is unnecessary.

jernejstrasner: :+1: would be handy to have "unnecessary cast" compiler warnings :)

ikesyo:
@jernejstrasner Thank you for the quick fixes, it looks great now! :sparkles: I'm gonna merge this.

@mdiep Would you take another look?

jernejstrasner:
@ikesyo thank you!

Source/CarthageKit/Xcode.swift
```diff
@@ -790,10 +804,36 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 		var argsForBuilding = argsForLoading
 		argsForBuilding.onlyActiveArchitecture = .No
 
-		var buildScheme = xcodebuildTask("build", argsForBuilding)
-		buildScheme.workingDirectoryPath = workingDirectoryURL.path!
+		func fetchDestination() -> SignalProducer<String?, ReactiveTaskError> {
```
mdiep: Can you add a comment describing why this is necessary and perhaps linking to #417?

Source/CarthageKit/Xcode.swift
```diff
@@ -790,10 +804,36 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
...
+			if sdk == .iPhoneSimulator {
+				let destinationLookup = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: [ "simctl", "list", "devices" ])
+				return launchTask(destinationLookup)
+					|> ignoreTaskData
+					|> map { data in
+						let string = NSString(data: data, encoding: NSStringEncoding(NSUTF8StringEncoding))!
+						let regex = NSRegularExpression(pattern: "-- iOS [0-9.]+ --\\n.*?\\(([0-9A-Z]{8}-([0-9A-Z]{4}-){3}[0-9A-Z]{12})\\)", options: nil, error: nil)!
```
mdiep: Maybe add a comment here as well noting that Xcode 7 includes an option to output as JSON?

mdiep:
Looks good. :+1: I'd like to see a couple comments since it's not obvious why this would be needed. I don't love the reggae, but it looks like it will work fine for now.

Thanks for reviewing, @ikesyo!

jernejstrasner:
@mdiep most definitely! Will add in a minute.
The regex seems like the best bet for now, considering is quite specific and it will gracefully fail if it can't parse the output. I was considering `NSScanner` but that would require too much code. JSON parsing can be added once the required Xcode version is 7.0.

ikesyo:
Fair enough. :ok_hand:

ikesyo:
I added an additional commit for comments tweaks (f970b97, fixed a typo @jdhealy commented), and merged this in master (9407a69).

Thank you again @jernejstrasner and @mdiep! :smile:

jdhealy:
:clap:

mdiep:
:tada:

jernejstrasner:
Glad I could help and thanks for all the comments! :clap:

PR #654, opened 2015-08-04
Accept binary frameworks of pre-release versions

ikesyo:
Resolves https://github.com/Carthage/Carthage/issues/431#issuecomment-127588964.


mdiep:
:sparkling_heart:

PR #655, opened 2015-08-04
Build schemes from workspaces, not projects

mdiep:
We were filtering out schemes that don’t actually sit in the given
project. That works great when you’re deciding what to build, but not
when you’re deciding where to build it from.

We should be building from the workspace (if there is one) to correctly pick up dependencies.

Move the filtering to happen only where we decide which schemes to
build.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -1009,17 +1009,15 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
 			|> filter { projects in !projects.isEmpty }
 			|> flatMap(.Merge) { (projects: [(ProjectLocator, [String])]) -> SignalProducer<(String, ProjectLocator), CarthageError> in
 				return SignalProducer(values: projects)
-					|> flatMap(.Concat) { (project: ProjectLocator, schemes: [String]) in
+					|> map { (project: ProjectLocator, schemes: [String]) in
 						// Only look for schemes that actually reside in the project
-						return SignalProducer(values: schemes)
-							|> filter { (scheme: String) -> Bool in
-								if let schemePath = project.fileURL.URLByAppendingPathComponent("xcshareddata/xcschemes/\(scheme).xcscheme").path {
-									return NSFileManager.defaultManager().fileExistsAtPath(schemePath)
-								}
-								return false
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -1015,6 +1009,18 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
 			|> filter { projects in !projects.isEmpty }
 			|> flatMap(.Merge) { (projects: [(ProjectLocator, [String])]) -> SignalProducer<(String, ProjectLocator), CarthageError> in
 				return SignalProducer(values: projects)
+					|> flatMap(.Concat) { (project: ProjectLocator, schemes: [String]) in
+						// Only look for schemes that actually reside in the project
```
robrix: In the project or in the workspace? I am a bit confused.

mdiep: We want to find a scheme that lives in a project (here) but build it through the workspace (below).

Source/CarthageKit/Xcode.swift
```diff
@@ -1015,6 +1009,18 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
...
+						return SignalProducer(values: schemes)
+							|> filter { (scheme: String) -> Bool in
+								if let schemePath = project.fileURL.URLByAppendingPathComponent("xcshareddata/xcschemes/\(scheme).xcscheme").path {
+									return NSFileManager.defaultManager().fileExistsAtPath(schemePath)
+								}
+								return false
+							}
+							|> collect
+							|> map { (project, $0) }
+					}
```
robrix: It seems like this `flatMap` is equivalent to a `map` using the Swift stdlib’s native `filter` method over `schemes`:

```swift
(project, filter(schemes) { scheme in
	project.fileURL.URLByAppendingPathComponent("xcshareddata/xcschemes/\(scheme).xcscheme").path != nil
})
```

Source/CarthageKit/Xcode.swift
robrix: :ok_hand:

robrix:
I’m :+1:—what say you, @joshaber?

joshaber:
I suppose you already gave your :+1:

PR #657, opened 2015-08-05
Add support for GitHub Enterprise

### LINKED ISSUE #319: Support GitHub Enterprise as a first-class citizen
Now that we support downloading binaries from GitHub Releases, there's a potentially significant difference between the `github` and `git` specifiers in a `Cartfile`.

Enterprise repositories can only be used over `git` right now, so we should find a way to make them more first-class.

ikesyo:
Resolves #319.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -140,7 +140,7 @@ public struct GitHubRepository: Equatable {
 	}
 
 	/// Matches an identifier of the form "owner/name".
-	private static let NWORegex = NSRegularExpression(pattern: "^(\\w+)/(\\w+)$", options: nil, error: nil)!
+	private static let NWORegex = NSRegularExpression(pattern: "^([\\-\\.\\w]+)/([\\-\\.\\w]+)$", options: nil, error: nil)!
 
 	/// Parses repository information out of a string of the form "owner/name"
 	/// for the github.com, or the form "http(s)://hostname/owner/name" for
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -144,8 +144,8 @@ public struct GitHubRepository: Equatable {
 	/// Enterprise instances.
 	public static func fromIdentifier(identifier: String) -> Result<GitHubRepository, CarthageError> {
 		// GitHub.com
-		let components = split(identifier, maxSplit: 2, allowEmptySlices: false) { $0 == "/" }
-		if components.count == 2 {
+		let components = split(identifier, maxSplit: 1, allowEmptySlices: false) { $0 == "/" }
+		if components.count == 2 && !components[0].hasPrefix("http") {
 			return .success(self(owner: components[0], name: components[1]))
 		}
 
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -120,7 +120,7 @@ public struct GitHubRepository: Equatable {
 
 	/// The URL that should be used for cloning this repository over HTTPS.
 	public var HTTPSURL: GitURL {
-		return GitURL("\(baseURL.description)/\(owner)/\(name).git")
+		return GitURL("\(baseURL)/\(owner)/\(name).git")
 	}
 
 	/// The URL that should be used for cloning this repository over SSH.
@@ -130,7 +130,7 @@ public struct GitHubRepository: Equatable {
 
 	/// The URL for filing a new GitHub issue for this repository.
 	public var newIssueURL: NSURL {
-		return NSURL(string: "\(baseURL.description)/\(owner)/\(name)/issues/new")!
+		return NSURL(string: "\(baseURL)/\(owner)/\(name)/issues/new")!
 	}
 
 	public init(baseURL: URL = .GitHub, owner: String, name: String) {
@@ -184,7 +184,7 @@ extension GitHubRepository: Printable {
 			return repository
 
 		case .Enterprise:
-			return "\(baseURL.description)/\(repository)"
+			return "\(baseURL)/\(repository)"
 		}
 	}
 }
@@ -292,7 +292,7 @@ extension GitHubRelease: Decodable {
 private typealias BasicGitHubCredentials = (String, String)
 
 private func loadCredentialsFromGit(forURL URL: GitHubRepository.URL) -> SignalProducer<BasicGitHubCredentials?, CarthageError> {
-	let data = "url=\(URL.description)".dataUsingEncoding(NSUTF8StringEncoding)!
+	let data = "url=\(URL)".dataUsingEncoding(NSUTF8StringEncoding)!
 	
 	return launchGitTask([ "credential", "fill" ], standardInput: SignalProducer(value: data))
 		|> flatMap(.Concat) { string -> SignalProducer<String, CarthageError> in
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -287,8 +287,8 @@ extension GitHubRelease: Decodable {
 
 private typealias BasicGitHubCredentials = (String, String)
 
-private func loadCredentialsFromGit() -> SignalProducer<BasicGitHubCredentials?, CarthageError> {
-	let data = "url=https://github.com".dataUsingEncoding(NSUTF8StringEncoding)!
+private func loadCredentialsFromGit(forURL URL: GitHubRepository.URL) -> SignalProducer<BasicGitHubCredentials?, CarthageError> {
+	let data = "url=\(URL.description)".dataUsingEncoding(NSUTF8StringEncoding)!
 	
 	return launchGitTask([ "credential", "fill" ], standardInput: SignalProducer(value: data))
 		|> flatMap(.Concat) { string -> SignalProducer<String, CarthageError> in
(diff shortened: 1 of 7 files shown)


Source/CarthageKit/GitHub.swift
```diff
@@ -249,12 +322,12 @@ private func loadCredentialsFromGit() -> SignalProducer<BasicGitHubCredentials?,
 }
 
 
-internal func loadGitHubAuthorization() -> SignalProducer<String?, CarthageError> {
+internal func loadGitHubAuthorization(forURL URL: GitHubRepository.URL) -> SignalProducer<String?, CarthageError> {
 	let environment = NSProcessInfo.processInfo().environment
 	if let accessToken = environment["GITHUB_ACCESS_TOKEN"] as? String {
```
ikesyo: It currently does not support the GitHub access token for GitHub Enterprise repositories (originally added in #605). I have no idea to make it successful now.

mdiep: The easiest solution would probably be to let this be a comma separated series of domains and tokens.

```
GITHUB_ACCESS_TOKEN = "github.com=XXXXXXXXXXXXX,myenterprise.foo=YYYYYYYYY"
```

We could still allow a token to be provided by itself for use with GitHub.com. This would give us backwards compatibility and make the simple case simple.

What do you think?

Source/CarthageKit/GitHub.swift
```diff
@@ -64,55 +64,142 @@ extension GitHubError: Decodable {
 	}
 }
 
-/// Describes a GitHub.com repository.
+/// Describes a GitHub.com or GitHub Enterprise repository.
 public struct GitHubRepository: Equatable {
+
+	/// Represents a URL for a GitHub website.
+	public enum URL: Equatable, Hashable, Printable {
```
mdiep: This seems like it might be better called `Host`?

mdiep: Actually, `Server` is probably better yet. That's the name that octokit.objc uses.

Source/CarthageKit/GitHub.swift
```diff
@@ -64,55 +64,142 @@ extension GitHubError: Decodable {
...
+		public var hashValue: Int {
+			return scheme.hashValue ^ hostname.hashValue
+		}
+
+		public var description: String {
+			return "\(scheme)://\(hostname)"
+		}
+	}
+
+	public let baseURL: URL
```
mdiep: I like `server` here as well.

Source/CarthageKit/GitHub.swift
```diff
@@ -64,55 +64,142 @@ extension GitHubError: Decodable {
...
+	return lhs.baseURL == rhs.baseURL && lhs.owner.caseInsensitiveCompare(rhs.owner) == .OrderedSame && lhs.name.caseInsensitiveCompare(rhs.name) == .OrderedSame
+}
+
+public func ==(lhs: GitHubRepository.URL, rhs: GitHubRepository.URL) -> Bool {
+	switch (lhs, rhs) {
+	case (.GitHub, .GitHub):
+		return true
+
+	case let (.Enterprise(la, lb), .Enterprise(ra, rb)):
+		return la == ra && lb == rb
```
mdiep: These should probably be case insensitive comparisons.

Source/CarthageKit/GitHub.swift
```diff
@@ -64,55 +64,142 @@ extension GitHubError: Decodable {
...
-			return .failure(CarthageError.ParseError(description: "invalid GitHub repository name \"\(NWO)\""))
+		// GitHub.com
+		let components = split(NWO, maxSplit: 2, allowEmptySlices: false) { $0 == "/" }
+		if components.count == 2 {
+			return .success(self(owner: components[0], name: components[1]))
+		}
+
+		// GitHub Enterprise
+		if let
+			URL = NSURL(string: NWO),
```
mdiep: `NWO` stands for _Name With Owner_. Since this method now takes a GitHub Enterprise URL, it's not really appropriate terminology. Something like `fromIdentifier` or `fromSpecifier` seems more appropriate.

Source/CarthageKit/GitHub.swift
```diff
@@ -64,55 +64,142 @@ extension GitHubError: Decodable {
...
+			scheme = URL.scheme,
+			host = URL.host,
+			var pathComponents = URL.pathComponents as? [String]
+			// The trailing slash of the host is included in the components.
+			where pathComponents.count >= 3
```
mdiep: Can we be more strict here?

Source/CarthageKit/GitHub.swift
```diff
@@ -64,55 +64,142 @@ extension GitHubError: Decodable {
...
 	}
 
-	public init(owner: String, name: String) {
+	public init(baseURL: URL = .GitHub, owner: String, name: String) {
+		self.baseURL = baseURL
 		self.owner = owner
 		self.name = name
 	}
 
 	/// Parses repository information out of a string of the form "owner/name".
```
mdiep: This documentation should be updated.

Source/CarthageKit/GitHub.swift
```diff
@@ -64,55 +64,142 @@ extension GitHubError: Decodable {
...
 	public static func fromNWO(NWO: String) -> Result<GitHubRepository, CarthageError> {
-		let components = split(NWO, maxSplit: 1, allowEmptySlices: false) { $0 == "/" }
-		if components.count < 2 {
-			return .failure(CarthageError.ParseError(description: "invalid GitHub repository name \"\(NWO)\""))
+		// GitHub.com
+		let components = split(NWO, maxSplit: 2, allowEmptySlices: false) { $0 == "/" }
```
mdiep: I think `maxSplit` should still be 1 here.

```
/// :param: maxSplit the maximum number of slices to return, minus 1.
```

mdiep:
Thanks for working on this! It will be great to have GitHub Enterprise support. :sparkling_heart:

Source/CarthageKit/GitHub.swift
ikesyo: `Server` seems good.

Source/CarthageKit/GitHub.swift
ikesyo: Oh, I didn't understand the meaning, thanks. :bow:

Source/CarthageKit/GitHub.swift
ikesyo: If `maxSplit` is 1, a input like `https://enterprise.local/ghe/desktop/git-error-translations2` would be split as `[ https:, /enterprise.local/ghe/desktop/git-error-translations2 ]`. So to avoid the input be treated as the github.com instance, we need to make `maxSplit` to 2, or handle enterprise instance cases at first.

Source/CarthageKit/GitHub.swift
ikesyo: I'll chose `fromIdentifier` going along with `ProjectIdentifier`.

ikesyo:
I've updated the points except for splitting and path components handling.

Source/CarthageKit/GitHub.swift
ikesyo: I tried rejecting Enterprise identifiers with the commit 66f9f53.

Source/CarthageKit/GitHub.swift
ikesyo: 576d34a?

Source/CarthageKit/GitHub.swift
ikesyo: I'm for that way. :smiley:

Source/CarthageKit/GitHub.swift
```diff
@@ -64,55 +64,144 @@ extension GitHubError: Decodable {
...
-		let components = split(NWO, maxSplit: 1, allowEmptySlices: false) { $0 == "/" }
-		if components.count < 2 {
-			return .failure(CarthageError.ParseError(description: "invalid GitHub repository name \"\(NWO)\""))
+	/// Parses repository information out of a string of the form "owner/name"
+	/// for the github.com, or the form "http(s)://hostname/owner/name" for
+	/// Enterprise instances.
+	public static func fromIdentifier(identifier: String) -> Result<GitHubRepository, CarthageError> {
+		// GitHub.com
+		let components = split(identifier, maxSplit: 1, allowEmptySlices: false) { $0 == "/" }
+		if components.count == 2 && !components[0].hasPrefix("http") {
```
ikesyo: This is not legal if the owner starts with http. 😓

mdiep: You could use a regular expression to check for `^(\w+)/(\w+)$`—or code that checks for the equivalent.

ikesyo:
:zap:

Source/CarthageKit/GitHub.swift
```diff
@@ -64,55 +64,149 @@ extension GitHubError: Decodable {
...
 		self.name = name
 	}
 
-	/// Parses repository information out of a string of the form "owner/name".
-	public static func fromNWO(NWO: String) -> Result<GitHubRepository, CarthageError> {
-		let components = split(NWO, maxSplit: 1, allowEmptySlices: false) { $0 == "/" }
-		if components.count < 2 {
-			return .failure(CarthageError.ParseError(description: "invalid GitHub repository name \"\(NWO)\""))
+	/// Matches an identifier of the form "owner/name".
+	private static let NWORegex = NSRegularExpression(pattern: "^(\\w+)/(\\w+)$", options: nil, error: nil)!
```
ikesyo: Sorry, this can't handle `.` character.

mdiep: Oh, you're right. This should work then: `^([\.\w]+)/([\.\w]+)$`.

ikesyo: `\-` is also needed.

ikesyo:
Fixed the regular expression.

mdiep:
This looks great! Can you add an example to [the documentation](https://github.com/Carthage/Carthage/blob/bb4582858f55d0318ee842c97af72473ecfaebc9/Documentation/Artifacts.md#example-cartfile)?

ikesyo:
Oh, I've forgotten that. :sweat_smile: I updated the documentation!

mdiep:
:sparkling_heart:

ikesyo:
:confetti_ball:

PR #659, opened 2015-08-05
Add a section to the README about using dSYM files

sharplet:
The main thing I want to make sure of here is that there is enough context to understand what this feature is for.

This is just the first thing I came up with — open to any kind of feedback!

### DIFF
--- a/Documentation/Artifacts.md
+++ b/Documentation/Artifacts.md
@@ -82,7 +82,7 @@ Although the `Cartfile.resolved` file is meant to be human-readable and diffable
 
 ## Carthage/Build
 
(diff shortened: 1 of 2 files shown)


mdiep:
Looks great. :metal: Thank you!

timonus:
Is this strictly necessary or does the copy-frameworks build phase outlined [here](https://github.com/Carthage/Carthage#if-youre-building-for-ios) also provide debugging symbols? This copy phase is preventing me from being able to make release builds https://github.com/Carthage/Carthage/issues/993.

mdiep:
`dSYM`s should only be required for symbolication

PR #669, opened 2015-08-12
Make a GitHub API request failure a warning and non-fatal

### LINKED ISSUE #465: Rate limiting in binary downloads should be a warning, not an error
https://github.com/Carthage/Carthage/pull/434 fixed our authentication with the GitHub API, and added proper error reporting for it.

Unfortunately, this has the effect that API rate limiting errors are now fatal, even though binary downloads shouldn't be strictly necessary for building. This can come up in automated workflows, or CI servers, where many requests might be made in a short period of time.

I still want to log rate limiting issues, but I think they should be warnings instead of errors.

ikesyo:
Resolves #465.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -78,8 +78,10 @@ public enum ProjectEvent {
 	/// there weren't any viable binaries after all.
 	case DownloadingBinaries(ProjectIdentifier, String)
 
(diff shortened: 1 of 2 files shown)


mdiep:
Awesome! :metal:

ikesyo:
Thanks! :v:

PR #674, opened 2015-08-13
Run `git submodule sync` recursively

### LINKED ISSUE #184: Recursive submodules are not automatically synced
We should probably be running `git submodule sync` (as well as `git submodule update`) within dependency repositories, but that currently doesn't happen.

ikesyo:
`--recursive` option for `submodule sync` is introduced in [git 1.8.1](https://github.com/git/git/blob/5d417842efeafb6e109db7574196901c4e95d273/Documentation/RelNotes/1.8.1.txt#L98) (but that is documented in [2.0.2](https://github.com/git/git/blob/ebc5da3208824e25a89672a3b91bd13629b215fe/Documentation/RelNotes/2.0.2.txt#L4-L5)).

Fixes #184.


mdiep:
Nice! :sparkles:

PR #675, opened 2015-08-14
Suppress NosharedFrameworkSchemes error in order to continue building dependencies

yasuoza:
Version 0.8 introduced `.NosharedFrameworkSchemes` error to skip building dependencies which has no scheme at all.

This was OK for normal build/update command, but with `platform` argument, this make carthage stop building when a dependency has no scheme for the given platform.

To solve above issue, just suppress `.NosharedFrameworkSchemes` error while building dependencies and let carthage continue building project's dependencies.

With this PR, carthage behavior will be same as 0.6 ~ 0.7.x.

---

Steps to reproduce the issue.

Cartfile:

```
github "Alamofire/Alamofire"
github "SwiftyJSON/SwiftyJSON"
github "TransitApp/SVProgressHUD"
```

```bash
$ carthage update --platform macosx
*** Fetching SVProgressHUD
*** Fetching SwiftyJSON
*** Fetching Alamofire
*** Checking out SwiftyJSON at "2.2.1"
*** Checking out SVProgressHUD at "1.1.3"
*** Checking out Alamofire at "1.3.1"
*** xcodebuild output can be found in /var/folders/cg/v15m2zq94p3czs9kltkbngmm0000gn/T/carthage-xcodebuild.DqrfOr.log
*** Building scheme "Alamofire OSX" in Alamofire.xcworkspace
Dependency "SVProgressHUD" has no shared framework schemes

If you believe this to be an error, please file an issue with the maintainers at https://github.com/TransitApp/SVProgressHUD/issues/new
```

With this PR,

```bash
$ carthage update --platform macosx
*** Fetching SVProgressHUD
*** Fetching SwiftyJSON
*** Fetching Alamofire
*** Checking out SVProgressHUD at "1.1.3"
*** Checking out SwiftyJSON at "2.2.1"
*** Checking out Alamofire at "1.3.1"
*** xcodebuild output can be found in /var/folders/cg/v15m2zq94p3czs9kltkbngmm0000gn/T/carthage-xcodebuild.MDwKTH.log
*** Building scheme "Alamofire OSX" in Alamofire.xcworkspace
*** Building scheme "SwiftyJSON OSX" in SwiftyJSON.xcworkspace
```

PR #679, opened 2015-08-15
Warn .NoSharedFrameworkSchemes error, not to error out

ikesyo:
This improves console output for that error, and make it possible to continue building other dependencies.

This is based on the discussion in #675. /cc @yasuoza

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -104,6 +104,10 @@ public struct BuildCommand: CommandType {
 	/// Returns a producer of producers, representing each scheme being built.
 	private func buildProjectInDirectoryURL(directoryURL: NSURL, options: BuildOptions) -> SignalProducer<BuildSchemeProducer, CarthageError> {
 		let project = Project(directoryURL: directoryURL)
+
+		var eventSink = ProjectEventSink(colorOptions: options.colorOptions)
+		project.projectEvents.observe(next: { eventSink.put($0) })
+
 		let buildProducer = project.loadCombinedCartfile()
 			|> map { _ in project }
 			|> catch { error in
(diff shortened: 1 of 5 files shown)


yasuoza:
This worked for me. Thank you!

But, I'm wondering it repetitious to print

```
(\n here)
If you believe this to be an error, please file an issue with the maintainers at ~
```

even when user passed `platform` argument, because user should have known what kind of platform(s) each framework supports.

Anyway, this worked.

Source/carthage/Build.swift
```diff
@@ -128,6 +132,17 @@ public struct BuildCommand: CommandType {
...
+					switch error {
+					case let .NoSharedFrameworkSchemes(project, _):
+						// Log that building the current project is being skipped.
+						eventSink.put(.SkippedBuilding(project, error.description))
+						return .empty
+
+					default:
+						return SignalProducer(error: error)
+					}
+				}
```
mdiep: This logic is duplicated here. Can we move it inside `buildInDirectory` instead of repeating it?

mdiep:
Just the one question. Otherwise things look good. :+1:

Source/carthage/Build.swift
ikesyo: Since the error case is repacked in `buildDependencyProject()` for building each dependencies, catching it in `buildInDirectory()` level is not the idea, I think. Additionally `buildInDirectory()` does not know about `Project` nor `ProjectEventSink`, so it seems difficult to send the event from the function atm. Do you have any idea to fix it?

mdiep: Ah, okay. That makes sense. :+1:

ikesyo:
Thanks! :sparkling_heart:

Issue #419, opened 2015-04-10
Submodule checkouts interact poorly with binary downloads


jspahrsummers:
If a project uses Carthage with the `--use-submodules` option and _also_ relies upon Carthage for building, binary downloads can complicate things.

The primary problem is: if a binary is downloaded, the submodule will be removed (and incompletely, at that). This means that _whether a submodule exists_ can change with every `carthage bootstrap`.

We should do one or both of the following things:

 1. Disable binary downloads by default when using submodules
 1. Still check out submodules even if binaries are available

mdiep:
This is complicated by the current code structure, which builds all checkout out dependencies.

paulyoung:
I just ran into this and was pretty confused.

PR #682, opened 2015-08-15
Completely remove git submodules when using prebuilt binaries

ikesyo:
Addresses the comment in #419:

> if a binary is downloaded, the submodule will be removed (__and incompletely, at that__)

This clarify, by deleting `.gitmodules` entries too, the fact that submodules are removed.


mdiep:
I think the correct behavior is to leave the submodules intact. So I'm not sure if removing them completely is an improvement. :confused:

ikesyo:
Sure, I'm gonna close this and revisit later. :bow:

And about #419,

> 1. Disable binary downloads by default when using submodules

seems sufficient for me.

PR #684, opened 2015-08-16
Copy dSYMs from downloaded archives to build folders

### LINKED ISSUE #671: dSYMs from precompiled binaries are not copied to Carthage/Build/
Currently, we host private frameworks built with `carthage archive` on github to speed up build times. Since Carthage 0.8, these uploads now include `.dSYM` files. However, when performing `carthage checkout`, the `.dSYM` files included within the `.zip` archive are not added to the Carthage/Build directory (unlike those that are built from source).

cc'ing @sharplet, as he was the one that worked on this feature.

I can take a stab at fixing this, if desired :+1:

erichoracek:
Roughly behaves as follows:

- For each framework included in the archive:
- Find its architecture UUIDs with `dwarfdump --uuid`
- Iterate through the contents of the archive and locate any items with the `com.apple.xcode.dsym` type identifier
- For each dSYM, find its architecture UUIDs with `dwarfdump --uuid`
- If a dSYM is located in the archive that matches the UUIDs to those in a downloaded framework, copy it to the same destination as that framework

Fixes #671

### DIFF
--- a/Source/CarthageKit/Errors.swift
+++ b/Source/CarthageKit/Errors.swift
@@ -192,7 +192,7 @@ extension CarthageError: Printable {
 			return "Invalid architecture: \(description)"
 
 		case let .InvalidUUIDs(description):
-			return "Invalid UUIDs: \(description)"
+			return "Invalid architecture UUIDs: \(description)"
 
 		case let .MissingEnvironmentVariable(variable):
 			return "Environment variable not set: \(variable)"
--- a/Source/CarthageKit/Errors.swift
+++ b/Source/CarthageKit/Errors.swift
@@ -46,6 +46,9 @@ public enum CarthageError: Equatable {
 	// An error occurred reading a framework's architectures.
 	case InvalidArchitectures(description: String)
 
+	// An error occurred reading a dSYM or framework's UUIDs.
+	case InvalidUUIDs(description: String)
+
 	/// The project is not sharing any framework schemes, so Carthage cannot
 	/// discover them.
 	case NoSharedFrameworkSchemes(ProjectIdentifier)
@@ -187,6 +190,9 @@ extension CarthageError: Printable {
 
 		case let .InvalidArchitectures(description):
 			return "Invalid architecture: \(description)"
+
+		case let .InvalidUUIDs(description):
+			return "Invalid UUIDs: \(description)"
 
 		case let .MissingEnvironmentVariable(variable):
 			return "Environment variable not set: \(variable)"
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -305,7 +305,7 @@ public final class Project {
 			|> then(checkoutResolvedDependencies())
 	}
 
-	/// Installs binaries for the given project, if available.
+	/// Installs binaries and symbols for the given project, if available.
 	///
 	/// Sends a boolean indicating whether binaries were installed.
 	private func installBinariesForProject(project: ProjectIdentifier, atRevision revision: String) -> SignalProducer<Bool, CarthageError> {
@@ -335,6 +335,9 @@ public final class Project {
 						|> flatMap(.Concat) { directoryURL in
 							return frameworksInDirectory(directoryURL)
 								|> flatMap(.Merge, self.copyFrameworkToBuildFolder)
+								|> flatMap(.Merge) { frameworkURL in
+									return self.copyDSYMToBuildFolderForFramework(frameworkURL, fromDirectoryURL: directoryURL)
+								}
 								|> on(completed: {
 									_ = NSFileManager.defaultManager().trashItemAtURL(checkoutDirectoryURL, resultingItemURL: nil, error: nil)
 								})
@@ -357,8 +360,8 @@ public final class Project {
 			}
 	}
 
-	/// Downloads any binaries that may be able to be used instead of a
-	/// repository checkout.
+	/// Downloads any binaries and symbols that may be able to be used instead 
+	/// of a repository checkout.
 	///
 	/// Sends the URL to each downloaded zip, after it has been moved to a
 	/// less temporary location.
@@ -409,7 +412,26 @@ public final class Project {
 			|> map { sdk in sdk.platform }
 			|> map { platform in self.directoryURL.URLByAppendingPathComponent(platform.relativePath, isDirectory: true) }
 			|> map { platformFolderURL in platformFolderURL.URLByAppendingPathComponent(frameworkURL.lastPathComponent!) }
-			|> flatMap(.Merge) { destinationFrameworkURL in copyFramework(frameworkURL, destinationFrameworkURL.URLByResolvingSymlinksInPath!) }
+			|> flatMap(.Merge) { destinationFrameworkURL in copyProduct(frameworkURL, destinationFrameworkURL.URLByResolvingSymlinksInPath!) }
+	}
+
+	/// Copies the DSYM matching the given framework and contained within the
+	/// given directory URL to the directory that the framework resides within.
+	///
+	/// If no dSYM is found for the given framework, completes with no values.
+	///
+	/// Sends the URL of the dSYM after copying.
+	public func copyDSYMToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+		return dSYMForFramework(frameworkURL, inDirectoryURL:directoryURL)
+			|> map { dSYMURL in
+				let destinationDirectoryURL = frameworkURL.URLByDeletingLastPathComponent!
+				let fileName = dSYMURL.lastPathComponent!
+				let destinationURL = destinationDirectoryURL.URLByAppendingPathComponent(fileName)
+				let resolvedDestinationURL = destinationURL.URLByResolvingSymlinksInPath!
+
+				return (dSYMURL, resolvedDestinationURL)
+			}
+			|> flatMap(.Merge) { (from, to) in copyProduct(from, to) }
 	}
 
 	/// Checks out the given project into its intended working directory,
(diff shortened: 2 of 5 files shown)


mdiep:
Do you think it's worth doing all that work? Or would it be better to just copy all the files from the binary release?

Source/CarthageKit/Errors.swift
```diff
@@ -188,6 +191,9 @@ extension CarthageError: Printable {
 		case let .InvalidArchitectures(description):
 			return "Invalid architecture: \(description)"
 
+		case let .InvalidUUIDs(description):
+			return "Invalid UUIDs: \(description)"
```
mdiep: This error isn't at all obvious. It'd be nice if it could mention a specific project and mention something about binary releases and which file has problems.

erichoracek:
I went down that route initially, but unfortunately you don't know for sure which directory to place the dSYM in if there's multiple dSYMs included in the archive (e.g. `ReactiveCocoa.framework` and `ReactiveCocoa.dSYM` for Mac & iOS). As far as I can tell, there's no guarantee that the directory structure of an archive will mirror that of the `Carthage/` directory. As such, you can't assume that a dSYM will be properly placed in a directory that matches the architecture without ensuring that they match at the UUID level.

Additionally, by performing this check, we ensure only valid dSYMs are copied to the Carthage directory. If an invalid dSYM is included in the archive, its architecture UUIDs will not match those of the included framework, and it will not be copied over.

Last of all, a very similar check is already in place for frameworks included in archives. A framework is not copied to the Build/Mac or Build/iOS folders unless it actually contains that architecture—it is _not_ inferred from the directory structure of the archive. I think it makes sense to follow this established convention.

sharplet:
If you use `carthage archive`, the relative paths into the `Carthage/Build/*` folder are preserved. It would definitely be simpler to just copy any binaries into the project, preserving the paths. This code is definitely more general, in a "be liberal in what you accept" kind of way.

erichoracek:
Agreed that it would be simpler, but it would break down if down the line precompiled binaries not created by `carthage archive` were able to be specified in the Cartfile, e.g. Crashlytics:

```
archive 'https://kit-downloads.fabric.io/ios/com.twitter.crashlytics.ios/3.1.0/com.twitter.crashlytics.ios-default.zip`
```

Source/CarthageKit/Errors.swift
erichoracek: The description printed is specified at https://github.com/Carthage/Carthage/pull/684/files#diff-f66644e8cfbcad3c3826395a690ddfbcR1307 — I'll beef it up a little.

mdiep:
> Last of all, a very similar check is already in place for frameworks included in archives. A framework is not copied to the Build/Mac or Build/iOS folders unless it actually contains that architecture—it is not inferred from the directory structure of the archive. I think it makes sense to follow this established convention.

Ah, okay. I didn't realize that. I'll give this another look unless @ikesyo beats me to it. :+1:

Source/CarthageKit/Project.swift
```diff
@@ -305,7 +305,7 @@ public final class Project {
 			|> then(checkoutResolvedDependencies())
 	}
 
-	/// Installs binaries for the given project, if available.
+	/// Installs binaries and symbols for the given project, if available.
```
ikesyo: “debug symbols” would be better?

Source/CarthageKit/Project.swift
```diff
@@ -357,8 +360,8 @@ public final class Project {
 			}
 	}
 
-	/// Downloads any binaries that may be able to be used instead of a
-	/// repository checkout.
+	/// Downloads any binaries and symbols that may be able to be used instead
```
ikesyo: As above.

Source/CarthageKit/Project.swift
```diff
@@ -409,7 +412,26 @@ public final class Project {
...
+
+	/// Copies the DSYM matching the given framework and contained within the
+	/// given directory URL to the directory that the framework resides within.
+	///
+	/// If no dSYM is found for the given framework, completes with no values.
+	///
+	/// Sends the URL of the dSYM after copying.
+	public func copyDSYMToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+		return dSYMForFramework(frameworkURL, inDirectoryURL:directoryURL)
+			|> map { dSYMURL in
```
ikesyo: The logic in the `map()` closure could be in the following `flatMap()`.

Source/CarthageKit/Project.swift
```diff
@@ -534,22 +556,39 @@ private func cacheDownloadedBinary(downloadURL: NSURL, toURL cachedURL: NSURL) -
 		}
 }
 
-/// Sends the URL to each framework bundle found in the given directory.
-private func frameworksInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+extension NSURL {
+	// The type identifier of the receiver, or an error if it was unable to be
+	// determined.
+	internal func typeIdentifier() -> Result<String, NSError> {
```
ikesyo: This should be `private` since used only in the file.

ikesyo: We can use computed property here.

Source/CarthageKit/Project.swift
```diff
@@ -534,22 +556,39 @@ private func cacheDownloadedBinary(downloadURL: NSURL, toURL cachedURL: NSURL) -
...
+		return try({ (error: NSErrorPointer) -> String? in
```
ikesyo: We can omit the parentheses and use a trailing closure as: `try { error in ... }`

Source/CarthageKit/Project.swift
```diff
@@ -534,22 +556,39 @@ private func cacheDownloadedBinary(downloadURL: NSURL, toURL cachedURL: NSURL) -
...
-			var typeIdentifier: AnyObject?
-			if URL.getResourceValue(&typeIdentifier, forKey: NSURLTypeIdentifierKey, error: nil) {
-				if let typeIdentifier: AnyObject = typeIdentifier {
-					if UTTypeConformsTo(typeIdentifier as! String, kUTTypeFramework) != 0 {
-						return true
-					}
-				}
+			switch URL.typeIdentifier() {
+			case let .Success(box):
+				return UTTypeConformsTo(box.value, typeIdentifier) != 0
```
ikesyo: We usualy insert a blank line between each cases.

Source/CarthageKit/Project.swift
```diff
@@ -559,6 +598,25 @@ private func frameworksInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL,
...
+/// Sends the URL to each dSYM found in the given directory
+private func dSYMsInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+	return filesInDirectory(directoryURL, "com.apple.xcode.dsym")
+}
+
+/// Sends the URL of the dSYM whose UUIDs match those of the given framework, or
+/// errors if there was an error parsing a dSYM contained within the directory.
+private func dSYMForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+	return combineLatest(UUIDsForFramework(frameworkURL), dSYMsInDirectory(directoryURL))
+		|> flatMap(.Merge) { (frameworkUUIDs, dSYMURL) in
```
ikesyo: We can omit the parameter list's parentheses.

Source/CarthageKit/Xcode.swift
```diff
@@ -1250,6 +1250,64 @@ public func architecturesInFramework(frameworkURL: NSURL) -> SignalProducer<Stri
 		}
 }
 
+/// Sends a set of UUIDs for each architecture present in the given framework.
+public func UUIDsForFramework(frameworkURL: NSURL) -> SignalProducer<Set<NSUUID>, CarthageError> {
+	return SignalProducer.try { () -> Result<NSURL, CarthageError> in
+			return binaryURL(frameworkURL)
+		}
+		|> flatMap(.Merge) { URL in UUIDsFromDwarfdump(URL) }
```
ikesyo: We can just pass the function as `flatMap(.Merge, UUIDsFromDwarfdump)`.

Source/CarthageKit/Xcode.swift
```diff
@@ -1250,6 +1250,64 @@ public func architecturesInFramework(frameworkURL: NSURL) -> SignalProducer<Stri
...
+}
+
+/// Sends a set of UUIDs for each architecture present in the given dSYM.
+public func UUIDsForDSYM(dSYMURL: NSURL) -> SignalProducer<Set<NSUUID>, CarthageError> {
+	return UUIDsFromDwarfdump(dSYMURL)
+}
+
+/// Sends a set of UUIDs for each architecture present in the given URL.
+private func UUIDsFromDwarfdump(URL: NSURL) -> SignalProducer<Set<NSUUID>, CarthageError> {
```
ikesyo: The parameter should be named the same as `UUIDsForDSYM()`?

Source/CarthageKit/Xcode.swift
```diff
@@ -1250,6 +1250,64 @@ public func architecturesInFramework(frameworkURL: NSURL) -> SignalProducer<Stri
...
+	let dwarfdumpTask = TaskDescription(launchPath: "/usr/bin/xcrun", arguments: [ "dwarfdump", "--uuid", URL.path!])
```
ikesyo: Could you add a space before `]`?

Source/CarthageKit/Xcode.swift
```diff
@@ -1250,6 +1250,64 @@ public func architecturesInFramework(frameworkURL: NSURL) -> SignalProducer<Stri
...
+
+				if let UUIDString = UUIDString as? String, let UUID = NSUUID(UUIDString: UUIDString) {
+					UUIDs.insert(UUID)
+				}
+
+				// Scan until a newline or end of file.
+				scanner.scanUpToCharactersFromSet(NSCharacterSet.newlineCharacterSet(), intoString: nil)
+			}
+
+			if UUIDs.count > 0 {
```
ikesyo: IMO `isEmpty` is more appropriate:

```swift
if UUIDs.isEmpty {
    return SignalProducer(error: .InvalidUUIDs(description: "Could not parse UUIDs using dwarfdump from \(URL.path!)"))
} else {
    return SignalProducer(value: UUIDs)
}
```

ikesyo:
Sorry for the nitpickings, it looks solid and great to me other than them! :smiley:

erichoracek:
Thanks for the review! Should be good to go—let me know if you have any additional feedback :smile:

Source/CarthageKit/Project.swift
```diff
@@ -559,6 +598,25 @@ private func frameworksInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL,
...
 
+/// Sends the URL to each dSYM found in the given directory
+private func dSYMsInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+	return filesInDirectory(directoryURL, "com.apple.xcode.dsym")
+}
+
+/// Sends the URL of the dSYM whose UUIDs match those of the given framework, or
+/// errors if there was an error parsing a dSYM contained within the directory.
+private func dSYMForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+	return combineLatest(UUIDsForFramework(frameworkURL), dSYMsInDirectory(directoryURL))
```
ikesyo: Always the last value of `dSYMsInDirectory` is used as `dSYMURL` (in my case it is a dSYM for Mac) because of the following reasons:

- `combineLatest` is used
- `UUIDsForFramework(frameworkURL)` takes some time
- `dSYMsInDirectory` sends all values in the meantime

As a result, coping dSYM for iOS always fails. :cry:

ikesyo: This worked:

```swift
return UUIDsForFramework(frameworkURL)
	|> flatMap(.Concat) { frameworkUUIDs in
		return dSYMsInDirectory(directoryURL)
			|> flatMap(.Merge) { dSYMURL in
				return UUIDsForDSYM(dSYMURL)
					|> filter { dSYMUUIDs in
						return dSYMUUIDs == frameworkUUIDs
					}
					|> map { _ in dSYMURL }
			}
	}
	|> take(1)
```

erichoracek: Good catch! I only tested it on an archive that had one dSYM, not two. The issue and fix both make sense—I'll update the diff.

ikesyo:
Thank you for the fixes! :shipit:

PR #686, opened 2015-08-17
Implement --no-checkout flag for bootstrap/update commands

### LINKED ISSUE #466: Support a --no-checkout flag for bootstrap/update?
I've previously been (in a thread I can't find) against adding another command that can resolve dependencies without checking them out. I still am against adding a command besides `update`, but it occurs to me that we could add a flag (disabled by default) that skips the checkout step.

This could be useful in cases where you want to perform some sort of custom checkout logic, a la the [`carthage-developer-checkouts` workflow](https://github.com/Carthage/workflows/blob/142231860c72b99a965af73dc649fdf330dcdf20/carthage-developer-checkouts).

ikesyo:
Resolves #466.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -295,14 +295,14 @@ public final class Project {
 	}
 
 	/// Updates the dependencies of the project to the latest version. The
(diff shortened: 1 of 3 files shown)


mdiep:
:metal:

PR #692, opened 2015-08-19
Disable binary downloads when using submodules

### LINKED ISSUE #419: Submodule checkouts interact poorly with binary downloads

ikesyo:
Fixes #419:

> 1. Disable binary downloads by default when using submodules

I'm not sure if f448fc5 is needed and d4fbb78 is sufficient.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -498,12 +498,21 @@ public final class Project {
 						let project = dependency.project
 						let revision = dependency.version.commitish
 
+						let submoduleFound = submodulesByPath[project.relativePath] != nil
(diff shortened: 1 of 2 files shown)


mdiep:
This looks good to me, but I'm curious if anyone else in @Carthage/carthage has an opinion about the behavior.

robb:
LGTM

mdiep:
:metal:

ikesyo:
:v:

PR #701, opened 2015-08-21
Revert 42ccb03

ikesyo:
From the conversation in the Carthage Slack channel, because embedding frameworks is handled by `carthage copy-frameworks`, using “Linked Frameworks and Libraries” section is appropriate for iOS application targets. I'm sorry for that confusion. :bow: 

See #681.


mdiep:
:smile:

PR #703, opened 2015-08-22
Safer optional chaining, fix for somewhat dangerous if-let  

Nirma:
This approach may be safer than what was previously written.


ikesyo:
Seems good, thank you for the contribution! :+1:

PR #707, opened 2015-08-24
Work around an indeterministic NSFileWriteFileExistsError when downloading binaries

### LINKED ISSUE #591: Race during bootstrap
We're seeing a race while bootstrapping project:

```
Failed to write to Carthage/Build/Mac: Error Domain=NSCocoaErrorDomain Code=516 "The file ���Mac��� couldn���t be saved in the folder ���Build��� because a file with the same name already exists." UserInfo=0x7fe07c145180 {NSFilePath=Carthage/Build/Mac, NSUnderlyingError=0x7fe07c144b90 "The operation couldn���t be completed. File exists"}
```

It seems like it should be fine if those directories already exist.

ikesyo:
Might fix #591, although I can't reproduce the issue and don't confirm that is fixed at my local machine. :astonished:

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -1159,7 +1159,15 @@ public func copyProduct(from: NSURL, to: NSURL) -> SignalProducer<NSURL, Carthag
 
 		let manager = NSFileManager.defaultManager()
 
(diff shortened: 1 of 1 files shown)


mdiep:
I don't like that we have to do this :disappointed:, but it's worth a shot.

PR #709, opened 2015-08-24
cleaner implementation of stripGitSuffix function

Nirma:
This may be a more simple way to achieve this function's goal of removing potential ".git" suffixes.

Hope it helps!

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -69,8 +69,7 @@ public struct GitURL: Equatable {
 	/// Strips any trailing .git in the given name, if one exists.
 	private func stripGitSuffix(string: String) -> String {
 		if string.hasSuffix(".git") {
-			let nsString = string as NSString
-			return nsString.substringToIndex(nsString.length - 4) as String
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Git.swift
```diff
@@ -68,12 +68,7 @@ public struct GitURL: Equatable {
 
 	/// Strips any trailing .git in the given name, if one exists.
 	private func stripGitSuffix(string: String) -> String {
-		if string.hasSuffix(".git") {
-			let nsString = string as NSString
-			return nsString.substringToIndex(nsString.length - 4) as String
-		} else {
-			return string
-		}
+		return string.stringByReplacingOccurrencesOfString(".git", withString: "")
```
mdiep: This will replace all occurrences of `.git`, which is definitely not what we want. I could have a repo named `i.love.github`, and this would change it to `i.lovehub`.

Nirma:
@mdiep Thank you so much for catching that!
I just updated it if you don't mind having a look again.

mdiep:
Looks good, thanks! :sparkles:

Issue #711, opened 2015-08-25
Carthage fails when home directory is not on startup volume


simX:
I recently upgraded to OS X 10.10, and I'm using Carthage 0.8.0, but my home directory is on my old partition.  So my home folder is located at `/Volumes/DifferentPartition/Users/simx/` .  My git repository is also on this other partition, as in `/Volumes/DifferentPartition/Users/simx/Development/SekritProject/`

In such a situation, carthage eventually fails during the downloading phase.  Here's the output of `carthage update --verbose`:

*** Cloning KIF
*** Cloning HockeySDK-iOS
*** Cloning stripe-ios
*** Cloning libPhoneNumber-iOS
*** Cloning ios-i18n
*** Cloning mixpanel-iphone
*** Cloning objc-TimesSquare
*** Cloning sskeychain
*** Cloning AFNetworking
*** Checking out stripe-ios at "v5.1.2"
*** Checking out mixpanel-iphone at "3c77e98c93422f99cc3506ddd23669691e6b5e03"
*** Downloading sskeychain at "Version 1.2.3"
*** Checking out libPhoneNumber-iOS at "0.8.7"
*** Checking out objc-TimesSquare at "978628c39fabc6c0c060536782f87a6c84012fb1"
*** Checking out ios-i18n at "v1.0.8"
*** Checking out AFNetworking at "eedc564b46cde9ae285d60fc393016f578258169"
*** Downloading HockeySDK-iOS at "Version 3.7.2"
*** Checking out HockeySDK-iOS at "3.7.2"
*** Checking out KIF at "v3.0.8"
*** Downloading sskeychain at "Version 1.2.3"
Error Domain=NSPOSIXErrorDomain Code=18 "The operation couldn’t be completed. Cross-device link"

simX:
Hm, maybe false alarm, this seems related to the soffes/sskeychain repo

ikesyo:
This is dupe of #706.

ikesyo:
Closed by #713.

pomozoff:
I've faced this issue:

```
[...]

PR #713, opened 2015-08-25
Use NSFileManager.moveItemAtURL() as a fallback of rename() system call when EXDEV error occurred

### LINKED ISSUE #706: "cross-device link" error on `carthage update` for some sources
When attempting to add `github "ReactiveCocoa/ReactiveCocoa"` to my project as a carthage dependency, running `carthage update` yields the following output:
```
*** Fetching ReactiveCocoa
*** Fetching Result
*** Fetching Box
*** Downloading ReactiveCocoa at "v3.0-beta.9"
*** Downloading Box at "1.2.2"
*** Checking out Result at "0.4.4"
Error Domain=NSPOSIXErrorDomain Code=18 "The operation couldn’t be completed. Cross-device link"
```
My system has home directories and the OS X installation on different hard drives.  This error does not appear to occur for all dependencies — `github "Alamofire/Alamofire"` seems to work fine, if I remove the ReactiveCocoa line from the Cartfile.

ikesyo:
Should fix #706 and #711.

The [manpage of `rename()`](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man2/rename.2.html) says:

> Both old and new must be of the same type (that is, both must be either directories or non-directories) and __must reside on the same file system__.

On the other hand, the documentation of `NSFileManager.moveItemAtURL()` says:

> If the source and destination of the move operation are not on the same volume, this method copies the item first and then removes it from its current location.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -573,10 +573,11 @@ private func cacheDownloadedBinary(downloadURL: NSURL, toURL cachedURL: NSURL) -
 			}
 		}
 		|> try { newDownloadURL in
-			if rename(downloadURL.fileSystemRepresentation, newDownloadURL.fileSystemRepresentation) == 0 {
+			var error: NSError?
+			if NSFileManager.defaultManager().moveItemAtURL(downloadURL, toURL: newDownloadURL, error: &error) {
 				return .success(())
 			} else {
-				return .failure(.TaskError(.POSIXError(errno)))
+				return .failure(.WriteFailed(newDownloadURL, error))
 			}
 		}
 }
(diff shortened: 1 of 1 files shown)


ikesyo:
Originally `moveItemAtURL()` was replaced by `rename()` in 48e480d422f232b748131151d0dc5ccb63a3b5f1. Going back to `moveItemAtURL()` is acceptable, or should we try `rename()` at first then fall back to `moveItemAtURL()` when `EXDEV` error occurred?

/cc @jspahrsummers

jashephe:
Hi — just wanted to check in:  is `NSFileManager.moveItemAtURL()` an acceptable fix, or is there a reason why `rename()` has to be used?

jspahrsummers:
rename() is atomic, and also replaces the destination if need be.

NSFileManager offers neither of these guarantees, unfortunately.

jspahrsummers:
Checking for EXDEV is a good idea. 👍

ikesyo:
Thank you for your inputs! :pray: I've updated the PR.

mdiep:
Looks good to me!

PR #714, opened 2015-08-25
remove needless optional cast

Nirma:
This may be a slightly more stable way to phrase this section.

Hope it helps!


mdiep:
Thanks for opening a pull request!

Unfortunately, I think forced unwrapping is appropriate in this case. That information should always be present. If it's not, it's appropriate for `carthage` to crash.

Nirma:
@mdiep Ok I completely see the value of that.
In that case if we used a forced cast it might be indicative of our intention.

I made a small change, would you mind taking a look again?

Thanks!

mdiep:
That's great—thanks!

PR #719, opened 2015-08-27
More concise phrasing for ProjectLocator description.

Nirma:
Thought this might be a simpler way to achieve the same thing. 

Hope it helps!


mdiep:
This is probably a relic of an older version of Swift. Thanks! :smile:

PR #720, opened 2015-08-27
Replace nested if statement with where clause

Nirma:
I tried to simplify this part to one if-let with a where clause testing for the same thing the outer if did.

Hope it helps!


mdiep:
Looks good, thanks!

PR #726, opened 2015-08-31
Accept multiple platforms in BuildOptions

### LINKED ISSUE #592: Multiple platforms
It would be great if there was a way to only build for a set of platforms (i.e. iOS & watchOS but not OS X). Even better, it would be great if the default was `--platform auto` and it looked at your project and figured out which platforms you needed.

ikesyo:
Resolves #592.

This allows `BuildOptions` to take comma-separated values for the `--platform` option as follows: `--platform ios,mac`

This is a breaking change to CarthageKit's public APIs.

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -224,6 +224,7 @@ public enum BuildPlatform: Equatable {
 	/// Build only for watchOS.
 	case watchOS
 
+	/// Build for multiple platforms within the list.
 	case Multiple([BuildPlatform])
 
 	/// The set of `Platform` corresponding to this setting.
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -202,7 +202,7 @@ public struct BuildOptions: OptionsType {
 	public static func evaluate(m: CommandMode) -> Result<BuildOptions, CommandantError<CarthageError>> {
 		return create
 			<*> m <| Option(key: "configuration", defaultValue: "Release", usage: "the Xcode configuration to build")
-			<*> m <| Option(key: "platform", defaultValue: .All, usage: "the platform to build for (one of ‘all’, ‘Mac’, or ‘iOS’)")
+			<*> m <| Option(key: "platform", defaultValue: .All, usage: "the platforms to build for (one of ‘all’, ‘Mac’, ‘iOS’, ‘watchOS’ or comma-separated values of the formers except for ‘all’)")
 			<*> m <| Option(key: "skip-current", defaultValue: true, usage: "don't skip building the Carthage project (in addition to its dependencies)")
 			<*> ColorOptions.evaluate(m)
 			<*> m <| Option(key: "verbose", defaultValue: false, usage: "print xcodebuild output inline")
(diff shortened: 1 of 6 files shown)


Source/carthage/Build.swift
```diff
@@ -224,24 +224,44 @@ public enum BuildPlatform {
 	/// Build only for watchOS.
 	case watchOS
 
-	/// The `Platform` corresponding to this setting.
-	public var platform: Platform? {
+	case Multiple([BuildPlatform])
```
mdiep: It seems weird to me to make this a `case` instead of using `[BuildPlatform]` in the caller. Is there a reason we can't do that?

mdiep:
This looks good. :+1: Just the one question.

Source/carthage/Build.swift
ikesyo: It seems that `ArgumentType` cannot handle array type, so we need a wrapper type something like:

```swift
public struct BuildPlatformList: ArgumentType {
	let buildPlatforms: [BuildPlatform]

	public static let name = "platform"

	public static func fromString(string: String) -> BuildPlatformList? {
		let commaSeparated = split(string, allowEmptySlices: false) { $0 == "," }

		let findBuildPlatform: String -> BuildPlatform? = { string in
			for (key, platform) in BuildPlatform.acceptedStrings {
				if string.caseInsensitiveCompare(key) == NSComparisonResult.OrderedSame {
					return platform
				}
			}
			return nil
		}

		switch commaSeparated.count {
		case 0:
			return nil

		case 1:
			return findBuildPlatform(commaSeparated[0]).map { BuildPlatformList(buildPlatforms: [ $0 ]) }

		default:
			var buildPlatforms = [BuildPlatform]()
			for string in commaSeparated {
				if let found = findBuildPlatform(string) where found != .All {
					buildPlatforms.append(found)
				} else {
					// Reject if an invalid value is included in the comma-
					// separated string.
					return nil
				}
			}
			return BuildPlatformList(buildPlatforms: buildPlatforms)
		}
	}
}
```

Is this the right direction?

ikesyo: I tried to implement that in 8d7d658.

mdiep: > It seems that `ArgumentType` cannot handle array type, so we need a wrapper type

Ah, okay! That seems like a good reason to use your original implementation. Sorry!

ikesyo: Thanks for the response. :bow: I'll revert the last commit soon.

ikesyo:
Done! :zap:

mdiep:
:sparkles:

VinceBurn:
first: thanks for this addition.

Could this be documented in the man page of Carthage, because the format is very specific. 
e.g. 
1 carthage update --platform iOS, watchOS
2 carthage update --platform "iOS, watchOS"
both don't work and the error message we get is not helping.

I actually had to found this thread to discover what was my problem.

I would suggest to
+ update man page, or
+ have a more forgiving parser (I'm not expecting 1. to work, but I would expect 2. to work

mdiep:
I opened #957 to cover (2).

PR #747, opened 2015-09-11
Strip .git suffix from GitHub repository name as GitURL does

ikesyo:
Addresses https://github.com/Carthage/Carthage/issues/745#issuecomment-139491947.

The following is the current output before this PR with the input `github "https://github.com/danielgindi/ios-charts.git"` or `github "danielgindi/ios-charts.git"`:

```sh
*** Cloning ios-charts.git
A shell task failed with exit code 128:
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -136,7 +136,7 @@ public struct GitHubRepository: Equatable {
 	public init(server: Server = .GitHub, owner: String, name: String) {
 		self.server = server
 		self.owner = owner
-		self.name = name
+		self.name = stripGitSuffix(name)
 	}
 
 	/// Matches an identifier of the form "owner/name".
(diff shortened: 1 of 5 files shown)


Source/CarthageKit/GitHub.swift
```diff
@@ -136,7 +136,7 @@ public struct GitHubRepository: Equatable {
 	public init(server: Server = .GitHub, owner: String, name: String) {
 		self.server = server
 		self.owner = owner
-		self.name = name
+		self.name = stripGitSuffix(name)
```
mdiep: I think it'd be better to call `stripGitSuffix` in the caller.

ikesyo: Thank you, I was unsure which is better.

ikesyo:
Done! :zap:

mdiep:
Looks great! :zap:

PR #752, opened 2015-09-13
Handle nonexistent directory in `isGitRepository`

bencochran:
After upgrading to 0.9, I started getting crashes like:

> *** Cloning SwiftCheck
*** Checking out SwiftCheck at "v0.3.0"
2015-09-13 03:05:31.289 carthage[56929:3098569] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'working directory doesn't exist.'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff9107d03c __exceptionPreprocess + 172
	1   libobjc.A.dylib                     0x00007fff9234176e objc_exception_throw + 43
	2   CoreFoundation                      0x00007fff9107ceed +[NSException raise:format:] + 205
	3   Foundation                          0x00007fff93eb71d6 -[NSConcreteTask launchWithDictionary:] + 582
	4   ReactiveTask                        0x0000000107769a45 _TFFFF12ReactiveTask10launchTaskFVS_15TaskDescriptionGV13ReactiveCocoa14SignalProducerGOS_9TaskEventCSo6NSData_OS_17ReactiveTaskError_U_FTGVSs6SinkOfGOS1_5EventGS3_S4__S5___CS1_19CompositeDisposable_T_U0_FTCS_P33_8622FA0C3FE9071537A42FE3BB905E344PipeS9__GS2_GS3_S4__S5__U_FTGS6_GS7_GS3_S4__S5___S8__T_ + 1253
	5   ReactiveTask                        0x000000010776348b _TPA__TFFFF12ReactiveTask10launchTaskFVS_15TaskDescriptionGV13ReactiveCocoa14SignalProducerGOS_9TaskEventCSo6NSData_OS_17ReactiveTaskError_U_FTGVSs6SinkOfGOS1_5EventGS3_S4__S5___CS1_19CompositeDisposable_T_U0_FTCS_P33_8622FA0C3FE9071537A42FE3BB905E344PipeS9__GS2_GS3_S4__S5__U_FTGS6_GS7_GS3_S4__S5___S8__T_ + 251

Tracked it down to 561c0851ffe41594350138bb7c09cd2543913ddf which left out a check for directory existence in `isGitRepository` before attempting to launch a git task in it.

So, here’s that.

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -394,6 +394,10 @@ public func resolveReferenceInRepository(repositoryFileURL: NSURL, reference: St
 /// Attempts to determine whether the given directory represents a Git
 /// repository.
 public func isGitRepository(directoryURL: NSURL) -> SignalProducer<Bool, NoError> {
+	if !NSFileManager.defaultManager().fileExistsAtPath(directoryURL.path!) {
+		return SignalProducer(value: false)
+	}
+	
 	return launchGitTask([ "rev-parse", "--git-dir", ], repositoryFileURL: directoryURL)
 		|> map { _ in true }
 		|> catch { _ in SignalProducer(value: false) }
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -397,7 +397,7 @@ public func isGitRepository(directoryURL: NSURL) -> SignalProducer<Bool, NoError
 	if !NSFileManager.defaultManager().fileExistsAtPath(directoryURL.path!) {
 		return SignalProducer(value: false)
 	}
-	
+
 	return launchGitTask([ "rev-parse", "--git-dir", ], repositoryFileURL: directoryURL)
 		|> map { _ in true }
 		|> catch { _ in SignalProducer(value: false) }


Source/CarthageKit/Git.swift
```diff
@@ -394,6 +394,10 @@ public func resolveReferenceInRepository(repositoryFileURL: NSURL, reference: St
 /// Attempts to determine whether the given directory represents a Git
 /// repository.
 public func isGitRepository(directoryURL: NSURL) -> SignalProducer<Bool, NoError> {
+	if !NSFileManager.defaultManager().fileExistsAtPath(directoryURL.path!) {
+		return SignalProducer(value: false)
+	}
+
 	return launchGitTask([ "rev-parse", "--git-dir", ], repositoryFileURL: directoryURL)
 		|> map { _ in true }
 		|> catch { _ in SignalProducer(value: false) }
```
ikesyo: That is weird, the error should be already handled here.

ikesyo:
Ah, I'm sorry, that might happen when

1. you specify `--use-submodules` option.
2. that is a first attempt of `carthage update/bootstrap/checkout` and `Carthage/Checkouts/reponame` does not exist.

(not confirmed yet).

Source/CarthageKit/Git.swift
```diff
@@ -394,6 +394,10 @@ public func resolveReferenceInRepository(repositoryFileURL: NSURL, reference: St
 /// Attempts to determine whether the given directory represents a Git
 /// repository.
 public func isGitRepository(directoryURL: NSURL) -> SignalProducer<Bool, NoError> {
+	if !NSFileManager.defaultManager().fileExistsAtPath(directoryURL.path!) {
+		return SignalProducer(value: false)
+	}
+
```
ikesyo: Instead of adding the check here, we should do it in [`addSubmoduleToRepository()`](https://github.com/Carthage/Carthage/blob/561c0851ffe41594350138bb7c09cd2543913ddf/Source/CarthageKit/Git.swift#L413) I think, because the other use cases are ensured that the given URL exists ([here](https://github.com/Carthage/Carthage/blob/561c0851ffe41594350138bb7c09cd2543913ddf/Source/CarthageKit/Git.swift#L453-L464) and [here](https://github.com/Carthage/Carthage/blob/561c0851ffe41594350138bb7c09cd2543913ddf/Source/CarthageKit/Project.swift#L712-L717)).

mdiep: I actually like it here. I'd rather make callers check that it doesn't exist if they care.

mdiep:
I'm going to go ahead and merge this, so I can release 0.9.1 with the fix. Thanks for the pull request!

ikesyo:
Thanks @bencochran and @mdiep

PR #760, opened 2015-09-16
Fix precompiled framework platform determination

### LINKED ISSUE #757: Precompiled watchOS binaries result in crashes during checkout
Currently, the logic that copies binaries from an archive does not properly handle watchOS binaries, instead crashing when it encounters one with failures during the copy. It currently attempts to infer its platform from the architecture, but this fails as both watchOS and iOS architectures have a prefix of `arm`. See https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/Project.swift#L406-L414

This currently affects the [ReactiveCocoa 4.0 alpha](https://github.com/ReactiveCocoa/ReactiveCocoa/releases/tag/v4.0-alpha.1), but would likely affect any precompiled binaries that include both watchOS and iOS binaries.

cc @neilpa @NachoSoto

erichoracek:
Currently platform determination of precompiled binaries depends on the architecture of the framework. However, with the addition of watchOS frameworks, this behavior is now incorrect, as both watchOS and iOS frameworks share the `arm` prefix on their architectures, resulting in the behavior documented in #757. To remedy this, instead of relying on architecture, we instead perform the following:

- For each downloaded framework, locate the included Info.plist (location differs in Mac OSX vs. iOS frameworks)
- Within the Info.plist, search for the DTSDKName key
- Strip the version numbers from the SDK name, e.g. `macosx10.10` → `macosx`
- Move the framework to the folder that matches the platform name, as before

Fixes #757

I've tested this with RAC 3.0 and the 4.0 alpha. Let me know if there's any additional cases worth looking into.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -639,8 +639,8 @@ private func platformForInfoPlist(plistURL: NSURL) -> SignalProducer<Platform, C
 		}
 		// Neither DTPlatformName nor CFBundleSupportedPlatforms can not be used 
 		// because Xcode 6 and below do not include either in Mac OSX frameworks.
-		|> tryMap { propertyList -> Result<String, CarthageError> in
-			if let sdkName = propertyList["DTSDKName"] as? String {
+		|> tryMap { plist -> Result<String, CarthageError> in
+			if let sdkName = plist["DTSDKName"] as? String {
 				return .success(sdkName)
 			}
 			return .failure(.InfoPlistParseFailed(plistURL: plistURL, reason: "the value for the DTSDKName key is not a string"))
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -408,12 +408,9 @@ public final class Project {
 	///
 	/// Sends the URL to the framework after copying.
 	private func copyFrameworkToBuildFolder(frameworkURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
-		return architecturesInFramework(frameworkURL)
-			|> filter { arch in arch.hasPrefix("arm") }
-			|> map { _ in SDK.iPhoneOS }
-			|> concat(SignalProducer(value: SDK.MacOSX))
+		return infoPlistForFramework(frameworkURL)
+			|> flatMap(.Merge) { infoPlistURL in platformsForInfoPlist(infoPlistURL) }
 			|> take(1)
-			|> map { sdk in sdk.platform }
 			|> map { platform in self.directoryURL.URLByAppendingPathComponent(platform.relativePath, isDirectory: true) }
 			|> map { platformFolderURL in platformFolderURL.URLByAppendingPathComponent(frameworkURL.lastPathComponent!) }
 			|> flatMap(.Merge) { destinationFrameworkURL in copyProduct(frameworkURL, destinationFrameworkURL.URLByResolvingSymlinksInPath!) }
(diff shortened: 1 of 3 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -1283,7 +1283,7 @@ private func stripArchitecture(frameworkURL: NSURL, architecture: String) -> Sig
 }
 
 /// Returns a signal of all architectures present in a given framework.
-public func architecturesInFramework(frameworkURL: NSURL) -> SignalProducer<String, CarthageError> {
+private func architecturesInFramework(frameworkURL: NSURL) -> SignalProducer<String, CarthageError> {
```
ikesyo: This is also used in `XcodeSpec.swift`, and might be used publically because this is a public API of `CarthageKit.framework`. So we should preserve this as public, I think.

Source/CarthageKit/Project.swift
```diff
@@ -609,6 +605,54 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
...
+			if let sdkName = propertyList["DTSDKName"] as? String {
+				return .success(sdkName)
+			}
+			return .failure(CarthageError.InfoPlistParseFailed(plistURL: plistURL, reason: "the value for the DTSDKName key is not a string"))
+		}
+		// Thus, the SDK name must be trimmed to match the platform name, e.g.
+		// macosx10.10 -> macosx
+		|> map { sdkName in sdkName.stringByTrimmingCharactersInSet(NSCharacterSet.letterCharacterSet().invertedSet) }
+		|> tryMap { platform in SDK.fromString(platform) }
+		|> map { sdk in sdk.platform }
```
ikesyo: This mapping can be done in `Result`: `|> tryMap { platform in SDK.fromString(platform).map { $0.platform } }`

Source/CarthageKit/Project.swift
```diff
@@ -609,6 +605,54 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
...
+private func platformForInfoPlist(plistURL: NSURL) -> SignalProducer<Platform, CarthageError> {
+	return SignalProducer(value: plistURL)
+		|> startOn(QueueScheduler(name: "org.carthage.CarthageKit.Project.platformForInfoPlist"))
+		|> tryMap { URL in
+			var error: NSError?
+			if let data = NSData(contentsOfURL: URL, options: nil, error: &error) {
+				return .success(data)
+			}
+			return .failure(CarthageError.ReadFailed(URL, error))
+		}
```
ikesyo: This might be simplified as follows:

```swift
SignalProducer.try { () -> Result<NSData, CarthageError> in
    var error: NSError?
    if let data = NSData(contentsOfURL: plistURL, options: nil, error: &error) {
        return .success(data)
    }
    return .failure(.ReadFailed(plistURL, error))
}
|> startOn(QueueScheduler(name: "org.carthage.CarthageKit.Project.platformForInfoPlist"))
```

Source/CarthageKit/Project.swift
```diff
@@ -609,6 +605,54 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
...
+		|> tryMap { (data: NSData) -> Result<AnyObject, CarthageError> in
+			var error: NSError?
+			let options = NSPropertyListReadOptions(NSPropertyListMutabilityOptions.Immutable.rawValue)
+			let plist: AnyObject? = NSPropertyListSerialization.propertyListWithData(data, options: options, format: nil, error: &error)
+			if let pist: AnyObject = plist {
+				return .success(pist)
+			}
+			return .failure(CarthageError.ReadFailed(plistURL, error))
```
ikesyo: We can omit the `CarthageError` enum name here.

Source/CarthageKit/Project.swift
```diff
@@ -609,6 +605,54 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
...
+		}
+		|> tryMap { plist -> Result<Dictionary<String, AnyObject>, CarthageError> in
+			if let plist = plist as? Dictionary<String, AnyObject> {
+				return .success(plist)
+			}
+			return .failure(CarthageError.InfoPlistParseFailed(plistURL: plistURL, reason: "the root plist object is not a dictionary of values by strings"))
```
ikesyo: Ditto.

Source/CarthageKit/Project.swift
```diff
@@ -609,6 +605,54 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
...
+		}
+		// Neither DTPlatformName nor CFBundleSupportedPlatforms can not be used 
+		// because Xcode 6 and below do not include either in Mac OSX frameworks.
+		|> tryMap { propertyList -> Result<String, CarthageError> in
+			if let sdkName = propertyList["DTSDKName"] as? String {
+				return .success(sdkName)
+			}
+			return .failure(CarthageError.InfoPlistParseFailed(plistURL: plistURL, reason: "the value for the DTSDKName key is not a string"))
```
ikesyo: Ditto.

Source/CarthageKit/Project.swift
```diff
@@ -609,6 +605,54 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
...
+		|> tryMap { (data: NSData) -> Result<AnyObject, CarthageError> in
+			var error: NSError?
+			let options = NSPropertyListReadOptions(NSPropertyListMutabilityOptions.Immutable.rawValue)
+			let plist: AnyObject? = NSPropertyListSerialization.propertyListWithData(data, options: options, format: nil, error: &error)
+			if let pist: AnyObject = plist {
+				return .success(pist)
+			}
+			return .failure(CarthageError.ReadFailed(plistURL, error))
+		}
+		|> tryMap { plist -> Result<Dictionary<String, AnyObject>, CarthageError> in
```
ikesyo: We prefer the shorthand form: `[String: AnyObject]`

Source/CarthageKit/Project.swift
```diff
@@ -609,6 +605,54 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
...
+			if let plist = plist as? Dictionary<String, AnyObject> {
```
ikesyo: Ditto.

ikesyo:
Thank you for the great work! Could you check the notes above? Looks good to me other than them. :smiley:

Source/CarthageKit/Project.swift
```diff
@@ -609,6 +605,54 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
 		}
 }
 
+/// Sends the URL for the Info.plist of the specified Framework.
+private func infoPlistForFramework(frameworkURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+	return filesInDirectory(frameworkURL, "com.apple.property-list")
+		|> filter { plist in plist.lastPathComponent == "Info.plist" }
```
ikesyo: I'm not confident the file name checking is OK for all the dynamic frameworks... :neutral_face:

neilpa: I'm assuming `filesInDirectory` is recursive. This probably isn't what you want. Instead you should just look for `Resources/Info.plist` directly since that does seem to be a hard requirement for frameworks. If it wasn't, I'm not sure how `dyld` would know how to load a framework.

https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/FrameworkAnatomy.html

neilpa: Although this looks like it might be slightly different for umbrella frameworks.

ikesyo: That's really helpful reference! I'm for to explicitly get `Resources/Info.plist` here :+1:

erichoracek: @neilpa The ReactiveCocoa iOS framework has `Info.plist` at the root, while the Mac framework has it within the `Resources` symlink. Should I check those two paths explicitly rather than a recursive search? As to your other comment, if you inspect the implementation of `filesInDirectory`, it skips nested frameworks.

erichoracek: @ikesyo What would you recommend as the best way to check? I'm not super familiar with all of the permutations of frameworks.

Source/CarthageKit/Xcode.swift
erichoracek: Ah ok—was unaware of whether specs had access to private methods. I'll move it back to public.

Source/CarthageKit/Project.swift
erichoracek: @neilpa To clarify: iOS frameworks do not appear to have the `Resources` directory by default, so I'm not sure that this will work in all cases. That document appears to only pertain to Mac OS X frameworks.

neilpa: I was concerned about multi-version frameworks -- `Versions/A/Resources/Info.plist` and `Versions/B/Resources/Info.plist`. However, I'm not sure I've ever seen someone use framework versioning as apple describes it in that document.

> The ReactiveCocoa iOS framework has Info.plist at the root, while the Mac framework has it within the Resources symlink. Should I check those two paths explicitly rather than a recursive search?

Yea, I would go with that approach.

ikesyo:
Seems good to go. :shipit:

erichoracek:
@ikesyo does any action need to be taken regarding the failing tests on CI?

ikesyo:
@erichoracek All the tests are passed on my local machine, so I'm gonna merge this. :wink:

PR #766, opened 2015-09-16
Checks if cache dependency folder is writable.

guidomb:
If not falls back to '~/.carthage' folder. Also the fall back
URL is constructed using the $HOME enviromental variable if
available otherwise by expanding the tilde.

This commit also fixes an issue with Homebrew testing carthage
in a sandboxed mode. For more info about the Homebrew issue check
https://github.com/Homebrew/homebrew/pull/43884

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -30,28 +30,26 @@ private let fallbackDependenciesURL: NSURL = {
 private let CarthageUserCachesURL: NSURL = {
 	let fileManager = NSFileManager.defaultManager()
 	
-	let URL: Result<NSURL, NSError> = try({ (error: NSErrorPointer) -> NSURL? in
+	let URLResult: Result<NSURL, NSError> = try({ (error: NSErrorPointer) -> NSURL? in
 		fileManager.URLForDirectory(NSSearchPathDirectory.CachesDirectory, inDomain: NSSearchPathDomainMask.UserDomainMask, appropriateForURL: nil, create: true, error: error)
 	}).flatMap { cachesURL in
-		if fileManager.isWritableFileAtPath(cachesURL.absoluteString!) {
-			return Result(value: cachesURL)
+		let dependenciesURL = cachesURL.URLByAppendingPathComponent(CarthageKitBundleIdentifier, isDirectory: true)
+		if fileManager.isWritableFileAtPath(dependenciesURL.absoluteString!) {
+			return Result(value: dependenciesURL)
 		} else {
 			let error = NSError(domain: CarthageKitBundleIdentifier, code: 0, userInfo: nil)
 			return Result(error: error)
 		}
 	}
 
-	switch URL {
-	case .Success:
+	switch URLResult {
+	case let .Success(URL):
 		NSFileManager.defaultManager().removeItemAtURL(fallbackDependenciesURL, error: nil)
-
+		return URL.value
 	case let .Failure(error):
 		NSLog("Warning: No Caches directory could be found or created: \(error.value.localizedDescription). (\(error.value))")
+		return fallbackDependenciesURL
 	}
-
-	let dependenciesURL = URL.value?.URLByAppendingPathComponent(CarthageKitBundleIdentifier, isDirectory: true)
-	return dependenciesURL ?? fallbackDependenciesURL
-	
 }()
 
 /// The file URL to the directory in which downloaded release binaries will be
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -44,12 +44,8 @@ private let CarthageUserCachesURL: NSURL = {
 				return Result(error: error)
 			}
 		} else {
-			return try { (error: NSErrorPointer) -> NSURL? in
-				let attributes = [NSFilePosixPermissions : NSNumber(short:755)]
-				fileManager.createDirectoryAtURL(dependenciesURL, withIntermediateDirectories: true, attributes:attributes, error: error)
-				return dependenciesURL
-			}
-		}		
+			return try { fileManager.createDirectoryAtURL(dependenciesURL, withIntermediateDirectories: true, attributes: [NSFilePosixPermissions : NSNumber(short:755)], error: $0) }.map { dependenciesURL }
+		}
 	}
 
 	switch URLResult {
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -34,12 +34,25 @@ private let CarthageUserCachesURL: NSURL = {
 		fileManager.URLForDirectory(NSSearchPathDirectory.CachesDirectory, inDomain: NSSearchPathDomainMask.UserDomainMask, appropriateForURL: nil, create: true, error: error)
 	}).flatMap { cachesURL in
 		let dependenciesURL = cachesURL.URLByAppendingPathComponent(CarthageKitBundleIdentifier, isDirectory: true)
-		if fileManager.isWritableFileAtPath(dependenciesURL.absoluteString!) {
-			return Result(value: dependenciesURL)
+		let dependenciesPath = dependenciesURL.absoluteString!
+		
+		let result: Result<NSURL, NSError>
+		if fileManager.fileExistsAtPath(dependenciesPath, isDirectory:nil) {
+			if fileManager.isWritableFileAtPath(dependenciesPath) {
+				result = Result(value: dependenciesURL)
+			} else {
+				let error = NSError(domain: CarthageKitBundleIdentifier, code: 0, userInfo: nil)
+				result = Result(error: error)
+			}
 		} else {
-			let error = NSError(domain: CarthageKitBundleIdentifier, code: 0, userInfo: nil)
-			return Result(error: error)
+			result = try { (error: NSErrorPointer) -> NSURL? in
+				let attributes = [NSFilePosixPermissions : NSNumber(short:755)]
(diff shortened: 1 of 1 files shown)


guidomb:
@mdiep Can you review this PR. I am not sure what to about that the error inside the flatMap in [this](https://github.com/Carthage/Carthage/pull/766/files#diff-3bd7f6198aa606c5ec6931e77cb3c42bR39) line. I didn't want to change the from `Result<NSURL, NSError>` to `Result<NSURL, CarthageError>` but I wasn't sure how to create a new instance of `NSError` for that case.

guidomb:
Mmmm I run the test locally and they passed.

guidomb:
@mdiep I think there is a problem with the tests in travis because [this](https://travis-ci.org/Carthage/Carthage/builds/80679495) other build is also failing with the same errors.

Source/CarthageKit/Project.swift
```diff
@@ -13,13 +13,33 @@ import ReactiveCocoa
...
+	let fileManager = NSFileManager.defaultManager()
+	
 	let URL: Result<NSURL, NSError> = try({ (error: NSErrorPointer) -> NSURL? in
-		NSFileManager.defaultManager().URLForDirectory(NSSearchPathDirectory.CachesDirectory, inDomain: NSSearchPathDomainMask.UserDomainMask, appropriateForURL: nil, create: true, error: error)
-	})
-
-	let fallbackDependenciesURL = NSURL.fileURLWithPath("~/.carthage".stringByExpandingTildeInPath, isDirectory:true)!
+		fileManager.URLForDirectory(NSSearchPathDirectory.CachesDirectory, inDomain: NSSearchPathDomainMask.UserDomainMask, appropriateForURL: nil, create: true, error: error)
+	}).flatMap { cachesURL in
+		if fileManager.isWritableFileAtPath(cachesURL.absoluteString!) {
```
mdiep: I think this should be checking the Carthage-specific directory of the caches directory. So this line:

```swift
let dependenciesURL = URL.value?.URLByAppendingPathComponent(CarthageKitBundleIdentifier, isDirectory: true)
```

should already have been executed.

Source/CarthageKit/Project.swift
```diff
@@ -29,7 +49,9 @@ private let CarthageUserCachesURL: NSURL = {
 		NSLog("Warning: No Caches directory could be found or created: \(error.value.localizedDescription). (\(error.value))")
 	}
 
-	return URL.value?.URLByAppendingPathComponent(CarthageKitBundleIdentifier, isDirectory: true) ?? fallbackDependenciesURL
+	let dependenciesURL = URL.value?.URLByAppendingPathComponent(CarthageKitBundleIdentifier, isDirectory: true)
+	return dependenciesURL ?? fallbackDependenciesURL
```
mdiep: While you're at it, would you mind moving these `return`s into the `switch` above? :pray:

guidomb:
@mdiep I made the changes you suggested and then I had to change things a little bit because I was seeing an error related with checking if the folder was writable. I do agree that we have to check for write permissions only for the Carthage-specific cache folder, `~/Library/Caches/org.carthage.CarthageKit/` but I order to that the folder has to exists.

That is why I first check if the Carthage-specific cache folder exits and then check its write permissions. If it does not exist I create it with `755` file permissions. What do you think about that?

What about the `NSError` that I am creating in case the folder does not have write permissions? Do you think we should have an enum with error codes or maybe use a `CarthageError`?

Finally all test pass in my local machine but after running `make install` in a project with the following `Cartfile` and `Cartfile.private` I get huge crash.

**Cartfile:**
```
github "ReactiveCocoa/ReactiveCocoa" ~> 3.0.0
github "CocoaLumberjack/CocoaLumberjack" ~> 2.0.1
github "Wolox/WLXBluetoothDevice" ~> 0.2.0
```

**Cartfile.private:**
```
github "specta/specta" ~> 1.0.0
github "specta/expecta" ~> 1.0.1
```

**Cartfile.resolved:**
```
github "robrix/Box" "1.2.2"
github "CocoaLumberjack/CocoaLumberjack" "2.0.1"
github "specta/expecta" "v1.0.3"
github "specta/specta" "v1.0.3"
github "antitypical/Result" "0.4.4"
github "ReactiveCocoa/ReactiveCocoa" "v3.0.0"
github "Wolox/WLXBluetoothDevice" "0.2.0"
```

**Error:**
```
carthage bootstrap --platform ios                                                                                                              
*** Checking out specta at "v1.0.3"
*** Checking out WLXBluetoothDevice at "0.2.0"
*** Downloading Box.framework binary at "1.2.2"
*** Checking out CocoaLumberjack at "2.0.1"
*** Checking out Result at "0.4.4"
*** Checking out expecta at "v1.0.3"
*** Downloading ReactiveCocoa.framework binary at "v3.0"
*** xcodebuild output can be found in /var/folders/gf/dslbq4t94zg44sz7wnk2lg2r0000gp/T/carthage-xcodebuild.1vdEcz.log
*** Building scheme "CocoaLumberjack" in Lumberjack.xcodeproj
fatal error: SDK count 3 in scheme CocoaLumberjack is not supported: file /Users/guidomb/Documents/projects/forks/Carthage/Source/CarthageKit/Xcode.swift, line 959

Caught signal triggered by the Swift runtime!
Illegal instruction: 4

Caught signal triggered by the Swift runtime!
Illegal instruction: 4

Caught signal triggered by the Swift runtime!
Illegal instruction: 4

Caught signal triggered by the Swift runtime!
Illegal instruction: 4

Caught signal triggered by the Swift runtime!
Illegal instruction: 4

```

I am not sure this is an error introduced by this patch

guidomb:
Well it seems that the crash was already happening before I submitted this patch. Check the latest commit in master https://github.com/Carthage/Carthage/commit/52249f46f370d8fbb82fdd71fe8b043b3b625061 and it also crashes.

If you think this patch is OK the only thing to decide is what to do about the failing tests in travis. @ikesyo had the same issue in https://github.com/Carthage/Carthage/pull/760.

mdiep:
> That is why I first check if the Carthage-specific cache folder exits and then check its write permissions. If it does not exist I create it with 755 file permissions. What do you think about that?

That seems great. :+1: 

> What about the NSError that I am creating in case the folder does not have write permissions? Do you think we should have an enum with error codes or maybe use a CarthageError?

I think it's fine as-is.

Source/CarthageKit/Project.swift
```diff
@@ -13,23 +13,56 @@ import ReactiveCocoa
...
+			if fileManager.isWritableFileAtPath(dependenciesPath) {
+				result = Result(value: dependenciesURL)
+			} else {
+				let error = NSError(domain: CarthageKitBundleIdentifier, code: 0, userInfo: nil)
+				result = Result(error: error)
+			}
+		} else {
+			result = try { (error: NSErrorPointer) -> NSURL? in
+				let attributes = [NSFilePosixPermissions : NSNumber(short:755)]
+				fileManager.createDirectoryAtURL(dependenciesURL, withIntermediateDirectories: true, attributes:attributes, error: error)
```
mdiep: I think we should be checking the return value here and creating an error if it fails.

Source/CarthageKit/Project.swift
```diff
@@ -13,23 +13,56 @@ import ReactiveCocoa
...
-	let fallbackDependenciesURL = NSURL.fileURLWithPath("~/.carthage".stringByExpandingTildeInPath, isDirectory:true)!
+	let fileManager = NSFileManager.defaultManager()
+	
+	let URLResult: Result<NSURL, NSError> = try({ (error: NSErrorPointer) -> NSURL? in
+		fileManager.URLForDirectory(NSSearchPathDirectory.CachesDirectory, inDomain: NSSearchPathDomainMask.UserDomainMask, appropriateForURL: nil, create: true, error: error)
+	}).flatMap { cachesURL in
+		let dependenciesURL = cachesURL.URLByAppendingPathComponent(CarthageKitBundleIdentifier, isDirectory: true)
+		let dependenciesPath = dependenciesURL.absoluteString!
+		
+		let result: Result<NSURL, NSError>
```
mdiep: Can we just `return` directly instead of assigning to this variable?

mdiep:
This is looking good. The failing test in travis doesn't need to be fixed in this PR, since it's not new here, but we should try to figure out what's going on there. I'd hate to release 0.9.2 and find out that those tests are failing on the homebrew CI.

Source/CarthageKit/Project.swift
guidomb: Sure

guidomb:
I have just committed fixes for the comments you made.

Apart from the failing tests in travis there is something wrong with `master` version of Carthage. As I said in a comment earlier, carthage crashes when doing `bootstrap` at least with the `Cartfile` I am using. Should I submit an issue?

mdiep:
> Apart from the failing tests in travis there is something wrong with master version of Carthage. As I said in a comment earlier, carthage crashes when doing bootstrap at least with the Cartfile I am using. Should I submit an issue?

Yes please!

guidomb:
Done https://github.com/Carthage/Carthage/issues/776

mdiep:
Looks good. :+1:

ikesyo:
> but we should try to figure out what's going on there. I'd hate to release 0.9.2 and find out that those tests are failing on the homebrew CI.

I might figure out why. Importing certificates phase is skipped because the secure data in `.travis.yml` is not available for PRs from forked repositories: http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests. So we don't able to build iOS device builds in the specs in that cases, the all failures happened when building iOS schemes.

mdiep:
@ikesyo Thanks—that makes sense!

PR #770, opened 2015-09-17
Remove useless specs from XcodeSpec.swift

ikesyo:
Those specs test `iOSSigningIdentitiesConfigured()` helper function, not the CarthageKit's APIs.


mdiep:
I agree: we don't need to test these. :+1:

PR #778, opened 2015-09-20
Works around the test failure in CI environment

ikesyo:
Addresses https://github.com/Carthage/Carthage/pull/766#issuecomment-141671784. This PR is submitted from my forked repo to confirm that the build ends successfully.


mdiep:
Seems better than having them fail. :+1:

PR #798, opened 2015-09-25
Avoid crashing with unknown SDKs for universal framework targets

### LINKED ISSUE #797: Crash during build
I'm seeing a crash using version 0.9.2 similar to #776, but I'm not getting any error messages. You can reproduce with Nimble:

```
github "Quick/Nimble" "master"
```

```
➜ ~/dev/scratch/carthage-test: carthage version
0.9.2
➜ ~/dev/scratch/carthage-test: carthage bootstrap --platform ios --verbose
*** Checking out nimble at "70b901e55bad427d211b93586a83e0e3bca32034"
*** Building scheme "Nimble-iOS" in Nimble.xcodeproj

Caught signal triggered by the Swift runtime!
Illegal instruction: 4

...
```

That is repeated a bunch of times and never finishes. I can build Nimble successfully with `xcodebuild`. The only thing that sticks out with that repo is the `SUPPORTED_PLATFORMS` setting on Nimble, but that may be unrelated:

```
SUPPORTED_PLATFORMS = "iphonesimulator iphoneos appletvsimulator appletvos"
```

Let me know if I can provide any other information

ikesyo:
Fixes #797. This changes the public `SDKFilterCallback` signature to return `Result<[SDK], CarthageError>`.

The output using Xcode 7.0 with this is as follows:

```sh
$ carthage build
*** xcodebuild output can be found in /var/folders/03/cr990fvs5g73tt4gq52h3j7r0000gn/T/carthage-xcodebuild.aQ45sV.log
*** Building scheme "Nimble-OSX" in Nimble.xcodeproj
*** Building scheme "Nimble-iOS" in Nimble.xcodeproj
A shell task failed with exit code 64:
xcodebuild: error: SDK "appletvsimulator" cannot be located.
```

or 

```sh
$ carthage build --platform ios
*** xcodebuild output can be found in /var/folders/03/cr990fvs5g73tt4gq52h3j7r0000gn/T/carthage-xcodebuild.a4PqQN.log
*** Building scheme "Nimble-iOS" in Nimble.xcodeproj
```

The above correctly ignores tvOS platform which is included in the universal framework target.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -887,15 +887,13 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 		}
 		|> flatMap(.Concat) { platform, sdks -> SignalProducer<(Platform, [SDK], Bool), CarthageError> in
 			let filterResult = sdkFilter(sdks: sdks, scheme: scheme, configuration: configuration, project: project)
-
-			if let filteredSDKs = filterResult.value where filteredSDKs.isEmpty {
-				return .empty
-			}
-
 			return SignalProducer(result: filterResult.map { filteredSDKs in
 				let isFiltered = filteredSDKs.count != sdks.count
 				return (platform, filteredSDKs, isFiltered)
 			})
+		}
+		|> filter { _, sdks, _ in
+			return !sdks.isEmpty
 		}
 		|> flatMap(.Concat) { platform, sdks, isFiltered in
 			let folderURL = workingDirectoryURL.URLByAppendingPathComponent(platform.relativePath, isDirectory: true).URLByResolvingSymlinksInPath!
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -785,13 +785,13 @@ private func mergeBuildProductsIntoDirectory(firstProductSettings: BuildSettings
 
 
 /// A callback function used to determine whether or not an SDK should be built
-public typealias SDKFilterCallback = (sdks: [SDK], scheme: String, configuration: String, project: ProjectLocator) -> [SDK]
+public typealias SDKFilterCallback = (sdks: [SDK], scheme: String, configuration: String, project: ProjectLocator) -> Result<[SDK], CarthageError>
 
 /// Builds one scheme of the given project, for all supported SDKs.
 ///
 /// Returns a signal of all standard output from `xcodebuild`, and a signal
 /// which will send the URL to each product successfully built.
-public func buildScheme(scheme: String, withConfiguration configuration: String, inProject project: ProjectLocator, #workingDirectoryURL: NSURL, sdkFilter: SDKFilterCallback = { $0.0 }) -> SignalProducer<TaskEvent<NSURL>, CarthageError> {
+public func buildScheme(scheme: String, withConfiguration configuration: String, inProject project: ProjectLocator, #workingDirectoryURL: NSURL, sdkFilter: SDKFilterCallback = { .success($0.0) }) -> SignalProducer<TaskEvent<NSURL>, CarthageError> {
 	precondition(workingDirectoryURL.fileURL)
 
 	let buildArgs = BuildArguments(project: project, scheme: scheme, configuration: configuration)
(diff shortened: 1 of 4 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -868,6 +869,10 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 		|> reduce([:]) { (var sdksByPlatform: [Platform: [SDK]], sdk: SDK) in
 			let platform = sdk.platform
 
+			if !platforms.isEmpty && !platforms.contains(platform) {
+				return sdksByPlatform
+			}
```
mdiep: Could this be done through the `sdkFilter` instead?

mdiep:
Can you help me understand what the actual fix was? I don't see where the failure case is.

ikesyo:
The problem is that calling `BuildSettings.loadWithArguments()` in `buildableSDKs()` with `-sdk appletvos` or `-sdk appletvsimulator` using Xcode 7.0's command line tools would fail with the error noted in https://github.com/Carthage/Carthage/issues/797#issuecomment-143090788. I tried to avoid resulting in a runtime crash with [this call](https://github.com/Carthage/Carthage/pull/798/files#diff-908d3140eab5f48b3e5f926de743ef52L112) in that case, and propagate the `CarthageError` instead.

Source/CarthageKit/Xcode.swift
ikesyo: Yep, that could be done. Thank you for pointing this out. :+1:

Source/CarthageKit/Xcode.swift
```diff
@@ -885,27 +885,36 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
...
-			
 			// TODO: Generalize this further?
-			switch sdksToBuild.count {
+			switch sdks.count {
 			case 0:
-				let isMissingSigningIdentities = !sdks.isEmpty
-				let identityAddendum = isMissingSigningIdentities ? " (you're missing one or more signing identities)" : ""
-				fatalError("No valid SDKs found to build\(identityAddendum)")
+				let filterlingAddendum = isFiltered ? " (you may be missing one or more signing identities)" : ""
+				fatalError("No valid SDKs found to build\(filterlingAddendum)")
```
ikesyo: This `case` is dead code now, because of [the change above](https://github.com/Carthage/Carthage/pull/798/files#diff-f66644e8cfbcad3c3826395a690ddfbcR891).

Source/CarthageKit/Xcode.swift
```diff
@@ -885,27 +885,36 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 			let values = map(sdksByPlatform) { ($0, $1) }
 			return SignalProducer(values: values)
 		}
-		|> flatMap(.Concat) { platform, sdks in
+		|> flatMap(.Concat) { platform, sdks -> SignalProducer<(Platform, [SDK], Bool), CarthageError> in
+			let filterResult = sdkFilter(sdks: sdks, scheme: scheme, configuration: configuration, project: project)
+
+			if let filteredSDKs = filterResult.value where filteredSDKs.isEmpty {
```
ikesyo: This is required to let platform filter work correctly. Without this, tvOS platform contained in a universal framework target with `--platform ios` would crash with reaching to [this line](https://github.com/Carthage/Carthage/pull/798/files#diff-f66644e8cfbcad3c3826395a690ddfbcR907), for example.

Source/CarthageKit/Xcode.swift
mdiep: I suppose that's okay. Do you think we should print something in that case? Move this message about maybe missing signing identities?

I would definitely remove this `case`.

Source/CarthageKit/Xcode.swift
mdiep: Can we do this with a `filter` after this `flatMap`?

Source/CarthageKit/Xcode.swift
```diff
@@ -1192,7 +1201,16 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
...
 				let initialValue = (project, scheme)
 
-				let buildProgress = buildScheme(scheme, withConfiguration: configuration, inProject: project, workingDirectoryURL: directoryURL, sdkFilter: sdkFilter)
+				let wrappedSDKFilter: SDKFilterCallback = { sdks, scheme, configuration, project in
+					if platforms.isEmpty {
+						return sdkFilter(sdks: sdks, scheme: scheme, configuration: configuration, project: project)
+					}
+
+					let filteredSDKs = sdks.filter { platforms.contains($0.platform) }
+					return sdkFilter(sdks: filteredSDKs, scheme: scheme, configuration: configuration, project: project)
```
mdiep: I found this block a little confusing to read. Could we do it something like this instead?

```
let filteredSDKs: [SDK]
if platforms.isEmpty {
    filteredSDKs = sdks
} else {
    filteredSDKs = sdks.filter { ... }
}
return sdkFilter(...)
```

That makes it more obvious IMO.

mdiep:
Sorry for the delay in reviewing!

Source/CarthageKit/Xcode.swift
ikesyo: Yep, that seems better. :+1:

Source/CarthageKit/Xcode.swift
ikesyo: Printing warning would need some more work, so I'll just remove them this time.

ikesyo: Done in 67dc9fa.

mdiep: I don't think we need the message—we already print a warning if we skip a build because the code sign identity is missing.

mdiep:
:boom:

PR #821, opened 2015-10-02
Revise phrasing in OS X getting started

mickeyreiss:
I think this rewording will clarify that `carthage update` itself will build each dependency (and sub-dependencies). Elsewhere in the readme, we use `then` to specify a subsequent action to be performed.


mdiep:
Thanks! :+1:

Issue #755, opened 2015-09-15
carthage copy-frameworks does not work for watchOS 2 target


MSchmidt:
With Xcode 7 and watchOS 2 it's necessary to include frameworks built for watchOS inside the watchOS target. Unfortunately the copy-frameworks script fails for this target with ``no identity found`` as described here: https://github.com/Carthage/Carthage/issues/587

For successful submission to the App Store it is needed to use this script or something else that slims the frameworks down to the required release architectures only.

NachoSoto:
Are you getting `Invalid segment Alignment` as well? I think that's due to something about `BITCODE` support. Did you try submitting without `BITCODE` enabled?

ikesyo:
Bitcode is requird, not optional for watchOS 2 apps, so disabling it should not work.

NachoSoto:
Oh good point.
I misread this actually, you're right @MSchmidt, frameworks need to be stripped for the `watchOS 2` target as well.

MSchmidt:
Yes ``Invalid segment Alignment`` is what I get. Along with some other errors but they all come down to the same issue. As a short term fix I've built the needed frameworks myself in Xcode with the release-ARM configuration and copied them into my project just for App Store Submission.

NachoSoto:
I'm not familiar with how `copy-frameworks` was implemented, but maybe someone in @Carthage/Carthage can shed some light?

jurvis:
@MSchmidt seems like I'm facing the same issue as you. mind sharing how to build the frameworks manually in Xcode? Thanks!

MSchmidt:
Sure no problem @jurvis - but keep in mind that this basically eliminates the whole point of Carthage :)

For each of your used Frameworks ($framework):
* open ``Carthage/Checkouts/$framework/$framework.xcodeproj``
* Change active scheme to ``$framework watchOS > watchOS Device`` (or name of currently connected iPhone)
* ``Product`` > ``Archive``
* go into your ``DerivedData`` folder (default is ``/Users/username/Library/Developer/Xcode/DerivedData``)
* within the derived data from your framework: ``./Build/Products/Release-watchos``
* copy ``$framework.framework`` to ``Carthage/Build/watchOS``

For your watchOS target use the frameworks as ``Embedded Binaries`` and you are good to go for App Store Submission. With this setup you cannot build for Watch Simulator anymore. Run ``carthage update`` to build simulator compatible versions of your frameworks again.

jurvis:
[...]

PR #829, opened 2015-10-06
Fix "no identity found" on `copy-frameworks` for watchOS Simulator

norio-nomura:
Xcode 7.0.1 set environments on building for watchOS Simulator:
```sh
CODE_SIGNING_ALLOWED=YES
EXPANDED_CODE_SIGN_IDENTITY=""
```
It cause launching `codesign` with empty identity option and error.

Xcode 7.1 beta 2 is same for watchOS Simulator.
But it is not happen for tvOS Simulator.

Is this related to #755?

### DIFF
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -47,7 +47,7 @@ public struct CopyFrameworksCommand: CommandType {
 private func codeSigningIdentity() -> SignalProducer<String?, CarthageError> {
 	return SignalProducer.try {
 		if codeSigningAllowed() {
-			return getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY").map { $0 }
+			return getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY").map { $0.isEmpty ? nil : $0 }
 		} else {
 			return .success(nil)
 		}


NachoSoto:
This is awesome, I just hit this problem myself :heart:

Source/carthage/CopyFrameworks.swift
```diff
@@ -47,7 +47,7 @@ public struct CopyFrameworksCommand: CommandType {
 private func codeSigningIdentity() -> SignalProducer<String?, CarthageError> {
 	return SignalProducer.try {
 		if codeSigningAllowed() {
-			return getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY").map { $0 }
+			return getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY").map { $0.isEmpty ? nil : $0 }
```
NachoSoto: Can you explain this? I noticed that Xcode was not setting this environment variable for me in one of my targets.
What would the behavior be with this change?

norio-nomura: Calling `codesign()` will be skipped on following: https://github.com/Carthage/Carthage/blob/master/Source%2FCarthageKit%2FXcode.swift#L1232
by `codesigningIdentity` is `nil`

NachoSoto: Ooooh that makes a lot of sense. I'm only seeing this problem when compiling for the simulator anyway (not on device or when archiving) so it seems like the correct fix :+1:

Thomvis:
I tested this in the set-up at work and can confirm that it works as advertised. Awesome :star2:!

mdiep:
Makes sense. :+1: Thanks for the PR! :sparkles:

norio-nomura:
:pray:

Issue #815, opened 2015-10-01
GITHUB_ACCESS_TOKEN environment variable has no effect with El Capitan


flovilmart:
The recently introduced GITHUB_ACCESS_TOKEN environment variable for private repositories has unfortunately no effet in El Capitan so the private github repositories respond with:

```fatal: could not read Username for 'https://github.com': terminal prompts disabled```

furyu-ikeda-sho:
`GITHUB_ACCESS_TOKEN`?

flovilmart:
that's what I meant...

ikesyo:
So is this dupe of #786?

flovilmart:
@ikesyo yeah you're right. I've investigated a bit the issue and that's pretty odd, in the carthage app code, accessing the env vars yields the proper values, but once in the CarthageKit code, it seems that nothing gets called...

mdiep:
We could switch things around to read those from `carthage`.

flovilmart:
That would defeat the purpose of the separation of concerns.

mdiep:
I don't think reading the environment variable has to be a CarthageKit concern.

flovilmart:
@mdiep ok then, Where do you suggest we set this property? as a static prop of Github.swift?

mdiep:
That should be fine.

flovilmart:
I'm gonna take care of it then!

flovilmart:
@mdiep question for you, has commandant the ability to run a command before each command? I'd load the proper environment in general as it may be used in multiple scripts (Update, Checkout etc...)

flovilmart:
So even with a public global var, it seems to keep not working. It's a pain to debug as when passing the environment vars from Xcode is works properly. May be it's related to my current configuration. tried in bash and zsh.

flovilmart:
See PR #834

mdiep:
Fixed by #891.

PR #834, opened 2015-10-07
Use GITHUB_ACCESS_TOKEN as auth for cloning github.com private repos in HTTPS

flovilmart:
When specifying GIT_AUTH environment variable in the form of either "domain.com=token" or "domain.com=username:pass" the value will be injected into the git url to provide headless authentication for build servers.

The default "token" or "username:password" will be applied to all github.com requests.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -119,7 +119,13 @@ public struct GitHubRepository: Equatable {
 
 	/// The URL that should be used for cloning this repository over HTTPS.
 	public var HTTPSURL: GitURL {
-		return GitURL("\(server)/\(owner)/\(name).git")
+		let gitAuth = parseGitHubAccessTokenFromEnvironment()
+
+		var serverAuth:String = ""
+		if let auth = gitAuth[server.hostname] {
+			serverAuth = "\(auth)@"
+		}
+		return GitURL("\(server.scheme)://\(serverAuth)\(server.hostname)/\(owner)/\(name).git")
 	}
 
 	/// The URL that should be used for cloning this repository over SSH.


flovilmart:
In reference of #815 

The problem was actually at the time Carthage tries to resolve the dependencies with https:// urls. With private repos it is possible to skip the login phase normally required with inserting the GITHUB_ACCESS_TOKEN for all github.com URLs.

Source/CarthageKit/Git.swift
```diff
@@ -157,7 +157,7 @@ public func cloneRepository(cloneURL: GitURL, _ destinationURL: NSURL, bare: Boo
 		arguments.append("--bare")
 	}
 
-	return launchGitTask(arguments + [ "--quiet", cloneURL.URLString, destinationURL.path! ])
+	return launchGitTask(arguments + [ "--quiet", cloneURL.URLStringWithGithubAccessToken, destinationURL.path! ])
```
mdiep: It'd be nice to bake this in earlier if possible, so that `Git.swift` can know as little as possible about `GitHub.swift`.

Source/CarthageKit/GitHub.swift
```diff
@@ -46,6 +46,16 @@ extension GitHubError: Hashable {
 	}
 }
 
+extension GitURL {
+	var URLStringWithGithubAccessToken:String {
+		if let url = NSURL(string:self.URLString) where url.host == GitHubRepository.Server.GitHub.hostname,
```
mdiep: This really needs to work with GitHub Enterprise repos as well. :confused:

mdiep:
Is this still an El Capitan-specific issue? Access tokens were working correctly before, right? What changed?

flovilmart:
I can't say if it was working before but the token wasn't used to actually fetch/clone the repo. It was supposedly using the keychain to pick up the credentials. Now that's not working for me, on El Capitan. 

I realized while trying to fix the issue that it works while setting the params in the run scheme options, but not from the terminal. 

How should one specify creds for https github urls?

Source/CarthageKit/GitHub.swift
flovilmart: That would be easy to add, given the credentials hashmap

Source/CarthageKit/Git.swift
flovilmart: So rename add another environment variable that could serve as the identity for the the git URL?

jdhealy:
> It was supposedly using the keychain to pick up the credentials. Now that's not working for me, on El Capitan.

Maybe the credential helper was affected by [System Integrity Protection](https://support.apple.com/en-us/HT204899)? [Step 3 of this GitHub help article instructs](https://help.github.com/articles/caching-your-github-password-in-git/) to put the helper executable in the same directory as the git executable. Try `which git-credential-osxkeychain`.

Maybe also check if the `credential.helper` config is set?
```shell
git config --get --global credential.helper
```

flovilmart:
$ which git-credential-osxkeychain
    /usr/local/bin/git-credential-osxkeychain

    $ git config --get --global credential.helper
    osxkeychain

flovilmart:
@jdhealy after investigation it seems that the error was related to improper configuration on my dev machine. I never use HTTPS but SSH based clones.

However, in the context of CI, one may want to build the carthage framework with it's dependencies so this fix could prove helpful (once added with all domains) in those scenarios.

mdiep:
Would you mind clarifying the current state of this and the bug that it's fixing? I'm a little confused based on some of the recent comments.

flovilmart:
The bug I experienced with HTTPS git update was not a bug but a 'wrong' config on my machine. I never use HTTPS, rather git+ssh 

The intended feature that it adds:
on a build server, letting Carthage clone/fetch private repositories would require adding the credentials to the keychain.
This proposed fix allows passing a GIT_AUTH environment variable that can take the form of either:

GIT_AUTH=token
GIT_AUTH=domain.com=token,otherdomain.com=othertoken

In order to provide proper authentication to the private repository instead of adding keychain username and passwords

Source/CarthageKit/Git.swift
```diff
@@ -13,8 +13,21 @@ import ReactiveTask
...
-	public let URLString: String
+	public var URLString: String {
+		let gitAuth = parseGitAuthFromEnvironment()
+
+		// Do not override auth if already provided in URL
+		if let url = NSURL(string: _URLString) where url.user == nil || url.password == nil,
+			let host = url.host,
+			let auth = gitAuth[host] {
+				return _URLString.stringByReplacingOccurrencesOfString("\(host)", withString: "\(auth)@\(host)")
+		}
```
mdiep: Sorry I wasn't clearer before, but I think the appropriate place to add this logic is inside `GitHubRepository.HTTPSURL`. Then all the logic about GitHub authentication tokens can stay inside `GitHub.swift`.

mdiep:
Thanks for the explanation! That helped a lot. :smile:

Source/CarthageKit/Git.swift
flovilmart: I'll move it around then :)

flovilmart:
@mdiep I rebased in 1 commit and updated the commit to reflect the simple change to add auth in `GitHub.HTTPSUrl`

Source/CarthageKit/GitHub.swift
```diff
@@ -119,7 +119,13 @@ public struct GitHubRepository: Equatable {
 
 	/// The URL that should be used for cloning this repository over HTTPS.
 	public var HTTPSURL: GitURL {
-		return GitURL("\(server)/\(owner)/\(name).git")
+		let gitAuth = parseGitHubAccessTokenFromEnvironment()
+
+		var serverAuth:String = ""
+		if let auth = gitAuth[server.hostname] {
+				 serverAuth = "\(auth)@"
```
mdiep: The indentation looks off on this line. Otherwise this looks great! :sparkles:

flovilmart:
rebased and pushed should the PR be on the master instead?

jdhealy:
Minor concern: could the modified URL (specifically, including the token) show up in any files written to disk? Mainly, `.git/config` including the private token, with `--use-submodules`.

flovilmart:
@jdhealy I don't believe we should store the token anywhere, and that it should be per environment basis. I just checked and when using --use-submodules the token doesn't get stored locally in .git/config

is that the behaviour you'd expect?

mdiep:
> rebased and pushed should the PR be on the master instead?

Yeah, that would probably be a good idea. :+1:

jdhealy:
@flovilmart Cool, seems like `git` rightfully treats that like private information. People do unintuitive things with git repos and their `.git` directories (e.g., checking them into other git repositories), so thankfully this PR doesn't make it any easier for them to accidentally publish a secret token.

LGTM :+1:

flovilmart:
@jdhealy I unfortunately started this branch out of the swift-2 branch. This branch also seems fully functional now. Any plan / schedule to merge swift-2 to master?

mdiep:
> Any plan / schedule to merge swift-2 to master?

We're waiting on our upstream dependencies to create Swift 2 releases. So I think it will be another week or two yet.

flovilmart:
I had no problems with Xcode 7 to build :)

carlossless:
+1

mdiep:
Swift 2 is on master now if you want to retarget this PR.

flovilmart:
@mdiep I need to close and reopen?

flovilmart:
See #891

PR #838, opened 2015-10-09
Add make command for building homebrew tarball

gfontenot:
This follows the same basic pattern as `prefix_install`, except that it
creates a tarball with the executable and CarthageKit in it.

These tarballs could be attached to releases along with the package and
CarthageKit, and installed via homebrew as pre-built binaries.

I'm unsure on if this removes the need for `prefix_install`.

Assuming this is acceptable, I can submit a PR for homebrew that uses this
tarball for installation. You can see a proof of concept of how this would work
[in my personal tap](https://github.com/gfontenot/homebrew-formulae).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,14 @@ installables: clean bootstrap
 	mv -f "$(CARTHAGE_EXECUTABLE)" "$(TEMPORARY_FOLDER)$(BINARIES_FOLDER)/carthage"
 	rm -rf "$(BUILT_BUNDLE)"
 
+homebrew: installables
+	mkdir -p "homebrew/Frameworks" "homebrew/bin"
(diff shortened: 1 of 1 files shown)


mdiep:
This looks fine to me if it's an acceptable output for Homebrew. But I don't know whether that's the case. Would this be an acceptable way to distribute Carthage via Homebrew?

gfontenot:
I have no idea. Maybe @mikemcquaid can lend some guidance?

mikemcquaid:
Might be worth just using the existing .pkg with Homebrew Cask?

mdiep:
The current approach seems to be working okay for now, so I'm going to close this out. I'm hopeful that we'll start to get a more reasonable upgrade path, like with Swift 2.1 -> 2.2, so that we'll experience less pain in upgrading.

I'm open to reconsidering this. But with my limited knowledge about Homebrew, I don't see how this would benefit us.

PR #839, opened 2015-10-10
Support embedding bitcode when ENABLE_BITCODE=YES

### LINKED ISSUE #535: BITCODE support
Reading the [release notes](http://adcdownload.apple.com/WWDC_2015/Xcode_7_beta/Xcode_7_beta_Release_Notes.pdf), I stumbled upon this:
```
ENABLE_BITCODE should be consistently turned on for all the targets. If you use a
library or framework provided by a third party, please contact the vendor for an updated version which
contains bitcode.
```

Given that `Carthage` is simply built on top of `xcodebuild` (yay!), it's likely that nothing will need to be changed. When compiling frameworks for dependencies using Xcode 7 (which hopefuly will have `ENABLE_BITCODE` set to `true` by default), it will do the right thing.

I figured I'd open an issue to keep track of this though :)

Happy Swifting!

ikesyo:
Resolves #535, many thanks to the discussion in that issue! :sparkling_heart:

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -108,6 +108,9 @@ public struct BuildArguments {
 	/// the native architecture.
 	public var onlyActiveArchitecture: OnlyActiveArchitecture = .NotSpecified
 
+	/// The build setting whether full bitcode should be embedded in the binary.
+	public var bitcodeGenerationMode: BitcodeGenerationMode = .None
+
 	public init(project: ProjectLocator, scheme: String? = nil, configuration: String? = nil, sdk: SDK? = nil) {
 		self.project = project
 		self.scheme = scheme
@@ -140,6 +143,7 @@ public struct BuildArguments {
 		}
 
 		args += onlyActiveArchitecture.arguments
+		args += bitcodeGenerationMode.arguments
 
 		return args
 	}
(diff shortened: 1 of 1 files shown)


mdiep:
@NachoSoto Does this approach make sense to you? I'm inexperienced with bitcode, so it's hard for me to say.

NachoSoto:
So if I understand this correctly, this PR basically adds `BITCODE_GENERATION_MODE = "bitcode"` iff the target has `ENABLED_BITCODE`, since `build` (as opposed to `archive`) doesn't add that automatically?

Makes sense to me! :+1: this is so much better, nobody will have to worry about that extra detail when creating a framework :star:

ikesyo:
> So if I understand this correctly, this PR basically adds BITCODE_GENERATION_MODE = "bitcode" iff the target has ENABLED_BITCODE, since build (as opposed to archive) doesn't add that automatically?

Yes, that's true. And _only when _`ENABLE_BITCODE=YES`_ is important to support testing libraries such as Quick and Nimble, since `XCTest` does not accept frameworks which contains bitcode (https://github.com/Quick/Quick/issues/343).

NachoSoto:
Yup! Perfect 👍🏻

conradev:
Awesome!

ikesyo:
:tada:

kean:
Hey.

>  since build (as opposed to archive) doesn't add that automatically?

Does this mean that Carthage builds frameworks using debug configuration that disables optimisations, enables a lot of debug tools like zombie objects etc?

NachoSoto:
No, `Carthage` uses `release` by default.

kean:
Looks like it uses 'release' conf. RAC is impossible -__-

```
<*> m <| Option(key: "configuration", defaultValue: "Release", usage: "the Xcode configuration to build")
```

kean:
Thanks @NachoSoto

PR #840, opened 2015-10-12
More direct way of returning string description.

Nirma:
Sorry for the nit-pick but I thought this might read a little cleaner.

Tests passed, build runs clean.

Hope it helps!


NachoSoto:
Yup, this should be better!

PR #841, opened 2015-10-12
Use native swift String version of enumerateLines

Nirma:
Thought it might be better to use the swift string version of the enumerateLines method.


mdiep:
Thanks!

PR #849, opened 2015-10-15
Added note to only use homebrew install on Xcode 6.x

aral:
This will save people on Xcode 7.x/El Cap some time so they don’t attempt to install via Homebrew, get a build error, search for the issue, and realise that it’s not supported.


mdiep:
Thanks! :sparkles:

PR #853, opened 2015-10-16
Copy bcsymbolmap files when building

kballard:
These files are necessary in order to archive a built product that includes bitcode.
This change makes it copy the bcsymbolmap files to the Carthage/Build/$plat folder,
as well as archiving them with `carthage archive` and installing them from prebuilt
archives.

Note, I have not actually tested the prebuilt archive stuff, as there are no prebuilt
archives with bcsymbolmap files (that I'm aware of) today.

This also changes `carthage copy-frameworks` to also copy the bcsymbolmap files
into the build products folder whenever an archive is done. It has a flag
`--copy-bcsymbolmaps` that can be used to copy them on other builds too if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
   $(SRCROOT)/Carthage/Build/iOS/ReactiveCocoa.framework
   ```
 
-  This script works around an [App Store submission bug](http://www.openradar.me/radar?id=6409498411401216) triggered by universal binaries.
+  This script works around an [App Store submission bug](http://www.openradar.me/radar?id=6409498411401216) triggered by universal binaries as well as ensures that necessary bitcode-related files are also copied.
 
 ##### Copying debug symbols for debugging and crash reporting
 
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
   $(SRCROOT)/Carthage/Build/iOS/ReactiveCocoa.framework
   ```
 
-  This script works around an [App Store submission bug](http://www.openradar.me/radar?id=6409498411401216) triggered by universal binaries as well as ensures that necessary bitcode-related files are also copied.
+  This script works around an [App Store submission bug](http://www.openradar.me/radar?id=6409498411401216) triggered by universal binaries and ensures that necessary bitcode-related files are copied when archiving.
 
 ##### Copying debug symbols for debugging and crash reporting
 
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -470,9 +470,13 @@ public final class Project {
 	/// Sends the URLs of the bcsymbolmap files after copying.
 	public func copyBCSymbolMapsToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
 		return BCSymbolMapsForFramework(frameworkURL, inDirectoryURL: directoryURL)
-			|> flatMap(.Merge) { source in
-				let target = frameworkURL.URLByDeletingLastPathComponent!.URLByAppendingPathComponent(source.lastPathComponent!)
-				return copyProduct(source, target.URLByResolvingSymlinksInPath!)
+			|> flatMap(.Merge) { fileURL in
+				let destinationDirectoryURL = frameworkURL.URLByDeletingLastPathComponent!
+				let fileName = fileURL.lastPathComponent!
+				let destinationURL = destinationDirectoryURL.URLByAppendingPathComponent(fileName, isDirectory: false)
+				let resolvedDestinationURL = destinationURL.URLByResolvingSymlinksInPath!
+				
+				return copyProduct(fileURL, resolvedDestinationURL)
 			}
 	}
 
@@ -636,27 +640,21 @@ private func cacheDownloadedBinary(downloadURL: NSURL, toURL cachedURL: NSURL) -
 }
 
 /// Sends the URL to each file found in the given directory conforming to the
-/// given type identifier.
-private func filesInDirectory(directoryURL: NSURL, #typeIdentifier: String) -> SignalProducer<NSURL, CarthageError> {
-	return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: NSDirectoryEnumerationOptions.SkipsHiddenFiles | NSDirectoryEnumerationOptions.SkipsPackageDescendants, catchErrors: true)
+/// given type identifier. If no type identifier is provided, all files are sent.
+private func filesInDirectory(directoryURL: NSURL, _ typeIdentifier: String? = nil) -> SignalProducer<NSURL, CarthageError> {
+	let producer = NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: .SkipsHiddenFiles | .SkipsPackageDescendants, catchErrors: true)
 		|> map { enumerator, URL in URL }
-		|> filter { URL in
-			return URL.typeIdentifier
-				.analysis(ifSuccess: { identifier in
-					return UTTypeConformsTo(identifier, typeIdentifier) != 0
-				}, ifFailure: { _ in false })
-		}
-}
-
-/// Sends the URL to each file found in the given directory that has the
-/// given file extension.
-///
-/// Note: If a proper UTI is defined for the file extension,
-///	      `filesInDirectory(_:typeIdentifier:)` is preferred.
-private func filesInDirectory(directoryURL: NSURL, #fileExtension: String) -> SignalProducer<NSURL, CarthageError> {
-	return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL, includingPropertiesForKeys: [], options: .SkipsHiddenFiles | .SkipsPackageDescendants, catchErrors: true)
-		|> map { enumerator, URL in URL }
-		|> filter { URL in URL.pathExtension == fileExtension }
+	if let typeIdentifier = typeIdentifier {
+		return producer
+			|> filter { URL in
+				return URL.typeIdentifier
+					.analysis(ifSuccess: { identifier in
+						return UTTypeConformsTo(identifier, typeIdentifier) != 0
+					}, ifFailure: { _ in false })
+			}
+	} else {
+		return producer
+	}
 }
 
 /// Sends the URL for the Info.plist of the specified Framework.
@@ -724,7 +722,7 @@ private func platformForInfoPlist(plistURL: NSURL) -> SignalProducer<Platform, C
 
 /// Sends the URL to each framework bundle found in the given directory.
 private func frameworksInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
-	return filesInDirectory(directoryURL, typeIdentifier: kUTTypeFramework as! String)
+	return filesInDirectory(directoryURL, (kUTTypeFramework as! String))
 		|> filter { URL in
 			// Skip nested frameworks
 			let frameworksInURL = URL.pathComponents?.filter { pathComponent in
@@ -736,7 +734,7 @@ private func frameworksInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL,
 
 /// Sends the URL to each dSYM found in the given directory
 private func dSYMsInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
-	return filesInDirectory(directoryURL, typeIdentifier: "com.apple.xcode.dsym")
+	return filesInDirectory(directoryURL, "com.apple.xcode.dsym")
 }
 
 /// Sends the URL of the dSYM whose UUIDs match those of the given framework, or
@@ -758,7 +756,8 @@ private func dSYMForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL: 
 
 /// Sends the URL to each bcsymbolmap found in the given directory.
 private func BCSymbolMapsInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
-	return filesInDirectory(directoryURL, fileExtension: "bcsymbolmap")
+	return filesInDirectory(directoryURL)
+		|> filter { URL in URL.pathExtension == "bcsymbolmap" }
 }
 
 /// Sends the URLs of the bcsymbolmap files that match the given framework and are
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -768,31 +768,20 @@ private func BCSymbolMapsForFramework(frameworkURL: NSURL, inDirectoryURL direct
 			if UUIDs.isEmpty {
 				return SignalProducer.empty
 			}
-			// conceptually, we just want to filter the URLs to find any that are in uuids.
-			// but we'd like to stop enumeration early when we find all the files.
-			// this also guarantees we don't send two URLs for bcsymbolmap files with the same UUID.
-			return SignalProducer { observer, disposable in
-				var remainingUUIDs = UUIDs
-				BCSymbolMapsInDirectory(directoryURL)
-					|> startWithSignal { signal, signalDisposable in
-						disposable += signalDisposable
-						signal.observe(Signal.Observer { event in
-							switch event {
-							case let .Next(value):
-								if let basename = value.value.lastPathComponent?.stringByDeletingPathExtension, fileUUID = NSUUID(UUIDString: basename) {
-									if remainingUUIDs.remove(fileUUID) != nil {
-										observer.put(event)
-										if remainingUUIDs.isEmpty {
-											sendCompleted(observer)
-										}
-									}
-								}
-							default:
-								observer.put(event)
+			return BCSymbolMapsInDirectory(directoryURL)
+				|> { (signal: Signal<NSURL, CarthageError>) -> Signal<NSURL, CarthageError> in
+					var remainingUUIDs = UUIDs
+					let count = remainingUUIDs.count
+					return signal
+						|> filter { fileURL in
+							if let basename = fileURL.lastPathComponent?.stringByDeletingPathExtension, fileUUID = NSUUID(UUIDString: basename) {
+								return remainingUUIDs.remove(fileUUID) != nil
+							} else {
+								return false
 							}
-						})
-					}
-			}
+						}
+						|> take(count)
+				}
 	}
 }
 
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -766,22 +766,23 @@ private func BCSymbolMapsForFramework(frameworkURL: NSURL, inDirectoryURL direct
 	return UUIDsForFramework(frameworkURL)
 		|> flatMap(.Merge) { UUIDs in
 			if UUIDs.isEmpty {
-				return SignalProducer.empty
+				return .empty
+			}
+			func filterUUIDs(signal: Signal<NSURL, CarthageError>) -> Signal<NSURL, CarthageError> {
+				var remainingUUIDs = UUIDs
+				let count = remainingUUIDs.count
+				return signal
+					|> filter { fileURL in
+						if let basename = fileURL.lastPathComponent?.stringByDeletingPathExtension, fileUUID = NSUUID(UUIDString: basename) {
+							return remainingUUIDs.remove(fileUUID) != nil
+						} else {
+							return false
+						}
+					}
+					|> take(count)
 			}
 			return BCSymbolMapsInDirectory(directoryURL)
-				|> { (signal: Signal<NSURL, CarthageError>) -> Signal<NSURL, CarthageError> in
-					var remainingUUIDs = UUIDs
-					let count = remainingUUIDs.count
-					return signal
-						|> filter { fileURL in
-							if let basename = fileURL.lastPathComponent?.stringByDeletingPathExtension, fileUUID = NSUUID(UUIDString: basename) {
-								return remainingUUIDs.remove(fileUUID) != nil
-							} else {
-								return false
-							}
-						}
-						|> take(count)
-				}
+				|> filterUUIDs
 	}
 }
 
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -367,10 +367,7 @@ public final class Project {
 								|> flatMap(.Merge, self.copyFrameworkToBuildFolder)
 								|> flatMap(.Merge) { frameworkURL in
 									return self.copyDSYMToBuildFolderForFramework(frameworkURL, fromDirectoryURL: directoryURL)
-										|> then(SignalProducer(value: frameworkURL))
-								}
-								|> flatMap(.Merge) { frameworkURL in
-									return self.copyBCSymbolMapToBuildFolderForFramework(frameworkURL, fromDirectoryURL: directoryURL)
+										|> then(self.copyBCSymbolMapsToBuildFolderForFramework(frameworkURL, fromDirectoryURL: directoryURL))
 								}
 								|> on(completed: {
 									_ = NSFileManager.defaultManager().trashItemAtURL(checkoutDirectoryURL, resultingItemURL: nil, error: nil)
@@ -471,21 +468,11 @@ public final class Project {
 	/// no values.
 	///
 	/// Sends the URLs of the bcsymbolmap files after copying.
-	public func copyBCSymbolMapToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
-		return UUIDsForFramework(frameworkURL)
-			|> flatMap(.Merge) { uuids in SignalProducer(values: uuids) }
-			|> flatMap(.Merge) { uuid in
-				let fileName = uuid.UUIDString.stringByAppendingPathExtension("bcsymbolmap")!
-				let fileURL = directoryURL.URLByAppendingPathComponent(fileName)
-				if fileURL.checkResourceIsReachableAndReturnError(nil) {
-					let destinationDirectoryURL = frameworkURL.URLByDeletingLastPathComponent!
-					let destinationURL = destinationDirectoryURL.URLByAppendingPathComponent(fileName)
-					let resolvedDestinationURL = destinationURL.URLByResolvingSymlinksInPath!
-					return copyProduct(fileURL, resolvedDestinationURL)
-				} else {
-					return SignalProducer.empty
-				}
-			}
+	public func copyBCSymbolMapsToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+		return BCSymbolMapsForFramework(frameworkURL)
+			|> map { fileURL in (target: fileURL, source: directoryURL.URLByAppendingPathComponent(fileURL.lastPathComponent!, isDirectory: false)) }
+			|> filter { $0.source.checkResourceIsReachableAndReturnError(nil) }
+			|> flatMap(.Merge) { copyProduct($0.source, $0.target.URLByResolvingSymlinksInPath!) }
 	}
 
 	/// Checks out the given project into its intended working directory,
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -367,6 +367,10 @@ public final class Project {
 								|> flatMap(.Merge, self.copyFrameworkToBuildFolder)
 								|> flatMap(.Merge) { frameworkURL in
 									return self.copyDSYMToBuildFolderForFramework(frameworkURL, fromDirectoryURL: directoryURL)
+										|> then(SignalProducer(value: frameworkURL))
+								}
+								|> flatMap(.Merge) { frameworkURL in
+									return self.copyBCSymbolMapToBuildFolderForFramework(frameworkURL, fromDirectoryURL: directoryURL)
 								}
 								|> on(completed: {
 									_ = NSFileManager.defaultManager().trashItemAtURL(checkoutDirectoryURL, resultingItemURL: nil, error: nil)
@@ -456,6 +460,31 @@ public final class Project {
 				let resolvedDestinationURL = destinationURL.URLByResolvingSymlinksInPath!
 
 				return copyProduct(dSYMURL, resolvedDestinationURL)
+			}
+	}
+	
+	/// Copies any *.bcsymbolmap files matching the given framework and contained
+	/// within the given directory URL to the directory that the framework
+	/// resides within.
+	///
+	/// If no bcsymbolmap files are found for the given framework, completes with
+	/// no values.
+	///
+	/// Sends the URLs of the bcsymbolmap files after copying.
+	public func copyBCSymbolMapToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+		return UUIDsForFramework(frameworkURL)
+			|> flatMap(.Merge) { uuids in SignalProducer(values: uuids) }
+			|> flatMap(.Merge) { uuid in
+				let fileName = uuid.UUIDString.stringByAppendingPathExtension("bcsymbolmap")!
+				let fileURL = directoryURL.URLByAppendingPathComponent(fileName)
+				if fileURL.checkResourceIsReachableAndReturnError(nil) {
+					let destinationDirectoryURL = frameworkURL.URLByDeletingLastPathComponent!
+					let destinationURL = destinationDirectoryURL.URLByAppendingPathComponent(fileName)
+					let resolvedDestinationURL = destinationURL.URLByResolvingSymlinksInPath!
+					return copyProduct(fileURL, resolvedDestinationURL)
+				} else {
+					return SignalProducer.empty
+				}
 			}
 	}
 
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -469,10 +469,11 @@ public final class Project {
 	///
 	/// Sends the URLs of the bcsymbolmap files after copying.
 	public func copyBCSymbolMapsToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
-		return BCSymbolMapsForFramework(frameworkURL)
-			|> map { fileURL in (target: fileURL, source: directoryURL.URLByAppendingPathComponent(fileURL.lastPathComponent!, isDirectory: false)) }
-			|> filter { $0.source.checkResourceIsReachableAndReturnError(nil) }
-			|> flatMap(.Merge) { copyProduct($0.source, $0.target.URLByResolvingSymlinksInPath!) }
+		return BCSymbolMapsForFramework(frameworkURL, inDirectoryURL: directoryURL)
+			|> flatMap(.Merge) { source in
+				let target = frameworkURL.URLByDeletingLastPathComponent!.URLByAppendingPathComponent(source.lastPathComponent!)
+				return copyProduct(source, target.URLByResolvingSymlinksInPath!)
+			}
 	}
 
 	/// Checks out the given project into its intended working directory,
@@ -636,7 +637,7 @@ private func cacheDownloadedBinary(downloadURL: NSURL, toURL cachedURL: NSURL) -
 
 /// Sends the URL to each file found in the given directory conforming to the
 /// given type identifier.
-private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> SignalProducer<NSURL, CarthageError> {
+private func filesInDirectory(directoryURL: NSURL, #typeIdentifier: String) -> SignalProducer<NSURL, CarthageError> {
 	return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: NSDirectoryEnumerationOptions.SkipsHiddenFiles | NSDirectoryEnumerationOptions.SkipsPackageDescendants, catchErrors: true)
 		|> map { enumerator, URL in URL }
 		|> filter { URL in
@@ -645,6 +646,17 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
 					return UTTypeConformsTo(identifier, typeIdentifier) != 0
 				}, ifFailure: { _ in false })
 		}
+}
+
+/// Sends the URL to each file found in the given directory that has the
+/// given file extension.
+///
+/// Note: If a proper UTI is defined for the file extension,
+///	      `filesInDirectory(_:typeIdentifier:)` is preferred.
+private func filesInDirectory(directoryURL: NSURL, #fileExtension: String) -> SignalProducer<NSURL, CarthageError> {
+	return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL, includingPropertiesForKeys: [], options: .SkipsHiddenFiles | .SkipsPackageDescendants, catchErrors: true)
+		|> map { enumerator, URL in URL }
+		|> filter { URL in URL.pathExtension == fileExtension }
 }
 
 /// Sends the URL for the Info.plist of the specified Framework.
@@ -712,7 +724,7 @@ private func platformForInfoPlist(plistURL: NSURL) -> SignalProducer<Platform, C
 
 /// Sends the URL to each framework bundle found in the given directory.
 private func frameworksInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
-	return filesInDirectory(directoryURL, kUTTypeFramework as! String)
+	return filesInDirectory(directoryURL, typeIdentifier: kUTTypeFramework as! String)
 		|> filter { URL in
 			// Skip nested frameworks
 			let frameworksInURL = URL.pathComponents?.filter { pathComponent in
@@ -724,7 +736,7 @@ private func frameworksInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL,
 
 /// Sends the URL to each dSYM found in the given directory
 private func dSYMsInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
-	return filesInDirectory(directoryURL, "com.apple.xcode.dsym")
+	return filesInDirectory(directoryURL, typeIdentifier: "com.apple.xcode.dsym")
 }
 
 /// Sends the URL of the dSYM whose UUIDs match those of the given framework, or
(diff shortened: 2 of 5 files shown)


ikesyo:
Is this a related thread? https://forums.developer.apple.com/thread/14729

kballard:
Yes. The error quoted by the thread poster is what this PR (plus the shell script phase) fixes.

pizthewiz:
I ran into the same missing bcsymbolmap error and using this PR paired with a run script build phase got things all sorted - thanks @kballard 🌮!

mdiep:
@ikesyo Would you mind reviewing this? I think you know more about bitcode than I do.

ikesyo:
@mdiep Sure, but that would be a few days later. :sweat_smile:

kballard:
Some context: Like dSYMs, archiving a build in Xcode copies all the `*.bcsymbolmap` files, one per architecture for every binary that has bitcode, and it puts them in the "BCSymbolMaps" folder in the archive. If you try to archive a build that has bitcode but there is no `*.bcsymbolmap` file for the binary, the archive fails with an error that looks like

> ```
"The archive did not contain <DVTFilePath:0x7f88ca959ef0:'/Users/foo/Library/Developer/Xcode/Archives/2015-08-17/Project 2015-08-17 14.46.29.xcarchive/BCSymbolMaps/DCF72CBF-7889-372D-B549-2CFA181C32CA.bcsymbolmap'> as expected."
```

And again, like dSYMs, Xcode expects these files to be in the built products directory. So what this PR does is it teaches Carthage to copy any relevant `*.bcsymbolmap` for every file it copies into the `Carthage/Builds/$plat/` folder. Similarly, it teaches it to archive the files, and, if I didn't screw up, it teaches it to copy them from a prebuilt archive that it's installing.

**Comment by NachoSoto** [comment]
This makes a lot of sense to me. Great job @kballard

**Comment by kballard** [comment]
I'm currently working on making `copy-frameworks` copy the `bcsymbolmap` files automatically for archive builds. If I finish it before this is reviewed I could push it to this PR, as I assume making it do that is non-controversial, although it would mean that the `copy-frameworks` command would become a permanent requirement rather than a workaround for the app store submission issue (though it seems like a good thing to have permanently anyway as ad-hoc builds wouldn't want to keep the dead weight of the simulator slices anyway).

**Comment by kballard** [comment]
Just pushed the changes to `carthage copy-frameworks`. It now has a flag `--copy-bcsymbolmaps` that defaults to enabled for archive builds (whenever `ACTION=install`), which causes it to copy any bcsymbolmap files that it finds for each framework it copies. This way everyone's existing `carthage copy-frameworks` build phase will suffice to make bitcode work.

File: Source/CarthageKit/Project.swift [hunk 1]
```diff
@@ -367,6 +367,10 @@ public final class Project {
 								|> flatMap(.Merge, self.copyFrameworkToBuildFolder)
 								|> flatMap(.Merge) { frameworkURL in
 									return self.copyDSYMToBuildFolderForFramework(frameworkURL, fromDirectoryURL: directoryURL)
+										|> then(SignalProducer(value: frameworkURL))
```
ikesyo: This can be changed to just `|> map { _ in frameworkURL }`.

ikesyo: Or we could do as the following:

```swift
return self.copyDSYMToBuildFolderForFramework(frameworkURL, fromDirectoryURL: directoryURL)
	|> then(self.copyBCSymbolMapToBuildFolderForFramework(frameworkURL, fromDirectoryURL: directoryURL))
```

Source/CarthageKit/Project.swift
```diff
@@ -458,6 +462,31 @@ public final class Project {
...
+	/// no values.
+	///
+	/// Sends the URLs of the bcsymbolmap files after copying.
+	public func copyBCSymbolMapToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+		return UUIDsForFramework(frameworkURL)
+			|> flatMap(.Merge) { uuids in SignalProducer(values: uuids) }
+			|> flatMap(.Merge) { uuid in
+				let fileName = uuid.UUIDString.stringByAppendingPathExtension("bcsymbolmap")!
+				let fileURL = directoryURL.URLByAppendingPathComponent(fileName)
+				if fileURL.checkResourceIsReachableAndReturnError(nil) {
```
ikesyo: Could you factor out the logic above into separate functions as same with `dSYMsInDirectory()` and `dSYMForFramework()`?

Source/CarthageKit/Xcode.swift
```diff
@@ -664,6 +666,31 @@ private func copyBuildProductIntoDirectory(directoryURL: NSURL, settings: BuildS
...
+}
+
+/// Finds any *.bcsymbolmap files for the built product and copies them into
+/// the given folder. Does nothing if bitcode is disabled.
+///
+/// Returns a signal that will send the URL after copying for each file.
+private func copyBCSymbolMapForBuildProductIntoDirectory(directoryURL: NSURL, settings: BuildSettings) -> SignalProducer<NSURL, CarthageError> {
+	if settings.bitcodeEnabled.value == true {
+		return SignalProducer(result: settings.builtProductsDirectoryURL &&& settings.executableURL)
+			|> flatMap(.Merge) { (products, exe) in
```
ikesyo: (`builtProductsURL` | `productsURL`) and `executableURL` are more appropriate.

Source/carthage/CopyFrameworks.swift
```diff
@@ -93,3 +129,7 @@ private func inputFiles() -> SignalProducer<String, CarthageError> {
 				|> flatten(.Concat)
 		}
 }
+
+private func buildAction() -> String? {
```
ikesyo: How about naming `isArchiveBuildAction` or something and returning `Bool` as with `codeSigningAllowed()`?

Source/carthage/CopyFrameworks.swift
```diff
@@ -18,29 +18,61 @@ public struct CopyFrameworksCommand: CommandType {
...
+										}
+									}
+							}
+					}
+					|> promoteErrors
+			}
+			|> waitOnCommand
+	}
+	
+	private func copyBCSymbolMapsForFramework(frameworkURL: NSURL, fromDirectory directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
```
ikesyo: Could this be written as a free function?

Source/carthage/CopyFrameworks.swift
```diff
@@ -18,29 +18,61 @@ public struct CopyFrameworksCommand: CommandType {
...
+		return UUIDsForFramework(frameworkURL)
+			|> flatMap(.Merge) { uuids in SignalProducer(values: uuids) }
+			|> map { uuid in directoryURL.URLByAppendingPathComponent(uuid.UUIDString).URLByAppendingPathExtension("bcsymbolmap") }
+			|> filter { url in url.checkResourceIsReachableAndReturnError(nil) }
+			|> flatMap(.Merge) { url in
+				return SignalProducer(result: builtProductsFolder())
+					|> flatMap(.Merge) { builtProducts in
```
ikesyo: `builtProductsDir`?

Source/carthage/Archive.swift
```diff
@@ -25,7 +25,15 @@ public struct ArchiveCommand: CommandType {
 					|> flatMap(.Concat) { platform in
 						let framework = platform.relativePath.stringByAppendingPathComponent(options.frameworkName).stringByAppendingPathExtension("framework")!
 						let dSYM = framework.stringByAppendingPathExtension("dSYM")!
-						return SignalProducer(values: [framework, dSYM])
+						if NSFileManager.defaultManager().fileExistsAtPath(framework) {
+							let frameworkURL = NSURL(fileURLWithPath: framework)!
+							let bcsymbolmapProducer = UUIDsForFramework(frameworkURL)
```
ikesyo: `bcsymbolmapsProducer` (plural)

Source/CarthageKit/Project.swift
```diff
@@ -458,6 +462,31 @@ public final class Project {
...
+	/// within the given directory URL to the directory that the framework
+	/// resides within.
+	///
+	/// If no bcsymbolmap files are found for the given framework, completes with
+	/// no values.
+	///
+	/// Sends the URLs of the bcsymbolmap files after copying.
+	public func copyBCSymbolMapToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+		return UUIDsForFramework(frameworkURL)
+			|> flatMap(.Merge) { uuids in SignalProducer(values: uuids) }
```
ikesyo: `UUIDs`

Source/CarthageKit/Project.swift
```diff
@@ -458,6 +462,31 @@ public final class Project {
...
+			|> flatMap(.Merge) { uuid in
```
ikesyo: `UUID`

Source/CarthageKit/Xcode.swift
```diff
@@ -664,6 +666,31 @@ private func copyBuildProductIntoDirectory(directoryURL: NSURL, settings: BuildS
...
+				return UUIDsFromDwarfdump(exe)
+					|> flatMap(.Merge) { uuids in SignalProducer(values: uuids) }
```
ikesyo: Ditto

Source/CarthageKit/Xcode.swift
```diff
@@ -664,6 +666,31 @@ private func copyBuildProductIntoDirectory(directoryURL: NSURL, settings: BuildS
...
+					|> map { uuid in products.URLByAppendingPathComponent(uuid.UUIDString).URLByAppendingPathExtension("bcsymbolmap") }
```
ikesyo: Ditto

Source/carthage/Archive.swift
```diff
@@ -25,7 +25,15 @@ public struct ArchiveCommand: CommandType {
...
+								|> flatMap(.Merge) { uuids in SignalProducer(values: uuids) }
```
ikesyo: Ditto

Source/carthage/Archive.swift
```diff
@@ -25,7 +25,15 @@ public struct ArchiveCommand: CommandType {
...
+								|> map { uuid in platform.relativePath.stringByAppendingPathComponent(uuid.UUIDString).stringByAppendingPathExtension("bcsymbolmap")! }
```
ikesyo: Ditto

Source/carthage/CopyFrameworks.swift
```diff
@@ -18,29 +18,61 @@ public struct CopyFrameworksCommand: CommandType {
...
+							}
+					}
+					|> promoteErrors
+			}
+			|> waitOnCommand
+	}
+	
+	private func copyBCSymbolMapsForFramework(frameworkURL: NSURL, fromDirectory directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+		return UUIDsForFramework(frameworkURL)
+			|> flatMap(.Merge) { uuids in SignalProducer(values: uuids) }
```
ikesyo: Ditto

Source/carthage/CopyFrameworks.swift
```diff
@@ -18,29 +18,61 @@ public struct CopyFrameworksCommand: CommandType {
...
+			|> map { uuid in directoryURL.URLByAppendingPathComponent(uuid.UUIDString).URLByAppendingPathExtension("bcsymbolmap") }
```
ikesyo: Ditto

Source/carthage/Archive.swift
```diff
@@ -25,7 +25,15 @@ public struct ArchiveCommand: CommandType {
 					|> flatMap(.Concat) { platform in
 						let framework = platform.relativePath.stringByAppendingPathComponent(options.frameworkName).stringByAppendingPathExtension("framework")!
 						let dSYM = framework.stringByAppendingPathExtension("dSYM")!
-						return SignalProducer(values: [framework, dSYM])
+						if NSFileManager.defaultManager().fileExistsAtPath(framework) {
```
ikesyo: Is this check required? The following might be an alternative:

```swift
UUIDsForFramework(frameworkURL)
	|> catch { _ in .empty }
```

Source/carthage/Archive.swift
```diff
@@ -25,7 +25,15 @@ public struct ArchiveCommand: CommandType {
...
+							let frameworkURL = NSURL(fileURLWithPath: framework)!
+							let bcsymbolmapProducer = UUIDsForFramework(frameworkURL)
+								|> flatMap(.Merge) { uuids in SignalProducer(values: uuids) }
+								|> map { uuid in platform.relativePath.stringByAppendingPathComponent(uuid.UUIDString).stringByAppendingPathExtension("bcsymbolmap")! }
+							return SignalProducer(values: [framework, dSYM]) |> concat(bcsymbolmapProducer)
```
ikesyo: Can you separate the operations into each line?

Source/CarthageKit/Project.swift
```diff
@@ -458,6 +462,31 @@ public final class Project {
...
+	
+	/// Copies any *.bcsymbolmap files matching the given framework and contained
+	/// within the given directory URL to the directory that the framework
+	/// resides within.
+	///
+	/// If no bcsymbolmap files are found for the given framework, completes with
+	/// no values.
+	///
+	/// Sends the URLs of the bcsymbolmap files after copying.
+	public func copyBCSymbolMapToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
```
ikesyo: `copyBCSymbolMaps` (plural)

ikesyo:
@kballard Thanks for your detailed explanation! :sparkling_heart:  The logic looks good. :+1:

Source/carthage/CopyFrameworks.swift
```diff
@@ -18,29 +18,61 @@ public struct CopyFrameworksCommand: CommandType {
...
+	let copyBCSymbolMaps: Bool
+	
+	static func create(copyBCSymbolMaps: Bool) -> CopyFrameworksOptions {
+		return self(copyBCSymbolMaps: copyBCSymbolMaps)
+	}
+	
+	static func evaluate(m: CommandMode) -> Result<CopyFrameworksOptions, CommandantError<CarthageError>> {
+		let actionIsArchive = buildAction() == "install" // archives use ACTION=install
+		return create
+			<*> m <| Option(key: "copy-bcsymbolmaps", defaultValue: actionIsArchive, usage: "copy the bcsymbolmap files for each product built with bitcode (enabled by default for archive builds)")
```
ikesyo: I'm not sure about adding the option is needed. How do you think? @mdiep and @NachoSoto

Source/carthage/Archive.swift
kballard: It's preferable to invoke `dwarfdump` and catch an error than it is to just check if the input file exists to begin with?

That would also squelch any other legitimate errors arising from `dwarfdump`, which may or may not be preferable.

Source/carthage/CopyFrameworks.swift
kballard: I wasn't sure if there's any situation in which someone might want to copy these for non-archive builds, or similarly if anyone would want to explicitly avoid copying them for archives for some reason (e.g. maybe they have a shell script phase after `copy-frameworks` that strips the bitcode out of all the frameworks).

Source/carthage/Archive.swift
kballard: I'm refactoring this a little, but I'm going to keep the file check here, because I don't like the idea of suppressing other errors arising from the `dwarfdump`.

Source/carthage/CopyFrameworks.swift
mdiep: I don't think we need to have an option. I'd rather leave things simple and add it later if someone requests it with a compelling reason.

Source/carthage/CopyFrameworks.swift
kballard: It could. I made it a method because I was using Build.swift as an example.

Source/carthage/CopyFrameworks.swift
kballard: After considering it some more, I think it's plausible that someone might be building an IPA file without using Xcode's `archive` command, which means they'd want a way to opt-in to copying the bcsymbolmaps.

Source/carthage/CopyFrameworks.swift
kballard: I'll go with `builtProductsURL`, like your [previous suggestion](https://github.com/Carthage/Carthage/pull/853/files#r42368919).

Source/carthage/CopyFrameworks.swift
kballard: The other reason I wanted the option was because it's a convenient place to document that `carthage copy-frameworks` will copy bcsymbolmaps on archive builds. Putting that in the command description makes an already-extremely-long description even more ridiculously long.

kballard:
PR updated. I left the option on `carthage copy-frameworks`, see [my comments](https://github.com/Carthage/Carthage/pull/853#discussion_r42420866) for a reason. I can remove it if you feel strongly about it.

README.md
```diff
@@ -67,7 +67,7 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
   $(SRCROOT)/Carthage/Build/iOS/ReactiveCocoa.framework
   ```
 
-  This script works around an [App Store submission bug](http://www.openradar.me/radar?id=6409498411401216) triggered by universal binaries.
+  This script works around an [App Store submission bug](http://www.openradar.me/radar?id=6409498411401216) triggered by universal binaries as well as ensures that necessary bitcode-related files are also copied.
```
NachoSoto: +1

kballard: According to the dictionary, "as well as" is defined, when used as a conjunction, as "and in addition", or "and also". Were I to write this as "This script works around an App Store submission bug and also ensures that necessary bitcode-related files are also copied", you'd probably think that looks fine.

In fact, I'll just go ahead and rephrase that. I think it sounds better written that way regardless.

kballard: Actually, that's a bit repetitive on the word "also". So maybe it doesn't sound as good.

mdiep: This does read a little funny to me. I'd write it like this:

> This script works around an [App Store submission bug](http://www.openradar.me/radar?id=6409498411401216) triggered by universal binaries and ensures that necessary bitcode-related files are copied.

kballard: Fair enough. I'll reword it.

kballard:
@ikesyo Any more comments on the new version?

Source/CarthageKit/Project.swift
```diff
@@ -458,6 +459,21 @@ public final class Project {
...
+	/// Copies any *.bcsymbolmap files matching the given framework and contained
+	/// within the given directory URL to the directory that the framework
+	/// resides within.
+	///
+	/// If no bcsymbolmap files are found for the given framework, completes with
+	/// no values.
+	///
+	/// Sends the URLs of the bcsymbolmap files after copying.
+	public func copyBCSymbolMapsToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+		return BCSymbolMapsForFramework(frameworkURL)
```
ikesyo: I think the implementation does not work for some archives those are not created from `carthage archive`, when a framework and its bcsymbolmap files are not in the same directory.

kballard: What archives are created that way? And I'm not sure how Carthage can support that anyway. If it doesn't know where to find the bcsymbolmaps, then it can't possibly install them, and the behavior will be the same as if the archive doesn't have bcsymbolmaps included at all.

kballard: On a similar note, Carthage already assumes the dSYM will be in the same folder as the framework, and that's not seen as a problem. Sure, if the dSYM is missing, Xcode won't fail to archive, but I think that's an argument in favor of ensuring your bcsymbolmap files are in the same folder as your framework, rather than, well, I'm not sure what the alternative is anyway, unless you want to search the entire archive for the file.

ikesyo: `dSYMsInDirectory()` is using `filesInDirectory()`, that traverse the given directory and its children.

kballard: Oh so it does. I never looked at the implementation of `filesInDirectory()` so I assumed it was only looking at the direct children.

Source/CarthageKit/Project.swift
```diff
@@ -728,6 +756,47 @@ private func dSYMForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL:
...
+					|> startWithSignal { signal, signalDisposable in
+						disposable += signalDisposable
+						signal.observe(Signal.Observer { event in
+							switch event {
+							case let .Next(value):
+								if let basename = value.value.lastPathComponent?.stringByDeletingPathExtension, fileUUID = NSUUID(UUIDString: basename) {
+									if remainingUUIDs.remove(fileUUID) != nil {
+										observer.put(event)
+										if remainingUUIDs.isEmpty {
+											sendCompleted(observer)
```
kballard: This seems really complicated. Is it really all necessary to get the effect I'm looking for (as documented in the comment a few lines up)?

Source/CarthageKit/Project.swift
```diff
@@ -631,6 +648,17 @@ private func filesInDirectory(directoryURL: NSURL, typeIdentifier: String) -> Si
 		}
 }
 
+/// Sends the URL to each file found in the given directory that has the
+/// given file extension.
+///
+/// Note: If a proper UTI is defined for the file extension,
+///	      `filesInDirectory(_:typeIdentifier:)` is preferred.
+private func filesInDirectory(directoryURL: NSURL, #fileExtension: String) -> SignalProducer<NSURL, CarthageError> {
+	return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL, includingPropertiesForKeys: [], options: .SkipsHiddenFiles | .SkipsPackageDescendants, catchErrors: true)
```
kballard: "bcsymbolmap" files have no defined UTI, and it turns out that if I generate a UTI from the file extension "bcsymbolmap" it doesn't match the UTI that Spotlight actually comes up with. If I generate a UTI that conforms to "public.data" then it matches, but I don't want to rely on that. Therefore, we need to test file extension directly instead.

kballard:
@ikesyo The new version should find bcsymbolmaps from anywhere in the archive. That said, I have no idea how to actually test this, because I don't know of any GitHub projects with archives that already include bcsymbolmap files.

Source/CarthageKit/Project.swift
ikesyo: This seems over-complicated to me too.

> Is it really all necessary to get the effect I'm looking for (as documented in the comment a few lines up)?

I would say no to that.

Source/CarthageKit/Project.swift
ikesyo: How about overloading `filesInDirectory(directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError>` or changing the existing one to `filesInDirectory(directoryURL: NSURL, typeIdentifier: String?) -> SignalProducer<NSURL, CarthageError>`, then filter the result from the caller?

Source/carthage/CopyFrameworks.swift
ikesyo: I came to the thought that would be unnecessary.

> maybe they have a shell script phase after copy-frameworks that strips the bitcode out of all the frameworks

I don't know whether there is a way to do that (stripping bitcode) or not.

> After considering it some more, I think it's plausible that someone might be building an IPA file without using Xcode's archive command, which means they'd want a way to opt-in to copying the bcsymbolmaps.

Taking it into consideration, we should always copy bcsymbolmaps regardless of the build action. Maybe that would not have any bad side effects?

> The other reason I wanted the option was because it's a convenient place to document that carthage copy-frameworks will copy bcsymbolmaps on archive builds. Putting that in the command description makes an already-extremely-long description even more ridiculously long.

This seems not to be a valid reason for adding the option / complexity.

Source/CarthageKit/Xcode.swift
```diff
@@ -664,6 +666,27 @@ private func copyBuildProductIntoDirectory(directoryURL: NSURL, settings: BuildS
...
+			return copyBCSymbolMapForBuildProductIntoDirectory(directoryURL, settings)
+				|> then(SignalProducer(value: url))
+		}
+}
+
+/// Finds any *.bcsymbolmap files for the built product and copies them into
+/// the given folder. Does nothing if bitcode is disabled.
+///
+/// Returns a signal that will send the URL after copying for each file.
+private func copyBCSymbolMapForBuildProductIntoDirectory(directoryURL: NSURL, settings: BuildSettings) -> SignalProducer<NSURL, CarthageError> {
```
ikesyo: Could you change to `copyBCSymbolMaps...` here too?

Source/CarthageKit/Project.swift
kballard: Do you have any suggestions for how to achieve the effect in a simpler way, then? I've never touched RAC before, so I'm sure there's plenty I'm missing, but I don't see any obvious way to easily handle this.

ikesyo: I'm sorry, I means that there would be no need to do the complicated thing.

ikesyo: Or using `|> take(UUIDs.count)` may be the way.

Source/CarthageKit/Project.swift
```diff
@@ -458,6 +459,22 @@ public final class Project {
...
+	///
+	/// If no bcsymbolmap files are found for the given framework, completes with
+	/// no values.
+	///
+	/// Sends the URLs of the bcsymbolmap files after copying.
+	public func copyBCSymbolMapsToBuildFolderForFramework(frameworkURL: NSURL, fromDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+		return BCSymbolMapsForFramework(frameworkURL, inDirectoryURL: directoryURL)
+			|> flatMap(.Merge) { source in
+				let target = frameworkURL.URLByDeletingLastPathComponent!.URLByAppendingPathComponent(source.lastPathComponent!)
+				return copyProduct(source, target.URLByResolvingSymlinksInPath!)
```
ikesyo: Could you imitate the style of `copyDSYMToBuildFolderForFramework()` above for consistency and readability?

Source/CarthageKit/Project.swift
kballard: That wouldn't solve the problem of finding two bcsymbolmap files for the same UUID. It works for `dSYMForFramework()` because that function only ever wants to send a single URL (that matches all the UUIDs), but this function is sending multiple URLs, one for each unique UUID.

Source/carthage/CopyFrameworks.swift
kballard: > I don't know whether there is a way to do that (stripping bitcode) or not.

There is. `xcrun bitcode_strip` can do it. See `man bitcode_strip` for details.

> Taking it into consideration, we should always copy bcsymbolmaps regardless of the build action. Maybe that would not have any bad side effects?

AFAIK it's completely unnecessary for normal development, so I'd rather not do any more work than necessary in a build phase that's executed on every single build during development. Re-copying the frameworks every time is bad enough.

kballard:
@ikesyo PR updated as per comments.

Source/CarthageKit/Project.swift
ikesyo: I don't understand why you are so concerned about the situation. :confused:

kballard: Because I want the code to behave correctly. If an archive accidentally has two copies of the same bcsymbolmap in different folders, carthage should still work right instead of failing to copy the bcsymbolmap for other architectures. The error case here is very opaque (Xcode giving a hard-to-read error about a missing file in the archive) and confusing, so we should strive to handle foreseeable edge cases.

ichibod:
So it looks like this will take care of the other issue I'm seeing with the `DVTFilePath`

kballard:
@ichibod The one that says something like this?

> ```The archive did not contain <DVTFilePath:0x7f88ca959ef0:'/Users/username/Library/Developer/Xcode/Archives/2015-08-17/Project 2015-08-17 14.46.29.xcarchive/BCSymbolMaps/DCF72CBF-7889-372D-B549-2CFA181C32CA.bcsymbolmap'> as expected.

That is the issue this takes care of, that's not some "other" issue. That's Xcode complaining that the bcsymbolmap wasn't found in the archive, which happened because the bcsymbolmap wasn't in the built products folder when generating the archive.

ichibod:
@kballard Correct yes. It just happened to be the next issue in my chain of issues of having to implement Carthage for the other framework that has apparently dropped support for git submodule (i.e. manual) installation.

Source/CarthageKit/Project.swift
ikesyo: If we should behave as you described, the following might works (I don't test it though):

```swift
return UUIDsForFramework(frameworkURL)
  |> flatMap(.Merge) { UUIDs in
    return BCSymbolMapsInDirectory(directoryURL)
      |> reduce([NSUUID: NSURL](minimumCapacity: UUIDs.count)) { URLsByUUID, fileURL in
        let fileUUID = fileURL.lastPathComponent?.stringByDeletingPathExtension.flatMap { NSUUID(UUIDString: $0) }
        if let fileUUID = fileUUID where UUIDs.contains(fileUUID) && URLsByUUID[fileUUID] == nil {
          URLsByUUID[fileUUID] = fileURL
        }
        return URLsByUUID
      }
  }
  |> flatMap(.Merge) { SignalProducer(values: $0.values) }
```

kballard: So there is a `reduce()`! I actually looked for one and didn't find it. Still, it has the downside of emitting a single value of all the URLs, which I find to be rather annoying, and you can only break it back out into individual URLs once the entire tree walk is done. Furthermore, there's no way to stop the tree walk early once you've found all the files, which just doesn't sit well with me.

ikesyo: So how about this, using `scan`, `takeWhile` and `takeLast`:

```swift
return UUIDsForFramework(frameworkURL)
  |> flatMap(.Merge) { UUIDs in
    let desiredCount = UUIDs.count
    return BCSymbolMapsInDirectory(directoryURL)
      |> scan([NSUUID: NSURL](minimumCapacity: desiredCount)) { URLsByUUID, fileURL in
        let fileUUID = fileURL.lastPathComponent?.stringByDeletingPathExtension.flatMap { NSUUID(UUIDString: $0) }
        if let fileUUID = fileUUID where UUIDs.contains(fileUUID) && URLsByUUID[fileUUID] == nil {
          URLsByUUID[fileUUID] = fileURL
        }
        return URLsByUUID
      }
      |> takeWhile { $0.count <= desiredCount }
      |> takeLast(1)
  }
  |> flatMap(.Merge) { SignalProducer(values: $0.values) }
```

kballard: This still has the downside where it waits for all of the URLs to be found before it sends any of them. Also, incidentally, it's almost certainly going to end up making a copy of the dictionary every time it mutates it, which is rather unfortunate.

ikesyo: > This still has the downside where it waits for all of the URLs to be found before it sends any of them.

That should not be true, because unlike `reduce`, `scan` sends values each time when the source signal send `.Next` events.

> Also, incidentally, it's almost certainly going to end up making a copy of the dictionary every time it mutates it, which is rather unfortunate (although in this case the dictionary will be so small that it's largely academic).

So if that might be a huge problem in Carthage, should we change all the usages of `scan` and `reduce` in the entire codebase? We are not tackling theoretical problem here.

kballard: Yeah, it's the `takeLast` that causes it.

> We are not tackling theoretical problem here.

Honestly, Carthage has a ton of inefficiencies everywhere. But that's not a particularly good justification for writing more inefficient code. And writing more efficient code here does not necessitate going back and fixing inefficiencies elsewhere.

ikesyo: Does it satisfies you? 

```swift
return UUIDsForFramework(frameworkURL)
  |> flatMap(.Merge) { UUIDs in
    let desiredCount = UUIDs.count
    var URLsByUUID = [NSUUID: NSURL](minimumCapacity: desiredCount)
    return BCSymbolMapsInDirectory(directoryURL)
      |> filter { fileURL in
        let fileUUID = fileURL.lastPathComponent?.stringByDeletingPathExtension.flatMap { NSUUID(UUIDString: $0) }
        if let fileUUID = fileUUID where UUIDs.contains(fileUUID) && URLsByUUID[fileUUID] == nil {
          URLsByUUID[fileUUID] = fileURL
          return true
        } else {
          return false
        }
      }
      |> take(desiredCount)
  }
```

ikesyo: > Honestly, Carthage has a ton of inefficiencies everywhere. But that's not a particularly good justification for writing more inefficient code. And writing more efficient code here does not necessitate going back and fixing inefficiencies elsewhere.

Pull requests are welcome.

kballard: > Does it satisfy you?

If I understand the `SignalProducer` operators correctly, any `SignalProducer` returned from the transform given to `flatMap` will only ever be started a maximum of once, right? Since only `flatten` ever actually gets ahold of those `SignalProducers` and it starts each producer it gets only once? As long as the `SignalProducer` returned by the transform is started at most once, then that code is safe and would work. If there's any way for the `SignalProducer` to be started multiple times, then they'd end up sharing the same dictionary and would misbehave.

If that assumption is valid, then I'd still probably tweak what you have to just subtracting elements from a copy of the `UUIDs` Set instead of building a dictionary, as the only info we need to persist is whether a particular UUID is valid.

Can you confirm that my assumption about `SignalProducers` returned from the transform given to `flatMap` is valid?

mdiep:
What's the status of this PR? I was going to wait for it to cut a release. But if it's going to be a while, then I'll go ahead and release `master` now.

kballard:
@mdiep As far as I'm concerned, it's good. I'm willing to rewrite the bit of code talked about in [this comment thread](https://github.com/Carthage/Carthage/pull/853#discussion_r42924549) if someone can confirm my stated assumption about RAC, but what I have there already should work perfectly fine. I also think this change is important enough that it needs to get into a release as soon as possible (it's a showstopper for anyone who's using Carthage and bitcode together as they can't archive their builds otherwise).

I haven't actually tested the case of installing a prebuilt archive that contains bitcode (because no github archives currently contain it), but I'm tempted to set up a dummy framework, push it to GitHub, and check that it works right now.

pizthewiz:
> important enough that it needs to get into a release as soon as possible

Yeah, I don't know how anyone using Carthage is able to upload and distribute Xcode 7.0+ compiled builds via the App Store/TestFlight without this and in my experience, disabling Bitcode was problematic as well and also caused issues at app install time.

kballard:
Ok, just tested it. `carthage archive` embedded the bcsymbolmaps correctly and using that framework in another project (and verifying that it used the prebuilt binary) successfully copied the bcsymbolmaps and archived correctly.

Source/CarthageKit/Project.swift
ikesyo: > If I understand the SignalProducer operators correctly, any SignalProducer returned from the transform given to flatMap will only ever be started a maximum of once, right? Since only flatten ever actually gets ahold of those SignalProducers and it starts each producer it gets only once?

The closure will be called for each `.Next` event from the source producer.

> As long as the SignalProducer returned by the transform is started at most once, then that code is safe and would work. If there's any way for the SignalProducer to be started multiple times, then they'd end up sharing the same dictionary and would misbehave.

No. The dictionary should not be shared because it is a closure-local variable, and per-`.Next` events, the closure is called afresh.

Source/carthage/CopyFrameworks.swift
ikesyo: I'd rather not to maintain the complexity... :worried:

Source/CarthageKit/Project.swift
kballard: Right, but the worry is that the dictionary is actually being captured by a `SignalProducer`, not just a `Signal`, and if that `SignalProducer` were to be started multiple times, the resulting signals would all misbehave. My belief is that, because the `SignalProducer` is immediately handed back to `flatMap`, the merge strategies only ever started the `SignalProducer` once and it's not accessible to anyone else. I just want confirmation that there's absolutely no way to have that `SignalProducer` object get started multiple times.

Source/carthage/CopyFrameworks.swift
kballard: Is adding one option really that complex? It seems pretty straightforward, and all the other commands (with the exception of the trivial `version`) already do all this.

Source/CarthageKit/Project.swift
ikesyo: > Right, but the worry is that the dictionary is actually being captured by a SignalProducer, not just a Signal, and if that SignalProducer were to be started multiple times, the resulting signals would all misbehave.

I'm not sure why that would misbehave.

kballard: Two signals that share the same dictionary? At best, one of the two signals would return nothing. Or both signals would return only a subset of the URLs. At worst, if they're executing concurrently, that's a threading violation (`Dictionary`, like all of the collections, is not safe to access concurrently from multiple threads if any of them mutate it).

ikesyo: [All signal events are handled serially](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/v3.0.0/Documentation/DesignGuidelines.md#events-are-serial).

kballard: Events in a single stream, i.e. events from a single `Signal`. But a `SignalProducer` is an object that can create multiple signals. These signals should be independent. But the proposed code here actually makes all the created signals share the same dictionary, which is unsafe.

kballard: All that said, I think that I can actually wrap the call to `filter()` in a closure that creates the dictionary there. It's a little confusing because `filter()` is actually a transform on `Signal`s that's being lifted to a `SignalProducer`, which is why I didn't think of it before. I'll see if I can whip that up.

ikesyo: > But the proposed code here actually makes all the created signals share the same dictionary, which is unsafe.

No. The signals created from a producer is fresh ones and don't share anything.

> These signals should be independent.

They are really independent.

https://github.com/ReactiveCocoa/ReactiveCocoa/blob/v3.0.0/Documentation/DesignGuidelines.md#each-produced-signal-may-send-different-events-at-different-times

ikesyo: > My belief is that, because the SignalProducer is immediately handed back to flatMap, the merge strategies only ever started the SignalProducer once and it's not accessible to anyone else. I just want confirmation that there's absolutely no way to have that SignalProducer object get started multiple times.

Anyway what you said above should be correct.

kballard: Signals would normally be independent. But in your code snippet, the `SignalProducer` returned by `BCSymbolMapsInDirectory(directoryURL) |> filter { ... } |> take(desiredCount)` has already captured the dictionary, even though it hasn't started yet. And if you start it several times, all of the resulting `Signal`s end up sharing the same dictionary. This is what had me worried. The version I'm just about to push fixes this by figuring out how to create the shared state at the point where the `SignalProducer` is started (and therefore if it's started multiple times the resulting `Signal`s will not share any state but will be truly independent).

kballard:
@ikesyo I just pushed a new version of the hotly-debated `BCSymbolMapsForFramework` implementation. I have not yet tested this version to make sure it works. I'm about to leave work, but I will test it once I get home.

Source/CarthageKit/Project.swift
ikesyo: Thank you, I finally understand your concern, sorry for the misunderstanding.

kballard:
Just tested, it works. Built and archived a framework with bitcode, uploaded to GitHub, extracted it in another project, and it successfully copied and installed the bcsymbolmap files.

Source/CarthageKit/Project.swift
```diff
@@ -728,6 +754,37 @@ private func dSYMForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL:
...
+/// Sends the URLs of the bcsymbolmap files that match the given framework and are
+/// located somewhere within the given directory.
+private func BCSymbolMapsForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+	return UUIDsForFramework(frameworkURL)
+		|> flatMap(.Merge) { UUIDs in
+			if UUIDs.isEmpty {
+				return SignalProducer.empty
+			}
+			return BCSymbolMapsInDirectory(directoryURL)
+				|> { (signal: Signal<NSURL, CarthageError>) -> Signal<NSURL, CarthageError> in
```
ikesyo: This seems a bit hacky, returning a new SignalProducer would be better to understand?

```swift
return UUIDsForFramework(frameworkURL)
  |> flatMap(.Merge) { UUIDs in
    if UUIDs.isEmpty {
      return .empty
    }
  
    return SignalProducer { observer, disposable in
      var remainingUUIDs = UUIDs
      BCSymbolMapsInDirectory(directoryURL)
        |> filter { fileURL in
          if let fileUUID = fileURL.lastPathComponent?.stringByDeletingPathExtension.flatMap { NSUUID(UUIDString: $0) } {
            return remainingUUIDs.remove(fileUUID) != nil
          } else {
            return false
          }
        }
        |> take(UUIDs.count)
        |> startWithSignal { signal, signalDisposable in
          disposable += signalDisposable
          signal.observe(observer)
        }
    }
  }
```

Source/CarthageKit/Project.swift
```diff
@@ -728,6 +754,37 @@ private func dSYMForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL:
...
+		|> filter { URL in URL.pathExtension == "bcsymbolmap" }
+}
+
+/// Sends the URLs of the bcsymbolmap files that match the given framework and are
+/// located somewhere within the given directory.
+private func BCSymbolMapsForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+	return UUIDsForFramework(frameworkURL)
+		|> flatMap(.Merge) { UUIDs in
+			if UUIDs.isEmpty {
+				return SignalProducer.empty
```
ikesyo: This could be just as `return .empty`.

Source/carthage/CopyFrameworks.swift
ikesyo: > Taking it into consideration, we should always copy bcsymbolmaps regardless of the build action. Maybe that would not have any bad side effects?

I would withdraw what I said above and say:

> I don't think we need to have an option. I'd rather leave things simple and add it later if someone requests it with a compelling reason.

kballard: Man, it makes me really uncomfortable to have the behavior differ based purely on the `ACTION` environment variable without any way to document this or override it, but if you feel it's that important...

Source/CarthageKit/Project.swift
kballard: I actually find the idea of mapping a signal easier to follow than creating a new `SignalProducer` and starting it immediately. I can extract the logic into a local function if that's easier to read:

```swift
return UUIDsForFramework(frameworkURL)
	|> flatMap(.Merge) { UUIDs in
		if UUIDs.isEmpty {
			return .empty
		}
		func filterUUIDs(signal: Signal<NSURL, CarthageError>) -> Signal<NSURL, CarthageError> {
			var remainingUUIDs = UUIDs
			let count = remainingUUIDs.count
			return signal
				|> filter { fileURL in
					if let basename = fileURL.lastPathComponent?.stringByDeletingPathExtension, fileUUID = NSUUID(UUIDString: basename) {
						return remainingUUIDs.remove(fileUUID) != nil
					} else {
						return false
					}
				}
				|> take(count)
		}
		return BCSymbolMapsInDirectory(directoryURL)
			|> filterUUIDs
}
```

kballard:
Updated as per PR feedback.

Source/carthage/CopyFrameworks.swift
```diff
@@ -31,19 +31,38 @@ public struct CopyFrameworksCommand: CommandType {
...
+									}
 								}
 						}
 				}
 				|> promoteErrors
 				|> waitOnCommand
-
 		case .Usage:
-			return .success(())
+			return .success()
```
ikesyo: We are passing empty tuple (`Void`) explicitly in the codebase, could you revert this change?

kballard: Oops, didn't notice that. I just let Xcode autocomplete. I'll fix that in the morning.

Source/CarthageKit/Project.swift
```diff
@@ -728,6 +754,38 @@ private func dSYMForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL:
...
+
+/// Sends the URLs of the bcsymbolmap files that match the given framework and are
+/// located somewhere within the given directory.
+private func BCSymbolMapsForFramework(frameworkURL: NSURL, inDirectoryURL directoryURL: NSURL) -> SignalProducer<NSURL, CarthageError> {
+	return UUIDsForFramework(frameworkURL)
+		|> flatMap(.Merge) { UUIDs in
+			if UUIDs.isEmpty {
+				return .empty
+			}
+			func filterUUIDs(signal: Signal<NSURL, CarthageError>) -> Signal<NSURL, CarthageError> {
```
ikesyo: Creating and lifting the signal operator seems to be tricky and I still prefer creating a SignalProducer as in [my comment](https://github.com/Carthage/Carthage/pull/853/#discussion_r43079186). I'd like to leave it to @mdiep.

ikesyo:
I'm OK except for https://github.com/Carthage/Carthage/pull/853/files#r43089238. Any thoughts on this PR? @mdiep and @NachoSoto

Source/CarthageKit/Project.swift
mdiep: This isn't necessarily how I'd do it, but I think this is fine for now.

mdiep:
Thanks for all the hard work @ikesyo and @kballard!

jdhealy:
:clap:

pepibumur:
:clap:

maciekish:
So I'm still facing this issue. I just installed `Carthage` today so I assume I'm on the latest version. I added Hue `github "hyperoslo/Hue"`. It runs and archives fine. When i submit it says it's missing Hue's `.bcsymbolmap` file (I could tell by `mdfind UUID` which pointed me to `Hue`) :(

I have followed all the `Carthage` installation instructions including adding the copy-script.

Please advise.

maciekish:
Here is an `Archive` `Post Action Script` (**not** a Build Phase Script) that fixes the issue temporarily

    cp "${SRCROOT}/Carthage/Build/iOS/"*.bcsymbolmap "${ARCHIVE_PATH}/BCSymbolMaps/

maciekish:
Apparently this is the issue I'm having #4624

mdiep:
@maciekish Please open a new issue if you're having problems. Thanks!

PR #861, opened 2015-10-23
Fix "fatal: destination path … already exists" on checkout

norio-nomura:
`git rev-parse --git-dir` returns not only relative path but also absolute path. When absolute path was returned, it caused "fatal: destination path" error.

```sh
% git rev-parse --git-dir
.git
% cd Carthage/Checkouts/Argo
% git rev-parse --git-dir
/Users/norio/Documents/workspace/github/Carthage/.git/modules/Carthage/Checkouts/Argo
% 
```

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -400,10 +400,9 @@ public func isGitRepository(directoryURL: NSURL) -> SignalProducer<Bool, NoError
 
 	return launchGitTask([ "rev-parse", "--git-dir", ], repositoryFileURL: directoryURL)
 		|> map { outputIncludingLineEndings in
-			let relativeGitDirectory = outputIncludingLineEndings.stringByTrimmingCharactersInSet(.newlineCharacterSet())
-			let gitDirectory = directoryURL.URLByAppendingPathComponent(relativeGitDirectory).path
+			let relativeOrAbsoluteGitDirectory = outputIncludingLineEndings.stringByTrimmingCharactersInSet(.newlineCharacterSet())
 			var isDirectory: ObjCBool = false
-			let directoryExists = gitDirectory.map { NSFileManager.defaultManager().fileExistsAtPath($0, isDirectory: &isDirectory) } ?? false
+			let directoryExists = NSFileManager.defaultManager().fileExistsAtPath(relativeOrAbsoluteGitDirectory, isDirectory: &isDirectory)
 			return directoryExists && isDirectory
 		}
 		|> catch { _ in SignalProducer(value: false) }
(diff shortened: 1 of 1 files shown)


mdiep:
Good catch!

Using a relative path makes me a little uneasy. Could we use `NSString.absolutePath` to test if it's absolute and make it absolute if it's not?

norio-nomura:
Oh, this PR still have a bug.
When `relativeOrAbsoluteGitDirectory` is relative, that is not based on current directory of `defaultManager()`.
It needs to making absolute path based on `directoryURL`.

norio-nomura:
Thanks for reviewing.
I changed to making absolute path as your review.

Source/CarthageKit/Git.swift
```diff
@@ -400,10 +400,15 @@ public func isGitRepository(directoryURL: NSURL) -> SignalProducer<Bool, NoError
...
 	return launchGitTask([ "rev-parse", "--git-dir", ], repositoryFileURL: directoryURL)
 		|> map { outputIncludingLineEndings in
-			let relativeGitDirectory = outputIncludingLineEndings.stringByTrimmingCharactersInSet(.newlineCharacterSet())
-			let gitDirectory = directoryURL.URLByAppendingPathComponent(relativeGitDirectory).path
+			let relativeOrAbsoluteGitDirectory = outputIncludingLineEndings.stringByTrimmingCharactersInSet(.newlineCharacterSet())
+			var absoluteGitDirectory: String
+			if (relativeOrAbsoluteGitDirectory as NSString).absolutePath {
+				absoluteGitDirectory = relativeOrAbsoluteGitDirectory
+			} else {
+				absoluteGitDirectory = (directoryURL.path! as NSString).stringByAppendingPathComponent(relativeOrAbsoluteGitDirectory)
```
ikesyo: `directoryURL.URLByAppendingPathComponent` would be better.

norio-nomura: Would it be as follows?:
```
absoluteGitDirectory = directoryURL.URLByAppendingPathComponent(relativeOrAbsoluteGitDirectory).path!
```

norio-nomura: It would be as follows:
```
var absoluteGitDirectory: String?
if (relativeOrAbsoluteGitDirectory as NSString).absolutePath {
	absoluteGitDirectory = relativeOrAbsoluteGitDirectory
} else {
	absoluteGitDirectory = directoryURL.URLByAppendingPathComponent(relativeOrAbsoluteGitDirectory).path
}
var isDirectory: ObjCBool = false
let directoryExists = absoluteGitDirectory.map { NSFileManager.defaultManager().fileExistsAtPath($0, isDirectory: &isDirectory) } ?? false
```

mdiep:
That looks great! Thanks!

PR #883, opened 2015-10-30
Remove deploy:on:all_branches from .travis.yml

ikesyo:
That has no effect for GitHub Releases.

http://docs.travis-ci.com/user/deployment/releases/ says:

> Please note that deploying GitHub Releases works only for tags, not for branches.


mdiep:
I had no idea this was configured or that it was a thing.

PR #874, opened 2015-10-28
Refactor copying bcsymbolmap files and others

ikesyo:
Extracts the common logic into `copyFileURLsFromProducer()` function.

PR #885, opened 2015-10-31
Implement copyFileURLsIntoDirectory() as a producer extension method

ikesyo:
As per @kballard's comment: https://github.com/Carthage/Carthage/pull/874#discussion_r43556939

Surely this seems better.


mdiep:
:+1:

PR #892, opened 2015-11-04
Add support for multiple subprojects of a workspace

younata:
This is a fairly pessimal way of doing this (doing the same work twice? bleh!), but it allows us to be far less opinionated about project structure.


mdiep:
It looks like there are some test failures. :relaxed:

younata:
Which is interesting, because I didn't even touch anywhere that might've stepped on those particular toes.

andymoe:
Yay! Thanks @younata

younata:
poke @mdiep thoughts?

mdiep:
Thanks for opening a pull request! It's always great to have new contributors. :sparkles:

I see that you extracted `projectsInDirectory` and `frameworkSchemesForProjects`. I appreciate the effort to clean up the code, but that makes it harder to review this change. Would you mind moving those sections back so we can focus on the behavioral changes in this PR?

Source/CarthageKit/Xcode.swift
```diff
@@ -1181,6 +1181,67 @@ public func parseSecuritySigningIdentities(securityIdentities securityIdentities
 		.ignoreNil()
 }
 
+public func projectsInDirectory(directoryURL: NSURL, configuration: String, platforms: Set<Platform> = []) -> SignalProducer<(ProjectLocator, [String]), CarthageError> {
+	let directoryLength = directoryURL.pathComponents!.count
+	return locateProjectsInDirectory(directoryURL)
+		.filter { project in
+			let fileURL = project.fileURL
+			let fileComponentsLength = fileURL.pathComponents!.count
+			return fileComponentsLength - directoryLength < 3 // within 2 subdirectories. Because we actually have a test for this.
```
mdiep: Which test is that? Is this why the tests failed when you first opened the PR?

younata: Apparently I'm dyslexic. I was going to point to https://github.com/Carthage/Carthage/blob/master/Source/CarthageKitTests/XcodeSpec.swift#L291-L298 and say that that's supposed to be testing that it's a subsubdirectory or earlier. My bad.

The tests earlier failed because the ReactiveCocoaLayout test fixture has an old version of Quick (0.6.0) committed to it's Carthage subfolder. Maybe it would be better to check if the project does not have a parent (or further up) "Carthage" folder? In either case, should probably remove that comment.

mdiep: So is this `filter` needed at all?

younata: It is not needed if we just update (or remove) the Carthage directories of the test fixtures.

However, it would probably be a good idea to cull out the Carthage directory - in case someone accidentally checks in their Carthage directory into source control.

In either case, it should be changed (or removed).

mdiep: I think ignoring projects found inside a `Carthage` directory at the top-level of the project directory seems like the right thing to do.

younata:
Ultimately went with punting on that (you're not supposed to check in the Carthage folder anyway, right? :p).

younata:
Figured I'd expand on why I punted:

In [XcodeSpec.it_Should_Build_For_All_Platforms](https://github.com/Carthage/Carthage/blob/master/Source/CarthageKitTests/XcodeSpec.swift#L62-L176), we're actually specifically testing that the dependencies for ReactiveCocoaLayout are built first - which, are [git submodules checked in to the Carthage folder](https://github.com/ReactiveCocoa/ReactiveCocoaLayout/tree/master/Carthage/Checkouts). Thankfully, in the test fixture, only Archimedes and ReactiveCocoa are kept.

However, each of these dependencies has their own "Externals" directory, which, prior to the most recent commit in this pull request, had an outdated version of Quick. (Of course, since then, both have moved over to git submodules checked in to their Carthage folders). I went ahead and removed the Quick folders in each dependency (as my punt, because [XcodeSpec.it_Should_Build_For_All_Platforms](https://github.com/Carthage/Carthage/blob/master/Source/CarthageKitTests/XcodeSpec.swift#L62-L176) actually contradicts the desire to just not search down Carthage folders - at least, that's how I interpreted it).

Source/CarthageKit/Xcode.swift
```diff
@@ -1181,6 +1181,36 @@ public func parseSecuritySigningIdentities(securityIdentities securityIdentities
 		.ignoreNil()
 }
 
+/// Finds all framework schemes for a directory.
+/// Returns a signal of all schemes and projects found. Might be an empty signal.
+public func frameworkSchemesForProjects(projects: [(ProjectLocator, [String])]) -> SignalProducer<(String, ProjectLocator), CarthageError> {
```
mdiep: I'd prefer to leave this code below and not extract it to a function.

Source/CarthageKit/Xcode.swift
```diff
@@ -1227,29 +1257,11 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
...
-						}
-					}
-					.concat(SignalProducer(error: .NoSharedFrameworkSchemes(.Git(GitURL(directoryURL.path!)), platforms)))
-					.take(1)
-					.flatMap(.Merge) { project, schemes in SignalProducer(values: schemes.map { ($0, project) }) }
+				let frameworkSchemes = frameworkSchemesForProjects(projects)
+				if frameworkSchemes.last() == nil {
+					return SignalProducer(error: .NoSharedFrameworkSchemes(.Git(GitURL(directoryURL.path!)), platforms))
+				}
+				return frameworkSchemes
```
mdiep: Instead of using `last`, making this function block and do work eagerly, it would be better to do it like so:

```swift
return frameworkSchemesForProjects(projects)
    .collect()
    .flatMap(.Merge) { schemes in
        if schemes.count > 0 {
            return SignalProducer(values: schemes)
        } else {
            return SignalProducer(error: .NoSharedFrameworkSchemes(.Git(GitURL(directoryURL.path!)), platforms))
        }
    }
```

(I haven't tried to compile that code, but it should at least give you the rough idea.)

mdiep:
> However, each of these dependencies has their own "Externals" directory, which, prior to the most recent commit in this pull request, had an outdated version of Quick (which used an older, outdated version of swift that failed to compile). (Of course, since then, both have moved over to git submodules checked in to their Carthage folders). I went ahead and removed the Quick folders in each dependency (as my punt, because XcodeSpec.it_Should_Build_For_All_Platforms actually contradicts the desire to just not search down Carthage folders - at least, that's how I interpreted it).

Hmm… that's tricky. Automatically building the submodules seems like a problem. In order to move forward with this, I think we'd need to avoid building any submodules (by getting those directories from Git) in addition to not building `Carthage/`.

younata:
@mdiep (Sorry about taking so long to get back to you).

With regards to filtering out dependencies: Carthage is a dependency manager, it doesn't really make sense to filter out any dependencies of a given framework. Additionally, there are a whole bunch of edge cases to account for that are really hard to figure out. Git submodules would be possible to filter out, the Carthage folder would be possible to filter out, however what's essentially impossible to filter out would be dependencies that aren't those. For example, if I just download the Quick-master.zip from github and place that in my project, instead of using at least git submodules, then that's pretty indistinguishable from a not-dependency. This is why I ultimately went with not even trying to filter out submodules or carthage folders.

Prior to this pull request, that third method of using dependencies would get ignored (which may or may not be desired)... hopefully. (For example, if you were bad at project structure, you might have a dependency.xcproj in the root directory, then if your $projectName.xc(proj|workspace) was lexicographically after the dependency.xcproj, it would only find the dependency.) However, I think this is actually a bonus feature of this pull request.

As for less worm-canny things, I squashed the commits all down into one, and inlined as much as I could, for easy reviewing. I would, however, like to do some serious refactoring of the specs (they run on the order of minutes on my very beefy machine, I'm used to on the order of seconds for my projects), as well as the actually implementation code (the fact that Xcode.swift is more than 1000 lines is a very big code smell. I would love to break that up into more manageable bits of code). However, that's very much a separate pull request.

Also, thanks for opening me up to ReactiveCocoa, it's a very powerful paradigm that I've not used before.

mdiep:
> (Sorry about taking so long to get back to you).

No worries! Open source takes time and patience from everyone. :smile:

> With regards to filtering out dependencies: Carthage is a dependency manager, it doesn't really make sense to filter out any dependencies of a given framework. Additionally, there are a whole bunch of edge cases to account for that are really hard to figure out. Git submodules would be possible to filter out, the Carthage folder would be possible to filter out, however what's essentially impossible to filter out would be dependencies that aren't those. For example, if I just download the Quick-master.zip from github and place that in my project, instead of using at least git submodules, then that's pretty indistinguishable from a not-dependency. This is why I ultimately went with not even trying to filter out submodules or carthage folders.

I think this a good place to remember the expression "don't let the perfect be the enemy of the good".

The first two cases—the Carthage folder and Git submodules—are very common cases. I'm concerned that recurring into these directories and building the projects in them will cause issues. 

The last case, random projects in the random directories, are impossible to filter out. But (1) that's almost certainly a very rare case, (2) it's not clear whether building or not building is correct, and (3) the behavior is unaffected by whether we exclude Carthage folders and Git submodules.

So I still think we need to ignore Carthage folders and Git submodules. :confused:

> I squashed the commits all down into one

FWIW, we don't normally squash commits. You don't need to do that in the future.

> I would, however, like to do some serious refactoring of the specs (they run on the order of minutes on my very beefy machine, I'm used to on the order of seconds for my projects), as well as the actually implementation code (the fact that Xcode.swift is more than 1000 lines is a very big code smell. I would love to break that up into more manageable bits of code). However, that's very much a separate pull request.

That would be great!!! :sparkles:

> Also, thanks for opening me up to ReactiveCocoa, it's a very powerful paradigm that I've not used before.

You're welcome! We think it's pretty great. I'm glad you like it. :smile:

younata:
I'm sorry. I was mistaken/out of date.

As of when I started this pull request, ignoring Carthage subdirectories (there were/are no fixtures w/ git submodules) would result in test failures. As of now (having rebased to fd02af5e7561893901f56dc030934ae912ca8d09), it appears to pass tests.

Ignoring Carthage is easy ("is directory a subdirectory of Carthage? Yes? IGNORED"). There's support for finding git submodules, though no test fixture includes it (yet another fixture to add :/). Is there a better way to use one signal producer to filter the contents of another that's not than the naive "wait until the first producer finishes, then use that to filter the other one?"

mdiep:
> Is there a better way to use one signal producer to filter the contents of another that's not than the naive "wait until the first producer finishes, then use that to filter the other one?"

No, you'll need to wait for the first one to finish so you know which values to filter out. You can do it with `collect` and `flatMap`, and it looks like you've done that.

Is this ready for review? Or are you still working on it?

younata:
> No, you'll need to wait for the first one to finish so you know which values to filter out. You can do it with collect and flatMap, and it looks like you've done that.

Darn. Was hoping for a method I overlooked that was something like `signalProducer.allValues()` which would do what you expect.

>Is this ready for review? Or are you still working on it?

As of cb81d1ca41eca6555c9efd7f22bb6823d669e4d8, yes.

Thanks for working with me on this!

Source/CarthageKit/Xcode.swift
```diff
@@ -189,10 +189,24 @@ extension BuildArguments: CustomStringConvertible {
 public func locateProjectsInDirectory(directoryURL: NSURL) -> SignalProducer<ProjectLocator, CarthageError> {
 	let enumerationOptions: NSDirectoryEnumerationOptions = [ .SkipsHiddenFiles, .SkipsPackageDescendants ]
 
+	let submodules = submodulesInRepository(directoryURL)
+		.map { return $0.path }
+		.collect()
+		.first()?.value ?? []
```
mdiep: `.first()` is usually discouraged outside of tests. The idiomatic way would be to `collect()` and then do the next step inside a `flatMap`.

Source/CarthageKit/Xcode.swift
```diff
@@ -189,10 +189,24 @@ extension BuildArguments: CustomStringConvertible {
...
+
 	return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL.URLByResolvingSymlinksInPath!, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: enumerationOptions, catchErrors: true)
+		.filter { _, URL in
+			let pathRelativeToDirectoryURL = URL.absoluteString.substringFromIndex(directoryURL.absoluteString.endIndex)
+			for submodulePath in submodules {
+				if pathRelativeToDirectoryURL.hasPrefix(submodulePath) {
```
mdiep: `hasPrefix` really isn't a safe way to do this. (Consider if `pathRelativeToDirectoryURL` is `foo-bar` and `submodulePath` is `foo`.) You'd need to use `NSURL.pathComponents` or `NSString.pathComponents`.

It'd be nice to add an extension to `NSURL` to test if one URL is a descendent of another URL.

Source/CarthageKit/Xcode.swift
```diff
@@ -189,10 +189,24 @@ extension BuildArguments: CustomStringConvertible {
...
+					return false
+				}
+			}
+			return !pathRelativeToDirectoryURL.hasPrefix("Carthage")
```
mdiep: Same thing goes here for the use of `hasPrefix`.

Source/CarthageKit/Xcode.swift
```diff
@@ -1220,9 +1236,17 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
 							return false
 						}
 					}
-					.concat(SignalProducer(error: .NoSharedFrameworkSchemes(.Git(GitURL(directoryURL.path!)), platforms)))
-					.take(1)
-					.flatMap(.Merge) { project, schemes in SignalProducer(values: schemes.map { ($0, project) }) }
+					.flatMap(FlattenStrategy.Merge) { project, schemes in
```
mdiep: I think it'd be clearer to keep the `.Concat` strategy here since the order could be considered important.

Source/CarthageKit/Xcode.swift
```diff
@@ -1211,6 +1225,8 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
 						}
 						return (project, containedSchemes)
 					}
+					// replace the filter and concat with something that accounts for multiple subprojects
+					// cause it errors out if even one of the results is that error...
```
mdiep: Unfortunately, I think this is going to be a prerequisite for merging this PR. Otherwise, we would likely break some projects that currently work with Carthage.

If you need help with this, I can try to steer you in the right direction.

mdiep:
I left a few more notes of things that need to be fixed up. Thanks for working on this, responding to the feedback, and adding test fixtures! I'm really excited to ship this. :smile:

Source/CarthageKit/Xcode.swift
younata: That was actually erroneously left in as a note to myself when I first started work on this.

younata:
Fixed all the line notes, added a nice little helper extension to NSURL that reports if a given URL is the subdirectory of the receiver (It's even tested, because yay tests!), moved the .first() usage to use a flatmap, which, yeah, is the more idiomatic (and better) way to do this (forgive me, still getting used to FRP).

Also rebased upon master, and fixed merge conflicts.

Carthage/Checkouts/Commandant
```diff
@@ -1 +1 @@
-Subproject commit fdb18a2602fcf6641faad87b4a1667995a68b706
+Subproject commit 186fc8d7937cbdced7bc9e90114079241d864f77
```
mdiep: It looks like you accidentally reset the Commandant submodule during a merge/rebase. Can you undo this change?

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -261,6 +261,17 @@ extension NSURL {
...
+
+	public func hasSubdirectory(possibleSubdirectory: NSURL) -> Bool {
+		guard self.scheme == possibleSubdirectory.scheme else { return false }
+
+		guard let pathComponents = self.pathComponents else { return false }
+		guard let otherPathComponents = possibleSubdirectory.pathComponents else { return false }
+
+		guard otherPathComponents.count >= pathComponents.count else { return false }
+
+		return Array(otherPathComponents[0..<pathComponents.count]) == pathComponents
```
mdiep: Can you rework this to use `if let … else …`? That should let us specify the `false` condition once instead of four times.

Source/CarthageKit/Xcode.swift
```diff
@@ -189,25 +189,38 @@ extension BuildArguments: CustomStringConvertible {
...
-			if let UTI = URL.typeIdentifier.value {
-				if (UTTypeConformsTo(UTI, "com.apple.dt.document.workspace")) {
-					matches.append(.Workspace(URL))
-				} else if (UTTypeConformsTo(UTI, "com.apple.xcode.project")) {
-					matches.append(.ProjectFile(URL))
+	return submodulesInRepository(directoryURL)
+		.map { return directoryURL.URLByAppendingPathComponent($0.path) }
+		.concat(SignalProducer<NSURL, CarthageError>(value: directoryURL.URLByAppendingPathComponent("Carthage")))
+		.collect()
+		.flatMap(.Merge) {directoriesToSkip in
```
mdiep: Can you add a space here before `directoriesToSkip`?

Source/CarthageKit/Xcode.swift
```diff
@@ -189,25 +189,38 @@ extension BuildArguments: CustomStringConvertible {
...
+				.filter { _, URL in
+					for directory in directoriesToSkip {
+						if directory.hasSubdirectory(URL) {
+							return false
+						}
+					}
+					return true
 				}
-			}
+				.reduce([]) { (var matches: [ProjectLocator], tuple) -> [ProjectLocator] in
```
mdiep: Can you outdent this step and the following steps? I try to keep the `flatMap` block as small as possible.

mdiep:
This is looking good to me! I just left a few more minor notes. Sorry for the delay in reviewing—I'm on vacation, so I'm less available ATM. :smile:

@ikesyo Would you be interested in taking a look at this as well? You've made some changes in this area before, and I think it'd be good to get another pair of :eyes: on it.

ikesyo:
Okay, I'll take another look. :eyes:

Source/CarthageKit/Xcode.swift
ikesyo: Could you restore `var matches = matches`? That is an intentional change in #998.

Source/CarthageKit/Xcode.swift
```diff
@@ -1223,9 +1236,17 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
...
 					}
-					.concat(SignalProducer(error: .NoSharedFrameworkSchemes(.Git(GitURL(directoryURL.path!)), platforms)))
-					.take(1)
-					.flatMap(.Merge) { project, schemes in SignalProducer(values: schemes.map { ($0, project) }) }
+					.flatMap(.Concat) { project, schemes in
+						return SignalProducer(values: schemes.map { ($0, project) })
+					}
+					.collect()
+					.flatMap(.Merge) { (schemes: [(String, ProjectLocator)]) -> SignalProducer<(String, ProjectLocator), CarthageError> in
+						if schemes.count > 0 {
```
ikesyo: `schemes.isEmpty` is preferable.

ikesyo:
Can we revert the change to `Source/CarthageKitTests/fixtures/ReactiveCocoaLayout.zip`?

Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -59,6 +59,39 @@ class XcodeSpec: QuickSpec {
...
+				expect(relativePaths).toNot(beEmpty())
+				let pathsStartingWithCarthage = relativePaths.filter { return $0.hasPrefix("Carthage/") }
+				expect(pathsStartingWithCarthage).to(beEmpty())
+			}
+
+			it("should not find anything that's listed as a git submodule") {
+				let multipleSubprojects = "SampleGitSubmodule"
+				let _directoryURL = NSBundle(forClass: self.dynamicType).URLForResource(multipleSubprojects, withExtension: nil)!
+
+				let _ = submodulesInRepository(_directoryURL)
```
ikesyo: This `let` can be omitted.

Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -59,6 +59,39 @@ class XcodeSpec: QuickSpec {
...
+					.on(next: {submodule in
+						print("submodule: \(submodule.path)")
+					}).wait()
```
ikesyo: Could you add a line break before `.wait()`?

Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -59,6 +59,39 @@ class XcodeSpec: QuickSpec {
...
+
+				let relativePaths = relativePathsForProjectsInDirectory(_directoryURL)
+				expect(relativePaths).to(equal(["SampleGitSubmodule.xcodeproj/"]))
```
ikesyo: Could you add spaces after `[` and before `]`?

Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -171,6 +204,40 @@ class XcodeSpec: QuickSpec {
...
+				.flatten(.Concat)
+				.ignoreTaskData()
+				.on(next: { (project, scheme) in
+					NSLog("Building scheme \"\(scheme)\" in \(project)")
+				})
+				.wait()
+
+			expect(result.error).to(beNil())
+
+			let expectedPlatformsFrameworks = [("iOS", "SampleiOSFramework"), ("Mac", "SampleMacFramework"), ("tvOS", "SampleTVFramework"), ("watchOS", "SampleWatchFramework")]
```
ikesyo: [Ditto](https://github.com/Carthage/Carthage/pull/892/files#r48115298)

Source/CarthageKit/Xcode.swift
```diff
@@ -189,25 +189,38 @@ extension BuildArguments: CustomStringConvertible {
...
-			var matches = matches
-			let (_, URL) = tuple
-			
-			if let UTI = URL.typeIdentifier.value {
-				if (UTTypeConformsTo(UTI, "com.apple.dt.document.workspace")) {
-					matches.append(.Workspace(URL))
-				} else if (UTTypeConformsTo(UTI, "com.apple.xcode.project")) {
-					matches.append(.ProjectFile(URL))
+	return submodulesInRepository(directoryURL)
+		.map { return directoryURL.URLByAppendingPathComponent($0.path) }
```
ikesyo: We usually omit the `return` for single line, single expression closure.

Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -59,6 +59,39 @@ class XcodeSpec: QuickSpec {
...
+					.collect()
+					.first()
+				expect(result?.error).to(beNil())
+				return result?.value ?? []
+			}
+
+			it("should not find anything in the Carthage Subdirectory") {
+				let relativePaths = relativePathsForProjectsInDirectory(directoryURL)
+				expect(relativePaths).toNot(beEmpty())
+				let pathsStartingWithCarthage = relativePaths.filter { return $0.hasPrefix("Carthage/") }
```
ikesyo: [Ditto](https://github.com/Carthage/Carthage/pull/892/files#r48115351)

Source/CarthageKit/FrameworkExtensions.swift
ikesyo: How about the following?

```swift
if scheme == possibleSubdirectory.scheme, let path = self.path, otherPath = possibleSubdirectory.path {
    return otherPath.hasPrefix(path)
}
return false
```

Source/CarthageKit/Xcode.swift
younata: Whoops. Must've gotten mixed up in an earlier rebase.

younata:
Fixed the code style issues.

Merged commits, and rebased upon master again

ikesyo:
> FWIW, we don't normally squash commits. You don't need to do that in the future.

:wink:

Source/CarthageKitTests/XcodeSpec.swift
ikesyo: This isn't updated.

ikesyo:
It seems that @mdiep's [this comment](https://github.com/Carthage/Carthage/pull/892#discussion_r48110395) is not addressed. So what it should be is:

```swift
	return submodulesInRepository(directoryURL)
		.map { directoryURL.URLByAppendingPathComponent($0.path) }
		.concat(SignalProducer<NSURL, CarthageError>(value: directoryURL.URLByAppendingPathComponent("Carthage")))
		.collect()
		.flatMap(.Merge) { directoriesToSkip in
			return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL.URLByResolvingSymlinksInPath!, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: enumerationOptions, catchErrors: true)
				.filter { _, URL in
					for directory in directoriesToSkip {
						if directory.hasSubdirectory(URL) {
							return false
						}
					}
					return true
				}
		}
		.reduce([]) { (matches: [ProjectLocator], tuple) -> [ProjectLocator] in
			var matches = matches
			let (_, URL) = tuple

			if let UTI = URL.typeIdentifier.value {
				if (UTTypeConformsTo(UTI, "com.apple.dt.document.workspace")) {
					matches.append(.Workspace(URL))
				} else if (UTTypeConformsTo(UTI, "com.apple.xcode.project")) {
					matches.append(.ProjectFile(URL))
				}
			}

			return matches
		}
		.map { $0.sort() }
		.flatMap(.Merge) { SignalProducer(values: $0) }
```

Source/CarthageKitTests/XcodeSpec.swift
younata: I'm going to assume that you mean to add a space between the square-brackets on line 71.

younata:
When it comes to code style issues, I actually prefer to squash commits, as that's better than 5 "fix code style" commits.

As for [this comment](https://github.com/Carthage/Carthage/pull/892#discussion_r48110395), it took me until this morning to realize that you wanted a close-curly-bracket moved up.

> Can we revert the change to Source/CarthageKitTests/fixtures/ReactiveCocoaLayout.zip?

This is actually the cause of the current test failures, actually. [Link for posteriority](https://travis-ci.org/Carthage/Carthage/builds/98068605).

What's happening here is that the dependency project ReactiveCocoa has references to outdated versions (pre-swift 2.0) of Quick and Nimble under their `External` directory. All projects that aren't listed in git submodules or under the Carthage directory are compiled (as part of this pull request), which leads to a test failure. My solution was to modify the dependency to just remove the references to quick and nimble. This shouldn't normally be an issue (any more than depending upon outdated versions of swift is), except for the fact that buildDependencyProject is specifically called for ReactiveCocoa in that particular test case.

Going to push a (not squashed) commit that re-introduces this change.

mdiep:
> As for this comment, it took me until this morning to realize that you wanted a close-curly-bracket moved up.

Sorry! I should have been clearer in my original comment. :relaxed:

This looks great to me! Once @ikesyo is happy with it, we can merge it!

Source/CarthageKitTests/XcodeSpec.swift
ikesyo: I mean omitting `return` here.

ikesyo:
> > Can we revert the change to Source/CarthageKitTests/fixtures/ReactiveCocoaLayout.zip?
>This is actually the cause of the current test failures, actually. Link for posterity.

My bad! :bow: Yes, those in the zip are not submodules. :astonished: Thank you for the re-fixing.

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -261,6 +261,13 @@ extension NSURL {
 
 		return .Failure(.ReadFailed(self, error))
 	}
+
+	public func hasSubdirectory(possibleSubdirectory: NSURL) -> Bool {
+		if scheme == possibleSubdirectory.scheme, let path = self.pathComponents, otherPath = possibleSubdirectory.pathComponents where path.count <= otherPath.count {
+			return Array(otherPath[0..<path.count]) == path
```
ikesyo: Just a curiosity, `hasPrefix` approach didn't work well or had performace downside?

mdiep: I [mentioned here](https://github.com/Carthage/Carthage/pull/892#discussion-diff-47987373) that `hasPrefix` isn't safe because it's hard to ensure that you're not matching `Carthage-blah` with `hasPrefix("Carthage")`.

ikesyo: Oh, I overlooked the comment. :astonished: Thanks!

ikesyo: So `0..<path.count` could be replaced with `path.indices`.

Source/CarthageKit/Xcode.swift
```diff
@@ -189,11 +189,25 @@ extension BuildArguments: CustomStringConvertible {
...
+				.filter { _, URL in
+					for directory in directoriesToSkip {
+						if directory.hasSubdirectory(URL) {
+							return false
+						}
+					}
+					return true
+				}
+		}
+		.reduce([ ]) { (matches: [ProjectLocator], tuple) -> [ProjectLocator] in
```
ikesyo: We use `[]` without an intermediate space for an empty array.

Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -59,6 +59,32 @@ class XcodeSpec: QuickSpec {
...
 		}
 
+		describe("locateProjectsInDirectory:") {
+			func relativePathsForProjectsInDirectory(directoryURL: NSURL) -> [String] {
+				let result = locateProjectsInDirectory(directoryURL)
+					.map { $0.fileURL.absoluteString.substringFromIndex(directoryURL.absoluteString.endIndex) }
+					.collect()
+					.first()
+				expect(result?.error).to(beNil())
+				return result?.value ?? [ ]
```
ikesyo: Ditto

Source/CarthageKitTests/FrameworkExtensionsSpec.swift
```diff
@@ -0,0 +1,25 @@
+import Quick
+import Nimble
+
+class FrameworkExtensionsSpec: QuickSpec {
+    override func spec() {
```
ikesyo: Can you use tab here?

Source/CarthageKitTests/FrameworkExtensionsSpec.swift
```diff
@@ -0,0 +1,25 @@
...
+
+				expect(subject.hasSubdirectory(subject)).to(beTruthy())
+				expect(subject.hasSubdirectory(unrelatedScheme)).to(beFalsy())
+				expect(subject.hasSubdirectory(parentDir)).to(beFalsy())
+				expect(subject.hasSubdirectory(immediateSub)).to(beTruthy())
+				expect(subject.hasSubdirectory(distantSub)).to(beTruthy())
+				expect(subject.hasSubdirectory(unrelatedDirectory)).to(beFalsy())
+			}
+		}
+    }
```
ikesyo: Can you use tab here?

Source/CarthageKit/Xcode.swift
```diff
@@ -189,11 +189,25 @@ extension BuildArguments: CustomStringConvertible {
 public func locateProjectsInDirectory(directoryURL: NSURL) -> SignalProducer<ProjectLocator, CarthageError> {
 	let enumerationOptions: NSDirectoryEnumerationOptions = [ .SkipsHiddenFiles, .SkipsPackageDescendants ]
 
-	return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL.URLByResolvingSymlinksInPath!, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: enumerationOptions, catchErrors: true)
-		.reduce([]) { (matches: [ProjectLocator], tuple) -> [ProjectLocator] in
+	return submodulesInRepository(directoryURL)
+		.map { directoryURL.URLByAppendingPathComponent($0.path) }
+		.concat(SignalProducer<NSURL, CarthageError>(value: directoryURL.URLByAppendingPathComponent("Carthage")))
```
ikesyo: `concat` can infer the generic parameters, so `<NSURL, CarthageError>` can be omitted.

Source/CarthageKit/Xcode.swift
```diff
@@ -189,11 +189,25 @@ extension BuildArguments: CustomStringConvertible {
...
+		.collect()
+		.flatMap(.Merge) { directoriesToSkip in
+			return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL.URLByResolvingSymlinksInPath!, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: enumerationOptions, catchErrors: true)
+				.filter { _, URL in
```
ikesyo: I thought that to filter URLs in this phase is inefficient. We could reduce the evaluations by just filtering out the matched `[ProjectLocator]`. So this func could be written as follows:

```swift
public func locateProjectsInDirectory(directoryURL: NSURL) -> SignalProducer<ProjectLocator, CarthageError> {
	let enumerationOptions: NSDirectoryEnumerationOptions = [ .SkipsHiddenFiles, .SkipsPackageDescendants ]

	return NSFileManager.defaultManager().carthage_enumeratorAtURL(directoryURL.URLByResolvingSymlinksInPath!, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: enumerationOptions, catchErrors: true)
		.reduce([]) { (matches: [ProjectLocator], tuple) -> [ProjectLocator] in
			var matches = matches
			let (_, URL) = tuple

			if let UTI = URL.typeIdentifier.value {
				if (UTTypeConformsTo(UTI, "com.apple.dt.document.workspace")) {
					matches.append(.Workspace(URL))
				} else if (UTTypeConformsTo(UTI, "com.apple.xcode.project")) {
					matches.append(.ProjectFile(URL))
				}
			}

			return matches
		}
		.flatMap(.Merge) { matches in
			return submodulesInRepository(directoryURL)
				.map { directoryURL.URLByAppendingPathComponent($0.path) }
				.concat(SignalProducer(value: directoryURL.URLByAppendingPathComponent("Carthage")))
				.collect()
				.map { directoriesToSkip in
					return matches.filter { match in
						for directory in directoriesToSkip {
							if directory.hasSubdirectory(match.fileURL) {
								return false
							}
						}
						return true
					}
				}
		}
		.map { $0.sort() }
		.flatMap(.Merge) { SignalProducer(values: $0) }
}
```

Source/CarthageKit/FrameworkExtensions.swift
younata: Didn't know about that. That's awesome.

Source/CarthageKit/Xcode.swift
younata: I feel like this is a case of premature optimization.

younata:
Amended previous commit with the style fixes as requested.

Source/CarthageKit/Xcode.swift
mdiep: I think the current version is probably preferable. As I understand it, retrieving the type identifier for the URL has a non-zero cost. But either way, it's probably inconsequential.

mdiep:
Merging this as-is. The [one addressed note](https://github.com/Carthage/Carthage/pull/892#discussion-diff-48115362) is very minor.

Thank you for all your hard work on this, @younata! I really appreciate all the fixes you made in response to feedback, and I'm really excited to ship this!

ikesyo:
:clap:

Source/CarthageKitTests/FrameworkExtensionsSpec.swift
```diff
@@ -0,0 +1,25 @@
...
+	override func spec() {
+		describe("NSURL Extensions") {
+			it("should figure out if a is a subdirectory of b") {
+				guard let subject = NSURL(string: "file:///foo/bar") else { return }
+
+				guard let unrelatedScheme = NSURL(string: "http:///foo/bar/baz") else { return }
+				guard let parentDir = NSURL(string: "file:///foo") else { return }
+				guard let immediateSub = NSURL(string: "file:///foo/bar/baz") else { return }
+				guard let distantSub = NSURL(string: "file:///foo/bar/baz/qux") else { return }
+				guard let unrelatedDirectory = NSURL(string: "file:///bar/bar/baz") else { return }
```
NachoSoto: These shouldn't just `return`, otherwise this is a false positive, since the test isn't running at all!

NachoSoto:
Looks like this may have broken `master`: #1029. I'm not familiar with this change, could some of you take a quick look and see if you can figure out why? :)

Issue #889, opened 2015-11-03
Invalid architecture UUIDs: Could not parse UUIDs using dwarfdump


Palleas:
**Disclaimer**: I've posted this issue in realm/realm-cocoa#2785 thinking it was Realm-related but it seems it may have something to do with Carthage's `copy-framework`

I'm getting the following error when I try to setup Realm in my iOS application using Carthage:

```
$ carthage update --use-submodules --platform ios
*** Fetching realm-cocoa
*** Fetching keychain-swift
*** Fetching SwiftyJSON
*** Fetching ReactiveCocoa
*** Fetching Result
*** Checking out Result at "0.6.0-beta.3"
*** Checking out SwiftyJSON at "2.3.0"
*** Checking out keychain-swift at "3.0.7"
*** Checking out ReactiveCocoa at "9af5f32b4f9f908ae5f26904fbaf49d0a9039e24"
*** Checking out realm-cocoa at "v0.96.2"
*** xcodebuild output can be found in /var/folders/tk/cw93k7v928qf21kppt47d_380000gn/T/carthage-xcodebuild.0HflvL.log
*** Building scheme "Result-iOS" in Result.xcodeproj
*** Building scheme "SwiftyJSON iOS" in SwiftyJSON.xcworkspace
*** Building scheme "KeychainSwift" in KeychainSwift.xcodeproj
*** Building scheme "iOS" in Realm.xcodeproj
Invalid architecture UUIDs: Could not parse UUIDs using dwarfdump from /Users/Romain/Library/Developer/Xcode/DerivedData/Realm-acrmeedomupvqhcepjsgrlqtlowg/Build/Products/Release-iphoneos/Realm.framework/Realm
```

Any idea what could be causing this? I've attached the [xcodebuild full log here](https://gist.github.com/Palleas/7567e783a235ef1dde54) and here is my Cartfile:

```
github "ReactiveCocoa/ReactiveCocoa" "swift2"
github "SwiftyJSON/SwiftyJSON" "2.3.0"
github "exchangegroup/keychain-swift" "3.0.7"
github "realm/realm-cocoa"
```

This doesn't seem to be an issue with Carthage because it really fails as soon as I add Realm in my Cartfile and run cartage update.

I'm using Xcode 7.1:  
```
xcodebuild -version
Xcode 7.1
Build version 7B91b
```

and Carthage 0.9.4:
```
carthage version
0.9.4
```

Thanks !

gilazo:
[...]

Issue #734, opened 2015-09-05
Facebook SDK


lucasderraugh:
I've added the facebook/facebook-ios-sdk github entry to my Cartfile, however I always get this as output:
```
*** Building scheme "Bolts" in FacebookSDK.xcworkspace
A shell task failed with exit code 1:
error: cannot parse the debug map for "/Users/username/Developer/Spring 2015/project/Carthage/Build/iOS/Bolts.framework/Bolts": The file was not recognized as a valid object file
```
Is there a way around this issue? Does it have something to do with Bolts being a submodule of the Facebook SDK? Thanks for the help.

tomco:
+1 
Having the same problem. 

Also completely unsure why Carthage is compiling Bolts:
```
*** Building scheme "MacBolts" in FacebookSDK.xcworkspace
*** Building scheme "Bolts" in FacebookSDK.xcworkspace
```
Since when opening `FacebookSDK.xcworkspace` no shared scheme for Bolts can be found.

collindonnell:
Also seeing this exact same issue.

nathanhosselton:
This is still an issue in 0.9.2 (in my case I am only building Bolts, not the entire Facebook SDK). Using xcodebuild works fine. So it seems to be an issue with Carthage even though Bolts is (as far as I know) the only framework currently exposing the issue.

nathanhosselton:
The "Bolts" file which is the point of failure is a binary file that is also generated (without issue) by the Mac version of Bolts. I'm not sure what the purpose of these files are or what specifically is generating them. I have mentioned this in the open issue at https://github.com/BoltsFramework/Bolts-iOS/issues/152.

lucasderraugh:
Looking into it more, it comes down to the fact that Bolts is a static library, not dynamic. I'm going to close this as a Carthage issue as it is not. [Bolts Issue](https://github.com/BoltsFramework/Bolts-iOS/issues/152)

sirvine:
FYI — It looks like Bolts is Carthage-ready now (https://github.com/BoltsFramework/Bolts-iOS/issues/152#issuecomment-167904234), but the FB SDK does not seem to be.

fatuhoku:
What's the latest on this?  Will the FBSDK ever be manageable by Carthage?

mdiep:
Can you open a new issue with information about what you're seeing? The specific issue here seems to have been addressed.

jamesone:
Any update on the FBSDK setup with carthage?

PR #894, opened 2015-11-05
Skip building static frameworks by checking Mach-O type

ikesyo:
Addresses #734 and #889.


Source/CarthageKit/Xcode.swift
```diff
@@ -733,26 +761,31 @@ private func shouldBuildProductType(productType: ProductType) -> Bool {
...
+
 /// Determines whether the given scheme should be built automatically.
 private func shouldBuildScheme(buildArguments: BuildArguments, _ forPlatforms: Set<Platform>) -> SignalProducer<Bool, CarthageError> {
 	precondition(buildArguments.scheme != nil)
 
 	return BuildSettings.loadWithArguments(buildArguments)
-		.flatMap(.Concat) { settings -> SignalProducer<ProductType, CarthageError> in
-			let productType = SignalProducer(result: settings.productType)
+		.flatMap(.Concat) { settings -> SignalProducer<(ProductType, MachOType), CarthageError> in
+			let typePair = SignalProducer(result: settings.productType &&& settings.machOType)
```
NachoSoto: Ooooh nice, I forgot this operator.

Source/CarthageKit/Xcode.swift
```diff
@@ -733,26 +761,31 @@ private func shouldBuildProductType(productType: ProductType) -> Bool {
...
 				return settings.buildSDKs
 					.filter { forPlatforms.contains($0.platform) }
-					.flatMap(.Merge) { _ in productType }
+					.flatMap(.Merge) { _ in typePair }
 					.flatMapError { _ in .empty }
 			}
 		}
-		.filter(shouldBuildProductType)
-		// If we find any framework target, we should indeed build this scheme.
+		.filter { shouldBuildProductType($0) && shouldBuildMachOType($1) }
```
NachoSoto: Not a big deal, but maybe split this into 2 `filter` calls?

NachoSoto:
Great idea!

Source/CarthageKit/Xcode.swift
```diff
@@ -468,6 +468,30 @@ public enum ProductType: String {
 	}
 }
 
+/// Describes the type of Mach-O files.
+/// See https://developer.apple.com/library/mac/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW73.
```
mdiep: $5 says this link won't work in a year. :smile:

Source/CarthageKit/Xcode.swift
mdiep: I think we should either (1) combine these into one function that takes a product type _and_ Mach-O type or (2) combine the two enums into a single enum that differentiates between dynamic and static frameworks.

I'd lean towards (2) but could be good with (1) if we think it's important for build setting properties to have a 1-to-1 correspondence.

mdiep:
I really like this idea! Thanks for working on it. :sparkling_heart:

Source/CarthageKit/Xcode.swift
NachoSoto: Ooooh I really like `(2)` as well.

Source/CarthageKit/Xcode.swift
ikesyo: I'll remove it. :sweat_smile:

mdiep: Sorry, you should leave it in. I was just joking because it's frustrating how Apple breaks these links all the time. :relaxed:

mdiep:
@ikesyo I think I'll create a new release after this is merged so we can get a working build with homebrew. Does that sound good to you?

ikesyo:
@mdiep Sounds good! :+1:

Source/CarthageKit/Xcode.swift
ikesyo: Oh, forgive my misunderstanding. :stuck_out_tongue_closed_eyes:

ikesyo:
How do you feel the changes in d8b22ec?

Source/CarthageKit/Xcode.swift
```diff
@@ -468,6 +468,52 @@ public enum ProductType: String {
...
+	case Staticlib = "staticlib"
+
+	/// Attempts to parse a Mach-O type from a string returned from `xcodebuild`.
+	public static func fromString(string: String) -> Result<MachOType, CarthageError> {
+		return Result(self.init(rawValue: string), failWith: .ParseError(description: "unexpected Mach-O type \"\(string)\""))
+	}
+}
+
+/// Describes the type of frameworks.
+private enum FrameworkType {
```
ikesyo: I'm not convinced this should be private or public.

mdiep:
This looks good. :+1:

ikesyo:
Thank you for the feedbacks. :sparkling_heart:

PR #910, opened 2015-11-10
Add “OSX” alias for --platform option

### LINKED ISSUE #909: Add alias for platform osx same as mac
Currently platform `osx` does not work:
```bash
> carthage update --no-use-binaries --platform ios,osx
Invalid value for '--platform': ios,osx
```

But this works:
```bash
 carthage update --no-use-binaries --platform ios,mac
```

ikesyo:
Resolves #909.


mdiep:
:sparkles:

PR #919, opened 2015-11-12
Remove the section on copying dSYM files

### LINKED ISSUE #918: Is the section on Copying dSYMs required anymore / outdated?
Following #853 Do the dysms need to be manually copied over anymore?

Abizern:
Closes #918

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,16 +71,6 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
 
   This script works around an [App Store submission bug](http://www.openradar.me/radar?id=6409498411401216) triggered by universal binaries and ensures that necessary bitcode-related files are copied when archiving.
 
-##### Copying debug symbols for debugging and crash reporting
-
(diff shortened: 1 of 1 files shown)


mdiep:
I think this should still be suggested for OS X builds, which don't require the use of `copy-frameworks`?

furyu-ikeda-sho:
> for OS X builds, which don't require the use of `copy-frameworks`?

That'a good point, and since the `copy-frameworks` doesn't copy dSYM files automatically as noted in #918, this is still required for iOS builds too.

ikesyo:
> for OS X builds, which don't require the use of `copy-frameworks`?

That'a good point, and since the `copy-frameworks` doesn't copy dSYM files automatically as noted in #918, this is still required for iOS builds too.

ikesyo:
Closing this, thank you for the PR though!

Abizern:
NP. That's the way this is supposed to work.

PR #923, opened 2015-11-16
Use guard syntax in Comparable implementation for ProjectLocator

ikesyo:


mdiep:
Interesting use of `guard`. :+1:

PR #931, opened 2015-11-20
Ensure that required git version is used

### LINKED ISSUE #774: Warn on old versions of git
Lets avoid issues like #744. I think 2.3.x is the current minimum for everything to work as expected.

ikesyo:
Resolves #774. How do you feel about the implementation?

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -11,6 +11,9 @@ import Result
 import ReactiveCocoa
 import ReactiveTask
 
+/// The git version Carthage requires at least.
+public let CarthageRequiredGitVersion = "2.3.0"
+
 /// Represents a URL for a Git remote.
 public struct GitURL: Equatable {
 	/// The string representation of the URL.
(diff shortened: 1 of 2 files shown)


Source/carthage/main.swift
```diff
@@ -9,8 +9,14 @@
 import CarthageKit
 import Commandant
 import Foundation
-import Result
+import ReactiveCocoa
 import ReactiveTask
+import Result
+
+guard ensureGitVersion().first()?.value == true else {
+	fputs("Carthage requires git \(CarthageRequiredGitVersion) or later. Please update your git version at first.\n", stderr)
```
mdiep: Can we drop the second sentence and just make it `Carthage requires git \(CarthageRequiredGitVersion) or later.`?

Source/CarthageKit/Git.swift
```diff
@@ -11,6 +11,9 @@ import Result
 import ReactiveCocoa
 import ReactiveTask
 
+/// The git version Carthage requires at least.
+public let CarthageRequiredGitVersion = "2.3.0"
```
NachoSoto: Awesome

NachoSoto:
:+1:

mdiep:
This is really great. :sparkles:

Source/carthage/main.swift
ikesyo: Fixed in 74ba78b. :zap:  `\n` is still needed as we already do [here](https://github.com/Carthage/Carthage/blob/ensure-git-version/Source/carthage/main.swift#L35).

ikesyo:
Thank you for the reviews! :v:

PR #932, opened 2015-11-20
Use `cp -Rf` instead of `cp -rf` in Makefile

norio-nomura:
`-r` option of `cp` copies symbolic links as normal files.
As the result of that, `CarthageKit.framework` that distributed by Homebrew contains a lot of duplicated files inside that.
`-R` option copies symbolic links as symbolic links.

On my testing, this fix reduces the size of `carthage-0.10.el_capitan.bottle.tar.gz` from 11.4MB to 3.5MB.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ installables: clean bootstrap
 
 prefix_install: installables
 	mkdir -p "$(PREFIX)/Frameworks" "$(PREFIX)/bin"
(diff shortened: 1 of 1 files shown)


ikesyo:
I checked `man cp` and that sais:

> Symbolic links are always followed unless the -R flag is set, in which case symbolic links are not followed, by default.  The -H or -L flags (in conjunction with the -R flag) cause symbolic links to be followed as described above.  The -H, -L and -P options are ignored unless the -R option is specified. In addition, these options override each other and the command's actions are determined by the last one specified.

and

> Historic versions of the cp utility had a -r option.  This implementation supports that option; however, its use is strongly discouraged, as it does not correctly copy special files, symbolic links, or fifo's.

so this seems to be the correct fix. :+1:

PR #935, opened 2015-11-21
Modify LD_RUNPATH_SEARCH_PATHS to prioritize Homebrew installation

### LINKED ISSUE #495: /Library/Frameworks takes priority over Homebrew installation
If you've installed Carthage from [the Homebrew formula](https://github.com/Homebrew/homebrew/blob/d1f70d44369de4d4998af351428f2b72e3dbf29d/Library/Formula/carthage.rb), you might expect that `carthage` would always use the version of CarthageKit installed from Homebrew.

Unfortunately, it seems like the executable always prioritizes loading frameworks from `/Library/Frameworks`, even though that's only populated from our package installer. This results in issues like #490.

We should update the formula and/or `Makefile` to prioritize custom framework locations over the “base” ones.

ikesyo:
Fixes #495.

This change prioritizes `@executable_path/../Frameworks` and `@executable_path/../Frameworks/CarthageKit.framework/Versions/Current/Frameworks` over `/Library/Frameworks`.

If pkg installation (or `make install`) is used, `/usr/local/bin/carthage` would use CarthageKit in `/Library/Frameworks` since there should be no `/usr/local/Frameworks/CarthageKit.framework`.

If homebrew installation is used, `/usr/local/bin/carthage`, which is really `/usr/local/Celler/carthage/$version/bin/carthage`, would use CarthageKit in `/usr/local/Celler/carthage/$version/Frameworks/CarthageKit.framework` by the updated priority.

I want detailed check for this. :wink:

### DIFF
--- a/Carthage.xcodeproj/project.pbxproj
+++ b/Carthage.xcodeproj/project.pbxproj
@@ -890,7 +890,7 @@
 			baseConfigurationReference = D0D1213419E878CC005E4BAA /* Mac-Application.xcconfig */;
 			buildSettings = {
 				INFOPLIST_FILE = Source/carthage/Info.plist;
-				LD_RUNPATH_SEARCH_PATHS = "@executable_path/. @executable_path/CarthageKit.framework/Versions/Current/Frameworks /Library/Frameworks /Library/Frameworks/CarthageKit.framework/Versions/Current/Frameworks $(inherited)";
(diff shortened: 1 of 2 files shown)


ikesyo:
I'm sorry, 

> If pkg installation (or `make install`) is used, `/usr/local/bin/carthage` would use CarthageKit in `/Library/Frameworks` since there should be no `/usr/local/Frameworks/CarthageKit.framework`.

is not right, since Homebrew would create a symlink at `/usr/local/Frameworks`.

I'm considering another way...

mdiep:
It seems like we should probably have different make targets  for the .pkg and for home brew that add/delete different rpaths.

Issue #790, opened 2015-09-23
Alamofire 2.0.2 failing to build on watchOS


cnoon:
Hey everyone,

I'm fairly certain I have everything in the Alamofire project configured correctly, but carthage seems to be a bit unhappy about building our `watchOS` target. I figured I'd try to pull in the experts to help get it sorted out. Here's all the info I can think to provide:

### Cartfile

```bash
github "Alamofire/Alamofire" "2.0.2"
```

### Carthage Version

```bash
$ carthage version
0.8.0
```

### Xcodebuild Version

```bash
$ xcodebuild -version
Xcode 7.0
Build version 7A220
```

### Carthage Update Output

```bash
$ carthage update --platform watchos

Build settings from command line:
    ONLY_ACTIVE_ARCH = NO
    SDKROOT = watchos2.0

=== BUILD TARGET Alamofire watchOS OF PROJECT Alamofire WITH CONFIGURATION Release ===

Check dependencies

... lots of build details ...
...code signing

** BUILD SUCCEEDED **

Build settings from command line:
    ONLY_ACTIVE_ARCH = NO
    SDKROOT = watchsimulator2.0

=== BUILD TARGET Alamofire watchOS OF PROJECT Alamofire WITH CONFIGURATION Release ===

Check dependencies
target 'Alamofire watchOS' has bitcode disabled (ENABLE_BITCODE = NO), but it is required for the 'watchos' platform
```

Please let me know if there are any other details I can provide that would be of use. More details about our filed issues can be found in [Alamofire #689](https://github.com/Alamofire/Alamofire/issues/689).

marcpalmer:
I am seeing exactly this problem with CocoaLumberjack's "swift_2.0" branch.

If I open the Lumberjack xcodeproj in Xcode 7.1 beta, I can build all the watchOS targets and they show bitcode "YES" in Build Settings. 

However Carthage seems to fail to carry the bitcode option through. xcode-select has been run in the shell to point at the 7.1 beta tools.

cnoon:
Apologies for not including this originally, but Alamofire also has bitcode enabled on the `watchOS` target.

guidomb:
Could this be related with https://github.com/antitypical/Result/issues/71 ?

jurvis:
[...]

PR #941, opened 2015-11-26
Specify destination for simulator builds

ikesyo:
This seems to be required for Xcode 7.2.

https://github.com/Carthage/Carthage/issues/790#issuecomment-159349939 should be the case.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -322,6 +322,17 @@ public enum SDK: String {
 		return Result(self.init(rawValue: string.lowercaseString), failWith: .ParseError(description: "unexpected SDK key \"\(string)\""))
 	}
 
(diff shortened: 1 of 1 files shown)


NachoSoto:
😵😵😵

This might actually work around http://www.openradar.appspot.com/22993940!!

NachoSoto:
Thus looks good! Awesome find @ikesyo, I didn't realize we had this implemented but only for iOS!

:shipit:

mdiep:
:clap:

NachoSoto:
Apple got back to me in [rdar://22993940](http://www.openradar.appspot.com/22993940):
> While we will investigate this behavioral change, we recommend using the `-destination` flag rather than the `-sdk` flag for this use case.

So looks like you got it right @ikesyo! Still they said they're going to look into it because they broke a previously working command...

PR #943, opened 2015-11-26
Workaround for carthage build error (XCode 7.1.1).

guyschaos:
xcodebuild -list always append "Plugin Console Open With Application: [NSApp mainMenu] returned nil" at last of schemes.


ikesyo:
Thanks for the PR! However, this should be happening because of the Xcode plugins you are using, not of Xcode itself. Please remove or fix the plugin which causes the problem on your side. We won't add a special fix like this (Xcode plugins are not officially supported).

ikesyo:
(https://github.com/inquisitiveSoft/Open-with-Application should be the cause.)

guyschaos:
Yes, problem solved. Thank you. 
You guys rock!

PR #944, opened 2015-11-27
Fix #851: Signal handler is not signal-safe

### LINKED ISSUE #851: Signal handler is not signal-safe
The signal handler found in `swift-is-crashy.c` is not implemented in a signal-safe manner. Specifically, neither `strsignal()` nor `fprintf()` are in the list of functions that are safe to be called from signal handlers (see `man sigaction` for the list). One potential result of this not being signal-safe is that it may deadlock when attempting to handle the signal (e.g. if either function allocates memory and the malloc lock is already held by something else).

cradnovich:
Only use write(), raise(), and a custom strnlen (probably unneeded) instead of fprintf.

https://github.com/Carthage/Carthage/issues/851


Source/carthage/swift-is-crashy.c
```diff
@@ -27,10 +28,29 @@ static const char * const exceptionMessage = "\n"
...
+		if (string[i] == '\0') {
+			return i;
+		}
+	}
 
+	return max;
+}
+
+static void uncaughtSignal(int zig, siginfo_t *info, void *context) {
+	size_t preludeLength = safe_strnlen(exceptionPrelude, 90); // 47 at this time
```
NachoSoto: Maybe add a comment to the top of the file to make it more obvious that these lengths have to always be greater?

mdiep: Can we not use `sizeof`?

cradnovich: Can't use `sizeof` for the `signalName` string that comes from `sys_siglist`, but `strlen` works for that.

PR #952, opened 2015-12-01
Fix merge order of frameworks to ensure that the device build should be the base (destination) one

### LINKED ISSUE #824: Apple does not allow submitting Apple TV apps with frameworks that include 'appletvsimulator' in SUPPORTED_PLATFORMS
I just filed this radar: http://www.openradar.me/22967941

This is a very similar scenario to the bug that lead to implementing `copy-framework` to strip extra architectures, except now, `AppleTV` apps should only have `appletvos` as `SUPPORTED_PLATFORMS`.

I think `Carthage` (and libraries) should continue building with both `appletvos` and `appletvsimulator`, obviously, so that these frameworks are usable in both simulator and device. However, would it be possible to make `copy-frameworks` remove `appletvsimulator` from the framework `.plist`, to workaround this issue?

there is always a way

### EDIT October 8th:
Apple closed my radar as a dupe, and after I asked for more information they replied the following:
> Xcode does not support frameworks which are fat across multiple platforms. A framework which specifies that it supports the simulator platform, but lacks the architectures for it, as you've constructed, is an invalid framework; it's a bug that iOS and watchOS validation does not catch this. We're marking your bug as a duplicate of the general bug we have about providing a better workflow for distributing built simulator frameworks. We appreciate the detailed bug report!

So yeah, we need to make `copy-frameworks` strip simulator platforms from `SUPPORTED_PLATFORMS`.

ikesyo:
Should fix #824.

The order of SDK list which comes form `xcodebuild -showBuildSettings | grep SUPPORTED_PLATFORMS` are different depending on the platform.

- iOS: `iphonesimulator iphoneos`
- watchOS: `watchsimulator watchos`
- tvOS: `appletvos appletvsimulator`

In tvOS case, the simulator build had been the base framework when merging with the device build. Thus, `CFBundleSupportedPlatforms` would be `AppleTVSimulator` and `DTPlatformName` would be `appletvsimulator` in the resulting Info.plist.

This fix ensures that the device build should be the base one when merging the two frameworks.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -299,7 +299,7 @@ extension Platform: CustomStringConvertible {
 }
 
 /// Represents an SDK buildable by Xcode.
-public enum SDK: String {
+public enum SDK: String, Comparable {
 	/// Mac OS X.
 	case MacOSX = "macosx"
 
@@ -352,6 +352,39 @@ public enum SDK: String {
 		case .MacOSX:
 			return .Mac
 		}
+	}
+}
+
+public func < (lhs: SDK, rhs: SDK) -> Bool {
+	switch (lhs, rhs) {
+	// Prefer Mac OS X SDK.
+	case (.MacOSX, _):
+		return true
+
+	case (_, .MacOSX):
+		return false
+
+	// Prefer device SDK over simulator SDK in same platform.
+	case (.iPhoneOS, .iPhoneSimulator):
+		return true
+
+	case (.iPhoneSimulator, .iPhoneOS):
+		return false
+
+	case (.watchOS, .watchSimulator):
+		return true
+
+	case (.watchSimulator, .watchOS):
+		return false
+
+	case (.tvOS, .tvSimulator):
+		return true
+
+	case (.tvSimulator, .tvOS):
+		return false
+
+	case _:
+		return true
 	}
 }
 
@@ -967,7 +1000,11 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 		}
 		.flatMap(.Concat) { platform, sdks -> SignalProducer<(Platform, [SDK]), CarthageError> in
 			let filterResult = sdkFilter(sdks: sdks, scheme: scheme, configuration: configuration, project: project)
-			return SignalProducer(result: filterResult.map { (platform, $0) })
+			return SignalProducer(result: filterResult.map { sdks in
+				// Ensure that a device SDK are ordered before than a simulator
+				// SDK of the same platform.
+				return (platform, sdks.sort())
+			})
 		}
 		.filter { _, sdks in
 			return !sdks.isEmpty
@@ -984,10 +1021,10 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 					}
 
 			case 2:
-				let firstSDK = sdks[0]
-				let secondSDK = sdks[1]
+				let deviceSDK = sdks[0]
+				let simulatorSDK = sdks[1]
 
-				return settingsByTarget(buildSDK(firstSDK))
+				return settingsByTarget(buildSDK(deviceSDK))
 					.flatMap(.Concat) { settingsEvent -> SignalProducer<TaskEvent<(BuildSettings, BuildSettings)>, CarthageError> in
 						switch settingsEvent {
 						case let .StandardOutput(data):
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +355,39 @@ public enum SDK: String {
 	}
 }
 
+public func < (lhs: SDK, rhs: SDK) -> Bool {
+	switch (lhs, rhs) {
```
ikesyo: We can just compare `description` or `rawValue` here instead.

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +355,39 @@ public enum SDK: String {
...
+		return false
+
+	case (.tvOS, .tvSimulator):
+		return true
+
+	case (.tvSimulator, .tvOS):
+		return false
+
+	case _:
+		return true
```
mdiep: I don't love returning true in this case. I'd prefer to establish a total order by specifying the order of different platforms. We could either sort by `self.platform.rawValue` here or compare the `rawValue` like you suggested above. If we do the latter, I think  it'd be good to add a comment explaining that simulators need to come last.

Source/CarthageKit/Xcode.swift
```diff
@@ -967,7 +1000,11 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 		}
 		.flatMap(.Concat) { platform, sdks -> SignalProducer<(Platform, [SDK]), CarthageError> in
 			let filterResult = sdkFilter(sdks: sdks, scheme: scheme, configuration: configuration, project: project)
-			return SignalProducer(result: filterResult.map { (platform, $0) })
+			return SignalProducer(result: filterResult.map { sdks in
+				// Ensure that a device SDK are ordered before than a simulator
+				// SDK of the same platform.
```
mdiep: This comment could use a little cleanup. :pray:

> Ensure that device SDKs are ordered before simulator SDKs of the same platform.

mdiep:
Good catch. :clap: I'd like to get this into the next release.

NachoSoto:
😮 good catch @ikesyo!!

Source/CarthageKit/Xcode.swift
```diff
@@ -984,10 +1021,10 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 					}
 
 			case 2:
-				let firstSDK = sdks[0]
-				let secondSDK = sdks[1]
+				let deviceSDK = sdks[0]
```
NachoSoto: This still seems fragile. How about a function in `SDK` like:
```swift
func splitSDKs(sdks: [SDK]) -> (simulators: [SDK], devices: [SDK]) {

}
```

Then here we can use that and do split `SDK.simulators.first!`

NachoSoto: Then we wouldn't need the `Comprable` implementation which IMO doesn't make much sense for a type like `SDK`.

ikesyo:
Thank you for the feedbacks! I removed `Comparable` implementations and added `SDK.splitSDKs` as @NachoSoto said.

mdiep:
Looks good to me, but I'll let @NachoSoto take another look.

Source/CarthageKit/Xcode.swift
```diff
@@ -326,6 +326,22 @@ public enum SDK: String {
...
+
+		sdks.forEach { sdk in
+			if sdk.isSimulator {
+				simulators.append(sdk)
+			} else {
+				devices.append(sdk)
+			}
+		}
+
+		return (simulators: simulators, devices: devices)
```
NachoSoto: How about:
```swift
return (
    simulators: sdks.filter { $0.isSimulator },
    devices: sdks.filter { !$0.isSimulator }
)
```

Source/CarthageKit/Xcode.swift
```diff
@@ -967,7 +983,9 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 		}
 		.flatMap(.Concat) { platform, sdks -> SignalProducer<(Platform, [SDK]), CarthageError> in
 			let filterResult = sdkFilter(sdks: sdks, scheme: scheme, configuration: configuration, project: project)
-			return SignalProducer(result: filterResult.map { (platform, $0) })
+			return SignalProducer(result: filterResult.map { sdks in
+				return (platform, sdks)
```
NachoSoto: Nit: maybe just leave this as one line? `return SignalProducer(result: filterResult.map { (platform, $0) })`

Source/CarthageKit/Xcode.swift
```diff
@@ -984,10 +1002,11 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 					}
 
 			case 2:
-				let firstSDK = sdks[0]
-				let secondSDK = sdks[1]
+				let (simulatorSDKs, deviceSDKs) = SDK.splitSDKs(sdks)
+				let deviceSDK = deviceSDKs.first!
```
NachoSoto: It might be useful to generate an error here. Maybe
```swift
guard let deviceSDK = deviceSDKs.first else { fatalError("Could not find device SDK in \(sdks)") }
guard let simulatorSDK = simulatorSDKs.first else { fatalError("Could not find simulator SDK in \(sdks)") }
```

NachoSoto:
Looks great! Just a few suggestions to slightly improve the implementation, but the whole idea is perfect.

ikesyo:
Updated! :zap:

NachoSoto:
Perfect! :heart_eyes:

NachoSoto:
Awesome job @ikesyo!

ikesyo:
:tada:

PR #953, opened 2015-12-01
Documentation for travis-ci framework uploads

flovilmart:
- Small addition to the README in order to document how to setup travis releases

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,8 +180,8 @@ It is possible to use travis-ci in order to build and upload your tagged release
 	osx_image: xcode7.1
 	env:
 		global: 
-		- XCODE_SCHEME=
-		- FRAMEWORK_NAME=
+		- XCODE_SCHEME=<THIS_IS_A_PLACEHOLDER_REPLACE_ME>
+		- FRAMEWORK_NAME=<THIS_IS_A_PLACEHOLDER_REPLACE_ME>
 	before_install:
 	- brew install carthage
 	before_script:
--- a/README.md
+++ b/README.md
@@ -216,13 +216,14 @@ It is possible to use travis-ci in order to build and upload your tagged release
 		on: 
 			repo: repo/repo
 		
-	And add `tags: true`:
-	
+	And add `tags: true` and `skip_cleanup: true`:
+		
+		skip_cleanup: true
 		on:
 			repo: repo/repo
 			tags: true
 	
-	That will let travis know to create a deployment when a new tag is pushed.
+	That will let travis know to create a deployment when a new tag is pushed and prevent travis to cleanup the generated zip file
 
 ### Declare your compatibility
 
--- a/README.md
+++ b/README.md
@@ -173,14 +173,26 @@ It is possible to use travis-ci in order to build and upload your tagged release
 
 1. [Install travis CLI](https://github.com/travis-ci/travis.rb#installation) with `gem install travis`
 1. [Setup](https://docs.travis-ci.com/user/getting-started/) travis-ci for your repository (Steps 1 and 2)
-1. Create `.travis.yml` file at the root of your repository based on that template. Set `XCODE_SCHEME` as well as `FRAMEWORK_NAME` to your correct values
+1. Create `.travis.yml` file at the root of your repository based on that template. Set `FRAMEWORK_NAME` to the correct value.
 
+	Replace PROJECT_PLACEHOLDER and SCHEME_PLACEHOLDER
+	
+	If you are using a *workspace* instead of a *project* remove the xcode_project line and uncomment the xcode_workspace line.
+	
+	The project should be in the format: MyProject.xcodeproj
+	
+	The workspace should be in the format: MyWorkspace.xcworkspace
+	
+	For mor informations you can visit [travis docs for objective-c projects](https://docs.travis-ci.com/user/languages/objective-c)
+	
 	```
 	language: objective-c
 	osx_image: xcode7.1
+	xcode_project: <PROJECT_PLACEHOLDER>
+	# xcode_workspace: <WORKSPACE_PLACEHOLDER>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -167,6 +167,55 @@ carthage archive YourFrameworkName
...
+1. [Setup](https://docs.travis-ci.com/user/getting-started/) travis-ci for your repository (Steps 1 and 2)
+1. Create `.travis.yml` file at the root of your repository based on that template. Set `XCODE_SCHEME` as well as `FRAMEWORK_NAME` to your correct values
+
+	```
+	language: objective-c
+	osx_image: xcode7.1
+	env:
+		global: 
+		- XCODE_SCHEME=
+		- FRAMEWORK_NAME=
```
robrix: Maybe we should put placeholders here, and call the placeholder values about in the above list instead.

flovilmart: What do you mean? something like:

```
XCODE_SCHEME=MyScheme
FRAMEWORK_NAME=MyFramework
```

And say Replace MyScheme by the name of your Scheme and MyFramework by the name of your Framework?

robrix: Yes, but I was thinking a much more obvious placeholder, in the vein of `<THIS_IS_A_PLACEHOLDER_REPLACE_ME>` ^_^

robrix: That said the README currently uses `YourFrameworkName` in the archive section, so we can probably go with that.

NachoSoto:
This is great!

README.md
flovilmart: I can still update with our suggestion

flovilmart:
@NachoSoto next step is `carthage setup travis`

README.md
```diff
@@ -167,6 +167,55 @@ carthage archive YourFrameworkName
...
+	env:
+		global: 
+		- XCODE_SCHEME=<THIS_IS_A_PLACEHOLDER_REPLACE_ME>
+		- FRAMEWORK_NAME=<THIS_IS_A_PLACEHOLDER_REPLACE_ME>
+	before_install:
+	- brew install carthage
+	before_script:
+	# update the dependencies for the project
+	# you can remove if you don't have dependencies
+	- carthage update
```
neilpa: This should probably default to `carthage bootstrap`

flovilmart: I never really got the difference between `bootstrap` and `update`

mdiep: It should definitely be `bootstrap`.

`update` updates `Cartfile.resolved` to the newest version of all dependencies and then `bootstrap`s.

neilpa: `bootstrap` syncs exactly what's in the `Cartfile.resolved`. `update` could pick a newer dependency if you've specified something like `>= 1.0` and there's a new version available.

README.md
```diff
@@ -167,6 +167,55 @@ carthage archive YourFrameworkName
...
+		- XCODE_SCHEME=<THIS_IS_A_PLACEHOLDER_REPLACE_ME>
+		- FRAMEWORK_NAME=<THIS_IS_A_PLACEHOLDER_REPLACE_ME>
+	before_install:
+	- brew install carthage
+	before_script:
+	# bootstrap the dependencies for the project
+	# you can remove if you don't have dependencies
+	- carthage bootstrap	
+	script:
+	- true
```
mdiep: Why is this needed? Can we remove this line and include `script:` in the commented-out lines below?

flovilmart: I'm not sure travis will like without the `script` instruction.

I can try

README.md
```diff
@@ -167,6 +167,55 @@ carthage archive YourFrameworkName
...
+	# uncomment those line to run tests:
+	# - xcodebuild -scheme $XCODE_SCHEME -destination 'platform=iOS Simulator,name=iPhone 6,OS=9.1' -sdk iphonesimulator test
```
mdiep: Actually, these lines shouldn't be needed at all, right? If the `script:` item is removed above, then Travis should automatically run the tests.

README.md
```diff
@@ -167,6 +167,55 @@ carthage archive YourFrameworkName
...
+1. [Install travis CLI](https://github.com/travis-ci/travis.rb#installation) with `gem install travis`
+1. [Setup](https://docs.travis-ci.com/user/getting-started/) travis-ci for your repository (Steps 1 and 2)
+1. Create `.travis.yml` file at the root of your repository based on that template. Set `XCODE_SCHEME` as well as `FRAMEWORK_NAME` to your correct values
+
+	```
+	language: objective-c
+	osx_image: xcode7.1
+	env:
+		global: 
+		- XCODE_SCHEME=<THIS_IS_A_PLACEHOLDER_REPLACE_ME>
```
mdiep: This should use the builtin `xcode_scheme` value. And I think we should demonstrate using `xcode_workspace` too.

README.md
flovilmart: So that forces the user to have tests and use the default configuration.
I am not sure about all that I'm checking now.

README.md
flovilmart: Actually we don't need it for building the framework. If we want to have multiple options, let's put some gists there for the non travis aficionados.

mdiep: Setting `xcode_scheme` and `xcode_workspace` should let us get rid of `script:` altogether IIUC.

flovilmart: As per: 
https://docs.travis-ci.com/user/languages/objective-c#Default-Test-Script

I'll update

mdiep:
:fireworks:

flovilmart:
Can't wait to have those in release :) 4 frameworks pending :)

flovilmart:
Actually, we should have also: 

`xcode_sdk: iphonesimulator9.1`

 for example as otherwise the build fails

mdiep:
Hmm… I didn't think that was necessary. But if it is, a PR would be great.

flovilmart:
Yes it is :) just had a bad surprise trying it out :)

PR #966, opened 2015-12-03
Allow multiple platform list to contain spaces

### LINKED ISSUE #957: Allow multiple platform list to contain spaces
`carthage update --platform "iOS, watchOS"` should work, but doesn't because of the ` ` after the `,`.

Reported in https://github.com/Carthage/Carthage/pull/726#issuecomment-161275805.

ikesyo:
Fixes #957.

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -308,12 +308,10 @@ extension BuildPlatform: ArgumentType {
 			.map(String.init)
 
 		let findBuildPlatform: String -> BuildPlatform? = { string in
-			for (key, platform) in self.acceptedStrings {
-				if string.caseInsensitiveCompare(key) == .OrderedSame {
-					return platform
-				}
-			}
-			return nil
+			return self.acceptedStrings.lazy
+				.filter { key, _ in string.caseInsensitiveCompare(key) == .OrderedSame }
+				.map { _, platform in platform }
+				.first
 		}
 
 		switch tokens.count {
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -303,7 +303,9 @@ extension BuildPlatform: ArgumentType {
 	]
 
 	public static func fromString(string: String) -> BuildPlatform? {
-		let commaSeparated = string.characters.split(allowEmptySlices: false) { $0 == "," }.map(String.init)
+		let tokens = string.characters
+			.split(allowEmptySlices: false) { $0 == "," || $0 == " " }
+			.map(String.init)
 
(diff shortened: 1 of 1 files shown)


Source/carthage/Build.swift
```diff
@@ -314,17 +316,17 @@ extension BuildPlatform: ArgumentType {
 			return nil
 		}
 
-		switch commaSeparated.count {
+		switch tokens.count {
 		case 0:
 			return nil
 
 		case 1:
-			return findBuildPlatform(commaSeparated[0])
```
NachoSoto: Maybe use `commaSeparated.first!`?

Source/carthage/Build.swift
```diff
@@ -314,17 +316,17 @@ extension BuildPlatform: ArgumentType {
...
+			return findBuildPlatform(tokens[0])
 
 		default:
 			var buildPlatforms = [BuildPlatform]()
```
NachoSoto: This could be replaced with:
```swift
let buldPlatforms = tokens
    .map(self.findBuildPlatform)
    .filter { $0 != .All }
```

Source/carthage/Build.swift
```diff
@@ -303,7 +303,9 @@ extension BuildPlatform: ArgumentType {
 	]
 
 	public static func fromString(string: String) -> BuildPlatform? {
-		let commaSeparated = string.characters.split(allowEmptySlices: false) { $0 == "," }.map(String.init)
+		let tokens = string.characters
+			.split(allowEmptySlices: false) { $0 == "," || $0 == " " }
+			.map(String.init)
 
 		let findBuildPlatform: String -> BuildPlatform? = { string in
 			for (key, platform) in self.acceptedStrings {
```
NachoSoto: This could be
```swift
return lazy(self.acceptedStrings)
    .filter { key, _ in string.caseInsensitiveCompare(key) }
    .map { _, platform return platform }
    .first
```

NachoSoto:
This looks good! Just a few optional suggestions :)

Source/carthage/Build.swift
ikesyo: That would break this:

```swift
// Reject if an invalid value is included in the comma-
// separated string.
return nil
```

Source/carthage/Build.swift
ikesyo: It seems that we are using `[0]` rather than `first!` in the codebase.

ikesyo:
Thank you for the suggestions! I've updated `findBuildPlatform`.

mdiep:
:clap:

NachoSoto:
:rocket:

PR #973, opened 2015-12-05
Avoid curried function syntax

ikesyo:
That will be removed in Swift 3.0, so let's avoid to use it.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -53,12 +53,8 @@ extension GitHubError: CustomStringConvertible {
 }
 
 extension GitHubError: Decodable {
-	public static func create(message: String) -> GitHubError {
-		return self.init(message: message)
-	}
-	
 	public static func decode(j: JSON) -> Decoded<GitHubError> {
-		return self.create
+		return self.init
 			<^> j <| "message"
 	}
 }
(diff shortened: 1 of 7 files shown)


mdiep:
:+1:

Source/CarthageKit/GitHub.swift
```diff
@@ -306,8 +304,10 @@ extension GitHubRelease: CustomStringConvertible {
 }
 
 extension GitHubRelease: Decodable {
-	public static func create(ID: Int)(name: String?)(tag: String)(draft: Bool)(prerelease: Bool)(assets: [Asset]) -> GitHubRelease {
-		return self.init(ID: ID, name: name, tag: tag, draft: draft, prerelease: prerelease, assets: assets)
+	public static func create(ID: Int) -> String? -> String -> Bool -> Bool -> [Asset] -> GitHubRelease {
+		return { name in { tag in { draft in { prerelease in { assets in
+			return self.init(ID: ID, name: name, tag: tag, draft: draft, prerelease: prerelease, assets: assets)
+		} } } } }
```
NachoSoto: Ouch :(

mdiep: It'd be nice if they'd build in some real support for currying. :confused:

gfontenot: fwiw, these `create` functions shouldn't be necessary anymore. If you use something like [Curry.framework][curry], you could use `curry(GitHubRelease.init)` instead. These `create` functions were only really necessary pre-swift 2.0.

[curry]: https://github.com/thoughtbot/Curry

ikesyo:
I know that framework, but `Curry` also uses curried function syntax currently, so it's not the solution here.

Source/CarthageKit/GitHub.swift
ikesyo: I know that framework, but Curry also uses curried function syntax currently, so it's not the solution here.

PR #972, opened 2015-12-05
Support specifying dependencies list in `build` / `checkout` / `update` commands

ikesyo:
Resolves #610. Supersedes @NachoSoto's work of #618 and #886.

#218 is also related:

> Along the same lines, `carthage update` could accept a specific list of dependency names to bump.

This PR introduces an argument list to `build`, `checkout` and `update` command ~~(someday `--exclude` option might be added or not)~~. That accepts dependency names as follows:

```bash
carthage update ReactiveTask Commandant --platform mac
```

For `build` command, only the given dependencies and their nested dependencies would be build~~, but their nested dependencies would not since we can't detect them from `Cartfile.resolved`~~.

For `checkout` command, only the given dependencies and their nested dependencies would be checked out.

For `update` command, only the versions of the given dependencies and their nested dependencies would be updated.

**TODO**

- [x] https://github.com/Carthage/Carthage/pull/972#discussion_r46896585
- [x] https://github.com/Carthage/Carthage/pull/972#discussion_r46896714
- [x] Change `--include` option to a positional argument.
- [x] Build nested dependencies as well.
- [x] Support this for `checkout` as well.
- [x] Unit tests.
- [x] [Support multiple arguments of dependency names](https://github.com/Carthage/Carthage/pull/972#issuecomment-164314514).

PR #974, opened 2015-12-05
Determine the build order when building the dependencies in Cartfile.resolved

### LINKED ISSUE #717: Sort Cartfile.resolved
Any operation that outputs Cartfile.resolved should save it using a line based sort.

This would really help with git diffs and merging. Currently the order of the file seems to be pretty arbitrary and random.

ikesyo:
~~The build order is not preserved in `Cartfile.resolved` now.~~

Trying to fix #717 based on the discussion: https://github.com/Carthage/Carthage/issues/717#issuecomment-135501479

### DIFF
--- a/Source/CarthageKit/Cartfile.swift
+++ b/Source/CarthageKit/Cartfile.swift
@@ -270,14 +270,18 @@ public struct Dependency<V: VersionType>: Equatable {
 	/// The version(s) that are required to satisfy this dependency.
 	public var version: V
 
-	public init(project: ProjectIdentifier, version: V) {
+	/// The dependencies of this dependency.
+	public let dependencies: Set<ProjectIdentifier>
+
+	public init(project: ProjectIdentifier, version: V, dependencies: Set<ProjectIdentifier> = []) {
 		self.project = project
 		self.version = version
+		self.dependencies = dependencies
 	}
 
 	/// Maps over the `version` in the receiver.
 	public func map<W: VersionType>(f: V -> W) -> Dependency<W> {
-		return Dependency<W>(project: project, version: f(version))
+		return Dependency<W>(project: project, version: f(version), dependencies: dependencies)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


Source/CarthageKit/Cartfile.swift
```diff
@@ -270,14 +270,18 @@ public struct Dependency<V: VersionType>: Equatable {
 	/// The version(s) that are required to satisfy this dependency.
 	public var version: V
 
-	public init(project: ProjectIdentifier, version: V) {
+	/// The dependencies of this dependency.
+	public let dependencies: Set<ProjectIdentifier>
```
ikesyo: This is intended to be used for https://github.com/Carthage/Carthage/pull/972#issuecomment-162215783.

mdiep:
I think we need to break this into 2 changes, separated by several months and releases, in order to keep `Cartfile.resolved`s compatible with recent versions of Carthage.

1. Don't rely on the order of `Cartfile.resolved`: calculate dependencies when needed.
2. Output `Cartfile.resolved` in lexicographical order.

We need to allow a significant period of time between (1) and (2) so that everyone can upgrade to a version that does (1) before we release a version that does (2). Otherwise we run into an Xcode-upgrade type scenario.

Source/CarthageKitTests/ResolverSpec.swift
```diff
@@ -84,11 +84,11 @@ class ResolverSpec: QuickSpec {
 
 			var generator = dependencies.generate()
 
-			// Dependencies should be listed in build order.
+			// Dependencies should be listed in lexical order of its project name.
```
mdiep: We should make sure we add new tests that the dependencies in `Cartfile.resolved` are sorted in build order.

ikesyo: Exactly. :bow:

ikesyo:
Thanks for the idea, that seems totally reasonable. :smiley: 

> 1. Don't rely on the order of `Cartfile.resolved`: calculate dependencies when needed.

I would update this PR to work on only the thing.

ikesyo:
It's ready. :zap:

mdiep:
Thanks for tackling this! :sparkling_heart:

schwa:
Thanks for this!

PR #978, opened 2015-12-07
listing possible platform parameters in help for the update command

yene:
Took me some time to find out where the platform parameters are documented (only under help build). 
I hope this helps other users.

Maybe `--platform` should also be explained in the Readme.


mdiep:
Thanks!

PR #979, opened 2015-12-07
Update README.md

ealeksandrov:
Add `brew update` step in Travis CI script. Otherwise it will install outdated Carthage version (it was 0.8 for me today) and will eventually lead to build fail.


NachoSoto:
Thanks!

PR #982, opened 2015-12-09
Change the positional argument of `directoryPath` to a named argument

ikesyo:
:warning: **THIS IS A BREAKING CHANGE** :warning: 

As discussed in https://github.com/Carthage/Carthage/pull/972#issuecomment-162650937.

That option is now named as `--project-directory`.


mdiep:
Thanks! :fireworks:

PR #988, opened 2015-12-11
Remove pre-0.4 migrations

### LINKED ISSUE #358: Remove migration from pre-0.4 Carthage versions
1.0 should constitute a clean break, so let's get rid of this cruft.

mdiep:
It’s been almost a year, so I think these are safe to remove.

Fixes #358.


NachoSoto:
:fire:

Issue #924, opened 2015-11-16
Correct framework DWARFs and symbols to workaround broken debugging 🔥


NachoSoto:
*This is a follow up to #832 with details on a possible solution.*

## The story:

We've all been aware since the very beginning that debugging projects that include pre-compiled (Carthage) frameworks has been very broken. As Apple *promised* that this got better in Xcode 7.2, a few people on #832 did the test (I still can't test it myself [because reasons](http://www.openradar.me/22993940)), and realized that it's still just as broken.

Upon further investigation, looks like [DWARF files use absolute paths everywhere](https://twitter.com/nachosoto/status/666309518775599104), based on when they were compiled. 

Using `dwarfdump`:
<img width="867" alt="screen shot 2015-11-16 at 09 37 05" src="https://cloud.githubusercontent.com/assets/685609/11190412/293eb692-8c4a-11e5-8d41-ec7dba10df1d.png">
<img width="1028" alt="screen shot 2015-11-16 at 09 43 53" src="https://cloud.githubusercontent.com/assets/685609/11190309/a61b9a00-8c49-11e5-89fd-dc2936c852f7.png">

As you can see, they're full of paths to **the derived data directory of the user who built the framework**.
Needless to say, it's not just a problem because we can't share these frameworks with other devs, but also for personal use, **given how extremely frequent one must clear derived data to work around Xcode**.

![what year is it](https://cloud.githubusercontent.com/assets/685609/11190448/59bf4002-8c4a-11e5-9c02-6ae9942d23b0.jpg)

This is why I finally decided to file a radar, because after **two years** this is still utterly broken: http://www.openradar.me/23551273.

Of course, I don't expect this to be fixed (ever?), so we need a solution.

#### Abandon `Carthage`:

We can go back to working like animals, adding all dependencies as submodules. However, **this just doesn't cut it, Apple**, when will you see this?
- Swift compilation times **are already insanely long**. *Khan Academy* was taking 15 minutes for a clean build before I changed `Swift` dependencies to pre-compiled frameworks.
- Nested dependencies: this solution would not be feasible if you have multiple dependencies that depend on a common one. This is one of the reasons why `CocoaPods` was born:

#### `CocoaPods`
*(Yeah, I went there, but I wanted to analyze this too)*

- Obviously deals with the dependency graph problem.
- It's *invasive* and very complex, which is why Carthage was born.
[...]

PR #989, opened 2015-12-12
Add "Known Issues" section

ratkins:
Warn about Apple's recommendation that Swift frameworks should be built from source on each machine they're used, suggest some work-arounds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,11 @@ Want to advertise that your project can be used with Carthage? You can add a com
 ```markdown
 [![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
 ```
+## Known issues
+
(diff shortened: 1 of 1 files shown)


mdiep:
Thanks! Looks good to me, but I'll give others some time to chime in.

README.md
```diff
@@ -240,6 +240,11 @@ Want to advertise that your project can be used with Carthage? You can add a com
...
 ```
+## Known issues
+
+As at Xcode 7.2, Apple recommends that ["Frameworks written in Swift should be compiled from source as part of the same project that depends on them to guarantee a single, consistent compilation environment. (22492040)"](https://developer.apple.com/library/ios/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html). Using Swift frameworks built on other machines will cause Xcode's [debugger to crash](https://github.com/Carthage/Carthage/issues/924) and other [strange build errors](https://github.com/Carthage/Carthage/issues/785). To avoid this, do not check your `Carthage/Build` directory into source control or use .framework release binaries. Instead, create a Run Script build phase in your Xcode project which calls `carthage build --no-use-binaries` (optionally wrap this in a check for the existence of the `Carthage/Build/` directory to avoid long build times due to recompiling dependencies unnecessarily.)
+
+Dupe [rdar://23551273](http://www.openradar.me/23551273) if you want Apple to fix the root cause of this problem.
```
NachoSoto: Ooooh great idea!

NachoSoto:
Awesome! :+1: 

Could you maybe change the wording at the beginning to  encourage people to open #924 if they want to read more details about why?

Otherwise this is perfect.

ratkins:
Done.

(And again, sorry for being an arse earlier.)

NachoSoto:
@ratkins no worries, I'm the first to lose my patience with all these bugs :D

NachoSoto:
:shipit: thank you!

PR #997, opened 2015-12-16
Specify concrete `Options` type instead of explicit `typealias`es

ikesyo:
Those can be inferred from the method signatures.


mdiep:
Huh. I didn't know that would work!

PR #1000, opened 2015-12-17
Making --no-use-binaries the default

cellularmitosis:
Hi,

Unfortunately our team wasn't aware of https://github.com/Carthage/Carthage/issues/924.  Hilarity ensued.

This pull request makes --no-use-binaries the default behavior of Carthage.  This should prevent others from getting bitten quite so hard by Apple's http://www.openradar.me/23551273.

### DIFF
--- a/Source/carthage/Checkout.swift
+++ b/Source/carthage/Checkout.swift
@@ -52,7 +52,7 @@ public struct CheckoutOptions: OptionsType {
 		return create
 			<*> m <| Option(key: "use-ssh", defaultValue: false, usage: "use SSH for downloading GitHub repositories")
 			<*> m <| Option(key: "use-submodules", defaultValue: false, usage: "add dependencies as Git submodules")
-			<*> m <| Option(key: "use-binaries", defaultValue: true, usage: "check out dependency repositories even when prebuilt frameworks exist, disabled if --use-submodules option is present" + useBinariesAddendum)
+			<*> m <| Option(key: "use-binaries", defaultValue: false, usage: "check out dependency repositories even when prebuilt frameworks exist, disabled if --use-submodules option is present" + useBinariesAddendum)
 			<*> ColorOptions.evaluate(m)
 			<*> m <| Option(key: "project-directory", defaultValue: NSFileManager.defaultManager().currentDirectoryPath, usage: "the directory containing the Carthage project")
 	}


Source/carthage/Checkout.swift
```diff
@@ -52,7 +52,7 @@ public struct CheckoutOptions: OptionsType {
 		return create
 			<*> m <| Option(key: "use-ssh", defaultValue: false, usage: "use SSH for downloading GitHub repositories")
 			<*> m <| Option(key: "use-submodules", defaultValue: false, usage: "add dependencies as Git submodules")
-			<*> m <| Option(key: "use-binaries", defaultValue: true, usage: "check out dependency repositories even when prebuilt frameworks exist, disabled if --use-submodules option is present" + useBinariesAddendum)
+			<*> m <| Option(key: "use-binaries", defaultValue: false, usage: "check out dependency repositories even when prebuilt frameworks exist, disabled if --use-submodules option is present" + useBinariesAddendum)
```
NachoSoto: I'm not familiar with this aspect of `Commandant`: will `no-use-binaries` be a valid option (though redundant) after this change?

NachoSoto:
I'm not against this, I think it's a good idea, but it's important to note that this is not enough, though. If you remove `derived data` after building, you'd have to build again.

Source/carthage/Checkout.swift
mdiep: Yes, it will still be valid.

mdiep:
I have really mixed feelings about this. In the majority of cases, I still use binaries—both locally and on CI. Depending on the dependencies, debugging them may not be important.

I'd love to hear if anyone has a strong opinion about this. /cc @Carthage/carthage

gfontenot:
One thing that strikes me as odd is that `--use-binaries` can't guarantee that Carthage will pull down binaries. This also doesn't really "fix" the issue with the crashing debugger, because of the reasons @NachoSoto has stated (checking in `Carthage/Build`, deleting Derived Data, etc).

cellularmitosis:
Hmm, I suppose it's also true that for someone who isn't on a team, they could be using binaries and still have working debugging.  So there are still legitimate use-cases for using binaries, despite the pitfalls.

mdiep:
After thinking about this some more, I'd rather not change the default. I'd be more open to printing a warning or something when Carthage downloads a binary.

zittix:
According to http://adcdownload.apple.com/Developer_Tools/Xcode_7.3_beta/Xcode_7.3_beta_Release_Notes.pdf this issue should now be solved with XCode 7.3 beta

NachoSoto:
It's not, they only made it not crash. See #924 for more details.

AndrewSB:
@mdiep can you talk about how you're using archives locally? My project's dependencies take about 15 minutes to build currently, so I'm looking for ways to cut down on that build time.

Prebuilding existing frameworks on my machine may save hours for my setup

mdiep:
The two options I'm aware of are https://github.com/guidomb/carthage_cache and https://github.com/146BC/RomeBuild.

There also been talk about local caching (#1222). But I'd also like to add a way to detect different versions for frameworks attached to GitHub Releases (#1354).

PR #1004, opened 2015-12-18
Make all Options structs into inner structs

mdiep:
I thought this would make the relationship between the `CommandType` and the `OptionsType` clearer.


ikesyo:
This makes much sense! :100:

NachoSoto:
Awesome

PR #1005, opened 2015-12-18
Support multiple frameworks in `archive` command

ikesyo:
This would be useful for some libraries, for example, which have a Objective-C framework and the corresponding Swift one (e.g. https://github.com/CocoaLumberjack/CocoaLumberjack).

```bash
carthage archive CocoaLumberjack CocoaLumberjackSwift
```

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -297,7 +297,7 @@ public func buildableSchemesInDirectory(directoryURL: NSURL, withConfiguration c
 
 /// Sends pairs of a scheme and a project, the scheme actually resides in
 /// the project.
-public func schemesToBuildOfProjectLocators(projects: [(ProjectLocator, [String])]) -> SignalProducer<[(String, ProjectLocator)], CarthageError> {
+public func schemesInProjects(projects: [(ProjectLocator, [String])]) -> SignalProducer<[(String, ProjectLocator)], CarthageError> {
 	return SignalProducer(values: projects)
 		.map { (project: ProjectLocator, schemes: [String]) in
 			// Only look for schemes that actually reside in the project
@@ -1268,7 +1268,7 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
 			// `.NoSharedFrameworkSchemes`.
 			.filter { projects in !projects.isEmpty }
 			.flatMap(.Merge) { (projects: [(ProjectLocator, [String])]) -> SignalProducer<(String, ProjectLocator), CarthageError> in
-				return schemesToBuildOfProjectLocators(projects)
+				return schemesInProjects(projects)
 					.flatMap(.Merge) { (schemes: [(String, ProjectLocator)]) -> SignalProducer<(String, ProjectLocator), CarthageError> in
 						if !schemes.isEmpty {
 							return .init(values: schemes)
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -270,7 +270,7 @@ public func schemesInProject(project: ProjectLocator) -> SignalProducer<String, 
 
 /// Finds schemes of projects or workspace, which Carthage should build, found
 /// within the given directory.
-public func buildableSchemesByProjectLocatorInDirectory(directoryURL: NSURL, withConfiguration configuration: String, forPlatforms platforms: Set<Platform> = []) -> SignalProducer<(ProjectLocator, [String]), CarthageError> {
+public func buildableSchemesInDirectory(directoryURL: NSURL, withConfiguration configuration: String, forPlatforms platforms: Set<Platform> = []) -> SignalProducer<(ProjectLocator, [String]), CarthageError> {
 	precondition(directoryURL.fileURL)
 
 	return locateProjectsInDirectory(directoryURL)
@@ -1256,7 +1256,7 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
 		// multiple times.
 		let (locatorBuffer, locatorObserver) = SignalProducer<(ProjectLocator, [String]), CarthageError>.buffer()
 
-		buildableSchemesByProjectLocatorInDirectory(directoryURL, withConfiguration: configuration, forPlatforms: platforms)
+		buildableSchemesInDirectory(directoryURL, withConfiguration: configuration, forPlatforms: platforms)
 			.startWithSignal { signal, signalDisposable in
 				disposable += signalDisposable
 				signal.observe(locatorObserver)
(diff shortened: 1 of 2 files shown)


ikesyo:
Or should we remove [this line](https://github.com/Carthage/Carthage/blob/0.11/Source/CarthageKit/Project.swift#L390) and recommend to attach multiple `.framework.zip` files to their GitHub releases as [CocoaLumberjack currently does](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/2.2.0)?

NachoSoto:
Yeah, I think I'd rather keep this simple and encourage running `carthage archive` separately for each framework

mdiep:
I would be :+1: with this, post #892 especially. It makes sense for the `.zip` to contain whatever would get built if you weren't using binaries.

After #892, maybe we could make the arguments to `archive` optional—defaulting to archiving all the frameworks from the current project.

ikesyo:
> defaulting to archiving all the frameworks from the current project

Does this mean [the frameworks built by `--no-skip-current`](https://github.com/Carthage/Carthage/blob/41783e92dd754097b06a1e995afccaf2b9804a55/Source/carthage/Build.swift#L154) or all the frameworks in `Carthage/Build`?

mdiep:
All the frameworks in the current project (`--no-skip-current`), excluding dependencies.

ikesyo:
I got it, thanks!

ikesyo:
@mdiep I've rebased and implemented https://github.com/Carthage/Carthage/pull/1005#issuecomment-165859451. :sparkles:

mdiep:
:sparkles:

I hope to catch up on Carthage PRs over the next few days.

Source/carthage/Archive.swift
```diff
@@ -61,10 +94,11 @@ public struct ArchiveCommand: CommandType {
 			.collect()
 			.flatMap(.Merge) { paths -> SignalProducer<(), CarthageError> in
 				if paths.isEmpty {
-					return SignalProducer(error: CarthageError.InvalidArgument(description: "Could not find any copies of \(options.frameworkName).framework. Make sure you're in the project’s root and that the framework has already been built using 'carthage build --no-skip-current'."))
+					let error = CarthageError.InvalidArgument(description: "Could not find any copies of \(frameworks). Make sure you're in the project’s root and that the frameworks has already been built using 'carthage build --no-skip-current'.")
```
mdiep: This should read _that the frameworks have already_. :relaxed:

Source/carthage/Archive.swift
```diff
@@ -61,10 +94,11 @@ public struct ArchiveCommand: CommandType {
 			.collect()
 			.flatMap(.Merge) { paths -> SignalProducer<(), CarthageError> in
 				if paths.isEmpty {
```
mdiep: This only errors if _none_ of the frameworks are found. We should error if any of the frameworks is missing.

Source/carthage/Archive.swift
```diff
@@ -14,34 +14,67 @@ import ReactiveCocoa
...
+				return ($0 as NSString).stringByAppendingPathExtension("framework")!
+			}
+		} else {
+			let directoryURL = NSURL.fileURLWithPath(options.directoryPath, isDirectory: true)
+			frameworks = locateProjectsInDirectory(directoryURL)
+				.flatMap(.Concat) { project in
+					return schemesInProject(project)
+						.flatMapError { error in
+							if case .NoSharedSchemes = error {
+								return .empty
```
mdiep: We should probably return the no shared schemes error here. Otherwise you get this:

```
$ carthage-dev archive
Could not find any copies of []. Make sure you're in the project’s root and that the frameworks has already been built using 'carthage build --no-skip-current'.
```

Source/carthage/Archive.swift
```diff
@@ -14,34 +14,67 @@ import ReactiveCocoa
...
+							let buildArguments = BuildArguments(project: project, scheme: scheme, configuration: "Release")
+							return BuildSettings.loadWithArguments(buildArguments)
+								.map { $0.wrapperName.value }
+								.ignoreNil()
+						}
+				}
+				.skipRepeats()
+				.collect()
+				.map { $0.sort() }
+				.first()?.value ?? []
```
mdiep: I'd really like to get rid of this `first`. Could we restructure things (and maybe break some things out into separate functions) so that this is a part of the signal chain below? Most of what's down below will probably need to be in a `flatMap`.

Source/carthage/Archive.swift
```diff
@@ -14,34 +14,67 @@ import ReactiveCocoa
...
+								return .empty
+							} else {
+								return .init(error: error)
+							}
+						}
+						.flatMap(.Merge) { scheme -> SignalProducer<String, CarthageError> in
+							let buildArguments = BuildArguments(project: project, scheme: scheme, configuration: "Release")
+							return BuildSettings.loadWithArguments(buildArguments)
+								.map { $0.wrapperName.value }
+								.ignoreNil()
```
mdiep: Can you document what this is doing? It's not obvious. And what's the `ignoreNil` for?

Source/carthage/Archive.swift
mdiep: Also, it'd be nice to join the frameworks with commas so we don't get the `[]`s in the output.

mdiep:
This looks mostly correct, but I left a found a couple things that should be fixed and I think we ought to clean up the implementation a little.

Thanks for doing this! It'll be a nice touch.

Source/carthage/Archive.swift
ikesyo: I replaced it with `flatMap`: 07cb80a

ikesyo:
Thank you for the review, I've addressed them! :zap: ([diff with white space ignored](https://github.com/Carthage/Carthage/pull/1005/files?w=1))

Source/CarthageKit/Xcode.swift
```diff
@@ -268,6 +268,62 @@ public func schemesInProject(project: ProjectLocator) -> SignalProducer<String,
 		.map { (line: String) -> String in line.stringByTrimmingCharactersInSet(NSCharacterSet.whitespaceCharacterSet()) }
 }
 
+/// Finds schemes of projects or workspace, which Carthage should build, found
+/// within the given directory.
+public func buildableSchemesByProjectLocatorInDirectory(directoryURL: NSURL, withConfiguration configuration: String, forPlatforms platforms: Set<Platform> = []) -> SignalProducer<(ProjectLocator, [String]), CarthageError> {
```
mdiep: Could we call this `buildableSchemesInDirectory`?

Source/CarthageKit/Xcode.swift
```diff
@@ -268,6 +268,62 @@ public func schemesInProject(project: ProjectLocator) -> SignalProducer<String,
...
+						return .init(error: error)
+					}
+				}
+				.map { (project, $0) }
+		}
+}
+
+/// Sends pairs of a scheme and a project, the scheme actually resides in
+/// the project.
+public func schemesToBuildOfProjectLocators(projects: [(ProjectLocator, [String])]) -> SignalProducer<[(String, ProjectLocator)], CarthageError> {
```
mdiep: And call this `schemesInProjects`? This doesn't really specify anything about what's build able.

mdiep:
Looks good apart from the function names.

ikesyo:
Thanks for the advices. :sparkles:

mdiep:
:sparkles:

PR #1007, opened 2015-12-20
Clean up `Dependency`

mdiep:


Source/CarthageKit/Cartfile.swift
```diff
@@ -277,18 +277,9 @@ public struct Dependency<V: VersionType>: Equatable {
 	/// The version(s) that are required to satisfy this dependency.
 	public var version: V
 
-	/// The dependencies of this dependency.
-	public let dependencies: Set<ProjectIdentifier>
```
ikesyo: Nice catch! This was not necessary from 40fa8eece5daf74f46b5ce198dd4ae5c028d2e8e.

ikesyo:
:fire:

PR #1009, opened 2015-12-22
[Specs] Replace some Nimble matchers with equivalent operators

ikesyo:
The same as https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2618.


Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -365,3 +365,11 @@ class XcodeSpec: QuickSpec {
 
 	}
 }
+
+// MARK: Helpers
+
+extension ObjCBool: Equatable {}
```
NachoSoto: Whoa this isn't in the Stdlib??

PR #1010, opened 2015-12-22
when collecting scheme names from xcodebuild -list output,  only cons…

sleifer:
…ider lines that start with at least two spaces

I noticed that due to some plugins I have for xcode, there is noise in the output of xcodebuild --list...

simeon@SimBook:AutoLayoutTools(master) $ xcodebuild -list
2015-12-22 08:40:49.581 xcodebuild[86481:12379480] XcodeFScript: app=xcodebuild
2015-12-22 08:40:49:599 xcodebuild[86481:e0b] BBUncrustifyPlugin Version 2.1.3 loaded
Information about project "AutoLayoutTools":
    Targets:
        AutoLayoutTools
        AutoLayoutToolsTests
        AutoLayoutToolsFramework
        AutoLayoutToolsLibrary

    Build Configurations:
        Debug
        Release

    If no build configuration is specified and -scheme is not passed then "Release" is used.

    Schemes:
        AutoLayoutSandbox
        AutoLayoutToolsFramework
        AutoLayoutToolsLibrary
2015-12-22 16:40:50 +0000 Plugin Console Open With Application: [NSApp mainMenu] returned nil

carthage thinks that last line (2015-12-22 16:40:50 +0000 Plugin Console Open With Application: [NSApp mainMenu] returned nil) is a scheme name and tried to build it, xcodebuild then complains and carthage exits

This commit only considers lines that start with two spaces as scheme names

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -254,7 +254,7 @@ public func schemesInProject(project: ProjectLocator) -> SignalProducer<String, 
 		}
 		.skipWhile { line in !line.hasSuffix("Schemes:") }
 		.skip(1)
-		.takeWhile { line in !line.isEmpty }
+		.takeWhile { line in !line.isEmpty && line.hasPrefix("  ") }
 		// xcodebuild has a bug where xcodebuild -list can sometimes hang
 		// indefinitely on projects that don't share any schemes, so
 		// automatically bail out if it looks like that's happening.


Source/CarthageKit/Xcode.swift
```diff
@@ -254,7 +254,7 @@ public func schemesInProject(project: ProjectLocator) -> SignalProducer<String,
 		}
 		.skipWhile { line in !line.hasSuffix("Schemes:") }
 		.skip(1)
-		.takeWhile { line in !line.isEmpty }
+		.takeWhile { line in !line.isEmpty && line.hasPrefix("  ") }
```
NachoSoto: This seems very fragile. Xcode could change the prefix format at any time and this would break :/

sleifer: Agreed, though it seemed less fragile than any ideas I'd thought of to detect the 'noise' directly.

mdiep:
I think the correct answer here is to not use Xcode plugins, which are unsupported. We can't be responsible for their noise—particularly not when we risk breaking things for the majority of users who don't use plugins.

steveriggins:
"I think the correct answer here is to not use Xcode plugins"  This is not very realistic. There are a number of useful plugins, including a plugin manager. If this change caused carthage to stop working with a particular xcode build, it would not be much work to make it compliant. 

"particularly not when we risk breaking things for the majority of users who don't use plugins."  Do you have data on this? In my office, 80% of the developers have one or more plugins installed. Dash is very popular, and while I am not an alcatraz fan, I know a number of people who use it.

mdiep:
Sorry, but Carthage can't be responsible for the handling the noise from plugins.

PR #1011, opened 2015-12-24
Update completion scripts

ikesyo:
/cc @yoichitgy @rpowelll

### DIFF
--- a/Source/Scripts/carthage-bash-completion
+++ b/Source/Scripts/carthage-bash-completion
@@ -28,8 +28,8 @@ _carthage() {
                 ;;
             bootstrap|update)
                 COMPREPLY=($(compgen -W '--configuration --platform --verbose --no-checkout \
-				    --no-build --use-ssh --use-submodules --no-use-binaries --color \
-			        --project-directory' -- ${cur}))
+                    --no-build --use-ssh --use-submodules --no-use-binaries --color \
+                    --project-directory' -- ${cur}))
                 return 0
                 ;;
             build)
--- a/Source/Scripts/carthage-bash-completion
+++ b/Source/Scripts/carthage-bash-completion
@@ -28,17 +28,18 @@ _carthage() {
                 ;;
             bootstrap|update)
                 COMPREPLY=($(compgen -W '--configuration --platform --verbose --no-checkout \
-                	--no-build --use-ssh --use-submodules --no-use-binaries --color' -- ${cur}))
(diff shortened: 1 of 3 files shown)


Source/Scripts/carthage-bash-completion
```diff
@@ -49,7 +50,7 @@ _carthage() {
     fi
 
     case "${prev}" in
-        --output)
+        --output|--project-directory)
```
yoichitgy: It might be better to use `_filedir -d` for `--project-directory` if it accepts only a directory.

```
        --project-directory)
            _filedir -d
            return 0
            ;;
```

Source/Scripts/carthage-zsh-completion
```diff
@@ -48,6 +48,7 @@ _carthage() {
                         '--no-build' \
                         '--color: :(auto always never)' \
                         '--verbose' \
+                        '--project-directory: :_files' \
```
yoichitgy: It might be better to use `_directories` for `--project-directory` if it accepts only a directory.

```
'--project-directory: :_directories' \
```

Source/Scripts/carthage-zsh-completion
```diff
@@ -58,6 +59,7 @@ _carthage() {
                         '--no-skip-current' \
                         '--color: :(auto always never)' \
                         '--verbose' \
+                        '--project-directory: :_files' \
```
yoichitgy: Same as above.

Source/Scripts/carthage-zsh-completion
```diff
@@ -67,6 +69,7 @@ _carthage() {
                         '--use-submodules' \
                         '--no-use-binaries' \
                         '--color: :(auto always never)' \
+                        '--project-directory: :_files' \
```
yoichitgy: Same as above.

yoichitgy:
Thanks @ikesyo for notifying me. I added comments to the bash and zsh scripts in case `--project-directory` needs auto completion of only directories. I already tried as I commented in my local environment, and they worked.

ikesyo:
@yoichitgy Thank you for the review! I've updated this as per your comments. :smiley:

yoichitgy:
@ikesyo I'll investigate the fish script for directory-only completion later if @rpowelll won't notice the mention.😃

yoichitgy:
@ikesyo You've already updated the fish script. I'm sorry for my late response.

PR #1013, opened 2015-12-28
Add Instruction for tvOS

toshi0383:
This is what I'm doing when installing library via carthage to tvOS apps.
I'm not sure tvOS has apple submission bug, but seems like Xcode cannot resolve dependencies correctly unless you set `carthage copy-framework` Run Script Phase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
 1. Run `carthage update`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one.
 1. On your application targets’ “General” settings tab, in the “Embedded Binaries” section, drag and drop each framework you want to use from the [Carthage/Build][] folder on disk.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,7 +50,7 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
 1. Run `carthage update`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one.
 1. On your application targets’ “General” settings tab, in the “Embedded Binaries” section, drag and drop each framework you want to use from the [Carthage/Build][] folder on disk.
 
-##### If you're building for iOS
+##### If you're building for iOS (for tvOS, read `iOS` as `tvOS`)
```
NachoSoto: Right, good point. This also applies for `watchOS`. Can you maybe change the wording to something like "if you're building for iOS, tvOS, or watchOS..."

toshi0383:
Thanks for review. Fixed !

NachoSoto:
Thank you!!

mdiep:
:tada:

PR #1015, opened 2015-12-29
Clarifies the error message to the 'expected submodule commit SHA in ls-tree output'

mz2:
This change clarifies the error message to the dreaded "expected submodule commit SHA in ls-tree output:" error such that you learn what kind of a command gave output that could not be parsed. 

Before this modification there can be a lot of guesswork involved in resolving this error especially if you run `carthage update` on a new project with an existing "legacy" .gitmodules file: I for instance encountered this error with a project with a huge set of existing submodules that I began migrating over to Carthage based dependencies (there were a few weird old submodule dependencies whose brokenness would not be picked up as a build error because they would not result in build errors).

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -370,7 +370,7 @@ public func submoduleSHAForPath(repositoryFileURL: NSURL, _ path: String, revisi
 			if components.count >= 3 {
 				return .Success(String(components[2]))
 			} else {
-				return .Failure(CarthageError.ParseError(description: "expected submodule commit SHA in output of \(task.joinWithSeparator(" ")) but encountered: \(string)"))
+				return .Failure(CarthageError.ParseError(description: "expected submodule commit SHA in output of task (\(task.joinWithSeparator(" "))) but encountered: \(string)"))
 			}
 		}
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Git.swift
```diff
@@ -369,7 +370,7 @@ public func submoduleSHAForPath(repositoryFileURL: NSURL, _ path: String, revisi
 			if components.count >= 3 {
 				return .Success(String(components[2]))
 			} else {
-				return .Failure(CarthageError.ParseError(description: "expected submodule commit SHA in ls-tree output: \(string)"))
+				return .Failure(CarthageError.ParseError(description: "expected submodule commit SHA in output of \(task.joinWithSeparator(" ")) but encountered: \(string)"))
```
NachoSoto: What do you think about wrapping the task in parentheses to make the message even clearer?

NachoSoto:
Great idea :+1:

mz2:
@NachoSoto good one, added parentheses too. Happy? :-)

NachoSoto:
:shipit:

NachoSoto:
Thanks!

PR #1023, opened 2016-01-06
Copy and strip dSYM files in copy-frameworks

### LINKED ISSUE #1021: Carthage dSYM warnings from iTunes Connect.
Builds submitted to iTunes Connect that contain Carthage-build dSYM files result in a warning email stating that there are symbols that "have no corresponding slice in any binary".

```
Dear developer,

We have discovered one or more issues with your recent delivery for "APP". Your delivery was successful, but you may wish to correct the following issues in your next delivery:

Too many symbol files - These symbols have no corresponding slice in any binary [0746DF61-3843-3328-AD2D-8059C8BBCA86.symbols, 0E0FBDE4-272D-37BA-AA5F-E1250228047D.symbols, 29B5A6D8-70DD-334B-912B-2466415AFBBC.symbols, 36A18BD3-F7DA-3777-AF77-5285BF7DDB73.symbols, 437FDB3A-75D0-3834-9A33-53A94438EBDA.symbols, 43EE6A78-3136-372A-97E5-8BC3D6ED640F.symbols, 53F43456-CF1C-394A-81D6-4A9E3C0AC1FB.symbols, 5DDED596-0638-3E16-9B78-02A34B77DD42.symbols, 61024AB5-86D2-3F9D-BD87-BA52CDBDF1CB.symbols, 611FDB23-4831-3159-A175-66166215B105.symbols, 64F6B87F-E911-3D10-AB67-92B7A1B2EC81.symbols, 66187D82-FDF1-3AFF-A70A-9A6D63106BA5.symbols, 68873B77-5924-38DD-A44A-0F0768720293.symbols, 7395DD4E-9FEC-3270-BCEB-A365A7BD83E4.symbols, 7CBE62C1-4709-3B02-B765-8EBDC754826E.symbols, 7DB9D88A-F345-3F8D-BA7E-DCED1D95C61F.symbols, 81EF8FD6-60A1-3A47-BBAE-8C155A76654E.symbols, 8EADDF7C-EA4E-38EA-B954-0A3F09D3AF60.symbols, 8F1ADEF8-A7FB-35CD-9689-D01CC54A9096.symbols, A438ED2A-B91F-3375-9DB8-FFA155E037AC.symbols, A629BEFA-7B3D-398B-BDC9-2F7485E9950D.symbols, A66B3059-0750-3E72-99A0-247945F6BAD1.symbols, A843167F-F738-3BBE-930B-F1204E05909C.symbols, AA82ECA0-401E-347C-9BE5-862D42038773.symbols, B7BBBCF7-9AF8-3810-942E-AAFD172EA1D9.symbols, C17C9FA2-1260-3636-9164-CE7FDCA0BE58.symbols, C5937794-A321-3FC5-ABFB-A99ACA5327B3.symbols, C5AC389F-F574-3DCC-81C3-AAFE2E32016D.symbols, CB21B3EF-DD6A-3658-9DD2-81D45F83CDD4.symbols, D2B8BFD6-4B20-3172-B68D-0959A50FC10E.symbols, D3092228-2205-3360-BC02-A9D5B10AF8A0.symbols, E23B7321-20ED-30FE-A761-FF1E8916763C.symbols, E434FED9-CBBF-32A3-93F2-89B572C326B5.symbols, EE9B984A-BDDC-3C62-954C-7DB032378063.symbols, EEF5CB9B-3389-3F8C-A22B-3B47E5137CB1.symbols, FA4BCA6E-5966-3E44-8AA7-E90F10DB7E67.symbols]

After you’ve corrected the issues, you can use Xcode or Application Loader to upload a new binary to iTunes Connect.

Regards,

The App Store team
```

After some digging (and help from @kballard) I found that every referenced symbol came from the dSYM file of a Carthage-built framework.

```
dSYMs -> dwarfdump -u ReactiveCocoa.framework.dSYM 
UUID: 66187D82-FDF1-3AFF-A70A-9A6D63106BA5 (i386) ReactiveCocoa.framework.dSYM/Contents/Resources/DWARF/ReactiveCocoa
UUID: A629BEFA-7B3D-398B-BDC9-2F7485E9950D (x86_64) ReactiveCocoa.framework.dSYM/Contents/Resources/DWARF/ReactiveCocoa
UUID: 2B798C5A-B24C-316A-80CF-544BA8F5F3C8 (armv7) ReactiveCocoa.framework.dSYM/Contents/Resources/DWARF/ReactiveCocoa
UUID: AECCAFB7-10AB-370F-8FA3-F79F2BE265D1 (arm64) ReactiveCocoa.framework.dSYM/Contents/Resources/DWARF/ReactiveCocoa
```

It is my understanding that `copy-frameworks` strips any simulator slices when `ACTION=install`. This should also be performed on dSYM files.

eliperkins:
iTunes Connect submissions will validate dSYMs and will warn for dSYMs that contains slices for architectures that are not contained in the app submission.

carthage copy-frameworks now copies dSYM files from the Carthage/Build directory for each framework.

Closes #1021


calebd:
Does `copyDebugSymbolsForFramework` need to run for normal builds?

calebd:
The old producers would always strip and would optionally copy bit code symbols. I like the use of `combineLatest` so that all involved producers run simultaneously. I think it would be cool to see all three operations as root level producers (instead of nesting the copy bit code producer inside the copy framework producer) and where the symbols producers would each only run in install mode.

```
if buildActionIsArchiveOrInstall() {
    return combineLatest([ copyFrameworksProducer, copyBitCodeProducer, copyDSYMsProducer ])
}
else {
    return copyFrameworksProducer
}
```

hyperspacemark:
> Does copyDebugSymbolsForFramework need to run for normal builds?

My understanding is that you would want these even in debug builds so that you can step into framework methods.

eliperkins:
@calebd 

> Does `copyDebugSymbolsForFramework` need to run for normal builds?

It doesn't _need_ to, really. But no harm in doing it. Using the old Build Phase instructions would do this as well anyway.

>  I like the use of combineLatest so that all involved producers run simultaneously. I think it would be cool to see all three operations as root level producers

Totally agree, just didn't want to overstep any boundaries of this PR. :)

calebd:
Makes sense! This is :+1:.

kballard:
We absolutely need to copy the dSYMs on all builds, not just archives, because we need to eliminate the build instructions telling people to have a Copy Files build phase to copy the dSYMs. Otherwise anyone who has such a build phase that runs after `carthage copy-frameworks` will end up just replacing the stripped dSYMs with unstripped ones. It's a lot simpler just to have `carthage copy-frameworks` copy the dSYMs anyway.

Source/CarthageKit/Xcode.swift
```diff
@@ -1321,6 +1330,13 @@ public func stripFramework(frameworkURL: NSURL, keepingArchitectures: [String],
 		.concat(sign)
 }
 
+/// Strips a dSYM from unexpected architectures.
+public func stripDSYM(dSYMURL: NSURL, keepingArchitectures: [String]) -> SignalProducer<(), CarthageError> {
+	return architecturesInPackage(dSYMURL)
+		.filter { !keepingArchitectures.contains($0) }
+		.flatMap(.Concat) { stripArchitecture(dSYMURL, $0) }
```
mdiep: This this repeats the code at the top of `stripFramework`, can we give it a more generic name and call it from there as well?

Source/CarthageKit/Xcode.swift
```diff
@@ -1390,64 +1406,69 @@ private func stripArchitecture(frameworkURL: NSURL, _ architecture: String) -> S
...
-						let components = architectures?
-							.componentsSeparatedByString(" ")
-							.filter { !$0.isEmpty }
-
-						if let components = components {
-							return SignalProducer(values: components)
-						}
-					}
+/// Returns a signal of all architectures present in a given binary.
+public func architecturesInBinary(binaryURL: NSURL) -> SignalProducer<String, CarthageError> {
```
mdiep: Why did you split this out from `architecturesInPackage`? It looks like it's still only called there.

Source/CarthageKit/Xcode.swift
```diff
@@ -1542,15 +1563,25 @@ private func UUIDsFromDwarfdump(URL: NSURL) -> SignalProducer<Set<NSUUID>, Carth
...
+	case .Framework?:
+		if let binaryName = bundle?.objectForInfoDictionaryKey("CFBundleExecutable") as? String {
+			return .Success(packageURL.URLByAppendingPathComponent(binaryName))
+		}
+	case .dSYM?:
+		if let binaryName = packageURL.URLByDeletingPathExtension?.URLByDeletingPathExtension?.lastPathComponent {
+			let binaryURL = packageURL.URLByAppendingPathComponent("Contents/Resources/DWARF/\(binaryName)")
+			return .Success(binaryURL)
+		}
+	default: break
```
mdiep: Can you replace this case with a `.None` that returns the failure below?

mdiep:
Looks good to me, but I left a few notes that I think will clean up the implementation.

@Carthage/carthage It'd be nice if someone else could verify the behavior of these changes against the app store submission process.

Source/CarthageKit/Xcode.swift
eliperkins: My motivation around not returning from this case was that since both `case .Framework?` and `case .dSYM?` only return based on some conditional inside of this, they could both return a failure as well. To complete the control flow here, we'd need to either explicitly `return .Failure` inside of each case:

```swift
	switch packageType {
	case .Framework?:
		if let binaryName = bundle?.objectForInfoDictionaryKey("CFBundleExecutable") as? String {
			return .Success(packageURL.URLByAppendingPathComponent(binaryName))
		}

		return .Failure(.ReadFailed(packageURL, nil))
	case .dSYM?:
		if let binaryName = packageURL.URLByDeletingPathExtension?.URLByDeletingPathExtension?.lastPathComponent {
			let binaryURL = packageURL.URLByAppendingPathComponent("Contents/Resources/DWARF/\(binaryName)")
			return .Success(binaryURL)
		}

		return .Failure(.ReadFailed(packageURL, nil))
	case .None: return .Failure(.ReadFailed(packageURL, nil))
	}
```

or use `fallthrough` to get to the `default` case:

```swift
	switch packageType {
	case .Framework?:
		if let binaryName = bundle?.objectForInfoDictionaryKey("CFBundleExecutable") as? String {
			return .Success(packageURL.URLByAppendingPathComponent(binaryName))
		}

		fallthrough
	case .dSYM?:
		if let binaryName = packageURL.URLByDeletingPathExtension?.URLByDeletingPathExtension?.lastPathComponent {
			let binaryURL = packageURL.URLByAppendingPathComponent("Contents/Resources/DWARF/\(binaryName)")
			return .Success(binaryURL)
		}

		fallthrough
	case .None: return .Failure(.ReadFailed(packageURL, nil))
	}
``` 

or break without a return as we have here.

I'm cool with changing this to return in the `.None` here, but it won't let us delete the final return at the end of the function too unless we do a little bit of other tweaking.

eliperkins:
:mailbox:

eliperkins:
Looks like CI choked on checking out ReactiveTask?
```bash
fatal: unable to access 'https://github.com/Carthage/ReactiveTask.git/': Failed to connect to github.com port 443: Operation timed out
Clone of 'https://github.com/Carthage/ReactiveTask.git' into submodule path 'Carthage/Checkouts/ReactiveTask' failed
```

Source/CarthageKit/Xcode.swift
mdiep: Oh, right. *facepalm*

eliperkins: Haha, no worries. I'm still not entirely pleased with how it reads either, so I'm open to feedback on better ways to set this flow up!

mdiep:
Looks good. Thanks!

Bogidon:
I just spent a decent amount of time before realizing that the changes 5ef07e3 added to the readme are geared for Carthage 0.14. I had Carthage 0.11 installed, and I'm sure other users aren't on the latest version.

Can I PR a section in the readme with instructions for adding the dSYMs if Carthage < 0.14?

calebd:
Shouldn't the readme reflect the current version of the repository? You can go back to a < 0.14 tag and see the recommended use case there.

Bogidon:
@calebd I can certainly see your argument. However, I personally don't check the version of Carthage very often (I am not involved in its development) so I was a little boggled when my builds were failing to upload. Also, the GitHub tagging system is rather hidden. At least it's not something I look at before reading the docs. Maybe I should.

An alternate solution might be to display the version number somewhere more prominent in the readme. Though it's true that the release badge is there ( GitHub release ).

mdiep:
If anything, I'd say we should add a note saying which version added that feature.

hyperspacemark:
> If anything, I'd say we should add a note saying which version added that feature.

That seems untenable. Do we annotate everything in the readme that has changed relatively recently? How far back do we go? Expecting developers to stay up to date with tools critical to their workflow by checking releases and following a change log doesn't seem unreasonable to me.

calebd:
At most this is a good case for a `CHANGELOG.md`.

Bogidon:
How about warning developers that new releases are available ala Cocoapods?
a95ecfe6-8011-445b-99a2-a508be73eb7e

mdiep:
> How about warning developers that new releases are available ala Cocoapods?

I like that idea. Care to open an issue?

Bogidon:
@mdiep will do.

PR #1024, opened 2016-01-07
Refactor using Result.init(attempt:)

ikesyo:


mdiep:
Nice. :+1:

NachoSoto:
Love it, and it infers the error type. If only the compiler would check the thrown error type too :P

PR #1028, opened 2016-01-07
Always clean before building

kballard:
Some types of project changes require a clean to take effect. Typically
this means any change that causes the built product to no longer include
a file/directory it used to include. When building on top of an existing
built product, the old product isn't deleted without cleaning, which
means the resulting built product will still contain the obsolete
file/directory.

If Carthage ever switches to passing a custom derived data path then it
can stop explicitly cleaning (because deleting the derived data
accomplishes the same thing).

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -224,8 +224,14 @@ public func locateProjectsInDirectory(directoryURL: NSURL) -> SignalProducer<Pro
 
 /// Creates a task description for executing `xcodebuild` with the given
 /// arguments.
+public func xcodebuildTask(tasks: [String], _ buildArguments: BuildArguments) -> Task {
+	return Task("/usr/bin/xcrun", arguments: buildArguments.arguments + tasks)
+}
+
+/// Creates a task description for executing `xcodebuild` with the given
+/// arguments.
 public func xcodebuildTask(task: String, _ buildArguments: BuildArguments) -> Task {
-	return Task("/usr/bin/xcrun", arguments: buildArguments.arguments + [ task ])
+	return xcodebuildTask([task], buildArguments)
 }
 
 /// Sends each scheme found in the given project.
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -228,6 +228,12 @@ public func xcodebuildTask(task: String, _ buildArguments: BuildArguments) -> Ta
 	return Task("/usr/bin/xcrun", arguments: buildArguments.arguments + [ task ])
 }
 
+/// Creates a task description for executing `xcodebuild` with the given
+/// arguments.
+public func xcodebuildTask(task: String, _ task2: String, _ buildArguments: BuildArguments) -> Task {
+	return Task("/usr/bin/xcrun", arguments: buildArguments.arguments + [ task, task2 ])
```
NachoSoto: This implementation is repeated from the other `xcodebuildTask` overload. What do you think about making the canonical implementation take an array of `tasks` instead? Then we can leave the overloads for simplicity.

NachoSoto:
Ooooh I actually thought it already did this. If we were using `archive` it wouldn't be necessary since it's implied, so I'm 100% :+1:.

Just a small note on the implementation!

Source/CarthageKit/Xcode.swift
kballard: I overloaded it because it's public API and I wasn't sure if there were any issues with changing public API like that. But I do prefer the array of tasks option, with an overload to take a single task for convenience.

I actually wish we could say `func xcodebuildTask(tasks: String..., _ buildArguments: BuildArguments) -> Task` but you can't use variadic args before an unnamed parameter :(

kballard:
@NachoSoto PR updated.

Source/CarthageKit/Xcode.swift
```diff
@@ -224,8 +224,14 @@ public func locateProjectsInDirectory(directoryURL: NSURL) -> SignalProducer<Pro
...
 /// arguments.
+public func xcodebuildTask(tasks: [String], _ buildArguments: BuildArguments) -> Task {
+	return Task("/usr/bin/xcrun", arguments: buildArguments.arguments + tasks)
+}
+
+/// Creates a task description for executing `xcodebuild` with the given
+/// arguments.
 public func xcodebuildTask(task: String, _ buildArguments: BuildArguments) -> Task {
-	return Task("/usr/bin/xcrun", arguments: buildArguments.arguments + [ task ])
+	return xcodebuildTask([task], buildArguments)
```
NachoSoto: Makes sense.

NachoSoto:
Awesome :shipit:

PR #1032, opened 2016-01-08
Fix crash when running any command that relies on Cartfile.resolved without a cache

BobElDevil:
Steps to Repro

1) Create a new git repo, with a new Cartfile 
2) Run `carthage update` to create Cartfile.resolved
3) Delete ~/Library/Caches/org.carthage.CarthageKit
4) Run `carthage bootstrap`

```
2016-01-08 10:47:26.285 carthage[90573:2631447] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'working directory doesn't exist.'
```

I believe this is fallout from #974, as I tracked it down to the fact that it's now trying to use `cartfileForDependency` on a git repo in the cache that doesn't exist yet (because nothing has been checked out). 

These changes make sure we clone or fetch the dependency before actually trying to read from it. 

The other thought I had would be to put this further up in the stream, and clone/fetch all dependencies when we first get the Cartfile.resolved listing.


mdiep:
Good catch! And thanks for the pull request!

Unfortunately, this results in fetching projects multiple times:

```
$ cat Cartfile
$ carthage update --no-build
*** Fetching ReactiveCocoa
*** Fetching ReactiveCocoa
*** Fetching Result
*** Fetching Result
*** Fetching Box
*** Fetching Box
*** Fetching ReactiveCocoa
*** Fetching Result
*** Fetching Box
*** Fetching Result
*** Fetching Box
*** Fetching Box
*** Checking out Box at "1.2.2"
*** Checking out Result at "0.4.5"
*** Checking out ReactiveCocoa at "v3.0.0"
```

I think `Project` should track which repositories it has fetched, to prevent refetching them.

But what makes this tricky is that each new `Resolver` should probably be allowed to re-fetch. Since we've designed `CarthageKit` to be a reusable component, it could be used in a long-lived process. As part of that, calls to something approximating `carthage update` could happen days apart. It would be nice if we weren't caching state beyond each call to the public `Project` methods.

Is that something you'd be interested in tackling here? If not, I can merge this and open a new issue to track that.

BobElDevil:
Yeah, I can take a look. I figured it was too easy to just wrap that call in a fetch request =) 

Since we don't want an in memory cache, I noticed that we already have a similar 'fetch if needed' check in checkoutOrCloneProject, where it checks for the commit existing first. I'm thinking if we pull that logic down a level into the `cloneOrFetchDependency` signal, and provide an optional commitish that it can check for if there, then we can route access to dependencies through this optional request.

mdiep:
That sounds like a great idea. :+1:

BobElDevil:
I think this is all set now. Starting without a cache, we get: 
```
$ cat Cartfile                   
$ carthage update --no-build     
*** Cloning ReactiveCocoa
*** Cloning Result
*** Cloning Box
*** Checking out Box at "1.2.2"
*** Checking out Result at "0.4.5"
*** Checking out ReactiveCocoa at "v3.0.0"
$ carthage update --no-build     
*** Fetching ReactiveCocoa
*** Fetching Result
*** Fetching Box
*** Checking out Box at "1.2.2"
*** Checking out Result at "0.4.5"
*** Checking out ReactiveCocoa at "v3.0.0"
```
Then, bootstrapping without a cache:
```
$ carthage bootstrap --no-build      
*** Cloning ReactiveCocoa
*** Cloning Result
*** Cloning Box
*** Fetching Result
*** Fetching Box
*** Checking out Box at "1.2.2"
*** Checking out Result at "0.4.5"
*** Checking out ReactiveCocoa at "v3.0.0"
$ carthage bootstrap --no-build  
*** Fetching Result
*** Fetching Box
*** Checking out Box at "1.2.2"
*** Checking out Result at "0.4.5"
*** Checking out ReactiveCocoa at "v3.0.0"
```

One thing I noticed that is odd is during bootstrap it was still fetching the sub-dependencies of the top level dependencies. Putting a breakpoint in it looks like it was calling the fetch method without a commit, because it was running 'versionsForDependency' as part of the Resolver logic. 

I think this is something additional related to re-calculating the dependency graph every time. The resolver should probably have a 'versions already known' mode. Because it just needs to figure out the order to build things, but the actual commits to use are specified in the Cartfile.resolved and so should only need to fetch commits that don't exist from that list.

Source/CarthageKit/Project.swift
```diff
@@ -295,6 +299,16 @@ public final class Project {
...
 	}
+	
+	/// Loads the Cartfile for the given dependency, at the given version.
+	private func cartfileForDependency(dependency: Dependency<PinnedVersion>) -> SignalProducer<Cartfile, CarthageError> {
+		let revision = dependency.version.commitish
+		return self.cloneOrFetchDependency(dependency.project, commitish: revision).flatMap(.Concat) { repositoryURL in
+			return contentsOfFileInRepository(repositoryURL, CarthageProjectCartfilePath, revision: revision)
+				.flatMapError { _ in .empty }
+				.attemptMap(Cartfile.fromString)
+		}
```
mdiep: We try to keep signal chains as flat as possible to increase readability.

```swift
return self.cloneOrFetchDependency(dependency.project, commitish: revision)
	.flatMap(.Concat) { repositoryURL in
		return contentsOfFileInRepository(repositoryURL, CarthageProjectCartfilePath, revision: revision)
	}
	.flatMapError { _ in .empty }
	.attemptMap(Cartfile.fromString)
```

Would you mind reformatting this accordingly?

Source/CarthageKit/Project.swift
```diff
@@ -866,54 +864,56 @@ private func repositoryURLForProject(project: ProjectIdentifier, preferHTTPS: Bo
...
-			}
+		do {
+			try fileManager.createDirectoryAtURL(CarthageDependencyRepositoriesURL, withIntermediateDirectories: true, attributes: nil)
+		} catch let error as NSError {
+			return .Failure(.WriteFailed(CarthageDependencyRepositoriesURL, error))
+		}
 
-			return .Success(repositoryURLForProject(project, preferHTTPS: preferHTTPS))
+		return .Success(repositoryURLForProject(project, preferHTTPS: preferHTTPS))
 		}
```
mdiep: The code inside this block should be indented another level so that it's indented past the closing `}`.

Source/CarthageKit/Project.swift
```diff
@@ -483,12 +497,11 @@ public final class Project {
 	/// Checks out the given project into its intended working directory,
 	/// cloning it first if need be.
 	private func checkoutOrCloneProject(project: ProjectIdentifier, atRevision revision: String, submodulesByPath: [String: Submodule]) -> SignalProducer<(), CarthageError> {
-		let repositoryURL = repositoryFileURLForProject(project)
-		let workingDirectoryURL = directoryURL.URLByAppendingPathComponent(project.relativePath, isDirectory: true)
-
-		let checkoutSignal = SignalProducer.attempt { () -> Result<Submodule?, CarthageError> in
+		return cloneOrFetchDependency(project, commitish: revision).flatMap(.Merge) { repositoryURL -> SignalProducer<(), CarthageError> in
```
mdiep: This `.flatMap` should be on the next line to match our coding style.

Source/CarthageKit/Project.swift
```diff
@@ -483,12 +497,11 @@ public final class Project {
...
+			let workingDirectoryURL = self.directoryURL.URLByAppendingPathComponent(project.relativePath, isDirectory: true)
+			let checkoutSignal = SignalProducer.attempt { () -> Result<Submodule?, CarthageError> in
```
mdiep: Rather than assigning to `checkoutSignal`, this signal producer can just be returned.

Source/CarthageKit/Project.swift
```diff
@@ -497,31 +510,24 @@ public final class Project {
...
-			})
-
-		return commitExistsInRepository(repositoryURL, revision: revision)
-			.promoteErrors(CarthageError.self)
-			.flatMap(.Merge) { exists -> SignalProducer<NSURL, CarthageError> in
-				if exists {
-					return .empty
-				} else {
-					return self.cloneOrFetchDependency(project)
+				.flatMap(.Merge) { submodule -> SignalProducer<(), CarthageError> in
```
mdiep: It looks like this `flatMap` and the following `on` can be moved outside of this `flatMap`.

Then this `flatMap` could be replaced with a simple `map` without using `SignalProducer.attempt` and `Result`.

mdiep:
Thanks for tackling this! I really like this approach.

I left a few comments of things to fix up—mostly related to our coding style.

BobElDevil:
Thanks for the code review. I updated the things you mentioned.

BobElDevil:
Just checking in to see if there's anything else that needs to be done for this?

Source/CarthageKit/Project.swift
```diff
@@ -866,54 +858,56 @@ private func repositoryURLForProject(project: ProjectIdentifier, preferHTTPS: Bo
...
+				} else {
+					// Either the directory didn't exist or it did but wasn't a git repository
+					// (Could happen if the process is killed during a previous directory creation)
+					// So we remove it, then clone
+					_ = try? fileManager.removeItemAtURL(repositoryURL)
+					return SignalProducer(value: (.Cloning(project), repositoryURL))
+						.concat(cloneRepository(remoteURL, repositoryURL).then(.empty))
+				}
 		}
+	}
```
mdiep: The indentation of this signal chain is off—it should be indented one more level.

Source/CarthageKit/Project.swift
```diff
@@ -866,54 +858,56 @@ private func repositoryURLForProject(project: ProjectIdentifier, preferHTTPS: Bo
...
+						return SignalProducer(value: (.Fetching(project), repositoryURL))
+							.concat(fetchRepository(repositoryURL, remoteURL: remoteURL, refspec: "+refs/heads/*:refs/heads/*").then(.empty))
+					}
+					// If we've already cloned the repo, check for the revision, possibly skipping an unnecessary fetch
+					if let commitish = commitish {
+						return commitExistsInRepository(repositoryURL, revision: commitish)
+							.promoteErrors(CarthageError.self)
+							.flatMap(.Merge) { exists -> SignalProducer<(ProjectEvent, NSURL), CarthageError> in
+								if exists {
+									return SignalProducer(value: (.SkippedFetching(project), repositoryURL))
```
mdiep: Rather than adding a `SkippedFetching` event and then ignoring it, could we change the signature of this method to return `SignalProducer<(ProjectEvent?, NSURL), CarthageError>` and then return `nil` for the project event here? That seems like it'd make a bit more sense.

mdiep:
Sorry for the delayed review! I've had a busy couple weeks. I left 2 more notes that should be easy to fix up. After those are fixed, this will be ready to go.

Source/CarthageKit/Project.swift
BobElDevil: Ah yeah, that makes more sense. Will do.

Source/CarthageKit/Project.swift
BobElDevil: I'm sorry not seeing where it's off. The flatMap is coming right after the top level attempt brace, and so it's lining up with that

mdiep: Instead of this:

```swift
	return SignalProducer.attempt { () -> Result<GitURL, CarthageError> in
		do {
			try fileManager.createDirectoryAtURL(CarthageDependencyRepositoriesURL, withIntermediateDirectories: true, attributes: nil)
			return .Failure(.WriteFailed(CarthageDependencyRepositoriesURL, error))
		}

		return .Success(repositoryURLForProject(project, preferHTTPS: preferHTTPS))
	}
	.flatMap(.Merge) { remoteURL -> SignalProducer<(ProjectEvent, NSURL), CarthageError> 
		...
	}
```

We would typically write this:

```swift
	return SignalProducer.attempt { () -> Result<GitURL, CarthageError> in
			do {
				try fileManager.createDirectoryAtURL(CarthageDependencyRepositoriesURL, withIntermediateDirectories: true, attributes: nil)
				return .Failure(.WriteFailed(CarthageDependencyRepositoriesURL, error))
			}

			return .Success(repositoryURLForProject(project, preferHTTPS: preferHTTPS))
		}
		.flatMap(.Merge) { remoteURL -> SignalProducer<(ProjectEvent, NSURL), CarthageError> 
			...
		}
```

BobElDevil: Oh I see, got it.

BobElDevil:
Alright, just pushed in those two fixes. Thanks for the reviews.

mdiep:
Looks great! Thank you!! :clap: :sparkles:

PR #1035, opened 2016-01-11
Fix picking up submodules in locateProjectsInDirectory()

### LINKED ISSUE #1029: Carthage incorrectly checking for parent repo submodules in checkouts
After installing the current master (1f9d588b41bee6dcf58ec0ee9e1908a50a6eea32), `carthage bootstrap` fails in my project with the error:

```
Parse error: expected submodule commit SHA in output of task (ls-tree -z HEAD PostmatesKit/PMKCore) but encountered: 
```

(`PostmatesKit/PMKCore` is a submodule in the current repo)

After hacking up Carthage to print some more info, it turned out it was running this `git ls-tree` command from within the `Carthage/Checkouts/CocoaLumberjack` folder.

I have no idea what it's trying to do with submodules, but it certainly shouldn't be looking for them from within a Checkouts folder, because those things aren't git repos, they're just working trees, so any git commands executed from within that folder end up using the info from the parent project.

ikesyo:
Correctly ignore submodules inside `--no-use-submodules` checkouts.

Should/might fix #1029.

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -375,21 +375,38 @@ public func submoduleSHAForPath(repositoryFileURL: NSURL, _ path: String, revisi
 		}
 }
 
-/// Returns each submodule found in the given repository revision, or an empty
-/// signal if none exist.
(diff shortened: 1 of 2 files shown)


kballard:
Just tested your branch against my project and it seems to run correctly.

mdiep:
Great job. :clap:

PR #1040, opened 2016-01-13
Fix parsing scp syntax in GitURL

### LINKED ISSUE #1012: Carthage fails to clone repos when git URLs have SCP-syntax without user name.
For example:

```Cartfile
# SCP-syntax WITHOUT username.
# This will fail with error: "The following dependencies are duplicates"
git "github.com:antitypical/Result.git"
git "github.com:antitypical/Assertions.git"



# SCP-syntax with username. This will work.
#git "git@github.com:antitypical/Result.git"
#git "git@github.com:antitypical/Assertions.git"

# HTTPS-based. This will work too.
#git "https://github.com/antitypical/Result.git"
#git "https://github.com/antitypical/Assertions.git"
```

It seems `GitURL.normalizedURLString` always return empty string for above case
(`NSURL(string: URLString)` generally parses SCP-syntax well).
I think such URLs should have either current logged-in username or "git" by default.

ikesyo:
Fixes #1012. This does not implement the request:

> I think such URLs should have either current logged-in username or "git" by default.

but you can clone or fetch the repo if you have the following entry in your `~/.ssh/config`:

```
Host github.com
  User git
```

/cc @inamiy


mdiep:
Could we add a test for this?

ikesyo:
Yes, I'm working on it.

ikesyo:
I tested this through `func ==(lhs: GitURL, rhs: GitURL) -> Bool` since `normalizedURLString` is `private`.

mdiep:
:sparkles:

ikesyo:
Thank you! :smiley:

inamiy:
Thanks! :clap:

PR #1051, opened 2016-01-18
Prune git tags on fetch

### LINKED ISSUE #1027: Carthage Continues to Build at a Tag Even After the Tag has been Deleted
Carthage got stuck building a dependency at a tag, even after I deleted that tag in the dependency's project. 

Steps to reproduce:
1. Add tag a to a dependent project
2. Add the dependency to a cartfile and specify no version or branch
3. Run carthage update
4. Build succeeds
5. Delete the tag in the dependent project
6. Run carthage update
7. Build succeeds

After running rm -rf ~/Library/Caches/org.carthage.CarthageKit/, the build fails as expected with "No tagged versions found for github".

Let me know if I can help.

ikesyo:
`--prune` option of `man git-fetch` says:

> Tags are not subject to pruning if they are fetched only because of the default tag auto-following or due to a --tags option. However, if tags are fetched due to an explicit refspec (either on the command line or in the remote configuration, for example if the remote was cloned with the --mirror option), then they are also subject to pruning.

Fixes #1027.


mdiep:
Wow. Nice find.

NachoSoto:
👏🏻

PR #1055, opened 2016-01-19
Increase xcodebuild list timeout

tallada:
On my old and very slow Mac 'carthage update' fails with a Cartfile that has RxSwift among its dependencies because 'xcodebuild -list' times out, increasing the interval to 60 enables it to continue building.

The error message 
```
Failed to discover shared schemes in project Rx.xcworkspace—either the project does not have any shared schemes, or xcodebuild never returned
```
 was also a bit confusing to me, I think the proposed message tweak saying that 'xcodebuild timed out' is more clear about what has happened.


mdiep:
This seems fine. :+1: Thanks for the pull request!

Issue #992, opened 2015-12-14
CodeSign error: entitlements are required for product type [...] Your Xcode installation may be damaged.


ratkins:
````
** BUILD FAILED **


The following build commands failed:
        Check dependencies
(1 failure)
CodeSign error: entitlements are required for product type 'Framework' in SDK 'Simulator - iOS 9.2'. Your Xcode installation may be damaged.
A shell task failed with exit code 65:
** BUILD FAILED **


The following build commands failed:
        Check dependencies
(1 failure)

*** xcodebuild output can be found in /var/folders/9w/h24bj5js3sdf64gpv33_htjm000087/T/carthage-xcodebuild.XtCxzJ.log
*** Building scheme "GCDWebServers (iOS)" in GCDWebServer.xcodeproj

````

We're getting the above error from our `carthage build` [Run Phase script](https://github.com/Carthage/Carthage#known-issues) when it runs on our build server. Anyone seen it or have any idea what could be causing it? I don't think our Xcode installation is damaged, it's a relatively fresh install.

OS X 10.10.5
OS X Server 5.0.15
Xcode 7.2

mdiep:
What version of `carthage` is on your build server?

ratkins:
0.9.4

mdiep:
I would try to update to the latest release (0.11) because it disabled code-signing for frameworks built with Carthage. (Frameworks are code-signed when copied anyways.) That will probably fix up your error.

ratkins:
I've just updated to 0.11.0 and get the same result.

mdiep:
Is there anything interesting in the xcodebuild output?

larsacus:
While trying to work around #924 I ran into what looks like the same issue as this. I just upgraded from 0.10 (seeing the issue) to 0.11 (still persists). I appear to be thwarted at every turn trying to get `po` to work again. :frowning: 

The thing that looks strange in the output (below) right before it fails it says that `CODE_SIGN_IDENTITY` is provided from the command line and is empty. Whereas if you look at the exported list of build variables, I can clearly see `export CODE_SIGN_IDENTITY="iPhone Developer"`.

### Build Script
<img width="742" alt="screen shot 2015-12-15 at 18 35 20" src="https://cloud.githubusercontent.com/assets/590216/11828478/a9dd5714-a35a-11e5-87cf-e8d20af83b89.png">

### Result
[...]

PR #1057, opened 2016-01-19
Adds UseSanitizedBuildSystemEnvironment=YES

flovilmart:
And allows override with Process info

proposed fix for #992

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -173,11 +173,11 @@ public struct BuildArguments {
 		// Frameworks get signed in the copy-frameworks action
 		args += [ "CODE_SIGNING_REQUIRED=NO", "CODE_SIGN_IDENTITY=" ]
 		
-		// If we're in xcode phase, sanitize
+		// Allow to force unsanitized environment
 		let environment = NSProcessInfo.processInfo().environment
-		if let xcodeVersionActual = environment["XCODE_VERSION_ACTUAL"] {
-			args += [ "-UseSanitizedBuildSystemEnvironment=YES" ]
-		}
+		let useSanitizedEnvironment = environment["XCODEBUILD_USE_SANITIZED_ENVIRONMENT"] ?? "YES"
+		args += [ "-UseSanitizedBuildSystemEnvironment=\(useSanitizedEnvironment)" ]
+
 		return args
 	}
 }
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -172,7 +172,12 @@ public struct BuildArguments {
 		// Disable code signing requirement for all builds
 		// Frameworks get signed in the copy-frameworks action
 		args += [ "CODE_SIGNING_REQUIRED=NO", "CODE_SIGN_IDENTITY=" ]
-
+		
+		// If we're in xcode phase, sanitize
+		let environment = NSProcessInfo.processInfo().environment
+		if let xcodeVersionActual = environment["XCODE_VERSION_ACTUAL"] as? String {
+			args += [ "-UseSanitizedBuildSystemEnvironment=YES" ]
+		}
 		return args
 	}
 }
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -175,7 +175,7 @@ public struct BuildArguments {
 		
 		// If we're in xcode phase, sanitize
 		let environment = NSProcessInfo.processInfo().environment
-		if let xcodeVersionActual = environment["XCODE_VERSION_ACTUAL"] as? String {
+		if let xcodeVersionActual = environment["XCODE_VERSION_ACTUAL"] {
 			args += [ "-UseSanitizedBuildSystemEnvironment=YES" ]
 		}
 		return args


Source/CarthageKit/Xcode.swift
```diff
@@ -172,7 +172,12 @@ public struct BuildArguments {
 		// Disable code signing requirement for all builds
 		// Frameworks get signed in the copy-frameworks action
 		args += [ "CODE_SIGNING_REQUIRED=NO", "CODE_SIGN_IDENTITY=" ]
-
+		
+		// If we're in xcode phase, sanitize
+		let environment = NSProcessInfo.processInfo().environment
+		if let xcodeVersionActual = environment["XCODE_VERSION_ACTUAL"] as? String {
```
flovilmart: I used an environment file that is pretty Xcode specific, should not be set in any other environment. Suggestions welcome

ikesyo:
Compiler complains:

```
/Users/travis/build/Carthage/Carthage/Source/CarthageKit/Xcode.swift:178:67: error: downcast from 'String?' to 'String' only unwraps optionals; did you mean to use '!'?
                if let xcodeVersionActual = environment["XCODE_VERSION_ACTUAL"] as? String {
```

Source/CarthageKit/Xcode.swift
mdiep: As I commented in https://github.com/Carthage/Carthage/issues/992#issuecomment-172999671, I think we should set this argument all the time unless we've encountered an actual problem from setting it.

flovilmart: I from time to times use external arguments. Also, if we compare to other build systems/dependency manager (for ex npm), they all inherit from the current environment. I have mixed feelings about that for that reason.

mdiep: Would you mind sharing what things you use external arguments for?

flovilmart: Mostly with XC_CONFIG_FILE to override bitcode settings in order to have all the frameworks build with the same settings regardless what the developer wrote :)

Source/CarthageKit/Xcode.swift
```diff
@@ -172,6 +172,10 @@ public struct BuildArguments {
 		// Disable code signing requirement for all builds
 		// Frameworks get signed in the copy-frameworks action
 		args += [ "CODE_SIGNING_REQUIRED=NO", "CODE_SIGN_IDENTITY=" ]
+		
+		// Allow to force unsanitized environment
+		let useSanitizedEnvironment = environment["XCODEBUILD_USE_SANITIZED_ENVIRONMENT"] ?? "YES"
```
ikesyo: Another complaint:

```
/Users/travis/build/Carthage/Carthage/Source/CarthageKit/Xcode.swift:177:33: error: use of unresolved identifier 'environment'
                let useSanitizedEnvironment = environment["XCODEBUILD_USE_SANITIZED_ENVIRONMENT"] ?? "YES"
```

Please build it on your local machine before pushing.

flovilmart: Sorry about that...

hyperspacemark:
@mdiep Is there anything holding this up? Looks like #992 is the last thing preventing 0.12 from being released.

mdiep:
This needs to be reviewed and tested. I haven't had time to do so yet.

kballard:
LGTM, though it might be a bit odd that saying `env XCODEBUILD_USE_SANITIZED_ENVIRONMENT= carthage build` will suppress sanitizing (i.e. it would be nice to treat an empty string the same as having the env var being unset).

I haven't tested this in Carthage directly, but I did test the `-UseSanitizedBuildSystemEnvironment=YES` flag and it does behave as documented.

flovilmart:
@mdiep you want support for XCODEBUILD_USE_SANITIZED_ENVIRONMENT= -> "NO" ?

mdiep:
We should either always specify `-UseSanitizedBuildSystemEnvironment=YES` (which I'd lean towards) or add a command line flag to Carthage that defaults to using it. I don't think an environment variable is the right way to go for this.

flovilmart:
I'd need a way to bypass the sanitized environment. I don't mind having and extra flag. That's actually pretty much just for myself XCODEBUILD_USE_SANITIZED_ENVIRONMENT=NO as no one seems to use custom configuration on the top of what carthage provides.

mdiep:
Would you mind switching it to a boolean command line flag? `--(no-)sanitized-builds` or something?

flovilmart:
@mdiep I definitely could, I'm not familiar with the command line parsing facility.

mdiep:
It should be pretty easy. `BuildCommand.Options` lets you define a property. Then you'll need to add a flag to the build commands to pass in whether or not builds should be sanitized.

kballard:
Seems to me that adding a flag to `BuildArguments` that has to be set is going to be tricky, because you have to propagate the flag all the way through the build process, and hope that you don't miss anything. You'd also want to sanitize the calls to `xcodebuild -list` and `xcodebuild -showBuildSettings` too for good measure (just in case the environment affects that), but the caller of `xcodebuild -list` at least doesn't have a good way of passing freeform arguments through, so e.g. `func schemesInProject(project: ProjectLocator)` would have to gain a new argument just to specify build sanitizing. And this all seems rather pointless, since what you actually want is to explicitly sanitize *all* calls to `xcodebuild` (or none of them) instead of controlling it on a call-by-call basis.

I'd suggest that you should use a static stored property on `BuildArguments` that can be set once after processing the command-line arguments, but AFAIK Carthage currently avoids any kind of global state, and if there's already multithreading going on before the command-line args have been parsed then it might not be a good idea to have global shared state (unless you add some form of synchronization around the property, but I don't know if Carthage has an easy way to do that at the moment).

flovilmart:
@mdiep  the architecture for the build and archive command begs for a refactoring as they both share most of the same options, I'll have to duplicate the command line args options code, that seems pretty inefficient at the moment.

kballard:
Given that this is blocking 0.12, which fixes an annoying iTunesConnect warning, I'd rather we just go ahead with the environment variable solution, and then consider doing a command-line flag later. Especially because very few people will want to not sanitize their build environment.

mdiep:
I'm just going to take this out of the 0.12 milestone for now.

flovilmart:
Agreed with @kballard, we should also add a common Options for the build and update before adding that flag to prevent even more code duplications.

Source/CarthageKit/Xcode.swift
```diff
@@ -172,6 +172,11 @@ public struct BuildArguments {
 		// Disable code signing requirement for all builds
 		// Frameworks get signed in the copy-frameworks action
 		args += [ "CODE_SIGNING_REQUIRED=NO", "CODE_SIGN_IDENTITY=" ]
+		
+		// Allow to force unsanitized environment
```
NachoSoto: Could you expand this comment to explain what `XCODEBUILD_USE_SANITIZED_ENVIRONMENT` is exactly?

flovilmart: Starting Xcode 7, the UseSanitizedBuildEnvironment defaults to YES, which overrides/disables any externally set environment variables.

This env var allow a user to override the default behavior (=YES) in order to force xcodebuild to use the current environment. 

Is that clear enough? I've added that a while ago, and don't really recall what the status was on the issue/resolution we wanted to apply to the original issue.

Any guidance appreciated

ikesyo:
Closing this due to inactivity. But feel free to succeed to this. We now have `BuildOptions` type so this should be easier than before.

PR #1071, opened 2016-01-29
RAC4.0

mdiep:


Cartfile.resolved
```diff
@@ -5,5 +5,5 @@ github "Quick/Quick" "v0.8.0"
 github "antitypical/Result" "1.0.1"
 github "jspahrsummers/xcconfigs" "ec5753493605deed7358dec5f9260f503d3ed650"
 github "Carthage/Commandant" "0.8.2"
-github "ReactiveCocoa/ReactiveCocoa" "v4.0.0-RC.1"
-github "Carthage/ReactiveTask" "0.9"
+github "ReactiveCocoa/ReactiveCocoa" "v4.0.0"
```
NachoSoto: Maybe use `~> 4.0.0` now?

Source/CarthageKit/Xcode.swift
```diff
@@ -1254,7 +1254,7 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
 	return SignalProducer { observer, disposable in
 		// Use SignalProducer.buffer() to avoid enumerating the given directory
 		// multiple times.
-		let (locatorBuffer, locatorObserver) = SignalProducer<(ProjectLocator, [String]), CarthageError>.buffer()
+		let (locatorBuffer, locatorObserver) = SignalProducer<(ProjectLocator, [String]), CarthageError>.buffer(Int.max)
```
NachoSoto: Good call!

Source/CarthageKit/Xcode.swift
```diff
@@ -1254,7 +1254,7 @@ public func buildInDirectory(directoryURL: NSURL, withConfiguration configuratio
...
 
 		buildableSchemesInDirectory(directoryURL, withConfiguration: configuration, forPlatforms: platforms)
```
NachoSoto: Actually, this can be accomplished using `replayLazily` now!

Cartfile.resolved
mdiep: This is the `Cartfile.resolved`, so it has to be a commit or tag. ☺️

Source/CarthageKit/Xcode.swift
mdiep: Good idea! But I think this would be better left to a separate PR.

Cartfile.resolved
NachoSoto: Oh of course :P

Source/CarthageKit/Xcode.swift
NachoSoto: Ok :)

ikesyo: > Is that function covered by tests though? This might not be exactly equivalent since it was previously eagerly started, so we should make sure it doesn't break anything.

I think this should be covered by the [XcodeSpec](https://github.com/Carthage/Carthage/blob/f0124fc1aa9c3537169d6a5ba6a0c552d61f6563/Source/CarthageKitTests/XcodeSpec.swift#L106-L112). :wink:

Issue #1084, opened 2016-02-02
carthage build --no-use-binaries fails on 0.12


ratkins:
````
$ carthage build --verbose --platform iOS --no-use-binaries
Unrecognized arguments: --no-use-binaries
````

This works in 0.11. If this is a deliberate change, it's not mentioned in the [release notes](https://github.com/Carthage/Carthage/releases/tag/0.12).

ikesyo:
`--no-use-binaries` is not a valid option for `build` command since before. That is for `update`, `bootstrap` and `checkout` commands. Those invalid options which are not parsed by a command now outputs the errors.

PR #1085, opened 2016-02-02
Clarify wording in Known Issues

ratkins:
My original advice here (calling carthage build with --no-use-binaries, as opposed to update) was actually [wrong](https://github.com/Carthage/Carthage/issues/1084), and due to command line parsing in 0.12 being more strict it actually fails.


README.md
```diff
@@ -245,7 +245,7 @@ Want to advertise that your project can be used with Carthage? You can add a com
...
+See [Carthage issue #924](https://github.com/Carthage/Carthage/issues/924) for background on the reasons, but as at Xcode 7.2, Apple recommends that ["Frameworks written in Swift should be compiled from source as part of the same project that depends on them to guarantee a single, consistent compilation environment. (22492040)"](https://developer.apple.com/library/ios/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html). Using Swift frameworks built on other machines will cause Xcode's [debugger](https://github.com/Carthage/Carthage/issues/832) to [crash](https://github.com/Carthage/Carthage/issues/924) and other [strange build errors](https://github.com/Carthage/Carthage/issues/785). To avoid this, do not check your `Carthage/Build` directory into source control or use .framework release binaries. Instead, create a Run Script build phase in your Xcode project which calls `carthage update --no-use-binaries`.
```
mdiep: I wouldn't recommend calling `carthage update` as part of the build process—that will update you to newer versions of the dependencies if available. Calling `carthage bootstrap` would be a better fit.

ratkins:
In that case I don't understand what `carthage bootstrap` does? `carthage help bootstrap` says "Check out and build the project's dependencies", with a `--no-checkout` option that promises to "skip the checking out of dependencies after updating", implying that `bootstrap` also does an update. I can't find a more detailed doc link.

(We chose to check in our `Carthage/Checkouts` directory so we could be reasonably sure any given checkout of our project would be internally consistent and buildable without Internet access*, and run `carthage build` on each machine individually because of #924. I still have no idea if this is a recommended approach but pre the [for-other-reasons-broken](https://forums.developer.apple.com/thread/34683) Xcode 7.2.1 it mostly worked.)

mdiep:
`bootstrap` is `checkout` + `build`. The `--no-checkout` option is a bit weird, but it's because `bootstrap` just uses the same options as `update`. `update` first updates dependencies, then runs `bootstrap`.

I wouldn't check in `Carthage/Checkouts` personally, but it is something we support. `Cartfile.resolved` should guarantee internal consistency by itself. Building without internet access is another issue.

ratkins:
Updated to suggest `bootstrap --no-use-binaries`, and add a pointer to the Xcode release notes/#942's update on the saga.

README.md
```diff
@@ -245,10 +245,9 @@ Want to advertise that your project can be used with Carthage? You can add a com
...
 
+Xcode 7.2.1 (with a [fix backported from 7.3 beta 1](http://adcdownload.apple.com/Developer_Tools/Xcode_7.3_beta_2/Xcode_7.3_beta_2_v2_Release_Notes.pdf)) claims to fix the debugger crash when using precompiled binaries, but the issue is not yet [completely solved](https://github.com/Carthage/Carthage/issues/924#issuecomment-179009508).
```
NachoSoto: Maybe underlying `crash`, and make it more obvious that the crash is partially fixed, but not actually debugging?

ratkins: Sorry, I don't grok the phrasing you want here? And I understand from the details in the link that the Xcode crash itself *is fixed*, but debugging doesn't necessarily work properly in all cases, is that understanding incorrect?

ikesyo:
See #1137.

PR #1086, opened 2016-02-02
Permit dependency nodes match those have incompatible version specifiers, but have the same proposed version

### LINKED ISSUE #765: Potential issue with determining mutually incompatible requirements
I'm receiving the following message but I'm not sure how to approach it since `c704777e4b6d3a5c2a4778a36fca393ff94b1ffa` is a commit on the `td-swift-2` branch of Argo.

```
Could not pick a version for github "thoughtbot/Argo", due to mutually incompatible requirements:
	"c704777e4b6d3a5c2a4778a36fca393ff94b1ffa"
	"td-swift-2"
```

### LINKED ISSUE #1080: Error build robrix/Either using Carthage 0.12 (0.11 works)
Using a Cartfile with only `github "robrix/Either"` I get the following output. 
```
> carthage update --platform iOS --no-use-binaries
*** Fetching Either
*** Fetching Prelude
Could not pick a version for github "robrix/Prelude", due to mutually incompatible requirements:
	"a2c66881a5df6ccdb46f16e39523982c05e565c8"
	"master"
```
Currently master is at the commit with the above hash so both point to the same commit.

ikesyo:
~~Fixes #1080.~~ Fixes #765.

### DIFF
--- a/Source/CarthageKit/Resolver.swift
+++ b/Source/CarthageKit/Resolver.swift
@@ -310,7 +310,15 @@ private struct DependencyGraph: Equatable {
 				} else {
 					return .Failure(CarthageError.RequiredVersionNotFound(node.project, newSpecifier))
 				}
-			} else {
+			} else if existingNode.proposedVersion != node.proposedVersion {
+				// The guard condition above is required for enabling to build a
+				// dependency graph in the cases such as: one node has a
+				// `.GitReference` specifier of a branch name, and the other has
+				// a `.GitReference` of a SHA which is the HEAD of that branch.
+				// If the specifiers are not the same but the nodes have the same
+				// proposed version, the graph should be valid.
+				//
+				// See https://github.com/Carthage/Carthage/issues/765.
 				let existingDependencyOf = edges
 					.filter { _, value in value.contains(existingNode) }
 					.map { $0.0 }
--- a/Source/CarthageKitTests/ResolverSpec.swift
+++ b/Source/CarthageKitTests/ResolverSpec.swift
@@ -178,10 +178,10 @@ class ResolverSpec: QuickSpec {
 		case .GitHub(GitHubRepository(owner: "jspahrsummers", name: "objc-build-scripts")):
 			cartfile = Cartfile.fromString("github \"jspahrsummers/xcconfigs\" ~> 1.0")
 
-		case .Git(GitURL("tmp/TestCartfileBranch")):
+		case .Git(GitURL("/tmp/TestCartfileBranch")):
 			cartfile = Cartfile.fromString("git \"https://enterprise.local/desktop/git-error-translations2.git\" \"development\"")
 
-		case .Git(GitURL("tmp/TestCartfileSHA")):
+		case .Git(GitURL("/tmp/TestCartfileSHA")):
 			cartfile = Cartfile.fromString("git \"https://enterprise.local/desktop/git-error-translations2.git\" \"8ff4393ede2ca86d5a78edaf62b3a14d90bffab9\"")
 
 		default:
--- a/Source/CarthageKitTests/ResolverSpec.swift
+++ b/Source/CarthageKitTests/ResolverSpec.swift
@@ -38,12 +38,17 @@ class ResolverSpec: QuickSpec {
 	}
 
 	override func spec() {
-		it("should resolve a Cartfile") {
-			let resolver = Resolver(
+		var resolver: Resolver!
+
+		beforeEach {
+			resolver = Resolver(
 				versionsForDependency: self.versionsForDependency,
 				cartfileForDependency: self.cartfileForDependency,
 				resolvedGitReference: self.resolvedGitReference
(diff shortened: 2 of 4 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -537,9 +537,9 @@ public final class Project {
 		let resolver = Resolver(
 			versionsForDependency: versionsForProject,
 			cartfileForDependency: cartfileForDependency,
-			resolvedGitReference: { _, refName in
-				// Dependencies should be fully resolved already.
-				return SignalProducer(value: PinnedVersion(refName))
+			resolvedGitReference: { project, reference in
+				return resolveReferenceInRepository(repositoryFileURLForProject(project), reference)
```
mdiep: These should be fully resolved already. Is this because we're wanting to get a SHA instead of a tag name? Could a comment here and below explaining why it's necessary?

Source/CarthageKit/Resolver.swift
```diff
@@ -347,7 +347,7 @@ private struct DependencyGraph: Equatable {
 				} else {
 					return .Failure(CarthageError.RequiredVersionNotFound(node.project, newSpecifier))
 				}
-			} else {
+			} else if existingNode.proposedVersion != node.proposedVersion {
```
mdiep: Can you add a comment here as well? I don't understand why this is necessary.

mdiep:
Thanks for tackling this! I release 0.12.1 as soon as it's merged, since quite a few people seem to be hitting #1080.

NachoSoto:
Awesome!!!
Would it be possible to cover this with tests though? 🤑

ikesyo:
I've added some comments. Do they make sense?

mdiep:
I'm a little confused about why the `github "robrix/Either"` Cartfile mentioned in #1080 is hitting this. Since Prelude is only listed in a single Cartfile in that case, it doesn't seem like it should be a problem.

I'll need to take a deeper look at this (tonight, hopefully), but I agree that a test case would be a great addition.

ikesyo:
Updated `ResolverSpec` as well. :sparkles: 

> Since Prelude is only listed in a single Cartfile in that case, it doesn't seem like it should be a problem.

That conflicts between `a2c66881a5df6ccdb46f16e39523982c05e565c8` of `Either`'s `Cartfile.resolved` and `master` of `Either`'s `Cartfile`. The call graph is:

`resolveDependenciesInResolvedCartfile`
=> `resolveDependenciesFromNodePermutations`
=> `graphPermutationsForEachNode`
=> `graphPermutationsForDependenciesOfNode`
=> `nodePermutationsForCartfile`
=> [`resolvedGitReference`](https://github.com/Carthage/Carthage/blob/0.12/Source/CarthageKit/Resolver.swift#L155)

At that point, the closures ([this](https://github.com/Carthage/Carthage/blob/0.12/Source/CarthageKit/Project.swift#L540-L543) and [this](https://github.com/Carthage/Carthage/blob/0.12/Source/CarthageKit/Project.swift#L589-L592)) are called. `.GitReference("a2c66881a5df6ccdb46f16e39523982c05e565c8")` and `.GitReference("master")` [are not compatible](https://github.com/Carthage/Carthage/blob/0.12/Source/CarthageKit/Version.swift#L327-L332).

mdiep:
I have an alternative approach that I'd like to propose. I wasn't able to finish it tonight, but I'll open a PR tomorrow.

ikesyo:
I'm looking forward to seeing it! :smiley:

ikesyo:
I've just rebased and this now should fix #765 only. :zap:

mdiep:
This makes sense. :+1: Could you add a test for this?

Also, it needs another merge/rebase.

ikesyo:
Thanks @mdiep! I re-rebased and added a test case for this. :zap:

Source/CarthageKitTests/ResolverSpec.swift
```diff
@@ -93,6 +101,24 @@ class ResolverSpec: QuickSpec {
 			expect(dependencies).notTo(contain(Dependency("git-error-translations2", "8ff4393ede2ca86d5a78edaf62b3a14d90bffab9")))
 		}
 
+		it("should resolve a Cartfile whose dependency is specified by both a branch name and a SHA which is the HEAD of that branch") {
+			let resolver = Resolver(
```
NachoSoto: This is repeated. Maybe extract to a `beforeEach`?

mdiep: I think that'd be a good idea.

mdiep:
Thanks! This looks good. But I think we should extract the resolver as suggested. :sparkles:

ikesyo:
Done in 24bd33a. :wink:

mdiep:
:zap::zap::zap:

ikesyo:
Thank you @mdiep and @NachoSoto! :pray:

PR #1099, opened 2016-02-04
Skip "Carthage/Checkouts", not "Carthage" directory

### LINKED ISSUE #1098: TTTAttributedLabel shared framework scheme not found
I am trying to integrate `TTTAttributedLabel` into my project using Carthage at version `0.12.0` but unfortunately Carthage does not find the shared framework scheme. There definitely is a shared framework scheme, so I don't know why this is happening.

Here's the issue in the `TTTAttributedLabel` project: https://github.com/TTTAttributedLabel/TTTAttributedLabel/issues/642

Also, it seems that carthage should build shared framework schemes from all `.xcodeproj` since version `0.12.0` so this should work I think.

ikesyo:
Fixes #1098.


mdiep:
:+1:

PR #1100, opened 2016-02-04
Don't use Resolver to determine build order

### LINKED ISSUE #1080: Error build robrix/Either using Carthage 0.12 (0.11 works)

mdiep:
Fixes #1080 as an alternative to #1086. I’m not sure if it addresses #765.

I think this a simpler approach to determining build order, and it sidesteps #1080. It’s not straightforward to test, but I did manually test it with some projects and things look good.

/cc @ikesyo

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -564,7 +564,7 @@ public final class Project {
 					if let deps = graph[left.project] where deps.contains(right.project) {
 						return false
 					}
-					return left.project.name < right.project.name
+					return left.project.name.localizedStandardCompare(right.project.name) == .OrderedAscending
 				}
 				return SignalProducer(values: sorted)
 			}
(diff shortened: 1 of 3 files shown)


ikesyo:
Seems legit. :100: 

> I’m not sure if it addresses #765.

I confirmed that this PR does not fix it. I believe [this change](https://github.com/Carthage/Carthage/pull/1086/files#diff-44b501a44ff85cb36672c57f4139191dR350) in #1086 should fix it.

PR #1103, opened 2016-02-06
Don't use a localized sort

mdiep:
@kballard [pointed out](https://github.com/Carthage/Carthage/pull/1100#commitcomment-15925646) that it’s not the right thing to do in this instance.


mdiep:
Self-merging so I can release. :see_no_evil: :hear_no_evil: :speak_no_evil:

PR #1108, opened 2016-02-09
Skip unspecified dependencies in buildOrderForResolvedCartfile

### LINKED ISSUE #1107: `carthage checkout --use-submodules Foo Bar` tries to check out all dependencies
With Carthage 0.13.0, if I run `carthage checkout --use-submodules Foo Bar` to try and check out two of my dependencies as submodules, it instead checks out _all_ of them as submodules.

I'm pretty sure this worked in 0.12.0, though I'm not inclined to reinstall that to check.

ikesyo:
Fixes #1107. This is missed in #1100. :sweat_smile:

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -522,7 +522,7 @@ public final class Project {
 			})
 	}
 	
-	public func buildOrderForResolvedCartfile(cartfile: ResolvedCartfile) -> SignalProducer<Dependency<PinnedVersion>, CarthageError> {
+	public func buildOrderForResolvedCartfile(cartfile: ResolvedCartfile, dependenciesToInclude: [String]? = nil) -> SignalProducer<Dependency<PinnedVersion>, CarthageError> {
 		typealias DependencyGraph = [ProjectIdentifier: Set<ProjectIdentifier>]
 		// A resolved cartfile already has all the recursive dependencies. All we need to do is sort
 		// out the relationships between them. Loading the cartfile will each will give us its
@@ -557,6 +557,7 @@ public final class Project {
 					}
 					return result
 				}
+
 				let sorted = cartfile.dependencies.sort { left, right in
 					if let deps = graph[right.project] where deps.contains(left.project) {
 						return true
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -522,7 +522,7 @@ public final class Project {
 			})
 	}
 	
-	public func buildOrderForResolvedCartfile(cartfile: ResolvedCartfile) -> SignalProducer<Dependency<PinnedVersion>, CarthageError> {
+	public func buildOrderForResolvedCartfile(cartfile: ResolvedCartfile, dependenciesToInclude: [String]? = nil) -> SignalProducer<Dependency<PinnedVersion>, CarthageError> {
```
NachoSoto: Having a double value representing "no dependencies" is weird. Maybe change this to `dependenciesToInclude: [String] = []`?

mdiep: I think it might make more logical sense to let the caller do the filtering instead.

If we feel that duplication is an issue, we could add a custom RAC operator that filters values based on a collection of values.

ikesyo: That is discussed before: https://github.com/Carthage/Carthage/pull/972#discussion_r46878963

If we would change this, we should change the other signatures too.

mdiep: Actually, my comment doesn't make sense because we need to include the dependencies of any projects we're interested in.

mdiep: I think an optional here makes more sense than using an empty list to represent "include all dependencies". But I can see how an empty list is closer to how the command line arguments work.

ikesyo: >> I think it might make more logical sense to let the caller do the filtering instead.
>
> Actually, my comment doesn't make sense because we need to include the dependencies of any projects we're interested in.

Yes, we need to offer the graph to the caller if the caller should do the filtering.

mdiep:
:+1:

ikesyo:
Thank you! :pray:

Issue #1056, opened 2016-01-19
What format does Carthage expect for binary release zip files?


dpassage:
Background: I'm trying to adapt our build system to directly build a binary release for attachment to a private Github repo, such that Carthage can manage the dependency. Currently, the zip file's contents look like this:

```
Carthage/
   Build/
      iOS/
         Foo.framework
```

I can add a reference to the `Cartfile`; it downloads the binary release without complaint, but the framework doesn't appear under the `Carthage` directory of my project.

I've tried a few combinations, including making sure the name of the framework matches the name of the repo (`Foo` in the example above) with no luck. I've compared our .zip file with the file from an open source project that does work for binary releases; things look pretty much the same.

Any guidance available?

mdiep:
I don't remember exactly what Carthage looks for, but I think @ikesyo will remember the details.

ikesyo:
Carthage would do the following:

1. Picks up any frameworks which does not have nested frameworks in a archive: https://github.com/Carthage/Carthage/blob/0.11/Source/CarthageKit/Project.swift#L718-L728
2. Copy a framework to the appropriate platform folder (`Carthage/Build/$platform`). The platform will be determined by the `DTSDKName` value of the `Info.plist` file in the framework (`Resources/Info.plist` for Mac frameworks, `Info.plist` for other platforms): https://github.com/Carthage/Carthage/blob/0.11/Source/CarthageKit/Project.swift#L437-L449

In conclusion, there is no expected directory layout Carthage will do. The problem would be that your frameworks

- have nested frameworks in `Frameworks` folder.
- don't contain `Info.plist` file.
- the plist file does not have `DTSDKName` value.

dpassage:
Thanks, @ikesyo! I'll let you know how it goes.

ikesyo:
@dpassage Do you have any updates on this?

dpassage:
@ikesyo Yes, and thanks again for the info.

Looks like Carthage can't find the plist file:

```shell
$ ls -l Info.plist Resources/Info.plist
ls: Info.plist: No such file or directory
ls: Resources/Info.plist: No such file or directory
```

It looks like the framework in question is putting its `Info.plist` file down in a bundle directory:

```shell
$ find . -name Info.plist
./Versions/A/Resources/FooBundle.bundle/Info.plist
```

[...]

PR #1110, opened 2016-02-09
Improve the logic to detect the platform for a framework bundle

ikesyo:
Use `NSBundle.objectForInfoDictionaryKey`.

As suggested in https://github.com/Carthage/Carthage/issues/1056#issuecomment-179444959.


mdiep:
Thanks! This is a great improvement. :clap:

NachoSoto:
Nice

Issue #1060, opened 2016-01-21
`carthage version` should return either the SHA that it was compiled with, or the git tag version


younata:
I might end up making a pull to do this, but leaving this issue in case someone else with more free time than I wants to take it on.

The basic idea is:

`carthage version` should return either the SHA that it was compiled at (e.g. at the current master, this would be `a8cf57478166165d75bdb0d62f9b61f6ab7ee904`), or the tag version that the SHA corresponds with (e.g. instead of `38cafc2b60df1d4a6d22cd26bdad865d46012414`, it would return `0.11`).

Ashton-W:
Should be closed by #1114

PR #1114, opened 2016-02-10
Carthage version should display output of git describe

younata:
Adds a run script command to building the CLI version of carthage that overwrites the `CFBundleShortVersionString` variable in the Info.plist file be equal to the output of `git describe --tags`.

This means that `carthage version` will output something like `0.13-12-g4859c8c` when we're not on a release tag. Building on a release tag will work as expected.

This, unfortunately, will thrash the Info.plist file every time carthage is built.

If/when accepted, this'll complete #1060.

### DIFF
--- a/Carthage.xcodeproj/project.pbxproj
+++ b/Carthage.xcodeproj/project.pbxproj
@@ -540,6 +540,7 @@
 				D0E7B63019E9C64500EDBA4D /* Resources */,
 				D0E7B65719E9C7C700EDBA4D /* Extract CLI Tool */,
 				D0AAAB5319FB0960007B24B3 /* Embed Frameworks */,
+				89539B3B1C6BE36C00632A1B /* Update CLI version */,
 			);
 			buildRules = (
 			);
@@ -625,6 +626,21 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
+		89539B3B1C6BE36C00632A1B /* Update CLI version */ = {
+			isa = PBXShellScriptBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "Update CLI version";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/bash;
+			shellScript = ". script/update-cli-version";
+			showEnvVarsInLog = 0;
+		};
 		C2265FAB1A4B86AC00158358 /* Check Xcode Version */ = {
 			isa = PBXShellScriptBuildPhase;
 			buildActionMask = 2147483647;
--- a/Carthage.xcodeproj/project.pbxproj
+++ b/Carthage.xcodeproj/project.pbxproj
@@ -534,13 +534,13 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = D0E7B64919E9C64600EDBA4D /* Build configuration list for PBXNativeTarget "carthage" */;
 			buildPhases = (
+				89539B3B1C6BE36C00632A1B /* Update CLI version */,
 				C2265FAB1A4B86AC00158358 /* Check Xcode Version */,
 				D0E7B62E19E9C64500EDBA4D /* Sources */,
 				D0E7B62F19E9C64500EDBA4D /* Frameworks */,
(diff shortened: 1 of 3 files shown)


script/update-cli-version
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+main()
+{
+    local version=$(git describe --tags)
+    /usr/libexec/PlistBuddy -c "Set CFBundleShortVersionString $version" Source/carthage/Info.plist
```
NachoSoto: `Source/carthage/Info.plist` should probably be relative to this script's path.
I'd recommend doing:
```bash
DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```

Then getting the path like so:
```bash
"$DIR/../Source/carthage/Info.plist"
```

Make sure you add quotes though in case there are spaces in the path!

Carthage.xcodeproj/project.pbxproj
```diff
@@ -625,6 +626,21 @@
...
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "Update CLI version";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/bash;
+			shellScript = ". script/update-cli-version";
```
NachoSoto: Ditto, this should be `"$SRCROOT/script/update-cli-version"`

script/update-cli-version
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
NachoSoto: Can you add `-e`? Otherwise non-zero status codes will be ignored!

younata:
Turns out, because we don't have tests for this at all (GRUMBLE), this doesn't even work. I have a local version that does work, but this should have been identified by some form of tests/CI.

Carthage.xcodeproj/project.pbxproj
younata: Considering that the style of every other 'run script' build phase in the entire workspace is of type `. script/command`, I'm going to push back on this.

younata:
And with that last commit, I think I'm done.

ikesyo:
> This, unfortunately, will thrash the Info.plist file every time carthage is built.

We can use “Preprocess Info.plist File” build setting in `carthage` target and `${TEMP_DIR}/Preprocessed-Info.plist` in `script/update-cli-version`, then place the run-script phase before the “Copy Bundle Resources” phase. Additionally, we need to use `carthage`'s bundle instead of `CarthageKit`'s one here: https://github.com/Carthage/Carthage/pull/1114/files#diff-a4bcf3d9ce938ce87e22341b3004830cR19

younata:
>We can use “Preprocess Info.plist File” build setting in carthage target and ${TEMP_DIR}/Preprocessed-Info.plist in script/update-cli-version, then place the run-script phase before the “Copy Bundle Resources” phase. Additionally, we need to use carthage's bundle instead of CarthageKit's one here: https://github.com/Carthage/Carthage/pull/1114/files#diff-a4bcf3d9ce938ce87e22341b3004830cR19

Doing this only seems to work on the second time carthage is built, even with update-cli-version getting run right after Target Dependencies are built.

ikesyo:
> Additionally, we need to use carthage's bundle instead of CarthageKit's one here

I'm sorry, we should not do that because the Info.plist of `carthage` target cannot be found both by pkg installation nor homebrew installation (there should be no `carthage.app`).

ikesyo:
This currently does not work for pkg installation (confirmed by `sudo make install`). Because the packaged `CarthageKit.framework` comes from `carthage.app/Contents/Frameworks` and the framework is embedded into the location by “Embed Frameworks” phase which is in before the `Update CLI version` phase, so the framework's Info.plist is not modified. You should put the phase before “Embed Frameworks” I think.

younata:
@ikesyo @NachoSoto Thoughts now-ish? Any remaining nits?

younata:
Rebased upon master.

younata:
poke @ikesyo @NachoSoto @mdiep thoughts?

mdiep:
I'll try to take a look at this tomorrow if @ikesyo or @NachoSoto doesn't beat me to it.

mdiep:
Looks good to me. :+1:

PR #1115, opened 2016-02-11
[WIP] Add a unit test target for the CLI, with some tests

younata:
That test is failing while we wait for a couple of pull requests of mine to go through

Namely https://github.com/Carthage/Commandant/pull/55 and https://github.com/Carthage/Carthage/pull/1114


Carthage.xcodeproj/project.pbxproj
```diff
@@ -744,6 +829,100 @@
...
+				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_IDENTITY = "-";
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(DEVELOPER_FRAMEWORKS_DIR)",
+					"$(inherited)",
+				);
+				GCC_C_LANGUAGE_STANDARD = gnu99;
+				INFOPLIST_FILE = Source/carthageTests/Info.plist;
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/../Frameworks";
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
```
NachoSoto: Can you remove this? we should use the inherited value from the `xcconfig` files.

Carthage.xcodeproj/project.pbxproj
```diff
@@ -744,6 +829,100 @@
...
+				PRODUCT_BUNDLE_IDENTIFIER = org.carthage.carthageTests;
+				PRODUCT_NAME = "$(TARGET_NAME)";
```
NachoSoto: Ditto.

Carthage.xcodeproj/project.pbxproj
```diff
@@ -744,6 +829,100 @@
...
+				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_IDENTITY = "-";
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(DEVELOPER_FRAMEWORKS_DIR)",
+					"$(inherited)",
+				);
+				GCC_C_LANGUAGE_STANDARD = gnu99;
+				INFOPLIST_FILE = Source/carthageTests/Info.plist;
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/../Frameworks";
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
```
NachoSoto: Ditto

Carthage.xcodeproj/project.pbxproj
```diff
@@ -744,6 +829,100 @@
...
+		8903EA8A1C6C3C18001C0189 /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				ALWAYS_SEARCH_USER_PATHS = NO;
+				BUNDLE_LOADER = "$(TEST_HOST)";
+				CLANG_CXX_LANGUAGE_STANDARD = "gnu++0x";
+				CLANG_CXX_LIBRARY = "libc++";
+				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_IDENTITY = "-";
+				FRAMEWORK_SEARCH_PATHS = (
```
NachoSoto: Ditto

Carthage.xcodeproj/project.pbxproj
```diff
@@ -744,6 +829,100 @@
...
+					"$(DEVELOPER_FRAMEWORKS_DIR)",
+					"$(inherited)",
+				);
+				GCC_C_LANGUAGE_STANDARD = gnu99;
+				INFOPLIST_FILE = Source/carthageTests/Info.plist;
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/../Frameworks";
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
```
NachoSoto: Ditto

Carthage.xcodeproj/project.pbxproj
```diff
@@ -761,6 +941,7 @@
 			buildSettings = {
 				GCC_NO_COMMON_BLOCKS = YES;
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				VALID_ARCHS = x86_64;
```
NachoSoto: Ditto

Source/carthage/Printer.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+
+public protocol Printer {
```
NachoSoto: Does this need to be `public`?

Source/carthage/Printer.swift
```diff
@@ -0,0 +1,40 @@
...
+	func println(object: Any)
+	func print(object: Any)
+}
+
+internal struct ThreadSafePrinter: Printer {
+	private static let outputQueue = { () -> dispatch_queue_t in
+		let queue = dispatch_queue_create("org.carthage.carthage.outputQueue", DISPATCH_QUEUE_SERIAL)
+		dispatch_set_target_queue(queue, dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0))
+
+		atexit_b {
```
NachoSoto: Can you use `defer` for this instead?

Source/carthageTests/VersionCommandSpec.swift
```diff
@@ -0,0 +1,62 @@
...
+			printer = MockPrinter()
+			subject = VersionCommand(printer: printer)
+		}
+
+		it("prints the output of git describe --tags on the carthage directory as the version") {
+			let versionString = runCommand("/usr/bin/git describe --tags")
+
+//			subject.run(NoOptions<CarthageError>())
+
+			expect(printer.objectToPrintln) == versionString
```
NachoSoto: Nice

Source/carthageTests/VersionCommandSpec.swift
```diff
@@ -0,0 +1,62 @@
...
+
+		beforeEach {
+			printer = MockPrinter()
+			subject = VersionCommand(printer: printer)
+		}
+
+		it("prints the output of git describe --tags on the carthage directory as the version") {
+			let versionString = runCommand("/usr/bin/git describe --tags")
+
+//			subject.run(NoOptions<CarthageError>())
```
NachoSoto: Did you mean to leave this commented out?

Source/carthageTests/VersionCommandSpec.swift
```diff
@@ -0,0 +1,62 @@
+import Quick
+import Nimble
+import CarthageKit
+import Commandant
+import Cocoa
+
+import carthage
+
+private class MockPrinter: Printer {
```
NachoSoto: Nit: make this class `final`?

Source/carthage/Printer.swift
younata: In order to make a Mock implementation of Printer, Printer needs to be public.

Source/carthageTests/VersionCommandSpec.swift
younata: Yes, because the version of Commandant that Carthage has a submodule doesn't have https://github.com/Carthage/Commandant/pull/55 merged in yet.

Was waiting for Carthage Master to update the Commandant submodule, though I can include updating the Commandant submodule as part of this pull request if desired.

Source/carthageTests/VersionCommandSpec.swift
younata: Probably something that'll happen when it gets moved to it's own file so that I can share the MockPrinter with other spec files.

Source/carthage/Printer.swift
younata: You'll note that I pretty much copied this verbatim from Extensions.swift (see https://github.com/Carthage/Carthage/blob/master/Source/carthage/Extensions.swift#L19-L49)

Carthage.xcodeproj/project.pbxproj
younata: Removing `FRAMEWORK_SEARCH_PATHS = ( "$(DEVELOPER_FRAMEWORKS_DIR)", "$(inherited)" );` causes test failures because it can't find the xctest dylib at runtime (CarthageKitTests also has overriden FRAMEWORK_SEARCH_PATHS, for that matter), `BUNDLE_LOADER="$(TEST_HOST)";` and the variable that sets the `TEST_HOST`also need to be in so that we can correctly link with carthage.app's code.

Source/carthage/Printer.swift
ikesyo: `internal` should be okay if we use `@testable import carthage` in `VersionCommandSpec.swift`

Source/carthageTests/VersionCommandSpec.swift
ikesyo: In the meantime, you can use `NoOptions<CarthageError>.evaluate(.Usage).value!` as said in https://github.com/Carthage/Commandant/pull/55#issuecomment-182709531.

Source/carthage/Printer.swift
NachoSoto: Exactly :)

younata: I think that a unifying style means that `@testable import` shouldn't be used at all - you're not using it at all in `CarthageKitTests`, so why use it here?

I'm personally of the opinion that, unless you have good reason to, using `@testable import` means that you're lazy. You should write your implementation code as if it'll be consumed by some other entity (say... the tests). That is, `internal` should actually mean "private to the module", and `private` means "private to the file", instead of `internal` just meaning "I didn't want to type `public` in front of everything, but I also want to test my code".

In my own code/projects, I do use `@testable import` to test things like `ThreadSafePrinter`, which are implementation details that I just want to make sure I get right (and because it's the right thing to do - though `ThreadSafePrinter` in particular would be nigh-impossible to actually test correctly). `VersionCommand` just needs some instance of `Printer` to print to. When everything is integrated together in `carthage.app`, it just happens to use a `ThreadSafePrinter` (and, if I could, I would test that in main.swift, `VersionCommand` is init'd up with a `ThreadSafePrinter`, because that's important too).

Source/carthageTests/VersionCommandSpec.swift
younata: Interestingly enough, it actually breaks (undefined symbol errors pop up, I spent about an hour trying to track them down before giving up) if I do that, presumably because of weird interactions with ReactiveCocoa and it not properly searching for the right frameworks. Which is weird because I'm essentially using the exact same xcconfig setup as CarthageKitTests, which does that everywhere.

If I were to go into the Commandant code and just make that initializer public, those errors go away. In addition, this'll be the new behavior, why not just use that?

For reference, these are the kinds of errors I get which I use your suggestion:

```
Undefined symbols for architecture x86_64:
  "ext.Result.Result.ResultType<A where A: Result.ResultType>.value.getter : A.Value?", referenced from:
      carthageTests.VersionSpec.(spec (carthageTests.VersionSpec) -> () -> ()).(closure #2) in VersionCommandSpec.o
  "direct generic type metadata pattern for Result.Result", referenced from:
      type metadata accessor for Result.Result<Commandant.NoOptions<CarthageKit.CarthageError>, Commandant.CommandantError<CarthageKit.CarthageError>> in VersionCommandSpec.o
  "protocol witness table for <A, B where B: Swift.ErrorType> Result.Result<A, B> : Result.ResultType in Result", referenced from:
      carthageTests.VersionSpec.(spec (carthageTests.VersionSpec) -> () -> ()).(closure #2) in VersionCommandSpec.o
```

ikesyo: > why not just use that?

Since a version including that is not released yet.

ikesyo: > For reference, these are the kinds of errors I get which I use your suggestion:

That is because you didn't add `import Result`.

younata: Well don't I feel like an idiot. I thought Result was a part of ReactiveCocoa.

mdiep:
The dependencies mentioned above have been merged into `master`. Are you still planning to work on this, @younata?

younata:
@mdiep I don't foresee having the time to do this any time over the next few months. I'll close this PR (will open it back up if I do find the time to work on it again).

PR #1116, opened 2016-02-11
Restore `var` bindings

ikesyo:
This was removed in #998.

The conclusion of SE-00003 is changed to keep `var` bindings in Swift 3: https://github.com/apple/swift-evolution/blob/faff3215518b71965499b5b5be3475e0be850a01/proposals/0003-remove-var-parameters.md


mdiep:
:+1:

PR #1134, opened 2016-02-18
Extract out the BuildArguments struct into it's own file

younata:
Also provide unit tests for BuildArguments' arguments property.

This is going to be part of a series of pull requests meant to clean up the code quality (de-spaghetti Carthage), improve test coverage, and improve test runtime (the tests should run on the order of seconds, not minutes).

Related to that above goal, this adds 20 tests to CarthageKitTests, and fully unit tests the BuildArguments struct.


Source/CarthageKit/Xcode.swift
```diff
@@ -83,106 +83,6 @@ extension ProjectLocator: CustomStringConvertible {
 	}
 }
 
-/// Configures a build with Xcode.
-public struct BuildArguments {
```
NachoSoto: It's fine now, but next time can you try to keep things like this in separate commits? That way it's easier to review, since I can look at the moved code in one commit, and tests in another. It also makes potentially needing to revert one or the other easier.

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,327 @@
...
+						"-configuration",
+						"exampleConfiguration",
+						"CODE_SIGNING_REQUIRED=NO",
+						"CODE_SIGN_IDENTITY="
+					]
+				}
+
+				describe("specifying the sdk") {
+					it("includes the sdk if it's not .MacOSX") {
+						for sdk: SDK in [.iPhoneOS, .iPhoneSimulator, .watchOS, .watchSimulator, .tvOS, .tvSimulator] {
```
NachoSoto: Maybe add a `static let allSDKs: Set<SDK>` to `SDK` so that this list isn't here?

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,327 @@
...
+							var subject = BuildArguments(project: workspace)
+							subject.sdk = sdk
```
NachoSoto: What do you think about putting `it` inside of the loop so that each one is a separate tests? Then yo can include the `sdk` name in the test name, and failures would be more clear since each test would only have one assertion.

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,327 @@
...
+								"/Foo/Bar/workspace.xcworkspace",
+								"-sdk",
+								sdk.rawValue,
+								"CODE_SIGNING_REQUIRED=NO",
+								"CODE_SIGN_IDENTITY="
+							]
+						}
+					}
+
+					it("does not include the sdk flag if .MacOSX is specified") {
```
NachoSoto: Awesome tests!

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,327 @@
...
+						"-workspace",
+						"/Foo/Bar/workspace.xcworkspace",
+						"-destination",
+						"exampleDestination",
+						"CODE_SIGNING_REQUIRED=NO",
+						"CODE_SIGN_IDENTITY="
+					]
+				}
+
+				describe("specifying onlyActiveArchitecture") {
```
NachoSoto: Awesome

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,327 @@
...
+						"-configuration",
+						"exampleConfiguration",
+						"CODE_SIGNING_REQUIRED=NO",
+						"CODE_SIGN_IDENTITY="
+					]
+				}
+
+				describe("specifying the sdk") {
+					it("includes the sdk if it's not .MacOSX") {
+						for sdk: SDK in [.iPhoneOS, .iPhoneSimulator, .watchOS, .watchSimulator, .tvOS, .tvSimulator] {
```
NachoSoto: Then this wouldn't be duplicated anymore.

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,327 @@
...
+							var subject = BuildArguments(project: project)
+							subject.sdk = sdk
+
+							expect(subject.arguments) == [
```
NachoSoto: I would do the same thing here.

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,327 @@
...
+						"-project",
+						"/Foo/Bar/project.xcodeproj",
+						"-destination",
+						"exampleDestination",
+						"CODE_SIGNING_REQUIRED=NO",
+						"CODE_SIGN_IDENTITY="
+					]
+				}
+
+				describe("specifying onlyActiveArchitecture") {
```
NachoSoto: This test is repeated from above. Can you extract the common bits into functions?

NachoSoto:
Thank you! These tests are awesome :star2:

younata:
Hey @NachoSoto tried to DRY up the build arguments spec a bit, without it feeling too out of place for a quick/nimble spec. Not terribly a fan of the "clojure with inout parameter" thing in `itCreatesBuildArguments`, but I do think it fits stylistically better than, say, a clojure taking a BuildArguments, and returning one.

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,97 @@
...
+					subject.sdk = .MacOSX
+				}
+			}
+
+			itCreatesBuildArguments("includes the destination if given", arguments: ["-destination", "exampleDestination"]) { (inout subject: BuildArguments) in
+				subject.destination = "exampleDestination"
+			}
+
+			describe("specifying onlyActiveArchitecture") {
+				itCreatesBuildArguments("includes ONLY_ACTIVE_ARCH=YES if it's set to true", arguments: ["ONLY_ACTIVE_ARCH=YES"]) { (inout subject: BuildArguments) in
```
NachoSoto: I think this is fine!

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,97 @@
...
+			itCreatesBuildArguments("includes the scheme if one is given", arguments: ["-scheme", "exampleScheme"]) { (inout subject: BuildArguments) in
+				subject.scheme = "exampleScheme"
+			}
+
+			itCreatesBuildArguments("includes the configuration if one is given", arguments: ["-configuration", "exampleConfiguration"]) { (inout subject: BuildArguments) in
+				subject.configuration = "exampleConfiguration"
+			}
+
+			describe("specifying the sdk") {
+				for sdk in SDK.allSDKs.subtract([.MacOSX]) {
```
NachoSoto: Perfect

Source/CarthageKitTests/BuildArgumentsSpec.swift
```diff
@@ -0,0 +1,97 @@
...
+					itCreatesBuildArguments("includes \(sdk) in the argument if specified", arguments: ["-sdk", sdk.rawValue]) { (inout subject: BuildArguments) in
```
NachoSoto: Awesome

NachoSoto:
Beautiful, thank you1

PR #1136, opened 2016-02-18
[WIP] Inject instances of NSURLSession in GitHub.swift

younata:
Also test our calls to the network using a faked-out network interface.

I found a bit of unused code while doing this. `fetchAllPages()` in GitHub.swift is only called twice - recursively in `fetchAllPages()`, and in `releaseForTag()`. According to the github api v3 docs, `GET /repos/:owner/:repo/releases/tags/:tag` (the underlying url in `releaseForTag()`) won't ever page. So, unless I'm mistaken, the logic to recursively fetch additional pages is entirely unnecessary. Removing it, however, is outside the scope of this pull request.

Anyway, this is another pull request meant to clean up the code, improve test coverage, and improve test runtime. It doesn't do the last (the only major improvements to the test runtime will come when I can fake out all calls to xcodebuild), but it certainly does the previous two.

### DIFF
--- a/Source/CarthageKitTests/Fakes/FakeURLSession.swift
+++ b/Source/CarthageKitTests/Fakes/FakeURLSession.swift
@@ -1,0 +1,41 @@
+import Foundation
+
+final class FakeDownloadTask: NSURLSessionDownloadTask {
+	var started = false
+	override func resume() {
+		self.started = true
+	}
+
+	var cancelled = false
+	override func cancel() {
+		self.cancelled = true
+	}
+}
+
+final class FakeDataTask: NSURLSessionDataTask {
+	var started = false
+	override func resume() {
+		self.started = true
+	}
+
+	var cancelled = false
+	override func cancel() {
+		self.cancelled = true
+	}
+}
+
+final class FakeURLSession: NSURLSession {
+	var downloadTaskWithRequestArgs: [(NSURLRequest, (NSURL?, NSURLResponse?, NSError?) -> Void, FakeDownloadTask)] = []
+	override func downloadTaskWithRequest(request: NSURLRequest, completionHandler: (NSURL?, NSURLResponse?, NSError?) -> Void) -> NSURLSessionDownloadTask {
+		let downloadTask = FakeDownloadTask()
+		self.downloadTaskWithRequestArgs.append((request, completionHandler, downloadTask))
+		return downloadTask
+	}
+
+	var dataTaskWithRequestArgs: [(NSURLRequest, (NSData?, NSURLResponse?, NSError?) -> Void, FakeDataTask)] = []
+	override func dataTaskWithRequest(request: NSURLRequest, completionHandler: (NSData?, NSURLResponse?, NSError?) -> Void) -> NSURLSessionDataTask {
+		let dataTask = FakeDataTask()
+		dataTaskWithRequestArgs.append((request, completionHandler, dataTask))
+		return dataTask
+	}
+}
--- a/Source/CarthageKitTests/GitHubSpec.swift
+++ b/Source/CarthageKitTests/GitHubSpec.swift
@@ -1,0 +1,216 @@
+import Quick
+import Nimble
+import Result
+import ReactiveCocoa
+import CarthageKit
+
+class GitHubSpec: QuickSpec {
+	override func spec() {
+
+		describe("releaseForTag:repository:authorizationHeaderValue:urlSession:") {
+			it("makes a request to the specified server") {
+				let gitHubEnterpriseRepository = GitHubRepository(server: GitHubRepository.Server.Enterprise(scheme: "https", hostname: "example.com"), owner: "Carthage", name: "Carthage")
+
+				let urlSession = FakeURLSession()
+				_ = releaseForTag("v1.0.0", gitHubEnterpriseRepository, "example", urlSession).start()
+
+				expect(urlSession.dataTaskWithRequestArgs.count) == 1
+				let (request, _, task) = urlSession.dataTaskWithRequestArgs[0]
+
+				expect(task.started) == true
+				expect(request.URL) == NSURL(string: "https://example.com/api/v3/repos/Carthage/Carthage/releases/tags/v1.0.0")
+				expect(request.valueForHTTPHeaderField("Accept")) == "application/vnd.github.v3+json"
+				expect(request.valueForHTTPHeaderField("User-Agent")) == "CarthageKit-unknown/unknown"
+				expect(request.valueForHTTPHeaderField("Authorization")) == "example"
+			}
+
+			describe("for a standard github repo") {
+				let gitHubRepository = GitHubRepository(owner: "Example", name: "Carthage")
+
+				var releaseProducer: SignalProducer<GitHubRelease, CarthageError>!
+				var urlSession: FakeURLSession!
+				var events: [Event<GitHubRelease, CarthageError>] = []
+
+				beforeEach {
+					urlSession = FakeURLSession()
+					releaseProducer = releaseForTag("v1.0.0", gitHubRepository, "example", urlSession)
+
+					events = []
+					releaseProducer.on(event: { event in
+						events.append(event)
+					}).start()
+				}
+
+				it("makes a request to github") {
+					expect(urlSession.dataTaskWithRequestArgs.count) == 1
+					let (request, _, task) = urlSession.dataTaskWithRequestArgs[0]
+
+					expect(task.started) == true
+					expect(request.URL) == NSURL(string: "https://api.github.com/repos/Example/Carthage/releases/tags/v1.0.0")
+					expect(request.valueForHTTPHeaderField("Accept")) == "application/vnd.github.v3+json"
+					expect(request.valueForHTTPHeaderField("User-Agent")) == "CarthageKit-unknown/unknown"
+					expect(request.valueForHTTPHeaderField("Authorization")) == "example"
+				}
+
+				context("when the request succeeds") {
+					it("returns a GitHubRelease object when given a valid GitHubRelease dictionary data") {
+						let (request, completion, _) = urlSession.dataTaskWithRequestArgs[0]
+
+						let responseDictionary = [
+							"id": 1,
+							"name": "example release",
+							"tag_name": "v1.0.0",
+							"draft": false,
+							"prerelease": false,
+							"assets": [
+								[
+									"id": 2,
+									"name": "example asset",
+									"content_type": "text/text",
+									"url": "https://example.com/asset"
+								]
+							]
+						]
+						let data = try? NSJSONSerialization.dataWithJSONObject(responseDictionary, options: [])
+						let response = NSHTTPURLResponse(URL: request.URL!, statusCode: 200, HTTPVersion: nil, headerFields: nil)
+						completion(data, response, nil)
+
+						let asset = GitHubRelease.Asset(ID: 2, name: "example asset", contentType: "text/text", URL: NSURL(string: "https://example.com/asset")!)
+						let release = GitHubRelease(ID: 1, name: "example release", tag: "v1.0.0", draft: false, prerelease: false, assets: [asset])
+
+						expect(events.count) == 2
+						expect(events[0].error).to(beNil())
+						expect(events[0].value) == release
+						expect(events[1] == Event.Completed) == true
+					}
+
+					it("returns nothing when given a valid json object, but not decodeable to a GitHubRelease object") {
+						let (request, completion, _) = urlSession.dataTaskWithRequestArgs[0]
+
+						let data = "{}".dataUsingEncoding(NSUTF8StringEncoding)
+						let response = NSHTTPURLResponse(URL: request.URL!, statusCode: 200, HTTPVersion: nil, headerFields: nil)
+						completion(data, response, nil)
+
+						expect(events.count) == 1
+						expect(events[0] == Event.Completed) == true
+					}
+
+					it("returns an error when not given a valid json object") {
+						let (request, completion, _) = urlSession.dataTaskWithRequestArgs[0]
+
+						let data = "example string".dataUsingEncoding(NSUTF8StringEncoding)
+						let response = NSHTTPURLResponse(URL: request.URL!, statusCode: 200, HTTPVersion: nil, headerFields: nil)
+						completion(data, response, nil)
+
+						expect(events.count) == 1
+						expect(events[0].value).to(beNil())
+						expect(events[0].error) == CarthageError.ParseError(description: "Invalid JSON in releases for tag v1.0.0")
+					}
+				}
+
+				context("when the request fails") {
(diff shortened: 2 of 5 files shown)


Source/CarthageKitTests/Fakes/FakeURLSession.swift
```diff
@@ -0,0 +1,41 @@
...
+		self.started = true
+	}
+
+	var cancelled = false
+	override func cancel() {
+		self.cancelled = true
+	}
+}
+
+final class FakeURLSession: NSURLSession {
```
NachoSoto: Is `NSURLSession` meant to be subclassed? I don't think so... so this is probably fragile and relying on `NSURLSession`'s implementation details (_insert note about `final` default here_).

Instead of mocking requests like this, I'd recommend creating a `protocol URLSessionType` that `NSURLSession` conforms to, and so that `FakeURLSession` does **not** inherit `NSURLSession`, but instead implements all the fake logic by itself.

NachoSoto:
Awesome! Excited to have test coverage of this part of the codebase.

Source/CarthageKitTests/Fakes/FakeURLSession.swift
mdiep: How about using something like `OHHTTPStubs` instead.

I'm pretty :-1: on mocking, but the code as written isn't amenable to better testing.

younata: The docs for NSURLSession don't mention anything about subclassing (though the NSURLSession.h header does mention that the `init(configuration:)` and `init(configuration:delegate:delegateQueue:)` methods aren't inheritable), and I've been using this method for faking out the network since NSURLSession was introduced.

Additionally, this allowed me to re-use the existing `carthage_downloadWithRequest()` and `rac_dataWithRequest()` methods without making any modifications to them.

younata: If I were to mock out the entire url loading mechanism, I'd use my own (MockHTTP)[https://github.com/younata/MockHTTP], but I'd much rather not have to go down to that level - it's entirely unnecessary (would easily add 10x more test setup - would you like to do that the next time we add another network service for Carthage to talk to?), and providing a fake url session is much cleaner because we're forcing the issue of needing to have a network interface injected by the caller, not leaving it ambiguous as to who's job that is.

younata: Additionally, I need to get you onboard with mocking, because on my list of refactors to make is injecting NSFileManager (probably with a `FileManager` protocol that `NSFileManager` conforms to) because Tests should not depend on the file structure at all. Furthermore, having the code littered with calls to `NSFileManager.defaultManager()` is bad practice. NSURLSession is only used in a single file, which made it much easier to get in.

I also have "remove calls to Task" on that list, because making calls to other processes is slow and untestable. However, that's much deeper in the stack of refactors I want to do.

NachoSoto: I'm :+1: on using `OHHTTPStubs` too.

mdiep: Let me take some time to look through this code and think about testing.

IMO mocking is a code smell. It works well enough for adding tests to existing code, but it's much better to structure code so that it requires little to no mocking. So I'm not against mocking, but I'd also prefer to find an alternative if possible.

But I _absolutely_ appreciate you spending time trying to refactor and test this code.

younata: > But I absolutely appreciate you spending time trying to refactor and test this code.

Squee! You're welcome, and I'm happy to do it.

> IMO mocking is a code smell. It works well enough for adding tests to existing code, but it's much better to structure code so that it requires little to no mocking. So I'm not against mocking, but I'd also prefer to find an alternative if possible.

I totally see your point - integration tests are awesome (Speaking of, I have yet another pull request in the works to add a complete integration test suite - using Bash scripts to test the CLI API to make sure that repos are correctly downloaded, built, updated, etc.) - but in a lot of cases (e.g. talking to the network, talking to the file system, talking to other processes in general), you don't want to have to actually do that in the unit test suite. Not only does it make the tests run slower (tests that take more than a few seconds to run are incredibly painful), but it also tends to lead to unused or incorrect code that doesn't get found out because "oh, I won't test that area because it's too painful".

As a related example, one of the things I want to do is essentially mock out calls to xcodebuild in CarthageKitTests. Obviously, we want to still test that frameworks and such get created, but that should be done in an integration test, not a unit test. I think the unit test would be better served to just test that `xcodebuild -project foo.xcodeproj -scheme bar [...]` gets sent to a task service (and also that the actual task service does the right thing). Testing that the framework actually gets built should be left to an integration test.

This way, I can create a new feature that might make a call to something painful to actually run (e.g. support a new way of building projects), test that it makes the right calls to the mocked services, then also run the integration tests to see that everything still works.

In addition, using mocks forces us to use dependency injection, which is a beautiful thing (why get up and go to the fridge and get a beer when I can just ask my robot to get me a beer?). It also forces us to clean up the mess of bare functions that (in mine and other pivots' experience) RX tends to lead to.

I have plenty of things I want to do for Carthage, and more coming along the way. I have an entire desk here at work that is just cards representing things I want to do in Carthage (by the way, how would you like me to communicate this? Individual pull requests for things as I do them might get annoying after a while. I can create an issue for everything, but then the graph of issues (these two refactors depend upon this refactor, this other refactor depends upon two separate refactors, etc.) will get lost. I could also create a single giant issue, and have each individual refactor as a item in a list on that issue. If you have another suggestion, then that's great too. I'd like to be able to get feedback on some proposed refactors before I actually do them.

I really like Carthage, and I want to help improve it and make it easier for others to improve. Thanks for working with me on this.

Source/CarthageKitTests/GitHubSpec.swift
```diff
@@ -0,0 +1,216 @@
...
+							expect(events.count) == 1
+							expect(events[0].value).to(beNil())
+							expect(events[0].error) == CarthageError.NetworkError(error)
+						}
+					}
+				}
+			}
+		}
+
+		fdescribe("downloadAsset:authorizationHeaderValue:urlSession:") {
```
younata: Oops.

younata:
Unfocused that focused describe, and rebased upon master.

Source/CarthageKitTests/Fakes/FakeURLSession.swift
mdiep: I've had a bad cold this week, so I haven't had time to give this the look it deserves. I'm busy all day tomorrow too, but this is definitely on my list (along with the rest of your issues/PRs). If I don't get to it on Sunday, I'll find time on Monday. I hope this doesn't spoil your enthusiasm!

mdiep: I try to avoid DI when possible because I find it can make it harder to understand or change code. I gave [a talk](https://github.com/mdiep/composition-talk/blob/master/The%20Composition%20and%20Decomposition%20of%20Software.md) that touched on this last year. But the gist is that I try to structure programs like this:

 1. A functional core of pure functions that can easily be unit tested with simple inputs and outputs
 2. A DI-compatible layer of functions that chain things together
 3. An outer layer that connects the different pieces

I see layer 2 as being optional, depending on the project in question. Integration tests serve to test layer 3, but can be much reduced in scope if the first two layers are well tested. (I think you and I are in agreement on that point at least.)

[This article](http://jerrydantonio.com/testing-concurrent-code/) about testing concurrent code in Ruby is similar in approach, but probably more directly applicable to this particular example.

The code that we're trying to test is here:

```swift
private func fetchAllPages(URL: NSURL, _ authorizationHeaderValue: String?) -> SignalProducer<NSData, CarthageError> {
	let request = createGitHubRequest(URL, authorizationHeaderValue)

	return NSURLSession.sharedSession().rac_dataWithRequest(request)
		.mapError(CarthageError.NetworkError)
		.flatMap(.Concat) { data, response -> SignalProducer<NSData, CarthageError> in
			let thisData: SignalProducer<NSData, CarthageError> = SignalProducer(value: data)

			if let HTTPResponse = response as? NSHTTPURLResponse {
				let statusCode = HTTPResponse.statusCode
				if statusCode > 400 && statusCode < 600 && statusCode != 404 {
					return thisData
						.attemptMap { data -> Result<AnyObject, CarthageError> in
							if let object = try? NSJSONSerialization.JSONObjectWithData(data, options: []) {
								return .Success(object)
							} else {
								return .Failure(.ParseError(description: "Invalid JSON in API error response \(data)"))
							}
						}
						.map { (dictionary: AnyObject) -> String in
							if let error: GitHubError = decode(dictionary) {
								return error.message
							} else {
								return (NSString(data: data, encoding: NSUTF8StringEncoding) ?? NSHTTPURLResponse.localizedStringForStatusCode(statusCode)) as String
							}
						}
						.flatMapError { (error: CarthageError) in
							return SignalProducer(value: error.description)
						}
						.attemptMap { message -> Result<NSData, CarthageError> in
							return Result.Failure(CarthageError.GitHubAPIRequestFailed(message))
						}
				}
				
				if let linkHeader = HTTPResponse.allHeaderFields["Link"] as? String {
					let links = parseLinkHeader(linkHeader)
					for (URL, parameters) in links {
						if parameters.contains("rel=\"next\"") {
							// Automatically fetch the next page too.
							return thisData.concat(fetchAllPages(URL, authorizationHeaderValue))
						}
					}
				}
			}

			return thisData
		}
}
```

It does a few things:

1. Creates a request
2. Makes the request
3. Interprets the result

We don't need to test 2—`rac_dataWithRequest` is provided by a framework and should be tested there. We only need to test 1 (`createGitHubRequest`) and 3, which is currently embedded in the function. If we extract 3 to a separate function, we could end up with something like this:

```swift
private func fetchAllPages<O: ObjectType>(URL: NSURL, _ authorizationHeaderValue: String?) -> SignalProducer<O, CarthageError> {
    let request = createGitHubRequest(URL, authorizationHeaderValue)
    return NSURLSession.sharedSession().rac_dataWithRequest(request)
        .mapError(CarthageError.NetworkError)
        .map(O.init)
        .flatMap(.Concat) { object in
            if (object.nextPage) {
                return SignalProducer(value: object).concat(fetchAllPages(object.nextPage, authorizationHeader))
            }
            return SignalProducer(value: object)
        }
}
```

Now we can unit test `createGitHubRequest`, `O.init` for our object types, `O.nextPage`, and `fetchAllPages` doesn't need much testing. We can probably survive without testing it because it'll be obvious if it's completely broken, or we can add one or two integration-style tests (I would use an HTTP stubbing library for this).

What do you think of that approach?

> I have plenty of things I want to do for Carthage, and more coming along the way. I have an entire desk here at work that is just cards representing things I want to do in Carthage (by the way, how would you like me to communicate this? Individual pull requests for things as I do them might get annoying after a while. I can create an issue for everything, but then the graph of issues (these two refactors depend upon this refactor, this other refactor depends upon two separate refactors, etc.) will get lost. I could also create a single giant issue, and have each individual refactor as a item in a list on that issue. If you have another suggestion, then that's great too. I'd like to be able to get feedback on some proposed refactors before I actually do them.

That's awesome! :zap::zap::zap:

I'm going to add you as a contributor to Carthage. I think opening an issue for everything upfront will be overwhelming. I think a single giant issue would work (using a task list), or you could open up separate issues but not open the issue until you're nearly ready to work on it. That way you can get feedback, but you don't open 100 issues this week. :smile:

younata: Is there a video for that talk? Would love to see the original talk, but the notes are really good information too.

> Now we can unit test createGitHubRequest, O.init for our object types, O.nextPage, and fetchAllPages doesn't need much testing. We can probably survive without testing it because it'll be obvious if it's completely broken, or we can add one or two integration-style tests (I would use an HTTP stubbing library for this).

So, essentially, you're proposing that we reduce the untested/untestable area of code to as small as possible. I'm a fan of that.

My main concern is having too many pure functions, then it becomes non-obvious how everything fits together. For example, when I want to show someone an example of terrible "clean" code, I show them `Xcode.swift`: Lots (boo!) of pure (yay!) functions, in addition to [this 200-line behemoth](https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/Xcode.swift#L868-L1043), and it's really hard to figure out what's going on because everything is (seemingly?) intermingled. And of course, with the way things are currently written, it is necessarily tested in an integration style.

> I'm going to add you as a contributor to Carthage. I think opening an issue for everything upfront will be overwhelming. I think a single giant issue would work (using a task list), or you could open up separate issues but not open the issue until you're nearly ready to work on it. That way you can get feedback, but you don't open 100 issues this week. :smile:

Awesome. Thank you! I opened https://github.com/Carthage/Carthage/issues/1147 to collect everything into an issue.

mdiep: > Is there a video for that talk?

Sadly, no. :disappointed: 

> So, essentially, you're proposing that we reduce the untested/untestable area of code to as small as possible. I'm a fan of that.

:zap:

> My main concern is having too many pure functions, then it becomes non-obvious how everything fits together. For example, when I want to show someone an example of terrible "clean" code, I show them Xcode.swift: Lots (boo!) of pure (yay!) functions, in addition to this 200-line behemoth, and it's really hard to figure out what's going on because everything is (seemingly?) intermingled. And of course, with the way things are currently written, it is necessarily tested in an integration style.

I definitely agree about the code as it stands today. It really lacks good structure. I _think_ that's the problem: that we have the wrong pure functions, not that we have too many of them. But that's just intuition and could easily be wrong.

And if you think the integration tests as you have them in this PR would be helpful in refactoring and cleaning things up so we can better test, then I will happily merge them! If nothing else, they might be a good starting place for a PR to cleanup the code: add the tests, refactor the code, unit test the code, then remove some or all of the integration tests.

younata: > I definitely agree about the code as it stands today. It really lacks good structure. I think that's the problem: that we have the wrong pure functions, not that we have too many of them. But that's just intuition and could easily be wrong.

I think it was more that it started as something good, but has since grown unchecked and hasn't been pruned back.

> And if you think the integration tests as you have them in this PR would be helpful in refactoring and cleaning things up so we can better test, then I will happily merge them! If nothing else, they might be a good starting place for a PR to cleanup the code: add the tests, refactor the code, unit test the code, then remove some or all of the integration tests.

Hmm, this is somewhat troubling. I'd like to move to something that can be expressed as `handleNetworkResults(fetchFromNetwork(GitHubRequest(githubRepo, authenticationParam)))` , but that's not really within the scope of this pull request.

In the meantime, though, I think it would be better to have this area of code be tested, but acknowledge that we want to change it.

Bottom line, yes, the current tests will help as it gets refactored, so please merge this. I'll open up a separate pull request (maybe as soon as later today) refactor this again.

mdiep:
:+1:

PR #1137, opened 2016-02-18
Remove --no-use-binaries from Known Issues

daniel-beard:
The option is does not work with the newest version.


NachoSoto:
Good point, `--no-use-binaries` is only an option for `build` and `bootstrap` now.

PR #1139, opened 2016-02-19
Fix dependency resolution when building

erichoracek:
As of #1100,  the `build` command uses a comparison sort to resolve ordering of dependencies to build. That change resulted in frameworks being built out of order in significantly complex dependency graphs, causing "module not found" errors.

To resolve this, [Kahn's topological sort](https://en.wikipedia.org/wiki/Topological_sorting) is used in place of a comparison sort to ensure that the dependency graph is built in the correct order.

In our project with 43 unique dependencies (including transitive dependencies), this change fixed "module not found" errors when using the build command in the most recent release. Additionally, since this is a generic algorithm, it can be easily tested (assuming there is a reasonable place to put it). It would appear #1100 removed the previous tests around build order.

This PR has some open questions:
- Is there another place to put the generic topological sort algorithm?
- Is it reasonable to fail out the `build` command when a cycle is detected during a topological sort?

### DIFF
--- a/Source/CarthageKit/Algorithms.swift
+++ b/Source/CarthageKit/Algorithms.swift
@@ -35,8 +35,7 @@
 /// Nodes that are equal from a topological perspective are sorted by the
 /// strict total order as defined by `Comparable`.
 public func topologicalSort<Node: Comparable>(var graph: Dictionary<Node, Set<Node>>) -> [Node]? {
-	var queue = graph
-		.filter { _, incomingEdges in incomingEdges.isEmpty }
+	var queue = graph.filter { _, incomingEdges in incomingEdges.isEmpty }
 		.map { node, _ in node }
 
 	queue.forEach { node in graph.removeValueForKey(node) }
@@ -51,7 +50,6 @@ public func topologicalSort<Node: Comparable>(var graph: Dictionary<Node, Set<No
 
 		for (node, incomingEdges) in graph where incomingEdges.contains(lastNode) {
 			let filteredIncomingEdges = incomingEdges.subtract([lastNode])
-
 			graph[node] = filteredIncomingEdges
 
 			if filteredIncomingEdges.isEmpty {
--- a/Source/CarthageKit/Algorithms.swift
+++ b/Source/CarthageKit/Algorithms.swift
@@ -34,11 +34,13 @@
 ///
 /// Nodes that are equal from a topological perspective are sorted by the
 /// strict total order as defined by `Comparable`.
-public func topologicalSort<Node: Comparable>(var graph: Dictionary<Node, Set<Node>>) -> [Node]? {
-	var queue = graph.filter { _, incomingEdges in incomingEdges.isEmpty }
+public func topologicalSort<Node: Comparable>(graph: Dictionary<Node, Set<Node>>) -> [Node]? {
+	var queue = graph
+		.filter { _, incomingEdges in incomingEdges.isEmpty }
 		.map { node, _ in node }
 
-	queue.forEach { node in graph.removeValueForKey(node) }
+	var workingGraph = graph
+	queue.forEach { node in workingGraph.removeValueForKey(node) }
 
 	var sorted: [Node] = []
 
@@ -48,16 +50,16 @@ public func topologicalSort<Node: Comparable>(var graph: Dictionary<Node, Set<No
 		let lastNode = queue.removeLast()
 		sorted.append(lastNode)
 
-		for (node, incomingEdges) in graph where incomingEdges.contains(lastNode) {
+		for (node, incomingEdges) in workingGraph where incomingEdges.contains(lastNode) {
 			let filteredIncomingEdges = incomingEdges.subtract([lastNode])
-			graph[node] = filteredIncomingEdges
+			workingGraph[node] = filteredIncomingEdges
 
 			if filteredIncomingEdges.isEmpty {
 				queue.append(node)
-				graph.removeValueForKey(node)
+				workingGraph.removeValueForKey(node)
 			}
 		}
 	}
 
-	return graph.isEmpty ? sorted : nil
+	return workingGraph.isEmpty ? sorted : nil
 }
--- a/Source/CarthageKit/Algorithms.swift
+++ b/Source/CarthageKit/Algorithms.swift
@@ -1,0 +1,65 @@
+//
+//  Algorithms.swift
+//  Carthage
+//
+//  Created by Eric Horacek on 2/19/16.
+//  Copyright © 2016 Carthage. All rights reserved.
+//
+
+/// Returns an array containing the topologically sorted nodes of the provided
+/// directed graph, or nil if the graph contains a cycle or is malformed.
+///
+/// The sort is performed using
+/// [Khan's Algorithm](https://en.wikipedia.org/wiki/Topological_sorting#Kahn.27s_algorithm).
+///
+/// The provided graph should be encoded as a dictionary where:
+/// - The keys are the nodes of the graph
+/// - The values are the set of nodes that the key node has a incoming edge from
+///
+/// For example, the following graph:
+/// ```
+/// A ◀─── B
+/// ▲      ▲
+/// │      │
+/// C ◀─── D
+/// ```
+/// should be encoded as:
+/// ```
+/// [ A: Set([B, C]), B: Set([D]), C: Set([D]), D: Set() ]
+/// ```
+/// and would be sorted as:
+/// ```
+/// [D, B, C, A]
+/// ```
+///
+/// Nodes that are equal from a topological perspective are sorted by the
+/// strict total order as defined by `Comparable`.
+public func topologicalSort<Node: Comparable>(var edges: Dictionary<Node, Set<Node>>) -> [Node]? {
+	var queue: [Node] = edges
+		.filter { _, edges in edges.isEmpty }
+		.map { node, _ in node }
+
+	queue.forEach { node in edges.removeValueForKey(node) }
+
+	var sorted: [Node] = []
+
+	while !queue.isEmpty {
+		queue.sortInPlace(>)
+
+		let lastNode = queue.removeLast()
+		sorted.append(lastNode)
+
+		for (node, inEdges) in edges {
+			guard inEdges.contains(lastNode) else { continue }
+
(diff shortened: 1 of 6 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,57 @@ public final class Project {
...
-						
-						result.updateValue(recursive, forKey: dependency)
-						
-						for (project, projects) in result {
-							if projects.contains(dependency) {
-								result.updateValue(projects.union(recursive), forKey: project)
-							}
+			.reduce([:]) { (working: DependencyGraph, next: DependencyGraph) in
+				var result = working
+				next.forEach { result.updateValue($1, forKey: $0) }
```
mdiep: Could you make these named parameters? I think it will make it more obvious how this works.

Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,57 @@ public final class Project {
...
+				return result
+			}
+			.flatMap(.Latest) { (graph: DependencyGraph) -> SignalProducer<Dependency<PinnedVersion>, CarthageError> in
+				// Open Question: Is there a place that a generic topological 
+				// sort should live?
```
mdiep: I think it'd be fine to make a new file. Maybe `Algorithms.swift`?

mdiep: I'd love to see some documentation for this once it's moved.

Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,57 @@ public final class Project {
...
+							nodeQueue.append(node)
 						}
 					}
-					return result
+
+					// There is a cycle in the graph. How should we communicate
+					// this?
+//					if sortedNodes.count != edges.count {
+//
+//					}
```
mdiep: Since it's the only way that this can fail, I think returning `[Node]?` would work. So long as we document that it's what `nil` means.

But `buildOrderForResolvedCartfile` should return a suitable `CarthageError`. It doesn't look like we have one now for a build cycle, so one would need to be added.

Source/CarthageKit/Project.swift
mdiep: I think we could actually use `mergeGraphs` here instead. It doesn't seem like this properly handles all cases of shared, recursive dependencies as-is.

Also, I had no idea you could use `updateValue` like this. :confused:

Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,57 @@ public final class Project {
...
+
+					return sortedNodes
 				}
 
+				let sortedProjects = topologicalSort(graph, isOrderedBefore: { $0.name > $1.name })
```
mdiep: Isn't the order of `isOrderedBefore` here backwards? Also, I think it should still be using a case insensitive comparison.

I'd really like to see this added as a method or property on `DependencyGraph` as well. `buildOrder`, maybe?

mdiep:
1. Sorry for introducing bugs. :disappointed: 
2. Thank you so much for opening a pull request to fix them! :zap::zap::zap:
3. This seems like a brilliant approach.
4. I wish I'd thought of it!

I left a few notes about the structure and some of the details. I still need to look over `topologicalSort` in detail; I'll try to do that later today.

Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,57 @@ public final class Project {
...
+			.reduce([:]) { (working: DependencyGraph, next: DependencyGraph) in
+				var result = working
+				next.forEach { result.updateValue($1, forKey: $0) }
+				return result
+			}
+			.flatMap(.Latest) { (graph: DependencyGraph) -> SignalProducer<Dependency<PinnedVersion>, CarthageError> in
+				// Open Question: Is there a place that a generic topological 
+				// sort should live?
+				func topologicalSort<Node: Equatable>(edges: Dictionary<Node, Set<Node>>, @noescape isOrderedBefore: (Node, Node) -> Bool) -> [Node] {
+					var nodeQueue: [Node] = edges.filter { $1.isEmpty }.map { return $0.0 }
```
mdiep: I'd prefer to use named variables in the blocks. I think that makes it easier to read the code. It will probably make sense to split the `filter` and `map` on separate lines with the added characters.

Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,57 @@ public final class Project {
...
+
+					var workingEdges = edges
```
mdiep: Naming this `edges` would probably make this clearer.

Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,57 @@ public final class Project {
...
+					var sortedNodes: [Node] = []
+
+					while !nodeQueue.isEmpty {
+						nodeQueue.sortInPlace(isOrderedBefore)
+
+						let lastNode = nodeQueue.removeLast()
+						sortedNodes.append(lastNode)
+
+						for (node, nodeEdges) in workingEdges {
+							guard nodeEdges.contains(lastNode) else { continue }
```
mdiep: This seems like a good candidate for `for … in … where … { … }`.

Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,57 @@ public final class Project {
...
+
+							let unsharedEdges = nodeEdges.subtract([lastNode])
+							workingEdges[node] = unsharedEdges
```
mdiep: It'd be better to use `Set.remove()` here since we're only removing a single element. And you should be able to do `workingEdges[node].remove(lastNode)` without the local variable.

Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,57 @@ public final class Project {
...
+
+							guard unsharedEdges.isEmpty else { continue }
+							nodeQueue.append(node)
```
mdiep: I'd invert this conditional and put the `append` inside an `if`.

mdiep:
Great stuff. :zap: I left some comments that I think would make the code more readable, but the implementation looks spot on.

If you're up for it, I'd love to see tests around `topologicalSort`. That's the real beauty of this PR—the code is far more testable. But honestly, I'd take this without the tests and add them later if you're not up for it.

Source/CarthageKit/Project.swift
```diff
@@ -540,38 +540,23 @@ public final class Project {
...
-						result.updateValue(recursive, forKey: dependency)
-						
-						for (project, projects) in result {
-							if projects.contains(dependency) {
-								result.updateValue(projects.union(recursive), forKey: project)
-							}
-						}
-					}
-					return result
+			.reduce([:]) { (working: DependencyGraph, next: DependencyGraph) in
```
erichoracek: @mdiep I couldn't figure out how to use `mergeGraphs` here—it would seem that there's two types of `DependencyGraph` in CarthageKit.

Source/CarthageKit/Algorithms.swift
```diff
@@ -0,0 +1,63 @@
...
+/// [ A: Set([B, C]), B: Set([D]), C: Set([D]), D: Set() ]
+/// ```
+/// and would be sorted as:
+/// ```
+/// [D, B, C, A]
+/// ```
+///
+/// Nodes that are equal from a topological perspective are sorted by the
+/// strict total order as defined by `Comparable`.
+public func topologicalSort<Node: Comparable>(var graph: Dictionary<Node, Set<Node>>) -> [Node]? {
```
ikesyo: `var` for function parameters is deperecated in Swift 2.2 and will be removed in Swift 3: https://github.com/apple/swift-evolution/blob/master/proposals/0003-remove-var-parameters.md

Source/CarthageKit/Project.swift
mdiep: Oh, that's right. This is just a `typealias` in this method. :stuck_out_tongue_winking_eye:

Source/CarthageKit/Algorithms.swift
```diff
@@ -0,0 +1,63 @@
...
+		.map { node, _ in node }
+
+	queue.forEach { node in graph.removeValueForKey(node) }
+
+	var sorted: [Node] = []
+
+	while !queue.isEmpty {
+		queue.sortInPlace(>)
+
+		let lastNode = queue.removeLast()
```
mdiep: Since we're thinking of this as a _queue_, we should sort with `<` and `removeFirst`.

Source/CarthageKit/Algorithms.swift
```diff
@@ -0,0 +1,63 @@
...
+		sorted.append(lastNode)
+
+		for (node, incomingEdges) in graph where incomingEdges.contains(lastNode) {
+			let filteredIncomingEdges = incomingEdges.subtract([lastNode])
```
mdiep: This could still use `remove` instead of `subtract`.

Source/CarthageKit/Cartfile.swift
```diff
@@ -211,6 +211,10 @@ public func ==(lhs: ProjectIdentifier, rhs: ProjectIdentifier) -> Bool {
 	}
 }
 
+public func <(lhs: ProjectIdentifier, rhs: ProjectIdentifier) -> Bool {
+	return lhs.name < rhs.name
```
mdiep: This should use a case insensitive comparison.

mdiep:
This is looking really great. Just a few more small issues.

Source/CarthageKit/Algorithms.swift
erichoracek: While I agree that it is semantically simpler, `removeFirst` is `O(self.count)` while `removeLast` is `O(1)`. Perhaps the performance tradeoff merits an implementation that is slightly harder to reason about?

mdiep: In that case, can you add a comment to that effect? I doubt that the performance would ever be an issue with the small sizes here, but I'm willing to go with that if it's documented.

erichoracek: After thinking it through, it is less a queue and more of a list. I've renamed it as such. I think it resolves this comment as well.

mdiep:
Awesome. :zap:

ikesyo:
a:clap:

PR #1140, opened 2016-02-19
Add an integration test suite

younata:
Adds an integration test suite that uses [bats](https://github.com/sstephenson/bats) to test the carthage CLI.

Currently only tests archive, have plans to test others.

If you'd rather not promote Ra, then feel free to suggest some other repo.

### DIFF
--- a/Source/CarthageIntegration/archive.bats
+++ b/Source/CarthageIntegration/archive.bats
@@ -12,13 +12,13 @@ teardown() {
 }
 
 @test "carthage archive errors unless carthage build --no-skip-current has been run" {
-    run /tmp/Carthage.dst/usr/local/bin/carthage archive
+    run carthage archive
     [ "$status" -eq 1 ]
     [ "$output" = "Could not find any copies of Ra.framework. Make sure you're in the project’s root and that the frameworks have already been built using 'carthage build --no-skip-current'." ]
 }
 
 @test "carthage archive after carthage build --no-skip-current produces a zipped framework of all frameworks" {
-    run /tmp/Carthage.dst/usr/local/bin/carthage build --no-skip-current
+    run carthage build --no-skip-current
     [ "$status" -eq 0 ]
     run carthage archive
     [ "$status" -eq 0 ]
--- a/Source/CarthageIntegration/archive.bats
+++ b/Source/CarthageIntegration/archive.bats
@@ -1,25 +1,26 @@
 #!/usr/bin/env bats
 
-@test "carthage archive errors unless carthage build --no-skip-current has been run" {
-    pushd $BATS_TMPDIR
+setup() {
+    cd $BATS_TMPDIR
     rm -rf Ra
-    git clone https://github.com/younata/Ra.git >&2
+    git clone https://github.com/younata/Ra.git
     cd Ra
+}
+
+teardown() {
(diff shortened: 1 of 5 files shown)


mdiep:
This is ***really** cool!!!

younata:
> This is really cool!!!

I think this is the highest praise I've gotten in a long time.

robrix:
Nice one, @younata!

robrix:
> If you'd rather not promote Ra, then feel free to suggest some other repo.

I think this is absolutely fine, personally :+1:

younata:
Ok, I think this works as a pretty decent smoke test for the common use-cases of carthage (maybe add a test for bootstrap?), other tests, if/when they come up, will probably just be additional pull requests.

I'm going to wait until https://github.com/Carthage/Carthage/pull/1150 is merged before I take the WIP flag off this. Reasonably certain that the only reason travis is red is because it's finding the bug that #1150 fixes.

younata:
@mdiep @robrix and whoever else: I think this is good to go for now. The only other command I can think of to test would be copy-frameworks, but I don't know of an efficient way to write tests for that that don't involve deep knowledge of how it works. Please let me know of any other commands that would be good for this PR.

Source/CarthageIntegration/archive.bats
```diff
@@ -0,0 +1,26 @@
...
+    [ "$status" -eq 1 ]
+    [ "$output" = "Could not find any copies of Ra.framework. Make sure you're in the project’s root and that the frameworks have already been built using 'carthage build --no-skip-current'." ]
+}
+
+@test "carthage archive after carthage build --no-skip-current produces a zipped framework of all frameworks" {
+    run carthage build --no-skip-current
+    [ "$status" -eq 0 ]
+    run carthage archive
+    [ "$status" -eq 0 ]
+    [ -e Ra.framework.zip ]
```
mdiep: Ideally, this would test a repository that has multiple frameworks and verify the contents of the `.zip`.

Source/CarthageIntegration/update.bats
```diff
@@ -0,0 +1,19 @@
...
+teardown() {
+    cd $BATS_TEST_DIRNAME
+}
+
+@test "carthage update builds everything" {
+    run carthage update --platform mac --no-use-binaries
+    [ "$status" -eq 0 ]
+    [ -e Carthage/Build/Mac/Quick.framework ]
+    [ -e Carthage/Build/Mac/Nimble.framework ]
+}
```
mdiep: Ideally this would build something that has (1) recursive dependencies and (2) both public and private dependencies.

mdiep:
This is great! I left a couple notes about how I think these could be improved, but they're a great start.

I think we could also test `checkout` (with and without submodules), `bootstrap`, and `build`. The only catch is that we need to make sure any building doesn't take too long because Travis limits build times.

younata:
> Ideally, this would test a repository that has multiple frameworks and verify the contents of the .zip.
> Ideally this would build something that has (1) recursive dependencies and (2) both public and private dependencies.
> I think we could also test checkout (with and without submodules), bootstrap, and build.

These are all very good ideas. I'll get to them sometime next week (I'm moving this weekend, and took today off to pack).

PR #1143, opened 2016-02-20
Added support for derivedDataPath.

bartoszj:
Hi,

I've added support for `-derivedDataPath` option for `xcodebuild`. With this option it would be possible to separate derived data folders between e.g. Jenkins jobs, or users could select own derived data folder and remove it after the frameworks have been built.


Source/carthage/Build.swift
```diff
@@ -17,23 +17,25 @@ public struct BuildCommand: CommandType {
...
-			} } } } } }
+				return self.init(configuration: configuration, buildPlatform: buildPlatform, derivedDataPath: derivedDataPath, skipCurrent: skipCurrent, colorOptions: colorOptions, verbose: verbose, directoryPath: directoryPath, dependenciesToBuild: dependenciesToBuild)
+			} } } } } } }
 		}
 
 		public static func evaluate(m: CommandMode) -> Result<Options, CommandantError<CarthageError>> {
 			return create
 				<*> m <| Option(key: "configuration", defaultValue: "Release", usage: "the Xcode configuration to build")
 				<*> m <| Option(key: "platform", defaultValue: .All, usage: "the platforms to build for (one of ‘all’, ‘Mac’, ‘iOS’, ‘watchOS’, 'tvOS', or comma-separated values of the formers except for ‘all’)")
+				<*> m <| Option(key: "derived-data", defaultValue: "", usage: "path to the custom derived data folder")
```
mdiep: This default value should be `nil` and `derivedDataPath` should be `String?`.

Source/carthage/Archive.swift
```diff
@@ -16,19 +16,21 @@ public struct ArchiveCommand: CommandType {
...
+			return { directoryPath in { derivedDataPath in { colorOptions in { frameworkNames in
+				return self.init(outputPath: outputPath, directoryPath: directoryPath, derivedDataPath: derivedDataPath, colorOptions: colorOptions, frameworkNames: frameworkNames)
+			} } } }
 		}
 
 		public static func evaluate(m: CommandMode) -> Result<Options, CommandantError<CarthageError>> {
 			return create
 				<*> m <| Option(key: "output", defaultValue: "", usage: "the path at which to create the zip file (or blank to infer it from the first one of the framework names)")
 				<*> m <| Option(key: "project-directory", defaultValue: NSFileManager.defaultManager().currentDirectoryPath, usage: "the directory containing the Carthage project")
+				<*> m <| Option(key: "derived-data", defaultValue: "", usage: "path to the custom derived data folder")
```
mdiep: It doesn't seem like this should be an argument to `carthage archive`. I don't see how it would affect the behavior.

Source/CarthageKit/Xcode.swift
```diff
@@ -171,14 +171,14 @@ public func schemesInProject(project: ProjectLocator) -> SignalProducer<String,
 
 /// Finds schemes of projects or workspaces, which Carthage should build, found
 /// within the given directory.
-public func buildableSchemesInDirectory(directoryURL: NSURL, withConfiguration configuration: String, forPlatforms platforms: Set<Platform> = []) -> SignalProducer<(ProjectLocator, [String]), CarthageError> {
+public func buildableSchemesInDirectory(directoryURL: NSURL, withConfiguration configuration: String, forPlatforms platforms: Set<Platform> = [], derivedDataPath: String? = nil) -> SignalProducer<(ProjectLocator, [String]), CarthageError> {
```
mdiep: I don't believe this function should take `derivedDataPath`—that shouldn't affect the output.

mdiep:
Thanks for the pull request!

I left a few notes that will require some changes. Leave a comment when you've addressed them and I'll take another look. :relaxed:

Source/carthage/Build.swift
bartoszj: When I use `nil` and `String` I receive an error:

```
Missing argument for --derived-data
```

According to the specification: 
> If this is nil, this option is always required.

Is there any workaround for this, or maybe I'm doing something wrong?

Source/carthage/Archive.swift
bartoszj: Yes, I messed this up. It will be removed in next commit.

Source/CarthageKit/Xcode.swift
bartoszj: Yes, next thing that I shouldn't change.

bartoszj:
@mdiep 
I've made some changes. Unfortunately I'm not very familiar with the code and I've no idea hot tow fix the default `nil` value for `derived-data`. If you can give me a hint, maybe I'll be able to fix it.

Source/carthage/Build.swift
mdiep: I'll take another look and get back to you. Sorry!

mdiep: As it stands, Commandant doesn't support that. I've [opened a PR](https://github.com/Carthage/Commandant/pull/58) that would make it possible. Let's give it a few days to see what comes of that.

bartoszj:
I've updated my code to use `Commandant` 0.9 and the optional values.

mdiep:
That's great! Thanks for updating it and for being patient!

PR #1149, opened 2016-02-22
Create a Network file to handle all network requests

younata:
Continuing https://github.com/Carthage/Carthage/pull/1136

This is a work in progress. I've spent about 3 hours on this so far, mostly trying to get OHHTTPStubs working (and being reminded why I hate trying to mock out at the URL Loading level). See [this gist](https://gist.github.com/younata/332fe5f7ae055b557738) for where I was going with that.

With regards to mocking the Foundation URL Loading system, I'm throwing my hands up. We'd gain some assurance guarding against NSURLSession changing if we go ahead and mock out at the URL Loading level, but I don't think it's worth doing - we'd still have to inject the NSURLSession (Surprise! `NSURLSession.sharedSession()` doesn't work with custom `NSURLProtocol`s!), and the little bit of extra guard we'd get over using `FakeURLSession` isn't worth the extra effort, IMO.

Feel free to tell me what I did wrong when I failed to properly mock out NSURLSession.


younata:
Taking the [WIP] flag off, assuming that travis agrees that this is good.

I encapsulated the URLSession behind a NetworkClient protocol/struct, and then proceeded to inject that uses a NetworkClient, down from Main, where URLSessionNetworkClient is first init'd. Doing this will make further testing out the `Project` struct down the line much easier.

In addition, I got rid of the FakeURLSession that mocked out the NSURLSession, in favor of using OHHTTPStubs as recommended previously. We still have to inject NSURLSession, because custom URL protocols aren't loaded into `NSURLSession.sharedSession()`, in test, I elected to inject one created with an ephemeral configuration to try to prevent any kind of caching behavior that is undesirable in test.

Next step is to start to test out much more of Project.swift. I'd love to get rid of `downloadAsset()` in GitHub.swift (which is only ever called from a private function in Project.swift), but I want to keep the tests around that (Which requires Project.swift to have tests backfilled).

@mdiep thoughts? (Anyone else?)

Source/CarthageKit/GitHub.swift
```diff
@@ -424,16 +424,17 @@ internal func loadGitHubAuthorization(forServer server: GitHubRepository.Server)
 
 /// Creates a request to fetch the given GitHub URL, optionally authenticating
 /// with the given credentials and content type.
-private func createGitHubRequest(URL: NSURL, _ authorizationHeaderValue: String?, contentType: String = APIContentType) -> NSURLRequest {
-	let request = NSMutableURLRequest(URL: URL)
-	request.setValue(contentType, forHTTPHeaderField: "Accept")
-	request.setValue(userAgent, forHTTPHeaderField: "User-Agent")
+public func createGitHubRequest(URL: NSURL, _ authorizationHeaderValue: String?, contentType: String = APIContentType) -> NSURLRequest {
```
mdiep: I don't think this should be `public`.

Source/CarthageKit/NetworkClient.swift
```diff
@@ -0,0 +1,54 @@
+import ReactiveCocoa
+
+
+public func createURLRequest(URL: NSURL, _ headers: [String: String] = [:]) -> NSURLRequest {
+	let urlRequest = NSMutableURLRequest(URL: URL)
+	urlRequest.allHTTPHeaderFields = headers
```
mdiep: Is this equivalent to what we had before? I'm not sure whether there were any default header fields that we were using.

Source/CarthageKitTests/Fakes/FakeNetworkClient.swift
```diff
@@ -0,0 +1,51 @@
+import CarthageKit
+import ReactiveCocoa
+
+// this file was generated by Xcode-Better-Refactor-Tools
+// https://github.com/tjarratt/xcode-better-refactor-tools
```
mdiep: Iiiiiinteresting.

Source/CarthageKit/GitHub.swift
```diff
@@ -463,64 +464,12 @@ private func parseLinkHeader(linkValue: String) -> [(NSURL, [String])] {
 	}
 }
 
-/// Fetches the given GitHub URL, automatically paginating to the end.
-///
-/// Returns a signal that will send one `NSData` for each page fetched.
-private func fetchAllPages(URL: NSURL, _ authorizationHeaderValue: String?, _ urlSession: NSURLSession) -> SignalProducer<NSData, CarthageError> {
```
mdiep: It looks like this functionality has been removed from this PR. :confused:

mdiep:
Sorry it took me so long to review this!

I left a few notes. The `fetchAllPages` bit is a pretty big deal. If we don't handle paginated responses, we will break some `Cartfile`s.

I'm also interested in breaking the GitHub-specific functionality out into a separate framework. I will take a look at that today and maybe set up a project to house it.

Source/CarthageKit/GitHub.swift
mdiep: Maybe we don't actually need this? I guess our only API usage right now is to fetch the release for a given tag, which isn't a paginated response. :flushed:

Source/CarthageKit/NetworkClient.swift
younata: Yes. But, if that's a fear, then it easily be changed to

```swift
for (key, value) in headers {
    urlRequest.setValue(value, forHTTPHeaderField: key)
}
```

Source/CarthageKit/GitHub.swift
younata: I agree. I think at the time, my reasoning was that I'd make it public, test it, then integrate that with the other tests for `releaseForTag`, and `downloadAsset` - which I did - but I never did the follow-up of "delete the tests for `createGitHubRequest` and make it private again".

Source/CarthageKit/NetworkClient.swift
mdiep: As long as you explicitly tested that the 2 are equivalent, I prefer what you wrote in the PR.

younata: I'll add a test to explicitly ensure this.

mdiep:
@younata Is there anything here that makes sense to keep after #1197?

Sorry to obsolete the work you did here! I hope you'll continue refactoring and increasing our test coverage.

mdiep:
I'm going to close this for now. :disappointed:

Issue #1145, opened 2016-02-22
Carthage checkout fails for dependency with submodules after #1125


erichoracek:
As of #1125, I'm seeing the following failure when attempting to `checkout` a repo that has submodules:

```
A shell task failed with exit code 128:
fatal: Not a valid object name <dependency tag>
```

This is due to the changes introduced in #1125, which uses the output of `git rev-parse --show-toplevel` in place of the project URL passed as a parameter to `submodulesInRepository(_:revision:)`.

On my machine, when I run `git rev-parse --show-toplevel` in any of the repos in `~/Library/Caches/org.carthage.CarthageKit/dependencies/`, I get a newline as the only output (no absolute path). As such, when the `submoduleSHAForPath(_:_:revision:)` method is called with an empty string, I get the above shell task failure, since Carthage is attempting to run
```
git ls-tree -z <dependency tag> Carthage/Checkouts/Dependency
```
in the current working directory, where no such tag exists.

To reproduce this locally, create a `Cartfile`:
```
github "Mantle/Mantle"
```
And run `carthage bootstrap --no-use-binaries` with latest.

On my machine, I'm using git 2.7.1. This reproduces regardless of whether the repo exists in the cache before the checkout begins.

cc @NachoSoto @bhargavg

bhargavg:
Hi @erichoracek, the issue is that, when running `carthage bootstrap --no-use-binaries`, the repositories are cloned as `bare` in `~/Library/Caches/org.carthage.CarthageKit/dependencies/` directory.

The `git rev-parse --show-toplevel` command will fail inside `bare` repositories (my bad).

Will raise a PR that fixes this.

PR #1150, opened 2016-02-23
Check for bare repo when running --show-toplevel

bhargavg:
As per Issue #1145, `rev-parse --show-toplevel` will fail (return empty path) when ran inside a bare repo, which caused `gitRootDirectoryForRepository` to return invalid root path (empty string)

Modified `gitRootDirectoryForRepository` to return repo path as it is if its bare, else return what `rev-parse --show-toplevel` says.

I've tested this with `github thoughtbot/Argo` and `github antitypical/Result`, working fine in both cases. 

But, for `github Mantle/Mantle` it is picking very old versions of `Nimble`, causing build failure.

Could somebody please confirm?

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -404,11 +404,21 @@ internal func gitmodulesEntriesInRepository(repositoryFileURL: NSURL, revision: 
 }
 
 /// Returns the root directory of the given repository
+///
+/// If in bare repository, return the passed repo path as the root
+/// else, return the path given by "git rev-parse --show-toplevel"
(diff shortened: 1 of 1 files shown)


erichoracek:
A Cartfile with Mantle as its only dependency is working for me with this change (no build failures). Thanks!

mdiep:
Looks good to me. Is there something else you think needs to be done? (This is marked as WIP.)

> But, for github Mantle/Mantle it is picking very old versions of Nimble, causing build failure.

Are you seeing https://github.com/Quick/Nimble/issues/213, perhaps?

bhargavg:
I was travelling, so, couldn't reply, sorry guys.

I added [WIP] because I was getting build failure, especially for Mantle. I realized now that it is because of a problem at my end and not with the carthage.

This can be merged @mdiep

Issue #1148, opened 2016-02-22
How to detect if running under Carthage


schwa:
Is there any environment variable we could use to detect if a build is running under Carthage?

I checked docs and couldn't find anything.

I'd like to modify the behavior of carthage frameworks builds - in this case I want to disable swiftlint from running inside a carthage module (due to bugs in swiftlint).

mdiep:
I don't think there's anything currently, but it sounds like a good idea. A pull request would be very welcome.

PR #1151, opened 2016-02-23
Add `CARTHAGE` build setting to xcodebuild arguments

Ashton-W:
Adding a build setting `CARTHAGE` makes it possible for framework projects to detect when they are being built by Carthage, as requested in #1148


mdiep:
@schwa Would this meet your needs?

schwa:
Looks great

mdiep:
Looks great. Thanks!

phatblat:
Very nice

PR #1158, opened 2016-02-26
Fix building with transitive dependencies

erichoracek:
Currently, deep transitive dependencies are not built first before a dependency when it is specified to the build command. This occurs when you specify a specific set of dependencies to the "build" command.

An example would be `carthage build A`, where `A` depends on `B` depends on `C`. I have observed that `C` is not always built before `B`.

### DIFF
--- a/Source/CarthageKit/Algorithms.swift
+++ b/Source/CarthageKit/Algorithms.swift
@@ -69,28 +69,34 @@ public func topologicalSort<Node: Comparable>(graph: Dictionary<Node, Set<Node>>
 /// Performs a topological sort on the provided graph with its output sorted to
 /// include only the provided set of nodes and their transitively incoming 
 /// nodes (dependencies).
+///
+/// Returns nil if the provided node is contained within the provided graph or
+/// if the provided graph has a cycle.
 public func topologicalSort<Node: Comparable>(graph: Dictionary<Node, Set<Node>>, nodes: Set<Node>) -> [Node]? {
-	if nodes.isEmpty { return topologicalSort(graph) }
+	guard !nodes.isEmpty else { return topologicalSort(graph) }
 
 	guard nodes.isSubsetOf(Set(graph.keys)) else { return nil }
 
 	// Ensure that the graph has no cycles, otherwise determining the set of 
-	// incoming nodes could infinitely recurse.
+	// transitive incoming nodes could infinitely recurse.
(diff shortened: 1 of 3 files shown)


Source/CarthageKit/Algorithms.swift
```diff
@@ -65,3 +65,38 @@ public func topologicalSort<Node: Comparable>(graph: Dictionary<Node, Set<Node>>
...
+/// Performs a topological sort on the provided graph with its output sorted to
+/// include only the provided set of nodes and their transitively incoming 
+/// nodes (dependencies).
+///
+/// Returns nil if any of the provided nodes are not contained within the
+/// provided graph or if the provided graph has a cycle.
+public func topologicalSort<Node: Comparable>(graph: Dictionary<Node, Set<Node>>, nodes: Set<Node>) -> [Node]? {
+	guard !nodes.isEmpty else { return topologicalSort(graph) }
+
+	guard nodes.isSubsetOf(Set(graph.keys)) else { return nil }
```
mdiep: It seems like this should be a `precondition` instead of a `guard`. It should indicate an error in Carthage's logic, rather than an error in `Cartfile`s, I think.

Source/CarthageKit/Algorithms.swift
```diff
@@ -65,3 +65,38 @@ public func topologicalSort<Node: Comparable>(graph: Dictionary<Node, Set<Node>>
...
+	// transitive incoming nodes could infinitely recurse.
+	guard let _ = topologicalSort(graph) else { return nil }
+
+	let relevantNodes = Set(nodes.flatMap { Set([$0]).union(transitiveIncomingNodes(graph, node: $0)) })
+	let irrelevantNodes = Set(graph.keys).subtract(relevantNodes)
+
+	var filteredGraph = graph
+	irrelevantNodes.forEach { node in filteredGraph.removeValueForKey(node) }
+	
+	return topologicalSort(filteredGraph)
```
mdiep: It doesn't seem like we should need to sort again. Can't we take the result of the earlier call to `topologicalSort` and remove the `irrelaventNodes` from the result?

mdiep:
I just realized that this never got reviewed/merged. :scream_cat: Sorry about that!

This is great. :zap: There are just a couple things I would change.

mdiep:
:sparkles:

PR #1174, opened 2016-03-04
Fixes #1172, duplicate dependency entries

### LINKED ISSUE #1172: 0.15 inserts duplicate entries with different versions in Cartfile.resolved
This did not happen in 0.14.

bhargavg:
As reported in #1172 , `carthage` is treating
`github "Owner/Repo"` and
`github "owner/repo"`
as two different dependencies causing duplicated entries in `Cartfile.resolved`.

Fixed it by adding case-insensitive hash generation in `GitHubRepository`


mdiep:
Good catch! Thanks!

PR #1177, opened 2016-03-08
[README] Update "Adding frameworks to unit tests or a framework" section

kumapo:
- iOS application targets now have the "Linked Frameworks and Libraries" section in their “General” settings

- Only unit test targets are missing the "Linked Frameworks and Libraries" section


kumapo:
Screenshot of an iOS application target:


A framework target screenshot:


Both of these now have "Linked Frameworks and Libraries" section.

But, a unit test target screenshot:


Only unit test targets are missing the section.

(All of screenshots are taken on Xcode 7.2.1 7C1002)

mdiep:
Looks fine to me. :+1: Thanks!

Issue #1079, opened 2016-02-01
copy-frameworks build phase fails with Carthage 0.12


dcaunt:
Installing from pkg attached to release.

Build fails in the Run Script Phase `/usr/local/bin/carthage copy-frameworks` with `Unrecognized arguments`, presumably since Commandant introduced error handling for unused/unrecognized args.

If I reinstall 0.11 the build passes. I'll try and debug this more when I have some free time but in the meantime is there any additional information I can provide to help figure this out?

ikesyo:
I can't reproduce it. Did you previously install carthage by homebrew? If so, running `brew uninstall carthage` would help?

dcaunt:
I have been installing from the pkg for the last few versions and `brew` says it's not installed. I even tried removing Carthage by deleting `/Library/Frameworks/CarthageKit.framework` and  `/usr/local/bin/carthage` before re-installing 0.12.

Is there a way to see the invocation of copy-frameworks without doing a custom build of Carthage?

jamieathans:
I can verify this error as well with 0.12.

I do install carthage via homebrew and running `brew uninstall carthage` before upgrading did not seem to help.

Reverted to 0.11 and build passes as stated by **dcaunt**.

dcaunt:
Still haven't figured this out yet, but my colleague tried with the same project and 0.12 works on their machine.

askielboe:
I think this has to do with how carthage 0.12 parses command line arguments.

Seems like the problem is that shell scripts in xcode are called with a shebang `#!`, which passes the path of the script as an argument to carthage.

No knowing what to do with the path, carthage fails with `Unrecognized arguments`.

ikesyo:
> Seems like the problem is that shell scripts in xcode are called with a shebang `#!`, which passes the path of the script as an argument to carthage.

I'm not sure what you are saying about. What is the contents of your script of the `Run Script` phase and the entire error message carthage outputs?

askielboe:
The contents of the temporary script Script.sh is:

    #!/usr/local/bin/carthage copy-frameworks

The error I get when I run it using `sh -c ./Script.sh` is:

    Unrecognized arguments: "./Script.sh"

Prior to v0.12 I would instead get as output (if run outside Xcode):

    Environment variable not set: SCRIPT_INPUT_FILE_COUNT

My point is that `#!` in Script.sh passes the path of the script as an argument to carthage. You can verify this by putting the following line in a Script.sh file:

[...]

PR #1190, opened 2016-03-15
Clarify where build script should be added

leppert:
This is intended to address https://github.com/Carthage/Carthage/issues/1079


mdiep:
Thanks!

PR #1193, opened 2016-03-18
Fix integration tests

### LINKED ISSUE #1184: Integration tests are failing
It seems to be because of changes to Ra. We should checkout a specific known revision, instead of relying on `master`.

https://travis-ci.org/Carthage/Carthage/builds/114789534

/cc @younata

ikesyo:
Fixes #1184.


mdiep:
Thanks! It's good to see another contribution from you! :smile:

PR #1202, opened 2016-03-22
Correctly fetch repository when the specified git object is a branch

### LINKED ISSUE #1191: Carthage doesn't read latest commit in local repository
Basically a general case for the corner-case https://github.com/Carthage/Carthage/issues/968.  Carthage is not reading the latest commit in a local git repository, and instead is repeatedly checking-out the commit which was available when first pulled.  I see this happening even without the *relative paths* condition.

I've tried deleting the Carthage.resolved, and deleting the entire Carthage folder in the consumer project, with no effect.

I've tried varying the branch tracking tag as the following, also with no effect:
```
git "file:///Users/i826181/xCodeProjects/myframework" "master"
git "file:///Users/i826181/xCodeProjects/myframework" "HEAD"
git "file:///Users/i826181/xCodeProjects/myframework" "head"
```

It seems that the only workaround is to delete the Carthage cache after each commit https://github.com/Carthage/Carthage/issues/968#issuecomment-197360787.

ikesyo:
Fixes #1191.

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -434,28 +434,39 @@ public func submodulesInRepository(repositoryFileURL: NSURL, revision: String = 
 		}
 }
 
+/// Determines whether any reference exists for the given pattern in the given
+/// repository.
+///
+/// If the specified file URL does not represent a valid Git repository, `false`
+/// will be sent.
+public func referenceExistsInRepository(repositoryFileURL: NSURL, pattern: String) -> SignalProducer<Bool, NoError> {
+	return ensureDirectoryExistsAtURL(repositoryFileURL)
+		.then(launchGitTask([ "show-ref", pattern ], repositoryFileURL: repositoryFileURL))
+		.then(.init(value: true))
+		.flatMapError { _ in .init(value: false) }
+}
+
 /// Determines whether the specified revision identifies a valid commit.
 ///
 /// If the specified file URL does not represent a valid Git repository, `false`
 /// will be sent.
 public func commitExistsInRepository(repositoryFileURL: NSURL, revision: String = "HEAD") -> SignalProducer<Bool, NoError> {
-	return SignalProducer { observer, disposable in
-		// NSTask throws a hissy fit (a.k.a. exception) if the working directory
-		// doesn't exist, so pre-emptively check for that.
-		var isDirectory: ObjCBool = false
-		if !NSFileManager.defaultManager().fileExistsAtPath(repositoryFileURL.path!, isDirectory: &isDirectory) || !isDirectory {
-			observer.sendNext(false)
-			observer.sendCompleted()
-			return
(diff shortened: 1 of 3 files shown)


Source/CarthageKit/Git.swift
```diff
@@ -434,28 +434,39 @@ public func submodulesInRepository(repositoryFileURL: NSURL, revision: String =
 		}
 }
 
+/// Determines whether any reference exists for the given pattern in the given
+/// repository.
+///
+/// If the specified file URL does not represent a valid Git repository, `false`
+/// will be sent.
+public func referenceExistsInRepository(repositoryFileURL: NSURL, pattern: String) -> SignalProducer<Bool, NoError> {
```
mdiep: Can we make this `internal`? I'd like to default to that and only make it `public` if necessary or requested.

Source/CarthageKit/Project.swift
```diff
@@ -884,17 +884,20 @@ public func cloneOrFetchProject(project: ProjectIdentifier, preferHTTPS: Bool, c
 							return SignalProducer(value: (.Fetching(project), repositoryURL))
 								.concat(fetchRepository(repositoryURL, remoteURL: remoteURL, refspec: "+refs/heads/*:refs/heads/*").then(.empty))
 						}
+
 						// If we've already cloned the repo, check for the revision, possibly skipping an unnecessary fetch
 						if let commitish = commitish {
-							return commitExistsInRepository(repositoryURL, revision: commitish)
+							return referenceExistsInRepository(repositoryURL, pattern: commitish)
+								.zipWith(commitExistsInRepository(repositoryURL, revision: commitish))
```
mdiep: I'd prefer to use the `zip` function here instead. I think that reads more clearly. But I'll accept this version if you prefer it.

```swift
return zip(
        referenceExistsInRepository(repositoryURL, pattern: commits),
        commitExistsInRepository(repositoryURL, revision: commitish)
    )
```

mdiep:
I'm very glad to see this! :metal:

I left a couple minor notes. Is this something we could add a test for? If it's too much work, we can skip it.

ikesyo:
@mdiep I added test cases for `cloneOrFetchProject` and slightly modified the function (and `repositoryFileURLForProject` as well) to make it testable. :boom:

mdiep:
:clap::clap::clap::clap::clap:

schwa:
This is hurting our development (we're forced to use an explicit SHA instead of a branch).

Is this going to be released soon?

mdiep:
This should be released soon. We're unfortunately waiting for Swift 2.2 compatibility in all of our dependencies. But they seem to be almost ready.

In the meantime, you can also do `git fetch url-to-repository` to update the local copy of the repository.

mdiep:
This was just released with 0.16: Tentacular Future.

PR #1197, opened 2016-03-20
Use Tentacle for the GitHub API

mdiep:
1 thing to note: This changes the deployment target from 10.9 to 10.10. I think that's reasonable, given that using 10.9 for deployment is pretty infeasible.

PR #1212, opened 2016-03-25
Use QueueScheduler.init(qos:name:)

ikesyo:
As of #1197, the deployment target is 10.10 now. So we should replace the deprecated `QueueScheduler.init(queue:name:)` with `QueueScheduler.init(qos:name:)`.


mdiep:
:sparkles:

PR #1238, opened 2016-04-03
Swift 2.2

ikesyo:
- [x] ReactiveTask
- [x] Tentacle: https://github.com/mdiep/Tentacle/pull/23


ikesyo:
Hmm, `carthage update` crashes immediately :scream:

> Caught signal triggered by the Swift runtime!
> Segmentation fault
> 
> Unfortunately, this is probably a bug in Swift and not Carthage. If
> this is preventing you from doing work, please file an issue and we'll
> do our best to work around it:
> https://github.com/Carthage/Carthage/issues/new
> 
> Please also consider filing a radar with Apple, containing the version
> of Carthage and any crash report found in Console.app.
> 
> [1]    97676 segmentation fault   update

ikesyo:
https://github.com/Carthage/Carthage/pull/1238#issuecomment-204922560: This should be the same as https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2751 happening on `SignalProducerType.permuteWith`.

ikesyo:
It should be ready now. :zap:

mdiep:
Looks great! Thanks for handling this! ⚡⚡⚡

PR #1241, opened 2016-04-05
[copy-frameworks] Ignore frameworks that do not support the current build architecture

coenert:
I have a (dynamic)framework that wraps static libs, but these binaries are not build for the simulator. It turns out that lipo throws an error if the result is an empty binary, in the copy-frameworks command. In this pr i added a function `shouldIgnoreFramework` that checks if the framework does support any of the valid architectures and if that returns false the i print a warning and the framework will be ignored.

### DIFF
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -27,11 +27,19 @@ public struct CopyFrameworksCommand: CommandType {
 
 				return combineLatest(SignalProducer(result: source), SignalProducer(result: target), SignalProducer(result: validArchitectures()))
 					.flatMap(.Merge) { (source, target, validArchitectures) -> SignalProducer<(), CarthageError> in
-						let copyFrameworks = copyFramework(source, target: target, validArchitectures: validArchitectures)
-						let copydSYMs = copyDebugSymbolsForFramework(source, validArchitectures: validArchitectures)
-						return combineLatest(copyFrameworks, copydSYMs)
-							.then(.empty)
-					}
+						return shouldIgnoreFramework(source, validArchitectures: validArchitectures)
+							.flatMap(.Concat) { shouldIgnore -> SignalProducer<(), CarthageError> in
+								if shouldIgnore {
+									carthage.println("warning: Ignoring \(frameworkName) because it does not support the current architecture\n")
+									return .empty
+								} else {
+									let copyFrameworks = copyFramework(source, target: target, validArchitectures: validArchitectures)
+									let copydSYMs = copyDebugSymbolsForFramework(source, validArchitectures: validArchitectures)
+									return combineLatest(copyFrameworks, copydSYMs)
+										.then(.empty)
+								}
+						}
+				}
 			}
 			.waitOnCommand()
 	}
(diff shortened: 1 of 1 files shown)


mdiep:
Can you give a bit more detail to help me understand your use case?

coenert:
Sure, I use a commercial sdk that only provides static libraries(fat binaries), but i use Carthage in all my projects. And because Carthage only supports dynamic frameworks, the only way to support the sdk is to wrap the static library(with headers) in a dynamic framework.

But thats not the issue, the issue is that the fat binaries do not include the simulator architecture. You can solve this issue in Xcode by setting the supported platforms setting to `iphoneos` and then xcode will ignore the framework if you build for iphonesimulator(Of course, you have to use `#if os()` to avoid build errors).

Carthage, on the other hand, still tries to copy the framework. One step of copying the framework is stripping the architectures using lipo: But if the only valid architecture is x86_64 and your framework does not have symbols for that architecture lipo throws the error "-remove's specified would result in an empty fat file". So i thought a good solution would be to, like xcode, ignore the framework.

Source/carthage/CopyFrameworks.swift
```diff
@@ -27,11 +27,19 @@ public struct CopyFrameworksCommand: CommandType {
...
 					.flatMap(.Merge) { (source, target, validArchitectures) -> SignalProducer<(), CarthageError> in
-						let copyFrameworks = copyFramework(source, target: target, validArchitectures: validArchitectures)
-						let copydSYMs = copyDebugSymbolsForFramework(source, validArchitectures: validArchitectures)
-						return combineLatest(copyFrameworks, copydSYMs)
-							.then(.empty)
-					}
+						return shouldIgnoreFramework(source, validArchitectures: validArchitectures)
+							.flatMap(.Concat) { shouldIgnore -> SignalProducer<(), CarthageError> in
+								if shouldIgnore {
+									carthage.println("warning: Ignoring \(frameworkName) because it does not support the current architecture\n")
```
mdiep: This should really have the bullets prefix that we use almost everywhere else. Can you add it and removing the `warning: ` prefix?

mdiep:
Thanks! That's helpful. I'm 👍 to add this, but would like to adjust the printed message to match the rest of the Carthage style.

Source/carthage/CopyFrameworks.swift
```diff
@@ -51,6 +59,19 @@ private func copyFramework(source: NSURL, target: NSURL, validArchitectures: [St
...
+private func shouldIgnoreFramework(framework: NSURL, validArchitectures: [String]) -> SignalProducer<Bool, CarthageError> {
+	return architecturesInPackage(framework)
+		.collect()
+		.map { architectures in
+			// Return all the architectures, present in the framework, that are valid.
+			validArchitectures.filter(architectures.contains)
+		}
+		.map { remainingArchitectures in
+			// If removing the useless architectures results in an empty fat file, wat means that the framework does not have a binary for the given architecture, ignore the framework.
+			remainingArchitectures.count == 0
```
ikesyo: `isEmpty` is preferable.

Source/carthage/CopyFrameworks.swift
coenert: I use `warning:` because Xcode shows all printed lines that begin with `warning:` as  a warning:
screen shot 2016-04-13 at 09 14 00

Instead of(using Carthage formatting):
screen shot 2016-04-13 at 09 19 19

mdiep: 👍

PR #1251, opened 2016-04-09
[archive] Accept a directory path input for `--output` option

ikesyo:


Source/carthage/Archive.swift
```diff
@@ -108,14 +108,41 @@ public struct ArchiveCommand: CommandType {
...
 						carthage.println(formatting.bullets + "Created " + formatting.path(string: outputPath))
 					})
 				}
 		}
 		.waitOnCommand()
 	}
+
+	/// Returns an appropriate output file path for the resulting zip file using
+	/// the given option and frameworks.
+	private func outputPathWithOptions(options: Options, frameworks: [String]) -> String {
```
mdiep: It seems like this could be a free function.

mdiep:
👍 Merging this despite the minor note so I can include it in the new release.

PR #1256, opened 2016-04-11
Extract outputPathWithOptions into a free function

ikesyo:
Addresses https://github.com/Carthage/Carthage/pull/1251#discussion_r59134080. :sparkles:


mdiep:
✨

PR #1277, opened 2016-04-18
Fix CI steps in readme

ealeksandrov:
TravisCI updated their VMs and `brew install carthage` now fails with error: "already installed".
I changed this to `brew outdated carthage || brew upgrade carthage`, so it will try to update only if carthage is outdated and no error will be produced.


ikesyo:
Seems legit. 👍 Thanks for the pull request!

Issue #1292, opened 2016-05-04
carthage bootstrap silently fails when attempted for a package which does not yet exist in Cartfile.resolved


mz2:
I'm not sure if this is intentional behaviour, but `carthage bootstrap` silently fails when attempted for a package which does not yet exist in Cartfile.resolved. I can understand that it would not behave normally if for instance the principle is that `carthage bootstrap` should never as a side effect modify `Cartfile.resolved`. But at least it should then give me a clear error message that I should first do something else (like `carthage update`?).

I'm happy to resolve this myself if you tell me what (if anything) should be done about it, and some pointers of the places I am expected to change.

mdiep:
We definitely shouldn't fail silently. An error message that says (1) that the dependency doesn't exist in `Cartfile.resolved` and (2) you may want to use `update` seems totally in order.

A PR would be great! I think you'll want to start by looking at [`Bootstrap.swift`](https://github.com/Carthage/Carthage/blob/master/Source/carthage/Bootstrap.swift).

ikesyo:
Closed by #1294.

PR #1294, opened 2016-05-05
Fix to https://github.com/Carthage/Carthage/issues/1292

mz2:
I'm not very sure if this fits your house style and whether I've missed something obvious, but here's an attempted fix to #1292. 

Also some shortcuts taken to force unwrap where some suitable errors should be thrown instead (on failing to read a string from Cartfile.resolved or to parse it into a ResolvedCartfile), wasn't sure what the intended style for such error cases is.

### DIFF
--- a/Source/carthage/Bootstrap.swift
+++ b/Source/carthage/Bootstrap.swift
@@ -27,21 +27,34 @@ public struct BootstrapCommand: CommandType {
 					return project.updateDependencies(shouldCheckout: options.checkoutAfterUpdate)
 				}
 				
+				let checkDependencies: SignalProducer<(), CarthageError>
 				if let depsToUpdate = options.dependenciesToUpdate {
-					let resolvedCartfileContents = try! String(contentsOfURL: project.resolvedCartfileURL, encoding: NSUTF8StringEncoding)
-					let resolvedCartfile = ResolvedCartfile.fromString(resolvedCartfileContents).value!
-					
-					let resolvedDependencyNames = resolvedCartfile.dependencies.map { $0.project.name }
-					if !Set(depsToUpdate).subtract(resolvedDependencyNames).isEmpty {
-						return SignalProducer<(), CarthageError>(error: CarthageError.UnresolvedDependencies([String](unresolvedDependencyNames).sort())
+					checkDependencies =
+						project
+						.loadResolvedCartfile()
+						.flatMap(.Concat) { resolvedCartfile -> SignalProducer<(), CarthageError> in
+							let resolvedDependencyNames = resolvedCartfile.dependencies.map { $0.project.name }
+							let unresolvedDependencyNames = Set(depsToUpdate).subtract(resolvedDependencyNames)
+							
+							if !unresolvedDependencyNames.isEmpty {
+								return SignalProducer<(), CarthageError>(error:.UnresolvedDependencies([String](unresolvedDependencyNames).sort()))
+							}
+							
+							return .empty
+						}
 					}
+					else {
+						checkDependencies = .empty
+					}
+				
+				let checkoutDependencies: SignalProducer<(), CarthageError>
+				if options.checkoutAfterUpdate {
+					checkoutDependencies = project.checkoutResolvedDependencies(options.dependenciesToUpdate)
+				} else {
+					checkoutDependencies = .empty
 				}
 				
-				if options.checkoutAfterUpdate {
-					return project.checkoutResolvedDependencies(options.dependenciesToUpdate)
-				} else {
-					return .empty
-				}
+				return checkDependencies.then(checkoutDependencies)
 			}
 			.then(options.buildProducer)
 			.waitOnCommand()
--- a/Source/carthage/Bootstrap.swift
+++ b/Source/carthage/Bootstrap.swift
@@ -30,12 +30,13 @@ public struct BootstrapCommand: CommandType {
 				if let depsToUpdate = options.dependenciesToUpdate {
 					let resolvedCartfileContents = try! String(contentsOfURL: project.resolvedCartfileURL, encoding: NSUTF8StringEncoding)
 					let resolvedCartfile = ResolvedCartfile.fromString(resolvedCartfileContents).value!
-					let resolvedDependencyNames = Set<String>(resolvedCartfile.dependencies.map { $0.project.name })
 					
-					for dep in depsToUpdate {
-						if !resolvedDependencyNames.contains(dep) {
-							return SignalProducer<(), CarthageError>(error:CarthageError.UnresolvedDependency(dep))
-						}
+					let resolvedDependencyNames = Set<String>(resolvedCartfile.dependencies.map { $0.project.name })
+					let updatedDependencyNames = Set<String>(depsToUpdate)
+					let unresolvedDependencyNames = updatedDependencyNames.subtract(resolvedDependencyNames)
+					
+					if unresolvedDependencyNames.count > 0 {
+						return SignalProducer<(), CarthageError>(error: CarthageError.UnresolvedDependencies([String](unresolvedDependencyNames).sort()))
 					}
 				}
 				
--- a/Source/carthage/Bootstrap.swift
+++ b/Source/carthage/Bootstrap.swift
@@ -29,23 +29,21 @@ public struct BootstrapCommand: CommandType {
 				
 				let checkDependencies: SignalProducer<(), CarthageError>
 				if let depsToUpdate = options.dependenciesToUpdate {
-					checkDependencies =
-						project
+					checkDependencies = project
 						.loadResolvedCartfile()
 						.flatMap(.Concat) { resolvedCartfile -> SignalProducer<(), CarthageError> in
-							let resolvedDependencyNames = resolvedCartfile.dependencies.map { $0.project.name }
-							let unresolvedDependencyNames = Set(depsToUpdate).subtract(resolvedDependencyNames)
+							let resolvedDependencyNames = resolvedCartfile.dependencies.map { $0.project.name.lowercaseString }
+							let unresolvedDependencyNames = Set(depsToUpdate.map { $0.lowercaseString }).subtract(resolvedDependencyNames)
 							
 							if !unresolvedDependencyNames.isEmpty {
-								return SignalProducer<(), CarthageError>(error:.UnresolvedDependencies([String](unresolvedDependencyNames).sort()))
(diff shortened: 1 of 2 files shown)


Source/carthage/Bootstrap.swift
```diff
@@ -26,7 +26,19 @@ public struct BootstrapCommand: CommandType {
 					carthage.println(formatting.bullets + "No Cartfile.resolved found, updating dependencies")
 					return project.updateDependencies(shouldCheckout: options.checkoutAfterUpdate)
 				}
-
+				
+				if let depsToUpdate = options.dependenciesToUpdate {
+					let resolvedCartfileContents = try! String(contentsOfURL: project.resolvedCartfileURL, encoding: NSUTF8StringEncoding)
+					let resolvedCartfile = ResolvedCartfile.fromString(resolvedCartfileContents).value!
```
mdiep: You can use `project.loadResolvedCartfile()` to load the cartfile.

Source/carthage/Bootstrap.swift
```diff
@@ -26,7 +26,19 @@ public struct BootstrapCommand: CommandType {
...
+					let resolvedDependencyNames = Set<String>(resolvedCartfile.dependencies.map { $0.project.name })
+					
+					for dep in depsToUpdate {
+						if !resolvedDependencyNames.contains(dep) {
+							return SignalProducer<(), CarthageError>(error:CarthageError.UnresolvedDependency(dep))
+						}
+					}
```
mdiep: This would be a good place to use `Set.subtract`.

Source/carthage/Bootstrap.swift
```diff
@@ -26,7 +26,19 @@ public struct BootstrapCommand: CommandType {
...
-
+				
+				if let depsToUpdate = options.dependenciesToUpdate {
+					let resolvedCartfileContents = try! String(contentsOfURL: project.resolvedCartfileURL, encoding: NSUTF8StringEncoding)
+					let resolvedCartfile = ResolvedCartfile.fromString(resolvedCartfileContents).value!
+					let resolvedDependencyNames = Set<String>(resolvedCartfile.dependencies.map { $0.project.name })
+					
+					for dep in depsToUpdate {
+						if !resolvedDependencyNames.contains(dep) {
+							return SignalProducer<(), CarthageError>(error:CarthageError.UnresolvedDependency(dep))
```
mdiep: You should be able to drop the `CarthageError` in `CarthageError.UnresolvedDependency(dep)`. Also, there should be a space after the `:`.

mdiep: To do that, you'll need to create a `SignalProducer` and use `flatMap` to execute the code below this `if`.

mdiep:
I left some notes about what you have. Let me know if you need help. 😄

Source/carthage/Bootstrap.swift
mz2: I was aware of that actually but I'm new enough to ReactiveCocoa that I quite honestly didn't know how to use the signal producer based flow that would introduce into this existing command handler. Would need your help to show how to do this, am afraid.

Source/carthage/Bootstrap.swift
mz2: Good point. Addressed via f0a37c0

Source/carthage/Bootstrap.swift
mz2: I'm afraid I'm very new to ReactiveCocoa and will need some help with both this and https://github.com/Carthage/Carthage/pull/1294#discussion_r62280905 above. Thanks!

mz2:
Thanks for the review comments – I addressed the set operation issue. With the other two requested changes I'm going to need to refer back to yourself.

Source/carthage/Bootstrap.swift
mz2: Namely I could not figure out how to make adding that signal producer there to executing the `if options.checkoutAfterUpdate` part in the case of no unresolved dependencies having been detected.

Source/carthage/Bootstrap.swift
```diff
@@ -26,7 +26,20 @@ public struct BootstrapCommand: CommandType {
...
 				}
-
+				
+				if let depsToUpdate = options.dependenciesToUpdate {
+					let resolvedCartfileContents = try! String(contentsOfURL: project.resolvedCartfileURL, encoding: NSUTF8StringEncoding)
+					let resolvedCartfile = ResolvedCartfile.fromString(resolvedCartfileContents).value!
+					
+					let resolvedDependencyNames = Set<String>(resolvedCartfile.dependencies.map { $0.project.name })
+					let updatedDependencyNames = Set<String>(depsToUpdate)
+					let unresolvedDependencyNames = updatedDependencyNames.subtract(resolvedDependencyNames)
```
mdiep: This can be simplified a bit further because:

1. You shouldn't need to specify the value type for the set
2. The parameter to `Set.subtract' doesn't need to be a set
3. `updatedDependencyNames` doesn't need to be its own variable

Something like this should suffice:

```swift
let resolvedDependencyNames = resolvedCartfile.dependencies.map { $0.project.name }
if !Set(depsToUpdate).subtract(resolvedDependencyNames).isEmpty {
   …
}
```

Source/carthage/Bootstrap.swift
mdiep: There's a few ways to do it, but I'd probably do it like this:

```swift
let checkDependencies: SignalProducer<(), CarthageError>
if let depsToUpdate = options.dependenciesToUpdate {
    checkDependencies = project
        .loadRespolvedDependencies()
        .flatMap(.Concat) { cartfile in
            if … {
                return .error(…)
            } else {
                return .empty()
            }
        }
} else {
    checkDependencies = .empty()
}

let checkoutDependencies: SignalProducer<(), CarthageError>
if options.checkoutAfterUpdate {
    checkoutDependencies = project.checkoutResolvedDependencies(options.dependenciesToUpdate)
} else {
    checkoutDependencies = .empty
}

return checkDependencies.then(checkoutDependencies)
```

⚠️ _I wrote this directly into my browser, so it may have errors._

Source/carthage/Bootstrap.swift
mz2: Oh, Swift you spoil us with your sweet standard library! Improved this according to your suggestion.

Source/carthage/Bootstrap.swift
mz2: Thanks – adapted based on the above! Change pushed now.

Source/carthage/Bootstrap.swift
mz2: Fix to this pushed.

mz2:
@mdiep I've addressed your comments, I think.

Source/carthage/Bootstrap.swift
```diff
@@ -26,12 +26,35 @@ public struct BootstrapCommand: CommandType {
 					carthage.println(formatting.bullets + "No Cartfile.resolved found, updating dependencies")
 					return project.updateDependencies(shouldCheckout: options.checkoutAfterUpdate)
 				}
-
+				
+				let checkDependencies: SignalProducer<(), CarthageError>
+				if let depsToUpdate = options.dependenciesToUpdate {
+					checkDependencies =
+						project
```
mdiep: Our coding style would put this on the previous line after the `=`.

Source/carthage/Bootstrap.swift
```diff
@@ -26,12 +26,35 @@ public struct BootstrapCommand: CommandType {
...
+						.loadResolvedCartfile()
+						.flatMap(.Concat) { resolvedCartfile -> SignalProducer<(), CarthageError> in
+							let resolvedDependencyNames = resolvedCartfile.dependencies.map { $0.project.name }
+							let unresolvedDependencyNames = Set(depsToUpdate).subtract(resolvedDependencyNames)
+							
+							if !unresolvedDependencyNames.isEmpty {
+								return SignalProducer<(), CarthageError>(error:.UnresolvedDependencies([String](unresolvedDependencyNames).sort()))
```
mdiep: I don't think you should need to specify the types here. This seems like it should be enough. (Note that I added a space after the `:`.)

```swift
return SignalProducer(error: .UnresolvedDependencies(unresolvedDependencyNames.sort()))
```

Source/carthage/Bootstrap.swift
```diff
@@ -26,12 +26,35 @@ public struct BootstrapCommand: CommandType {
...
+							}
+							
+							return .empty
+						}
+					}
+					else {
+						checkDependencies = .empty
+					}
```
mdiep: The indentation of these last 4 lines off. They should be at the same level as the `if`.

Source/carthage/Bootstrap.swift
```diff
@@ -26,12 +26,35 @@ public struct BootstrapCommand: CommandType {
...
-
+				
+				let checkDependencies: SignalProducer<(), CarthageError>
+				if let depsToUpdate = options.dependenciesToUpdate {
+					checkDependencies =
+						project
+						.loadResolvedCartfile()
+						.flatMap(.Concat) { resolvedCartfile -> SignalProducer<(), CarthageError> in
+							let resolvedDependencyNames = resolvedCartfile.dependencies.map { $0.project.name }
+							let unresolvedDependencyNames = Set(depsToUpdate).subtract(resolvedDependencyNames)
```
mdiep: One thing we forgot to account for: case differences. Dependency names should be case insensitive.

mdiep:
The structure looks great! I left a few comments about code style things and one about a bug that we'll need to fix.

Source/carthage/Bootstrap.swift
mz2: Good point.

Source/carthage/Bootstrap.swift
mz2: Oops, good one. I was unsure about that actually, forgot to ask.

Source/carthage/Bootstrap.swift
mz2: Fixed.

Source/carthage/Bootstrap.swift
mz2: Done

mz2:
@mdiep I handled case insensitivity by lowercasing both operands to that set comparison, hope that's an OK solution? Other points should be addressed by d1770ba3cd5307eb90557145f21b4d28b63fb238 too.

mdiep:
This is great! Thanks!

Any chance you'd like to do this for the other commands that take a list of dependencies? ☺️

PR #1301, opened 2016-05-12
Update ReadMe to add note about @loader_path/Frameworks

coreyfloyd:
This path must be present for Carthage to load frameworks in unit test targets, but may not be present in older projects.

I discovered this while using Carthage to add libraries to the unit test target here:
https://github.com/wikimedia/wikipedia-ios/pull/695


mdiep:
Thanks!

PR #1302, opened 2016-05-12
Fixes issue with archive command trying to include xctest files

BernardGatt:
When running the archive command on projects with test targets that have "Run" enabled in the build scheme, as below:

Issue

This error is shown:
Could not find any copies of SnapKit OSX Tests.xctest, SnapKit iOS Tests.xctest, SnapKit.framework. Make sure you're in the project’s root and that the frameworks have already been built using 'carthage build --no-skip-current'.

Apart from SnapKit this issue is happening on PromiseKit as well.

This fix adds a filter to filter out anything that is not a framework.

### DIFF
--- a/Source/carthage/Archive.swift
+++ b/Source/carthage/Archive.swift
@@ -64,14 +64,14 @@ public struct ArchiveCommand: CommandType {
 					return BuildSettings.loadWithArguments(buildArguments)
 				}
 				.flatMap(.Concat) { settings -> SignalProducer<String, CarthageError> in
-					if let wrapperName = settings.wrapperName.value {
+					if let wrapperName = settings.wrapperName.value where settings.productType.value == .Framework {
 						return .init(value: wrapperName)
 					} else {
 						return .empty
 					}
 				}
 				.collect()
-				.map { Array(Set($0.filter{ $0.hasSuffix(".framework") })).sort() }
+				.map { Array(Set($0)).sort() }
 		}
 
 		return frameworks.flatMap(.Merge) { frameworks -> SignalProducer<(), CarthageError> in
--- a/Source/carthage/Archive.swift
+++ b/Source/carthage/Archive.swift
@@ -71,7 +71,7 @@ public struct ArchiveCommand: CommandType {
 					}
 				}
 				.collect()
-				.map { Array(Set($0)).sort() }
+				.map { Array(Set($0.filter{ $0.hasSuffix(".framework") })).sort() }
 		}
 
 		return frameworks.flatMap(.Merge) { frameworks -> SignalProducer<(), CarthageError> in


mdiep:
I'm either misunderstanding or unable to reproduce the issue. Can you share steps to reproduce it?

BernardGatt:
Clone [Snapkit](https://github.com/SnapKit/SnapKit) and run `carthage build --no-skip-current` and after that `carthage archive` the archive will fail because the archive will try to look for xctest files.

Source/carthage/Archive.swift
```diff
@@ -71,7 +71,7 @@ public struct ArchiveCommand: CommandType {
 					}
 				}
 				.collect()
-				.map { Array(Set($0)).sort() }
+				.map { Array(Set($0.filter{ $0.hasSuffix(".framework") })).sort() }
```
mdiep: Rather than use `filter` on the `Array`, you could use `filter` directly on the signal chain before the `collect()`.

```swift
.filter { $0.hasSuffix(".framework") }
.collect()
.map { Array(Set($0)).sort() }
```

But I think it'd be better to reuse the existing logic for whether a `BuildSettings` represents something that Carthage should build. There's a private `shouldBuildFrameworkType` function in `Xcode.swift` that we could make `internal`. 

Then this signal chain could use it and the `settings.frameworkType`—just like `shouldBuildScheme` does—to filter out settings that don't represent dynamic frameworks.

This will keep the code working in spite of unforeseen changes down the line and deduplicate logic.

mdiep:
That was helpful, thanks! And thank you for the PR!

We should definitely fix this, but I left a comment about how I think the implementation should be improved. Thanks!

Source/carthage/Archive.swift
ikesyo: > There's a private `shouldBuildFrameworkType` function in `Xcode.swift` that we could make `internal`. 

Since that is in `CarthageKit` module, and the code here is in `carthage` module, it should be `public` to reuse it.

mdiep: facepalm

JeanAzzopardi:
I updated the Pull Request to check the product type by using the productType enum instead of string comparison.

I took a look at the comment by @mdiep and the issue isn't occurring when calling buildableSchemesInDirectory. 

Instead, it occurs when checking for `settings.wrapperName` in [the lines below](https://github.com/Carthage/Carthage/blob/master/Source/carthage/Archive.swift#L66-L72):

```
.flatMap(.Concat) { settings -> SignalProducer<String, CarthageError> in
	if let wrapperName = settings.wrapperName.value {
		return .init(value: wrapperName)
	} else {
		return .empty
	}
}
```

mdiep:
Looks great! Thanks!

Issue #1039, opened 2016-01-13
Add build command option that allows specification of custom toolchain.


calebd:
`xcodebuild` has a `-toolchains` argument that can be used to specify a custom toolchain. This is handy now that Xcode 7.3 allows the use of open source Swift toolchains.

tkrajacic:
Is there actually a way of how to build using the open source snapshots with Carthage at the moment? Carthage doesn't seem to respect the environment variable for setting the toolchain it seems.

mdiep:
No, I think an argument would need to be added to `carthage` for that.

calebd:
Trying to wade through the code to build this. Currently you only either provide `TOOLCHAINS=swift` or you don't. There isn't much else to customize. I had initially thought there should be a `--toolchains swift` argument but now I think we should have a `--use-swift-toolchains` or other similarly named boolean argument instead.

mdiep:
We should probably follow the form of `xcodebuild` and use `--toolchain NAME`.

tkrajacic:
Is anyone working on this?

Carthage uses `xcodebuild` right?

Sorry if that's a stupid question but can't we just make `-toolchain` a passthrough argument then?

mdiep:
> Is anyone working on this?

Not that I know of.

> Carthage uses xcodebuild right?

Right.

> Sorry if that's a stupid question but can't we just make -toolchain a passthrough argument then?

Yes! Pull requests welcome. ☺️

tkrajacic:
It's my first pull request on any open source project, so be gentle ;)

I don't know much about _ReactiveCocoa_ btw, so it was a bit of a pain to follow flow of execution. Hope I didn't miss anything.

mdiep:
> It's my first pull request on any open source project, so be gentle ;)

That's great! Congratulations! 😄

PR #1303, opened 2016-05-12
Add `--toolchain` option

tkrajacic:
Adds the ability to specify a custom Swift toolchain via a `--toolchain` option on the `build`, `update`, and `archive` commands.

E.g.: `carthage build --toolchain org.swift.3020160509a`

**Todo**
- [x] Make toolchain parameter default to `nil` where desired
- [x] tests?

**Resolves issue** #1039

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -1084,7 +1084,7 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
 /// places its build product into the root directory given.
 ///
 /// Returns producers in the same format as buildInDirectory().
-public func buildDependencyProject(dependency: ProjectIdentifier, _ rootDirectoryURL: NSURL, withConfiguration configuration: String, platforms: Set<Platform> = [], toolchain: String?, derivedDataPath: String? = nil, sdkFilter: SDKFilterCallback = { .Success($0.0) }) -> SignalProducer<BuildSchemeProducer, CarthageError> {
+public func buildDependencyProject(dependency: ProjectIdentifier, _ rootDirectoryURL: NSURL, withConfiguration configuration: String, platforms: Set<Platform> = [], toolchain: String? = nil, derivedDataPath: String? = nil, sdkFilter: SDKFilterCallback = { .Success($0.0) }) -> SignalProducer<BuildSchemeProducer, CarthageError> {
 	let rootBinariesURL = rootDirectoryURL.URLByAppendingPathComponent(CarthageBinariesFolderPath, isDirectory: true).URLByResolvingSymlinksInPath!
 	let rawDependencyURL = rootDirectoryURL.URLByAppendingPathComponent(dependency.relativePath, isDirectory: true)
 	let dependencyURL = rawDependencyURL.URLByResolvingSymlinksInPath!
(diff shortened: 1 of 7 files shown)


tkrajacic:
Since I haven't updated the tests, and the function arguments require the `toolchain` parameter at the moment, ci fails of course.
I'll update that after initial feedback if the general approach is fine.

Source/CarthageKit/Xcode.swift
```diff
@@ -171,14 +171,14 @@ public func schemesInProject(project: ProjectLocator) -> SignalProducer<String,
...
-public func buildableSchemesInDirectory(directoryURL: NSURL, withConfiguration configuration: String, forPlatforms platforms: Set<Platform> = []) -> SignalProducer<(ProjectLocator, [String]), CarthageError> {
+public func buildableSchemesInDirectory(directoryURL: NSURL, withConfiguration configuration: String, forPlatforms platforms: Set<Platform> = [], toolchain: String?) -> SignalProducer<(ProjectLocator, [String]), CarthageError> {
 	precondition(directoryURL.fileURL)
 
 	return locateProjectsInDirectory(directoryURL)
 		.flatMap(.Concat) { project -> SignalProducer<(ProjectLocator, [String]), CarthageError> in
 			return schemesInProject(project)
 				.flatMap(.Merge) { scheme -> SignalProducer<String, CarthageError> in
-					let buildArguments = BuildArguments(project: project, scheme: scheme, configuration: configuration)
+					let buildArguments = BuildArguments(project: project, scheme: scheme, configuration: configuration, toolchain: toolchain)
```
mdiep: Is this actually needed? It doesn't seem like the toolchain would affect whether we'd build the scheme.

Source/carthage/Archive.swift
```diff
@@ -14,19 +14,21 @@ import ReactiveCocoa
...
-			} } }
+		static func create(outputPath: String?) -> String? -> String -> ColorOptions -> [String] -> Options {
+			return { toolchain in { directoryPath in { colorOptions in { frameworkNames in
+				return self.init(toolchain: toolchain, outputPath: outputPath, directoryPath: directoryPath, colorOptions: colorOptions, frameworkNames: frameworkNames)
+			} } } }
 		}
 
 		public static func evaluate(m: CommandMode) -> Result<Options, CommandantError<CarthageError>> {
 			return create
+				<*> m <| Option<String?>(key: "toolchain", defaultValue: nil, usage: "the toolchain to build with")
```
mdiep: `carthage archive` doesn't build anything, so it shouldn't need the `toolchain` option.

mdiep:
This is a great first step! ☺️

tkrajacic:
I removed the unnecessary `toolchain` parameters. They were remnants from when I was hunting down compile errors ;)
I also added a default `nil` to the `BuildArgument` constructor and removed all explicit `nil` passing.

tkrajacic:
I didn't add the default `nil` to `buildDependencyProject(...)` and `buildInDirectory(...)` because _CarthageKit_ always passes in the toolchain from a higher level command.

When adding an expectation to the _BuildArgumentsSpec_ though I saw that the test don't do that and call those methods directly.

Does it make sense to add the default `nil` value to the `toolchain` parameter as well on `buildDependencyProject(...)` and `buildInDirectory(...)`? In my mind it doesn't make sense to leave out the toolchain anywhere downstream because once it has been supplied as an argument it must be used everywhere (all build products need to be compiled with the same toolchain).

mdiep:
Let's add `nil` defaults to all the `public` functions that take a toolchain. That way we don't break compatibility for CarthageKit.

Also, I missed this yesterday, but `carthage bootstrap` needs a `--toolchain` argument as well.

This is shaping up great!

tkrajacic:
Doesn't `bootstrap` already work with all of `update`s commands?
It reuses the `UpdateCommand.Options`, and running `bootstrap --toolchain xxx` did work when I tried it.

mdiep:
> Doesn't bootstrap already work with all of updates commands?
> It reuses the UpdateCommand.Options, and running bootstrap --toolchain xxx did work when I tried it.

Oh, duh!

mdiep:
This is great! Thanks so much for the pull request! ✨

NachoSoto:
What's the expected value for this parameter? I'm trying "swift-DEVELOPMENT-SNAPSHOT-2016-09-06-a", as well as the full path, but Carthage is ignoring it and using the default toolchain.

tkrajacic:
For the tests I had `org.swift.3020160509a` but I don't immediately recall how I found this. (I am already in bed over here so I can't check right now)

I'll see what I can find in the morning.

ikesyo:
> What's the expected value for this parameter?

The `CFBundleIdentifier` value of `Info.plist` located at `/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2016-MM-dd-a.xctoolchain/`.

Issue #1328, opened 2016-06-01
Broken dependency graph.


calebd:
`carthage update --no-build` fails in different ways with different combinations of dependencies.

Some combinations cause it to show a "dependency cycle" error. Some combinations cause it to to say "no version satisfies the requirement". Some combinations cause it to select incorrect (out of date) versions.

Clearing any Carthage and Xcode state has no effect.
- Carthage version: 0.16.2
- Xcode version: 7.3.1

calebd:
This is possibly related to #1171.

JaviSoto:
Sharing the contents of your Cartfile if possible would be really useful to try to reproduce 👍

calebd:
@JaviSoto I know. It is fairly complex and almost entirely private.

calebd:
Well. I guess it can't hurt.

## App Cartfile

```
github "TTTAttributedLabel/TTTAttributedLabel"
github "soffes/SSPullToRefresh"
github "JonasGessner/JGProgressHUD"
github "venmo/Static" ~> 1.1.0
github "AFNetworking/AFNetworking" ~> 3.1.0

github "hodinkee/Everest"
github "hodinkee/Escapement"
github "hodinkee/Sterling" ~> 1.3.0
github "hodinkee/Syndicate" ~> 1.4.2
github "hodinkee/Boutique" ~> 1.4.4
github "hodinkee/Tourbillon" ~> 1.1.0
```

## Everest Cartfile

```
github "hodinkee/Iris" ~> 1.1
github "hodinkee/Alexander" ~> 3.1
github "ReactiveCocoa/ReactiveCocoa" ~> 4.1
```

## Escapement Cartfile

```
github "hodinkee/Alexander" ~> 3.1
```

## Sterling Cartfile

```
github "hodinkee/Alexander" ~> 3.1
github "ReactiveCocoa/ReactiveCocoa" ~> 4.1
github "hodinkee/Hyperdrive" ~> 0.3.0
```

## Syndicate Cartfile

```

github "AliSoftware/Dip" ~> 4.1
github "AFNetworking/AFNetworking" ~> 3.1

github "hodinkee/Everest"
github "hodinkee/Pneumatic" ~> 1.2
```

## Boutique Cartfile

```
github "AliSoftware/Dip" ~> 4.1
github "shopify/mobile-buy-sdk-ios" == 1.2.4

github "hodinkee/Everest" ~> 0.67
```

## Tourbillon Cartfile

```
[...]

PR #1335, opened 2016-06-06
Improve error messaging around malformed dependency graphs

erichoracek:
Ref #1328

This should hopefully help out with some confusion caused by malformed dependency graphs being reported as having cycles.


mdiep:
I think we should keep this error limited to cycles. I don't think Carthage logic errors should be user-facing errors.

PR #1344, opened 2016-06-14
Fix carthage update silently fails when attempted for a package which does not yet exist in Cartfile

ktakayama:
same as #1292 on `update` command.


mdiep:
This is great! Thanks! ✨

PR #1345, opened 2016-06-14
Add support for completion dependency name with zsh

ktakayama:
updated zsh completion file.
Thanks.


ktakayama:
I forgot to support branch name and more other cases.

PR #1346, opened 2016-06-14
Add support for completion dependency name with zsh

ktakayama:
Updated zsh completion file.
Thanks.

### DIFF
--- a/Source/Scripts/carthage-zsh-completion
+++ b/Source/Scripts/carthage-zsh-completion
@@ -18,6 +18,12 @@ __subcommands() {
     _values Commands $commands
 }
 
+__carthage_dependencies() {
+   local -a _dependencies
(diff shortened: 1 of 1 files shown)


Source/Scripts/carthage-zsh-completion
```diff
@@ -18,6 +18,12 @@ __subcommands() {
     _values Commands $commands
 }
 
+__carthage_dependencies() {
+   local -a _dependencies
+   _dependencies=( $(cat Cartfile | grep git | awk -F '"' '{print $2}' | sed -e 's/.*\///' | sed -e 's/\.git$//') )
```
mdiep: This `grep` should probably match only against the beginning of the line so that it doesn't match comments that contain `git`.

ktakayama: I didn't think about this. thaks!

ktakayama:
fix comment issue.

mdiep:
Looks good AFAICT. Thanks!

PR #1353, opened 2016-06-18
Refactor parts of Project.swift and Xcode.swift

mdiep:
I did this on the plane the other day. ☺️

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -176,7 +176,7 @@ public final class Project {
 		let cartfileURL = directoryURL.URLByAppendingPathComponent(CarthageProjectCartfilePath, isDirectory: false)
 		let privateCartfileURL = directoryURL.URLByAppendingPathComponent(CarthageProjectPrivateCartfilePath, isDirectory: false)
 
-		let isNoSuchFileError = { (error: CarthageError) -> Bool in
+		func isNoSuchFileError(error: CarthageError) -> Bool {
 			switch error {
 			case let .ReadFailed(_, underlyingError):
 				if let underlyingError = underlyingError {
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -176,7 +176,7 @@ public final class Project {
 		let cartfileURL = directoryURL.URLByAppendingPathComponent(CarthageProjectCartfilePath, isDirectory: false)
 		let privateCartfileURL = directoryURL.URLByAppendingPathComponent(CarthageProjectPrivateCartfilePath, isDirectory: false)
 
-		let isNoSuchFileError = { (error: CarthageError) -> Bool in
+		func isNoSuchFileError(error: CarthageError) -> Bool {
```
NachoSoto: Much better.

NachoSoto:
Awesome 🌟

NachoSoto:
you should simplify

Source/CarthageKit/Project.swift
```diff
@@ -189,31 +189,20 @@ public final class Project {
 				return false
 			}
 		}
-
-		let cartfile = SignalProducer.attempt {
-				return Cartfile.fromFile(cartfileURL)
-			}
-			.flatMapError { error -> SignalProducer<Cartfile, CarthageError> in
-				if isNoSuchFileError(error) && NSFileManager.defaultManager().fileExistsAtPath(privateCartfileURL.path!) {
```
ikesyo: Loading `Cartfile` should be error out if both `Cartfile` and `Cartfile.private` dont't exist, and the error is successfully ignored in [`buildProjectInDirectoryURL` if `--no-skip-current` option is specified](https://github.com/Carthage/Carthage/blob/0.17/Source/carthage/Build.swift#L143-L149).

Thus removing the latter condition breaks the option appeared in #1366.

Issue #1308, opened 2016-05-16
Add a solution for vending binary-only frameworks


mdiep:
Some commercial frameworks are vended as binaries only.

tcurdt:
This issue does not capture the intend very well. This original request has only little to do with commercial vs non-commercial. And maybe at least reference https://github.com/Carthage/Carthage/issues/722 ?

grangej:
We need to be able to support specifying direct path for a binary resource that works exactly as the way the github does exact we specify the full path to the binary. 

I have checked out romebuild etc this does NOT do what we need it to do, it only caches the whole module, so if we have to rebuild the module because it was updated we again have to build all the sub modules even though they were already cached.

This is maddening. It takes 20 mins for me to do a stupid simple change.

flovilmart:
@mdiep 

PROPOSED SPEC:

For the Cartfile:

`http "<URL>"`

- keyword `http` would indicate a direct download of the file a the URL
- URL should be enclosed in double quotes
- version, branches and other trailing chars after the URL should be considered faulty configuration
- the file hosted at the URL should be a zip file
- the zip structure should be produced by `carthage build --no-skip-current && carthage archive`
- the zip file should be uploaded to the host as is

- Ideally the download, cache, copy, extraction should be exactly the code in place for Github releases
- If the file is unavailable, the download fails or any other error occur, carthage should fail
- the commands `carthage update` and `carthage install` should 'dumbly' fetch the latest versions on the host. it's up to the framework provider to handle versioning of it's frameworks.

- Suggestions for a 'clean' hosting:

`http "https://my.domain.com/release/0.0.1-alpha/framework.zip"`
`http "https://my.domain.com/release/1.0.0/framework.zip"`
`http "http://my.domain.com/1.0.0/framework.zip"`

http and https should be supported 

Suggestions?

sstadelman:
@flovilmart This looks reasonable.  The `carthage archive` product would work.  

I would propose the alternate keyword `src`, vs. `http`.  `src` has the nice precedent in HTML of containing the URI of the required resource, be it a [script, image, or frame content](https://www.w3.org/TR/html4/index/attributes.html).  Web User Agents support both fully-qualified and relative URI variants.  In the Cartfile domain, a relative URI would be nonsensical, and it would be obvious to developers to use the fully-qualified URL.

That said, `http` is pretty intuitive as well.

Moved from #722 

> 
[...]

PR #1363, opened 2016-06-23
WIP: Adds support and tests for http keyword in cartfile (CarthageKit)

flovilmart:
proposed resolution for #1308

### DIFF
--- a/Source/CarthageKitTests/ResolverSpec.swift
+++ b/Source/CarthageKitTests/ResolverSpec.swift
@@ -53,12 +53,13 @@ class ResolverSpec: QuickSpec {
 			let testCartfile: Cartfile = self.loadTestCartfile("TestCartfile")
 			let producer = resolver.resolveDependenciesInCartfile(testCartfile)
 			let dependencies = self.orderedDependencies(producer)
-			expect(dependencies.count) == 8
+			expect(dependencies.count) == 9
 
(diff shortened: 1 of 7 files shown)


Source/CarthageKitTests/ResolverSpec.swift
```diff
@@ -53,12 +53,13 @@ class ResolverSpec: QuickSpec {
...
 			let producer = resolver.resolveDependenciesInCartfile(testCartfile)
 			let dependencies = self.orderedDependencies(producer)
-			expect(dependencies.count) == 8
+			expect(dependencies.count) == 9
 
 			var generator = dependencies.generate()
 
 			// Dependencies should be listed in build order.
 			expect(generator.next()) == Dependency("Mantle", "1.3.0")
+			expect(generator.next()) == Dependency("MyFramework", "3.0.0")
```
brentleyjones: Shouldn't this be "0.0.1" instead of "3.0.0"?

flovilmart: It should actually be nothing, this version gets resolved magically by some tests side effects that spits it as the highest.

As there is no plan yet to leverage a structured URL for those libraries, we can't really infer the version.

ikesyo:
Superseded by #1760.

Issue #1260, opened 2016-04-11
Should carthage outdated include incompatible dependencies?


dcaunt:
- `carthage version`: 0.16
- `xcodebuild -version`: Xcode 7.3 Build version 7D175
- Are you using `--no-build`? No
- Are you using `--no-use-binaries`? Yes
- Are you using `--use-submodules`? No

**Cartfile**

```
github "Quick/Quick" ~> 0.9.1
github "Quick/Nimble" ~> 3.2.0
```

**Carthage Output**

```
The following dependencies are outdated:
Quick "v0.9.1" -> "v0.9.2"
```

The Carthage `outdated` command introduced in #598 is great.

Do you think it should also listed dependencies which are newer, but not compatible with the version specified in the Cartfile? Given the above Cartfile, I'm not notified about the incompatible newer major version of Quick/Nimble which is 4.0.0, but chances are I'd be interested to know that it exists.

mdiep:
I definitely think `outdated` should include that information somehow. The current implementation was a great first step, but I think there's a lot of room for improvement.

jaybowong:
How's this going? 

The carthage folder is added to gitignore in our project, and in this way we can reduce the size of our git repo. We need to set the very specific versions in Cartfile to make sure every developer has the same code. It would be really great to know about new versions with `carthage outdated` as __expected__. It's especially useful when the number of dependencies is 10 or more.

mdiep:
> How's this going?

No one is working on it AFAIK. Pull requests welcome. 😄 

> We need to set the very specific versions in Cartfile to make sure every developer has the same code

This shouldn't matter. The `Cartfile.resolved` always has specific versions. The developers need only run `carthage bootstrap`.

ikesyo:
Resolved by #2254.

dcaunt:
Thanks @iv-mexx ! ✨

PR #1369, opened 2016-06-24
[WIP] Include incompatible dependencies in `outdated` output

mattprowse:
Proposed fix for #1260.


Source/CarthageKit/Project.swift
```diff
@@ -353,34 +353,77 @@ public final class Project {
...
+	/// them out into the project's working directory.
+	private func latestDependencies() -> SignalProducer<[Dependency<SemanticVersion>], CarthageError> {
+		return loadResolvedCartfile()
+			.map {
+				return SignalProducer<Dependency<PinnedVersion>, CarthageError>(values: $0.dependencies)
+					.map {dependency in
+						return dependency.project
+					}
+			}
+			.flatten(.Merge)
```
mdiep: This could use `flatMap` instead of `map` and `flatten`.

Source/CarthageKit/Project.swift
```diff
@@ -353,34 +353,77 @@ public final class Project {
...
+					.map {(versions: [SemanticVersion]) -> Dependency<SemanticVersion>? in
+						if let latestVersion = versions.maxElement() {
+							return Dependency(project: projectIdentifier, version: latestVersion)
+						} else {
+							return nil
+						}
+					}
+					.ignoreNil()
+			}
+			.flatten(.Merge)
```
mdiep: Same here.

Source/CarthageKit/Project.swift
```diff
@@ -353,34 +353,77 @@ public final class Project {
...
+	/// Attempts to determine the latest version (whether satisfiable or not)
+	/// of the project's Carthage dependencies.
+	///
+	/// This will fetch dependency repositories as necessary, but will not check
+	/// them out into the project's working directory.
+	private func latestDependencies() -> SignalProducer<[Dependency<SemanticVersion>], CarthageError> {
+		return loadResolvedCartfile()
+			.map {
+				return SignalProducer<Dependency<PinnedVersion>, CarthageError>(values: $0.dependencies)
+					.map {dependency in
```
mdiep: There should be a space after the `{` here.

Source/CarthageKit/Project.swift
```diff
@@ -353,34 +353,77 @@ public final class Project {
...
+					}
+			}
+			.flatten(.Merge)
+			.map { (projectIdentifier: ProjectIdentifier) -> SignalProducer<Dependency<SemanticVersion>, CarthageError> in
+				let versions = self.versionsForProject(projectIdentifier).collect().map { (pinnedVersions: [PinnedVersion]) -> [SemanticVersion] in
+					return pinnedVersions.flatMap { SemanticVersion.fromPinnedVersion($0).value }
+				}
+				
+				return versions
+					.map {(versions: [SemanticVersion]) -> Dependency<SemanticVersion>? in
```
mdiep: Here too.

Source/CarthageKit/Project.swift
```diff
@@ -353,34 +353,77 @@ public final class Project {
...
+		return loadResolvedCartfile()
+			.map {
+				return SignalProducer<Dependency<PinnedVersion>, CarthageError>(values: $0.dependencies)
+					.map {dependency in
+						return dependency.project
+					}
+			}
+			.flatten(.Merge)
+			.map { (projectIdentifier: ProjectIdentifier) -> SignalProducer<Dependency<SemanticVersion>, CarthageError> in
+				let versions = self.versionsForProject(projectIdentifier).collect().map { (pinnedVersions: [PinnedVersion]) -> [SemanticVersion] in
```
mdiep: This should be formatted according to our conventions, where each new SignalProducer operator is on a new line.

```swift
let versions = self.versionsForProject(projectIdentifier)
    .collect()
    .map { (pinnedVersions: [PinnedVersion]) -> [SemanticVersion] in
        ..
    }
```

Source/CarthageKit/Project.swift
```diff
@@ -353,34 +353,77 @@ public final class Project {
...
+					return pinnedVersions.flatMap { SemanticVersion.fromPinnedVersion($0).value }
+				}
+				
+				return versions
+					.map {(versions: [SemanticVersion]) -> Dependency<SemanticVersion>? in
+						if let latestVersion = versions.maxElement() {
+							return Dependency(project: projectIdentifier, version: latestVersion)
+						} else {
+							return nil
+						}
```
mdiep: This `if`/`else` could be replaced with a `map`.

Source/CarthageKit/Project.swift
mdiep: Should there be a `uniqueValues()` here or something to prevent a project from getting sent twice?

(If `A` and `B` both depend on `C`, I think it will get sent twice?)

mdiep:
I left a few notes. Comment when those are fixed up and I'll take another look.

I'm excited to have this!

mattprowse:
Thanks for the feedback!

I've fixed up the formatting issues, switched from using `map()` + `flatten()` to `flatMap()`, and restructured things to get rid of the if/else in `latestDependencies()`. The restructure of `latestDependencies()` also allowed me to get rid of some intermediate variables that were previously required to keep Xcode happy, so that's nice.

Regarding the necessity of `uniqueValues()`, I'm fairly certain this can be left out because the list of dependencies comes from `resolvedCartfile()` which (if I'm understanding things correctly) only returns each dependency once, because at this point the work of resolving dependencies has already been done. I did some quick testing and this seemed to be the case, though I'm obviously not as familiar with Carthage and it's possible that I'm wrong about this. If so, let me know and I'll add it in.

Source/CarthageKit/Project.swift
```diff
@@ -353,34 +353,71 @@ public final class Project {
...
+					.map { dependency in
+						return dependency.project
+					}
+			}
+			.flatMap(.Merge) { (projectIdentifier: ProjectIdentifier) -> SignalProducer<Dependency<SemanticVersion>, CarthageError> in
+				return self.versionsForProject(projectIdentifier)
+					.collect()
+					.map { (pinnedVersions: [PinnedVersion]) -> SemanticVersion? in
+						return pinnedVersions
+							.flatMap { SemanticVersion.fromPinnedVersion($0).value }
```
mdiep: I think it'd be simpler to map to `SemanticVersion`s before `collect()`ing.

```swift
return self.versionsForProject(projectIdentifier)
    .map { SemanticVersion.fromPinnedVersion($0).value }
    .ignoreNil()
    .collect()
    .map { $0.maxElement() }
    .ignoreNil()
    .map { Dependency(project: projectIdentifier, version: $0) }
```

Source/CarthageKit/Project.swift
```diff
@@ -353,34 +353,71 @@ public final class Project {
...
 					currentDependenciesDictionary[dependency.project] = dependency
 				}
-
-				return updatedDependencies.flatMap { updated -> OutdatedDependency? in
-					if let resolved = currentDependenciesDictionary[updated.project] where resolved.version != updated.version {
-						return (resolved, updated)
+				
+				var latestDependenciesDictionary = [ProjectIdentifier: SemanticDependency]()
+				for latestDependency in latestDependencies {
+					if let latestVersion = latestDependency.version.pinnedVersion, currentVersion = currentDependenciesDictionary[latestDependency.project] where currentVersion.version != latestVersion {
```
mdiep: Can you move this `where` clause to the `if` below inside `satisfiableDepndencies.flatMap`? I think that's a more readable approach.

mdiep:
Okay, there are two problems that I've discovered:

1. It doesn't work with dependencies that have no tagged versions.

2. It prints results for dependencies that are using a branch when there's nothing new on the branch. (Because it's not on the latest version.)

I think we need to fix those things up before we can release this.

But this is really neat!

mattprowse:
Changes made as requested! I also changed the variable name in the loop that builds `currentDependenciesDictionary`, just so that it's consistent with the other two loops in `outdatedDependencies()`.

I'll work on a fix for the two problems you found and comment again once I've got it done!

ikesyo:
Closing this in favor of #2254.

Issue #1366, opened 2016-06-23
`carthage build` for projects without dependencies requires Cartfile.resolved


nlutsenko:
* `carthage version`: 0.17
* `xcodebuild -version`: 7.3.1 (7D1014)
* Are you using `--no-build`? Yes
* Are you using `--no-use-binaries`? Not relevant.
* Are you using `--use-submodules`? Not relevant.

**Cartfile**
```
Doesn't exist :P
```

**Carthage Output**
```
Failed to read file or folder at /Users/nlutsenko/***/Cartfile.resolved: Error Domain=NSCocoaErrorDomain Code=260 "The file “Cartfile.resolved” couldn’t be opened because there is no such file." UserInfo={NSFilePath=/Users/nlutsenko/***/Cartfile.resolved, NSUnderlyingError=0x7f86c2837d80 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
```

Hey guys,

Looks like 0.17 now requires for Cartfile.resolved to exist at all times for `build` command, which is not quite ideal for projects that don't have dependencies.

If that's intentional - feel free to close this. If not - I would love to raise this concern, since this just blowed up few of our OSS projects at FB. E.g. https://github.com/facebook/SocketRocket

Repro steps on 0.17:
```
git clone git@github.com:facebook/SocketRocket.git
cd SocketRocket
carthage build --no-skip-current

*** Failed to read file or folder at ...
```

Behavior with 0.16.2:
```
git clone git@github.com:facebook/SocketRocket.git
cd SocketRocket
carthage build --no-skip-current

Building scheme "SocketRocket-iOS" in SocketRocket.xcodeproj
Building scheme "SocketRocket-tvOS" in SocketRocket.xcodeproj
Building scheme "SocketRocket-OSX" in SocketRocket.xcodeproj
```

ikesyo:
It seems that this is affected by the changes in #1353. Sorry for the inconvenience! :bow:

mdiep:
I'm so sorry! This is totally my fault.

I opened #1367 to revert that, and we'll get it merged and released tomorrow. (I'll also try to add a test soon so this doesn't happen again. But I don't want the test to hold up the fix.)

nlutsenko:
Awesome, guys, thanks for jumping on this so quickly. 
A test around this sounds great, since it will validate that this behavior doesn't break I. Future!

mdiep:
[...]

PR #1371, opened 2016-06-25
Add a test for #1366

mdiep:


Source/CarthageKitTests/ProjectSpec.swift
```diff
@@ -16,43 +16,69 @@ import Tentacle
 class ProjectSpec: QuickSpec {
 	override func spec() {
 		let directoryURL = NSBundle(forClass: self.dynamicType).URLForResource("CartfilePrivateOnly", withExtension: nil)!
```
ikesyo: Looks like this should be moved into `it("should load a combined Cartfile when only a Cartfile.private is present")` now.

ikesyo:
Looks good, just a minor note. 👍

mdiep:
Done! ✨

ikesyo:
🎉 🚢

PR #1377, opened 2016-06-29
Strip “Headers” directory in copy-frameworks

### LINKED ISSUE #1337: Strip framework headers during copy-frameworks`
We have some private/internal frameworks with headers that we would prefer are not copied alongside the framework binary into the app bundle. As far as I can tell, there is not reason why anyone would want to keep headers when embedding frameworks, since they provide no value to the application itself, and take up some extra space in the bundle as well. As such, should `carthage copy-frameworks` strip headers by default? If so, is this something that would be reasonable to attempt in a PR?

ikesyo:
Resolves #1337.


mdiep:
Awesome. ✨ 

This looks good. Just waiting on Travis to catch up. 👍

PR #1380, opened 2016-07-02
Strip PrivateHeaders directory from copied frameworks

erichoracek:
#1377 stoped copying the `Headers` directory, but didn't include the `PrivateHeaders` directory, which is still being copied.

Ref #1377 #1337


NachoSoto:
At this point I wonder if it'd be better to remove the concrete methods and have a `Set<String>` of directories to strip to avoid the duplication.

mdiep:
Thanks!

We can clean this separately.

Issue #1365, opened 2016-06-23
Unnecessary builds


alejandro-isaza-hs:
* `carthage version`: 0.16.2
* `xcodebuild -version`: 7.3.1

When executing `carthage build --platform iOS --verbose SDWebImage` carthage always cleans and builds from scratch:

```
$ carthage build --platform iOS --verbose SDWebImage 
*** Building scheme "WebImage" in SDWebImage.xcworkspace
/usr/bin/xcrun xcodebuild -workspace /Users/alejandroi/dev/HootsuiteComposer/Carthage/Checkouts/SDWebImage/SDWebImage.xcworkspace -scheme WebImage -configuration Release -sdk iphoneos ONLY_ACTIVE_ARCH=NO BITCODE_GENERATION_MODE=bitcode CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean build
```

Note the **clean**.

 For projects with many dependencies this means that running `carthage update` takes forever. A related ticket is #432 but it was closed for #459 which doesn't change the fact that carthage is executing a `clean` every time.

alejandro-isaza-hs:
This seems to be the culprit https://github.com/Carthage/Carthage/blob/242288351852a9c0118ce9befec95a40d2cc3c7a/Source/CarthageKit/Xcode.swift#L935

mdiep:
This was an intentional change made in #1028. You can read about the rationale there.

I don't think getting rid of `clean` is the right thing to do, but there are probably some other measures we can take. You might be interested in #1222.

alejandro-isaza-hs:
What about adding an option to skip cleaning if you know there is no folder changes. `--no-clean`?

mdiep:
That's a decent idea, but I think I'd rather see effort be put into #1222. I don't really want people to get in the habit of `--no-clean` since it's a little dangerous.

PR #1382, opened 2016-07-06
Add --no-clean option to update and build commands

mark-urbanthings:
Added to speed up build and update commands allows for incremental builds rather than a full clean and rebuild each time. Default behaviour is as before.

We are using carthage to manage a number of internal components and update frequently and hence the desire to have this option added as time for full rebuilds under Carthage is excessive when regular part of development process.


mdiep:
As discussed in #1365, I don't think this is the correct approach for us to take.

PR #1386, opened 2016-07-09
Fixes issue with archive command adding absolute path directories to zip

BernardGatt:
The latest release 0.17.2 fixed an issue with the archive command --project-directory option but it also introduced an issue that all the generated .zip archives include the absolute path to the .framework files.

To reproduce this issue on 0.17.2 just run the archive command and extract the .zip contents.


BernardGatt:
Can anyone look at this PR? It's very important as archiving is not working on the latest version.

mdiep:
Sorry, I've been mostly away from the computer over the weekend. This is on my list of things to get to.

Source/carthage/Archive.swift
```diff
@@ -103,6 +103,8 @@ public struct ArchiveCommand: CommandType {
 						.map { ($0 as NSString).lastPathComponent }
 						.filter { $0.hasSuffix(".framework") }
 
+					let relativePaths = paths.map { $0.stringByReplacingOccurrencesOfString(options.directoryPath, withString: ".") }
```
mdiep: This makes it so that we convert the relative paths to absolute paths above and then convert them back to relative paths here.

It would be better to pass the relative paths all the way through and convert them to absolute paths when necessary.

mdiep:
There's one note above that I think we should fix.

I'd also like to add a test for this if possible. Would you be willing to take a look at doing that? An integration test might be a suitable here. (We already have some set up in `Source/CarthageIntegration`.)

Source/carthage/Archive.swift
BernardGatt: Do you think this is a better solution?

carthage-archive-issue

BernardGatt:
An integration test would be perfect for this scenario, we could run 2 tests, 1 with --project-directory flag and one without and then check the contents of the zip.

I'll try and put some time on it ASAP but I really believe this PR should be merged as this issue is quite critical.

Source/carthage/Archive.swift
mdiep: Yup! Or we could pass both the relative and absolute paths together in a tuple. That might be clearest.

mdiep:
An integration test sounds perfect. I'd like to get this in the next release either way, but it's waiting on dual Xcode 7/8 support anyway. (#1442)

BernardGatt:
@mdiep Fix updated, can you please have a look?

I'll have a look at the integration tests next.

Source/carthage/Archive.swift
```diff
@@ -78,31 +78,39 @@ public struct ArchiveCommand: CommandType {
...
 				.flatMap(.Merge) { platform -> SignalProducer<String, CarthageError> in
 					return SignalProducer(values: frameworks).map { framework in
-						return (options.directoryPath as NSString).stringByAppendingPathComponent((platform.relativePath as NSString).stringByAppendingPathComponent(framework))
+						return (platform.relativePath as NSString).stringByAppendingPathComponent(framework)
 					}
 				}
-				.filter { relativePath in NSFileManager.defaultManager().fileExistsAtPath(relativePath) }
+				.collect()
+				.flatMap(.Merge) { relativePaths -> SignalProducer<(relativePath: String, absolutePath: String), CarthageError> in
+					return SignalProducer(values: relativePaths).map { relativePath in
```
mdiep: You can move this map out of the `flatMap` and before the `collect`—that will make it quite a bit simpler.

mdiep: Oh, it looks like you don't need the `collect` at all.

mdiep:
Code looks good to me apart from that one note. Still trying to test, but having a bit of submodule trouble atm.

mdiep:
GitHub isn't reporting any merge conflicts, but we'll need to merge `master` into this branch since #1442 was merged.

BernardGatt:
Looking at it now.

BernardGatt:
Just merged master into this branch.

Should we merge this PR and then create another one adding the integration tests?

mdiep:
Just tested it and it works great! Thanks for the PR!

Another PR that adds a test would be ✨.

BernardGatt:
cheers :)

PR #1391, opened 2016-07-18
[carthage] Accept `appletv(os|simulator)` for `--platform` option

ikesyo:
The possible SDK values we pass to `xcodebuild -sdk` are not `tvos` nor `tvsimulator`.

Maybe we should remove the `"tvsimulator": .tvOS` case?


mdiep:
👍🏻

PR #1399, opened 2016-07-22
Use plain xcodebuild instead of xctool

ikesyo:
xctool won't support Xcode 8 and will be deprecated. See https://github.com/facebook/xctool/issues/704 for the details.


mdiep:
👍

PR #1419, opened 2016-08-05
Per-dependency, per-commitish DerivedData

### LINKED ISSUE #459: Build with private, per-project DerivedData
`carthage build` could use its own private DerivedData folder for each project built, and separate out those folders for each project as well.

This could theoretically allow us to parallelize builds (to the extent allowed by the dependency graph), since they'd no longer be modifying the same global location. It might also alleviate rebuilding issues like #432.

We could also purge this private folder automatically whenever we build a different dependency version, which could obviate the need for a [`carthage clean` command](https://github.com/Carthage/Carthage/pull/427).

### LINKED ISSUE #1285: Derived Data Location
Is there any reason Carthage couldn't put its derived data in a local folder instead of the Xcode default? Clearing global derived data breaks Xcode when products are compiled outside of a given Xcode project and I don't remember to use the new build flag every time. If that is not a valid option, could a Cartfile option or environment variable be used that would allow the behavior to be automatic?

ikesyo:
Would resolve #459, resolve #1285.


Source/CarthageKit/Xcode.swift
```diff
@@ -932,7 +932,7 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 							argsForBuilding.bitcodeGenerationMode = .Bitcode
 						}
 
-						var buildScheme = xcodebuildTask(["clean", "build"], argsForBuilding)
+						var buildScheme = xcodebuildTask("build", argsForBuilding)
```
NachoSoto: Given how non-resiliant Xcode is to bad incremental compilation, what would a user have to do to force a clean? I wonder if we should add `carthage clean`

NachoSoto:
I approve of this change. Let's kill the shared mutable cache!
Just one question!

Source/CarthageKit/Xcode.swift
ikesyo: Or `carthage build --clean` option?

mdiep:
This is on my todo list, but it's going to take me a few more days to get to it. Sorry!

Source/CarthageKit/Xcode.swift
NachoSoto: Yes, that would work too.

Source/CarthageKit/Project.swift
```diff
@@ -675,19 +681,46 @@ public final class Project {
...
 					return .empty
 				}
 
-				return buildDependencyProject(dependency.project, self.directoryURL, withOptions: options, sdkFilter: sdkFilter)
+				var options = options
+				let cleanDerivedDataIfNeeded: SignalProducer<(), CarthageError>
+
+				if options.derivedDataPath != nil {
+					cleanDerivedDataIfNeeded = .empty
+				} else {
```
mdiep: Why are we treating these two cases differently?

Source/CarthageKit/Project.swift
```diff
@@ -675,19 +681,46 @@ public final class Project {
...
+
+					let fileManager = NSFileManager.defaultManager()
+					cleanDerivedDataIfNeeded = fileManager
+						.carthage_enumeratorAtURL(derivedDataPerDependency, includingPropertiesForKeys: [], options: [ .SkipsSubdirectoryDescendants ], catchErrors: true)
+						.flatMap(.Concat) { _, URL -> SignalProducer<(), CarthageError> in
+							if URL != derivedDataVersioned {
+								_ = try? fileManager.removeItemAtURL(URL)
+							}
+							return .empty
+						}
```
mdiep: What's the intent here?

mdiep:
Could you write up how this is supposed to work? I'm not understanding the vision.

Source/CarthageKit/Project.swift
ikesyo: I was thinking that we should opt out the behavior when the `--derived-data` option is specified, but may not need to do that.

Source/CarthageKit/Project.swift
ikesyo: If there is an existing derived data dir for the specific commitish of the dependency, and we are trying to build the dependency with a different commitish, we should remove (clean) the directory and do a clean build instead of adding `clean` action to `xcodebuild` command.

ikesyo: But if we use a per-commitish directory, we would not need to delete other commitish directories. That just consumes more disk spaces but would not hurt our builds.

ikesyo:
> Could you write up how this is supposed to work? I'm not understanding the vision.

We would use a per-project, per-dependency, per-commitish derived data directory for our builds (e.g. `~/Library/Caches/org.carthage.CarthageKit/DerivedData/ReactiveCocoa/Result/2.1.3`). In this way, switching to another tag or commit or branch should not corrupt the build artifacts. Maybe we could omit the project level (`~/Library/Caches/org.carthage.CarthageKit/DerivedData/Result/2.1.3`) if we preserve different commitish directories for a dependency: https://github.com/Carthage/Carthage/pull/1419#discussion_r76242103

ikesyo:
How does it look like now?

ikesyo:
Since #1489 has been merged, we don't need to remove the `clean` frag anymore: ad8aa03

@mdiep Could you take another look for this? 🙏

mdiep:
Sorry @ikesyo! I'm embarrassed by how long this took me. 💔 

This looks great. It also fixes an issue I'd seen and been meaning to fix with the `--derived-data` flag.

PR #1420, opened 2016-08-05
Add BuildOptions to share common options between `build` and `update` commands

ikesyo:
Arose in https://github.com/Carthage/Carthage/pull/1057#issuecomment-177572209 and https://github.com/Carthage/Carthage/pull/1057#issuecomment-177612122.


NachoSoto:
Nice refactor :+1:

PR #1421, opened 2016-08-06
Symlink checkout folders

### LINKED ISSUE #467: Dependencies’ Carthage checkouts are not populated when checking out parent projects
If a dependency project B uses Carthage to add projects like [jspahrsummers/xcconfigs](https://github.com/jspahrsummers/xcconfigs), and _is not using submodules_, this can cause issues when B is integrated in a parent project A.

Specifically, `Carthage/Checkouts/xcconfigs` will be created if you run `carthage bootstrap` within project B, but not when run from project A. Since project B requires those files to build, the overall build will fail.

There are a couple of workarounds for this right now:

1. Commit affected checkouts directly into the Git repository, so they’re always available when Carthage clones it.
1. Skip Carthage for the affected checkouts, and add them as submodules outside of the `Carthage` folder.
1. Enable `--use-submodules` when using Carthage in project B.

All of these have certain drawbacks, and it feels like having to structure a workflow to work around a flaw in the tool.

What if we symlinked `Carthage/Checkouts` into every dependency, much like `Carthage/Build`? That wouldn’t solve this problem for private dependencies (because those won’t get checked out by the parent), but would allow other checkouts to be shared. If we wanted to do that, we would need to think about the behavior in the face of existing checkout folders.

BobElDevil:
fixes #467

Basic approach was to perform the checkout symlink at the same time that we do the build symlink. I split out the two symlinking events into their own methods, and chained them together in the build method. 

For the checkouts, as discussed in the issue, instead of just symlinking the "Checkouts" folder itself I loop through the directories inside and symlink those, if something isn't already there, to prevent blowing away submodules or other possibilities like an existing symlink to a local development repo.


mdiep:
This is on my todo list, but it's going to take me a few more days to get to it. Sorry!

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,85 +1080,152 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
 ///
 /// Returns producers in the same format as buildInDirectory().
 public func buildDependencyProject(dependency: ProjectIdentifier, _ rootDirectoryURL: NSURL, withOptions options: BuildOptions, sdkFilter: SDKFilterCallback = { .Success($0.0) }) -> SignalProducer<BuildSchemeProducer, CarthageError> {
-	let rootBinariesURL = rootDirectoryURL.URLByAppendingPathComponent(CarthageBinariesFolderPath, isDirectory: true).URLByResolvingSymlinksInPath!
 	let rawDependencyURL = rootDirectoryURL.URLByAppendingPathComponent(dependency.relativePath, isDirectory: true)
 	let dependencyURL = rawDependencyURL.URLByResolvingSymlinksInPath!
 
 	let schemeProducers = buildInDirectory(dependencyURL, withOptions: options, sdkFilter: sdkFilter)
```
mdiep: It doesn't look like there's any reason for this to be in a local variable anymore.

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,85 +1080,152 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+						return error
+					}
+			}
+		}
+}
+
+/// Creates symlink between the dependency build folder and the root build folder
+///
+/// Returns a signal indicating success
+private func symlinkBuildPathForDependencyProject(dependency: ProjectIdentifier, rootDirectoryURL: NSURL) -> SignalProducer<Bool, CarthageError> {
```
mdiep: This can return a `SignalProducer<(), CarthageError>` since value isn't needed.

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,85 +1080,152 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
 			}
+		}
+		return .Success(true)
+	}
+}
+
+/// Creates symlink between the dependency checkouts and the root checkouts
+///
+/// Returns a signal indicating success
+private func symlinkCheckoutPathForDependencyProject(dependency: ProjectIdentifier, rootDirectoryURL: NSURL) -> SignalProducer<Bool, CarthageError> {
```
mdiep: This can also return a `SignalProducer<(), CarthageError>`.

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,85 +1080,152 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+///
+/// Returns a signal indicating success
+private func symlinkBuildPathForDependencyProject(dependency: ProjectIdentifier, rootDirectoryURL: NSURL) -> SignalProducer<Bool, CarthageError> {
+	return SignalProducer.attempt {
+		let rootBinariesURL = rootDirectoryURL.URLByAppendingPathComponent(CarthageBinariesFolderPath, isDirectory: true).URLByResolvingSymlinksInPath!
+		let rawDependencyURL = rootDirectoryURL.URLByAppendingPathComponent(dependency.relativePath, isDirectory: true)
+		let dependencyURL = rawDependencyURL.URLByResolvingSymlinksInPath!
+
+		do {
+			try NSFileManager.defaultManager().createDirectoryAtURL(rootBinariesURL, withIntermediateDirectories: true, attributes: nil)
```
mdiep: It seems worth adding a `let fileManager = NSFileManager.defaultManager()` to remove all these long, redundant calls.

mdiep: I guess you just moved this code. 😄 But if you don't updating that, it'd be appreciated!

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,85 +1080,152 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+		do {
+			try NSFileManager.defaultManager().createDirectoryAtURL(dependencyCheckoutsURL, withIntermediateDirectories: true, attributes: nil)
+		} catch let error as NSError {
+			if !(error.domain == NSCocoaErrorDomain && error.code == NSFileWriteFileExistsError) {
+				return .Failure(.WriteFailed(dependencyCheckoutsURL, error))
+			}
+		}
+		return .Success(true)
+	}
+	.flatMap(.Merge) { _ -> SignalProducer<Bool, CarthageError> in
```
mdiep: Indentation here seems a bit wonky. Subsequent operations should be indented one level deeper than the `return`.

```swift
return SignalProducer<(), CarthageError>.attempt {
        …
        return …
    }
    .flatMap(.Merge) {
```

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,85 +1080,152 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+
+		return NSFileManager.defaultManager().carthage_enumeratorAtURL(rootCheckoutsURL, includingPropertiesForKeys: [NSURLIsDirectoryKey], options: [.SkipsSubdirectoryDescendants, .SkipsHiddenFiles])
```
mdiep: I think it'd be better to only create symlinks for the dependencies of this project.

So given these `Cartfile`s:

```
// App Cartfile
github "my/A"
github "my/B"
```

```
// my/A Cartfile
github "my/B"
```

```
// my/B Cartfile
```

You'd end up with this directory structure:

```
Carthage/
  Checkouts/
    A/
      Carthage/
        Checkouts/
          B -> ../../../B
    B/
       Carthage/
         Checkouts/
```

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,85 +1080,152 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+			.attemptMap { _, checkoutURL -> Result<Bool, CarthageError> in
+				var isDirectory: AnyObject?
 				do {
-					try NSFileManager.defaultManager().createSymbolicLinkAtURL(dependencyBinariesURL, withDestinationURL: rootBinariesURL)
+					try checkoutURL.getResourceValue(&isDirectory, forKey: NSURLIsDirectoryKey)
+					if isDirectory == nil {
```
mdiep: How about doing something like this:

```swift
let isDirectory: Bool
do {
    var value: AnyObject?
    try checkoutURL.getResourceValue(&value, forKey: NSURLIsDirectoryKey)
    if let value = value {
        isDirectory = value.boolValue
    } else {
        return .Failure(…)
    }
}
```

mdiep:
Left some initial comments. I'm not doing reviewing, but I'm out of time for tonight. ☺️

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,85 +1080,152 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+					return .Failure(.ReadFailed(checkoutURL, error))
 				}
 
-				do {
-					try NSFileManager.defaultManager().createSymbolicLinkAtPath(dependencyBinariesURL.path!, withDestinationPath: linkDestinationPath)
-				} catch let error as NSError {
-					return .Failure(.WriteFailed(dependencyBinariesURL, error))
+				if let directory = isDirectory?.boolValue where directory {
+					let dependencySubCheckoutURL = dependencyCheckoutsURL.URLByAppendingPathComponent(checkoutURL.lastPathComponent!)
+					let foo = (CarthageProjectCheckoutsPath as NSString).stringByAppendingPathComponent(checkoutURL.lastPathComponent!)
```
BobElDevil: oh just realized I left a debugging variable in here... I'll rename this to not be `foo` =)

Source/CarthageKit/Xcode.swift
BobElDevil: done

Source/CarthageKit/Xcode.swift
BobElDevil: done

Source/CarthageKit/Xcode.swift
BobElDevil: done

Source/CarthageKit/Xcode.swift
BobElDevil: sure, made a variable for this method and the other one

Source/CarthageKit/Xcode.swift
BobElDevil: updated

Source/CarthageKit/Xcode.swift
BobElDevil: done

Source/CarthageKit/Xcode.swift
BobElDevil: Still working on this. At this point in the code path the dependency information is not visible (the DAG is generated privately in Project.swift), so I'll have to add an argument to this method to take a list of dependency names as well and then move the call up a level into where project starts the build

BobElDevil:
As I was trying to work through symlinking only the sub dependencies I realized it makes more sense to have that code right where we are performing the checkout rather than in the build phase, so we have access to the graph, so I think it makes sense to move the implementation (and test) over to the Project where this is occurring.

BobElDevil:
Changed my mind on the location of the implementation, since most of the context of what was happening / needed is still with regards to the 'build'. So I just forwarded the necessary info into the build method so it could only symlink the sub dependencies

Source/CarthageKit/Project.swift
```diff
@@ -251,6 +251,16 @@ public final class Project {
 		}
 	}
 
+	/// Produces the sub dependencies of the given dependency
+	func subDependenciesForDependency(dependency: Dependency<PinnedVersion>) -> SignalProducer<Set<ProjectIdentifier>, CarthageError> {
```
mdiep: We can just call this `dependenciesForDependency`, I think.

Source/CarthageKit/Project.swift
```diff
@@ -589,12 +599,10 @@ public final class Project {
 		// dependencies before the projects that depend on them.
 		return SignalProducer<Dependency<PinnedVersion>, CarthageError>(values: cartfile.dependencies)
 			.flatMap(.Merge) { (dependency: Dependency<PinnedVersion>) -> SignalProducer<DependencyGraph, CarthageError> in
-				return self.cartfileForDependency(dependency)
-					.map { (cartfile: Cartfile) -> DependencyGraph in
-						return [ dependency.project: Set(cartfile.dependencies.map { $0.project }) ]
+				return self.subDependenciesForDependency(dependency)
+					.map { set in
+						[dependency.project: set]
```
mdiep: Can we use a semantic name here, like `dependencies`?

Source/CarthageKit/Xcode.swift
```diff
@@ -1079,88 +1079,164 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+				.collect()
+				.attemptMap { _ in
+					return .Success()
 				}
 		}
 }
 
+private func relativeLinkDestinationForDependency(dependency: ProjectIdentifier, subdirectory: String) -> String {
+	let dependencySubdirectoryPath = (dependency.relativePath as NSString).stringByAppendingPathComponent(subdirectory)
+	let componentsForGettingTheHellOutOfThisRelativePath = Array(count: (dependencySubdirectoryPath as NSString).pathComponents.count - 1, repeatedValue: "..")
```
mdiep: 😆

Source/CarthageKit/Xcode.swift
```diff
@@ -1079,88 +1079,164 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
-				let componentsForGettingTheHellOutOfThisRelativePath = Array(count: (dependencyBinariesRelativePath as NSString).pathComponents.count - 1, repeatedValue: "..")
-
-				// Directs a link from, e.g., /Carthage/Checkouts/ReactiveCocoa/Carthage/Build to /Carthage/Build
-				let linkDestinationPath = componentsForGettingTheHellOutOfThisRelativePath.reduce(CarthageBinariesFolderPath) { trailingPath, pathComponent in
-					return (pathComponent as NSString).stringByAppendingPathComponent(trailingPath)
-				}
+/// Creates symlink between the dependency checkouts and the root checkouts
+///
+/// Returns a signal indicating success
+private func symlinkCheckoutPathForDependencyProject(dependency: ProjectIdentifier, subDependencies: Set<ProjectIdentifier>, rootDirectoryURL: NSURL) -> SignalProducer<(), CarthageError> {
```
mdiep: _Path_ should be plural in this name since it's creating multiple links.

Source/CarthageKit/Xcode.swift
```diff
@@ -1079,88 +1079,164 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+							return .Failure(.ReadFailed(checkoutURL, nil))
+						}
+					} catch let error as NSError {
+						return .Failure(.ReadFailed(checkoutURL, error))
+					}
 
-					default:
-						return error
+					let dependencyName = checkoutURL.lastPathComponent!
+					let hasDependency = subDependencyNames.contains(dependencyName)
```
mdiep: This seems backwards. Rather than enumerating `/Checkouts/` and skipping dependencies we don't want, we should enumerate the dependencies we do want and look for their directories.

Source/CarthageKit/Xcode.swift
```diff
@@ -1079,88 +1079,164 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+
+					if isDirectory && hasDependency {
```
mdiep: The `isDirectory` check can probably move to a `SignalProducer.filter` after enumerating dependencies.

Source/CarthageKit/Xcode.swift
```diff
@@ -1079,88 +1079,164 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+								return .Failure(.WriteFailed(checkoutURL, error))
+							}
+						}
 					}
+					return .Success()
+				}
+				.collect()
+				.attemptMap { _ in
+					return .Success()
 				}
```
mdiep: You could probably use `.then(.empty)` instead of `.collect().attemptMap()`

Source/CarthageKit/Xcode.swift
```diff
@@ -1079,88 +1079,164 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
-	let rootBinariesURL = rootDirectoryURL.URLByAppendingPathComponent(CarthageBinariesFolderPath, isDirectory: true).URLByResolvingSymlinksInPath!
+public func buildDependencyProject(dependency: ProjectIdentifier, _ rootDirectoryURL: NSURL, subDependencies: Set<ProjectIdentifier>, withOptions options: BuildOptions, sdkFilter: SDKFilterCallback = { .Success($0.0) }) -> SignalProducer<BuildSchemeProducer, CarthageError> {
 	let rawDependencyURL = rootDirectoryURL.URLByAppendingPathComponent(dependency.relativePath, isDirectory: true)
 	let dependencyURL = rawDependencyURL.URLByResolvingSymlinksInPath!
 
-	let schemeProducers = buildInDirectory(dependencyURL, withOptions: options, sdkFilter: sdkFilter)
-	return SignalProducer.attempt { () -> Result<SignalProducer<BuildSchemeProducer, CarthageError>, CarthageError> in
+	return symlinkBuildPathForDependencyProject(dependency, rootDirectoryURL: rootDirectoryURL)
+		.concat(symlinkCheckoutPathForDependencyProject(dependency, subDependencies: subDependencies, rootDirectoryURL: rootDirectoryURL))
```
mdiep: This should probably happen as part of `carthage checkout`, not `carthage build`?

BobElDevil:
Thanks for the feedback! I'll try to get to these sometime this weekend

BobElDevil:
Pushed up the latest changes. Moved the symlinks into when we perform the checkout instead of build.

mdiep:
I'll review this again ASAP. Sorry for the long review times—this has been a very busy season with all the changes for Swift 3! (Plus changes for ReactiveCocoa 5)

Source/CarthageKit/Project.swift
```diff
@@ -639,33 +653,87 @@ public final class Project {
...
+	/// Creates symlink between the dependency checkouts and the root checkouts
+	private func symlinkCheckoutPathsForDependencyProject(dependency: ProjectIdentifier, subDependencies: Set<ProjectIdentifier>, rootDirectoryURL: NSURL) -> SignalProducer<(), CarthageError> {
+		let rootCheckoutsURL = rootDirectoryURL.URLByAppendingPathComponent(CarthageProjectCheckoutsPath, isDirectory: true).URLByResolvingSymlinksInPath!
+		let rawDependencyURL = rootDirectoryURL.URLByAppendingPathComponent(dependency.relativePath, isDirectory: true)
+		let dependencyURL = rawDependencyURL.URLByResolvingSymlinksInPath!
+		let dependencyCheckoutsURL = dependencyURL.URLByAppendingPathComponent(CarthageProjectCheckoutsPath, isDirectory: true).URLByResolvingSymlinksInPath!
+		let subDependencyNames = subDependencies.map { $0.name }
+		let fileManager = NSFileManager.defaultManager()
+
+		let symlinksProducer =  SignalProducer(values: Array(subDependencyNames))
```
ikesyo: An extra space before `SignalProducer`

ikesyo: You should not need to wrap the `subDependencyNames` in an extra `Array`.

BobElDevil: Ah, I had it in my head it was a set because the original subDependencies was. Done.

Source/CarthageKit/Project.swift
```diff
@@ -680,19 +748,22 @@ public final class Project {
...
-							// to continue building other dependencies.
-							self._projectEventsObserver.sendNext(.SkippedBuilding(dependency.project, error.description))
-							return .empty
-
-						default:
-							return SignalProducer(error: error)
-						}
+				return self.dependenciesForDependency(dependency)
+					.flatMap(.Merge) { subDependencies -> SignalProducer<BuildSchemeProducer, CarthageError> in
+						return buildDependencyProject(dependency.project, self.directoryURL, subDependencies: subDependencies, withOptions: options, sdkFilter: sdkFilter)
```
mdiep: It doesn't look like `subDependencies` really need to be passed to `buildDependencyProject` anymore, so a lot of this can be reverted/simplified.

mdiep:
This looks good to me. Just the one note about reverting one part of the change. Also, this will need a merge from `master` to pick up the Swift 2.3 compatibility changes.

Sorry again about the long review time! I was traveling last week and had less free time than I expected. But we should be able to get this merged this week!

Source/CarthageKit/Project.swift
BobElDevil: good point, will put the signature back as it was.

BobElDevil:
@mdiep Removed the unnecessary parameter and merged with the swift 2.3 compatibility

mdiep:
Looks great! Thanks!

PR #1425, opened 2016-08-10
Make a simple package manager for Linux as well?

mirrorinf:
Carthage performs much better than the native swift package manager with Xcode. So, why not create the same experience on Linux?
The implementation could be just like on OS X: Carthage just do the work of compiling, providing the user a compiled framework. And the IDE or editor plugins could help to integrate it.
For Linux compatibility, Carthage can create a specific format of package definition. There could a automatic tool to generate description from Xcode build schemes.


NachoSoto:
Carthage is made FOR Xcode, it wouldn't make sense to reinvent the SPM. If there's anything you'd like to improve on the SPM for Linux I'd recommend sending a proposal there instead.

PR #1429, opened 2016-08-11
Redundant update fetches

BobElDevil:
I don't believe there's an issue that tracks this. This is my attempt to reduce the amount of times we fetch a dependency.

This started occurring with the fix for #1202 (which was needed because I caused a regression for branches in my previous fetch optimization =/ )

What this change does is write out a custom config to the cache repo that specifies the last time carthage itself ran a fetch. Then we skip fetching if it occurred within the last minute. This avoids adding any state to Carthage in case of long running CarthageKit usage, but still optimizes the majority of cases where we're just looping through the dependencies and fetching them all during resolution.

The other approach I thought of that could be more accurate, but would require a lot more plumbing through the system, would be to maintain state, but only created at the point where the calls originate, e.g., a 'fetchedProjects' array that gets propagated through all the calls. I avoided this approach since it seemed like a lot of code change just to close up a 60 second window that likely wouldn't be hit,  but putting that alternative out there.

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -136,6 +136,34 @@ extension Submodule: CustomStringConvertible {
 	}
 }
 
+/// Struct to encapsulate global fetch interval cache
+public struct FetchCache {
+	/// Amount of time before a git repository is fetched again. Defaults to 1 minute
+	public static var fetchCacheInterval: NSTimeInterval = 60.0
+
+	private static var lastFetchTimes: [GitURL : NSTimeInterval] = [:]
+
+	internal static func clearFetchTimes() {
+		lastFetchTimes.removeAll()
+	}
+
+	internal static func needsFetch(forURL url: GitURL) -> Bool {
+		guard let lastFetch = lastFetchTimes[url] else {
+			return true
+		}
+
+		let difference = NSDate().timeIntervalSince1970 - lastFetch
+
+		return !(0...fetchCacheInterval).contains(difference)
+	}
+
+	private static func updateLastFetchTime(forURL url: GitURL?) {
+		if let url = url {
+			lastFetchTimes[url] = NSDate().timeIntervalSince1970
+		}
+	}
+}
+
 /// Shells out to `git` with the given arguments, optionally in the directory
 /// of an existing repository.
 public func launchGitTask(arguments: [String], repositoryFileURL: NSURL? = nil, standardInput: SignalProducer<NSData, NoError>? = nil, environment: [String: String]? = nil) -> SignalProducer<String, CarthageError> {
@@ -181,6 +209,7 @@ public func cloneRepository(cloneURL: GitURL, _ destinationURL: NSURL, bare: Boo
 	}
 
 	return launchGitTask(arguments + [ "--quiet", cloneURL.URLString, destinationURL.path! ])
+		.on(completed: { FetchCache.updateLastFetchTime(forURL: cloneURL) })
 }
 
 /// Returns a signal that completes when the fetch completes successfully.
@@ -201,6 +230,7 @@ public func fetchRepository(repositoryFileURL: NSURL, remoteURL: GitURL? = nil, 
 	}
 
 	return launchGitTask(arguments, repositoryFileURL: repositoryFileURL)
+		.on(completed: { FetchCache.updateLastFetchTime(forURL: remoteURL) })
 }
 
 /// Sends each tag found in the given Git repository.
--- a/Source/CarthageKitTests/ProjectSpec.swift
+++ b/Source/CarthageKitTests/ProjectSpec.swift
@@ -6,7 +6,7 @@
 //  Copyright (c) 2014 Carthage. All rights reserved.
 //
 
-import CarthageKit
+@testable import CarthageKit
 import Foundation
 import Nimble
 import Quick
@@ -104,8 +104,11 @@ class ProjectSpec: QuickSpec {
 				return cloneOrFetchProject(projectIdentifier, preferHTTPS: false, destinationURL: cacheDirectoryURL, commitish: commitish)
 			}
 
-			func assertProjectEvent(commitish commitish: String? = nil, action: ProjectEvent? -> ()) {
+			func assertProjectEvent(commitish commitish: String? = nil, clearFetchTime: Bool = true, action: ProjectEvent? -> ()) {
 				waitUntil { done in
+					if clearFetchTime {
+						FetchCache.clearFetchTimes()
+					}
 					cloneOrFetch(commitish: commitish).start(Observer(
 						completed: done,
 						next: { event, _ in action(event) }
@@ -161,6 +164,14 @@ class ProjectSpec: QuickSpec {
 				addCommit()
 
 				assertProjectEvent(commitish: commitish) { expect($0).to(beNil()) }
+			}
+
+			it ("should not fetch twice in a row, even if no commitish is given") {
+				// Clone first
+				expect(cloneOrFetch().wait().error).to(beNil())
+
+				assertProjectEvent { expect($0?.isFetching) == true }
+				assertProjectEvent(clearFetchTime: false) { expect($0).to(beNil())}
 			}
 		}
 	}
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -904,6 +904,10 @@ public func cloneOrFetchProject(project: ProjectIdentifier, preferHTTPS: Bool, d
 				.flatMap(.Merge) { isRepository -> SignalProducer<(ProjectEvent?, NSURL), CarthageError> in
 					if isRepository {
 						let fetchProducer: () -> SignalProducer<(ProjectEvent?, NSURL), CarthageError> = {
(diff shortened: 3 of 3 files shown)


mdiep:
Thanks for looking into this. ✨ 

I don't love writing out the fetch date to disk. Let me think on it. It's definitely an easier fix and could tide us over until we can clean up the structure to be more amenable to a better solution.

mdiep:
I don't feel good about adding on-disk data. Here are some alternatives that I'd feel better about:

* Adding a global shared in-process cache with a global variable that lets users of CarthageKit dictate how much time should pass between fetches of a repo.

* Adding a global shared in-process cache and passing how much time should pass between between fetches as a parameter to CarthageKit functions that fetch.

* Use the modification time of `FETCH_HEAD` from the git repository on disk to determine when the last fetch occurred.

Do any of those sound good and doable to you?

BobElDevil:
My first idea was to use the FETCH_HEAD timestamp (that's what homebrew does) but was a wary of it because it's relying on a side effect of how git works to identify when to fetch. That's why I ended up writing a custom property, but I get why it's not desirable. 

I think the first option is the best here. The second I think would end up adding parameters to a lot of different areas due to how we currently fetch on-demand. I'll work on modifying my current solution to use in memory timestamps instead of the disk.

BobElDevil:
Pushed up changes for using a global variable to track the timestamps instead of disk. Also adds the ability to set the expiration timeout (defaults to 1 minute)

Source/CarthageKit/Git.swift
```diff
@@ -57,6 +57,30 @@ public struct GitURL: Equatable {
 		}
 	}
 
+	/// Amount of time before a git repository is fetched again. Defaults to 1 minute
+	public static var fetchExpirationTime: NSTimeInterval = 60.0
```
mdiep: This should probably have _interval_ in the name instead of _time_.

`fetchCacheInterval`?

Source/CarthageKit/Git.swift
```diff
@@ -57,6 +57,30 @@ public struct GitURL: Equatable {
...
+
+	/// Cache of fetch times used to reduce redundant fetches
+	private static var lastFetchTimes: [GitURL : NSTimeInterval] = [:]
+
+	public static func clearFetchTimes() {
+		lastFetchTimes.removeAll()
+	}
+
+	public var needsFetch: Bool {
```
mdiep: I'd prefer to not make this and `clearFetchTimes` `public`.

mdiep:
Just a couple small things. 👍🏻

Source/CarthageKit/Git.swift
BobElDevil: done

Source/CarthageKit/Git.swift
BobElDevil: Sure, I made them internal. Also had to add a `@testable` import to the tests

NachoSoto:
Please squash the commits into one. A commit message like "Review fixes" will not be relevant in the future without this context.

Source/CarthageKit/Git.swift
```diff
@@ -57,6 +57,30 @@ public struct GitURL: Equatable {
 		}
 	}
 
+	/// Amount of time before a git repository is fetched again. Defaults to 1 minute
+	public static var fetchCacheInterval: NSTimeInterval = 60.0
```
NachoSoto: I don't think I understand why this mutable property belongs to `GitURL`, which represents a URL...

Source/CarthageKit/Git.swift
```diff
@@ -57,6 +57,30 @@ public struct GitURL: Equatable {
...
+
+	/// Cache of fetch times used to reduce redundant fetches
+	private static var lastFetchTimes: [GitURL : NSTimeInterval] = [:]
```
NachoSoto: Ditto.

Source/CarthageKit/Git.swift
```diff
@@ -57,6 +57,30 @@ public struct GitURL: Equatable {
...
+	}
+
+	var needsFetch: Bool {
+		guard let lastFetch = GitURL.lastFetchTimes[self] else {
+			return true
+		}
+
+		let difference = NSDate().timeIntervalSince1970 - lastFetch
+
+		return !(0...GitURL.fetchCacheInterval).contains(difference)
```
NachoSoto: This is extremely inefficient. Looking over an array of N elements just to see if `difference` is contained in the interval.

Source/CarthageKitTests/ProjectSpec.swift
```diff
@@ -162,6 +165,14 @@ class ProjectSpec: QuickSpec {
 
 				assertProjectEvent(commitish: commitish) { expect($0).to(beNil()) }
 			}
+
+			it ("should not fetch twice in a row, even if no commitish is given") {
```
NachoSoto: `commitish`?

Source/CarthageKit/Git.swift
BobElDevil: I put it here because it is the internal implementation detail of the non static 'needsFetch' and 'updateFetchTime', but I can move it outside of the struct

NachoSoto: Yeah, it would have to be outside, but also not a global property that's _implicitly_ used everywhere. Instead, please make that explicit by making it a parameter to whatever code uses this.

Source/CarthageKit/Git.swift
BobElDevil: it's not an array it's a Range, which is optimized for checks such as this

Source/CarthageKit/Git.swift
BobElDevil: So as discussed above due to the nature of how we fetch on demand there's many different paths into these fetch calls. So is making it a parameter to the funcs of the GitURL, but still a static var accessed in the two methods below (clone/fetch) ok, or should it get pulled out higher. At what point should it switch from a parameter to the global being passed in

Source/CarthageKitTests/ProjectSpec.swift
BobElDevil: it's the previous git parlance used and the name of the parameter into the clone/fetch methods. Basically it can be a commit sha or some other 'commit-ish' ref like a tag or branch

BobElDevil:
Just pushed an update. I moved all the new logic into a FetchCache struct (so it namespaced the static variables). It still keys off the GitURL in the dictionary but isn't tied directly to it anymore. 

Also squashed it all down into one commit

mdiep:
The fix looks good to me. Do you have have steps to reproduce the issue?

ffried:
@mdiep It should be reproducible by using the setup described in #1432 an performing carthage update.

mdiep:
This branch doesn't seem to be working with that repository. C still gets fetched twice. 😕

BobElDevil:
That's what I get for creating a reproduction setup without testing it. The zip I gave you tried to use relative paths for the cartfile, so when it tried to do the recursive fetch the relative path was based off the cache location, and the fetch failed (maybe that's another bug? Not sure if relative paths are even supposed to be supported for file based dependencies).

If you manually go in and change the Cartfile references in repoA and repoB to use absolute paths, then you should see the current master do the 3 fetches, and this branch fetch 2 times.  I promise I tested and it worked with my real setup (which I can't share because of private repos) when I wrote this =)

mdiep:
It looks like the fetching fails for the local URLs, even though it doesn't get reported as a failure. 😕 

Tested again with absolute URLs and that works!

Source/CarthageKit/Git.swift
```diff
@@ -136,6 +136,34 @@ extension Submodule: CustomStringConvertible {
 	}
 }
 
+/// Struct to encapsulate global fetch interval cache
+public struct FetchCache {
+	/// Amount of time before a git repository is fetched again. Defaults to 1 minute
+	public static var fetchCacheInterval: NSTimeInterval = 60.0
```
NachoSoto: This is still a mutable global 👎👎👎

Source/CarthageKit/Git.swift
```diff
@@ -136,6 +136,34 @@ extension Submodule: CustomStringConvertible {
...
+
+	private static var lastFetchTimes: [GitURL : NSTimeInterval] = [:]
```
NachoSoto: Ditto. Why not make this local to each cache?

Source/CarthageKit/Git.swift
```diff
@@ -136,6 +136,34 @@ extension Submodule: CustomStringConvertible {
 	}
 }
 
+/// Struct to encapsulate global fetch interval cache
+public struct FetchCache {
+	/// Amount of time before a git repository is fetched again. Defaults to 1 minute
```
NachoSoto: This doesn't have value semantics since the entire interface is static. Very misleading: every user of this type is actually implicitly coupled now 👎

mdiep: Yeah, it's not great.

This is a short-term fix for a fairly serious issue. It's not pretty, but it'll work until we can refactor to a design that makes more sense.

It's technical debt, pure and simple. But I think it's the good kind of debt: the kind that lets you achieve an important short-term goal.

NachoSoto: 👍

PR #1431, opened 2016-08-12
Ensure that we handle symbolic links correctly when filtering file paths for projects/workspaces

rasharab:
Carthage’s behavior w/regards to locating projects/workspaces was dependent on the location of the directory the repository was cloned into.

In particular, with macs, if you clone a repository under the /tmp folder, that in fact is a symbolic link to /private/tmp, and will result in incorrect behavior when Carthage attempts to elide paths.

The crux of the problem was the usage of NSFileManager enumeration in locateProjectsInDirectory in conjuction with exclusion paths returned from invoking git commands.
The NSFileManager paths begin with “/private/tmp” while the git commands produce paths that begin with “/tmp”.

The cleanest solution is to use NSFileManager functionality (getRelationship), that tells us correctly if a path is indeed a child of another even in the case of symlinks.

The carthage NSURL extension (hasSubdirectory) does not correctly handle said cases, and it’s really not the appropriate approach in these instances.

FYI, we noticed this bug with several of our customers, as we place the customers' repositories under /tmp, while customers obviously would place them under their home directory.

The tickets often would ask us why there were unable to reproduce the issues mentioned above, and this is why.

### DIFF
--- a/Source/CarthageKit/FrameworkExtensions.swift
+++ b/Source/CarthageKit/FrameworkExtensions.swift
@@ -297,6 +297,37 @@ extension NSFileManager {
 			observer.sendCompleted()
 		}
 	}
+	
+	/// Handle the case where the otherURL is actually a symlink and the directoryURL is not.
+	/// For example, on OSX, the /tmp directory is a symlink to /private/tmp.
+	/// Any usage of NSFileManager enumeration operations will return /private/tmp,
+	/// while many commands used by carthage to determine directory paths will return /tmp.
+	/// Simply using hasSubDirectory will not work in these instances.
+	public func carthage_hasSubdirectory(directoryURL:NSURL, otherURL:NSURL) -> Result<Bool, CarthageError> {
+		let nsURLRelationshipPtr = UnsafeMutablePointer<NSURLRelationship>.alloc(1)
+		defer { nsURLRelationshipPtr.destroy() }
+		
+		// getRelationship will produce an NSError if either of the paths does not exist
+		// If that's the case, we default to the old behavior
+		if (!self.fileExistsAtPath(directoryURL.path!) || !self.fileExistsAtPath(otherURL.path!)) {
+			return .Success(directoryURL.hasSubdirectory(otherURL));
+		}
+		
+		do {
+			try self.getRelationship(nsURLRelationshipPtr,
+			   ofDirectoryAtURL: directoryURL,
+			   toItemAtURL: otherURL)
+		} catch let error as NSError {
+			return .Failure(.ReadFailed(otherURL, error));
+		}
+		
+		switch nsURLRelationshipPtr.memory {
+		case NSURLRelationship.Contains, NSURLRelationship.Same:
+			return .Success(true);
+		default:
+			return .Success(false);
+		}
+	}
 }
 
 /// Creates a counted set from a sequence. The counted set is represented as a
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -88,17 +88,20 @@ extension ProjectLocator: CustomStringConvertible {
 /// Sends all matches in preferential order.
 public func locateProjectsInDirectory(directoryURL: NSURL) -> SignalProducer<ProjectLocator, CarthageError> {
 	let enumerationOptions: NSDirectoryEnumerationOptions = [ .SkipsHiddenFiles, .SkipsPackageDescendants ]
+	let defaultNSFileManager = NSFileManager.defaultManager();
 
 	return gitmodulesEntriesInRepository(directoryURL, revision: nil)
 		.map { directoryURL.URLByAppendingPathComponent($0.path) }
 		.concat(value: directoryURL.URLByAppendingPathComponent(CarthageProjectCheckoutsPath))
 		.collect()
 		.flatMap(.Merge) { directoriesToSkip in
-			return NSFileManager.defaultManager()
+			return defaultNSFileManager
 				.carthage_enumeratorAtURL(directoryURL.URLByResolvingSymlinksInPath!, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: enumerationOptions, catchErrors: true)
 				.map { _, URL in URL }
 				.filter { URL in
-					return !directoriesToSkip.contains { $0.hasSubdirectory(URL) }
+					return !directoriesToSkip.contains {
+						return defaultNSFileManager.carthage_hasSubdirectory($0, otherURL: URL).value == true;
+					}
 				}
 		}
 		.map { URL -> ProjectLocator? in
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ Unless you’re specifically improving something about the command-line experien
 After checkout, you can run the following command from the cloned directory, and then open the workspace in Xcode:
 
 ```bash
-./script/bootstrap
+make bootstrap
 ```
 
 Then, to install your development copy of Carthage (and any local changes you've made) on your system, and test with your own repos:


Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -297,6 +297,34 @@ extension NSFileManager {
...
 	}
+	
+	/// Handle the case where the otherURL is actually a symlink and the directoryURL is not.
+	/// For example, on OSX, the /tmp directory is a symlink to /private/tmp.
+	/// Any usage of NSFileManager enumeration operations will return /private/tmp,
+	/// while many commands used by carthage to determine directory paths will return /tmp.
+	/// Simply using hasSubDirectory will not work in these instances.
+	public func carthage_hasSubdirectory(directoryURL:NSURL, otherURL:NSURL) -> Bool {
+		let nsURLRelationshipPtr = UnsafeMutablePointer<NSURLRelationship>.alloc(1)
+		defer { nsURLRelationshipPtr.destroy(); }
```
NachoSoto: No ; please

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -297,6 +297,34 @@ extension NSFileManager {
...
+			                                                   toItemAtURL: otherURL)
+		
+			switch nsURLRelationshipPtr.memory {
+			case NSURLRelationship.Contains, NSURLRelationship.Same:
+				return true;
+			default:
+				return false;
+			}
+		} catch {
+			/// Hit this when directoryURL or otherURL does not exist.
```
NachoSoto: Can you catch **only** this particular error then?

Source/CarthageKit/Xcode.swift
```diff
@@ -98,7 +98,9 @@ public func locateProjectsInDirectory(directoryURL: NSURL) -> SignalProducer<Pro
 				.carthage_enumeratorAtURL(directoryURL.URLByResolvingSymlinksInPath!, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: enumerationOptions, catchErrors: true)
 				.map { _, URL in URL }
 				.filter { URL in
-					return !directoriesToSkip.contains { $0.hasSubdirectory(URL) }
+					return !directoriesToSkip.contains {
+						return NSFileManager.defaultManager().carthage_hasSubdirectory($0, otherURL: URL);
```
NachoSoto: Can you create a local constant for this file manager to reuse it above too and remove the duplication?

Source/CarthageKit/FrameworkExtensions.swift
rasharab: Will do.

Source/CarthageKit/FrameworkExtensions.swift
rasharab: I'll see what I can do here.

-Rashin

Source/CarthageKit/Xcode.swift
rasharab: Will do.

-Rashin

rasharab:
I've submitted another pass that hopefully addresses your points. :)

-Rashin

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -297,6 +297,37 @@ extension NSFileManager {
 			observer.sendCompleted()
 		}
 	}
+	
+	/// Handle the case where the otherURL is actually a symlink and the directoryURL is not.
+	/// For example, on OSX, the /tmp directory is a symlink to /private/tmp.
+	/// Any usage of NSFileManager enumeration operations will return /private/tmp,
+	/// while many commands used by carthage to determine directory paths will return /tmp.
+	/// Simply using hasSubDirectory will not work in these instances.
+	public func carthage_hasSubdirectory(directoryURL:NSURL, otherURL:NSURL) -> Result<Bool, CarthageError> {
```
NachoSoto: Spaces after `:`

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -297,6 +297,37 @@ extension NSFileManager {
...
+		let nsURLRelationshipPtr = UnsafeMutablePointer<NSURLRelationship>.alloc(1)
+		defer { nsURLRelationshipPtr.destroy() }
+		
+		// getRelationship will produce an NSError if either of the paths does not exist
+		// If that's the case, we default to the old behavior
+		if (!self.fileExistsAtPath(directoryURL.path!) || !self.fileExistsAtPath(otherURL.path!)) {
```
NachoSoto: No parentheses around the condition

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -297,6 +297,37 @@ extension NSFileManager {
...
+			return .Success(directoryURL.hasSubdirectory(otherURL));
+		}
+		
+		do {
+			try self.getRelationship(nsURLRelationshipPtr,
```
NachoSoto: Can you our each parameter in it's own line?

Source/CarthageKit/Xcode.swift
```diff
@@ -88,17 +88,20 @@ extension ProjectLocator: CustomStringConvertible {
 /// Sends all matches in preferential order.
 public func locateProjectsInDirectory(directoryURL: NSURL) -> SignalProducer<ProjectLocator, CarthageError> {
 	let enumerationOptions: NSDirectoryEnumerationOptions = [ .SkipsHiddenFiles, .SkipsPackageDescendants ]
+	let defaultNSFileManager = NSFileManager.defaultManager();
```
NachoSoto: No `;`

NachoSoto: I would just call this `fileManager`.

Source/CarthageKit/Xcode.swift
```diff
@@ -88,17 +88,20 @@ extension ProjectLocator: CustomStringConvertible {
...
 		.collect()
 		.flatMap(.Merge) { directoriesToSkip in
-			return NSFileManager.defaultManager()
+			return defaultNSFileManager
 				.carthage_enumeratorAtURL(directoryURL.URLByResolvingSymlinksInPath!, includingPropertiesForKeys: [ NSURLTypeIdentifierKey ], options: enumerationOptions, catchErrors: true)
 				.map { _, URL in URL }
 				.filter { URL in
-					return !directoriesToSkip.contains { $0.hasSubdirectory(URL) }
+					return !directoriesToSkip.contains {
+						return defaultNSFileManager.carthage_hasSubdirectory($0, otherURL: URL).value == true;
```
NachoSoto: No `;`

NachoSoto: Also I don't think we should ignore errors here, should we?

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -297,6 +297,37 @@ extension NSFileManager {
...
+	/// while many commands used by carthage to determine directory paths will return /tmp.
+	/// Simply using hasSubDirectory will not work in these instances.
+	public func carthage_hasSubdirectory(directoryURL:NSURL, otherURL:NSURL) -> Result<Bool, CarthageError> {
+		let nsURLRelationshipPtr = UnsafeMutablePointer<NSURLRelationship>.alloc(1)
+		defer { nsURLRelationshipPtr.destroy() }
+		
+		// getRelationship will produce an NSError if either of the paths does not exist
+		// If that's the case, we default to the old behavior
+		if (!self.fileExistsAtPath(directoryURL.path!) || !self.fileExistsAtPath(otherURL.path!)) {
+			return .Success(directoryURL.hasSubdirectory(otherURL));
```
NachoSoto: No `;`

mdiep:
Superseded by #1459.

Issue #1427, opened 2016-08-10
Update to README: Linking dynamic frameworks to embedded frameworks requires as well linking them to the app target


evgeniyd:
See #1401

As @mdiep stated:

> Carthage doesn't support nested frameworks.
> 
> Nesting frameworks doesn't let you reuse those frameworks. For instance, if `A.framework`and `B.framework` both depend on `Other.framework`, then neither of them can nest `Other.framework`—otherwise you might end up with 2 different versions, and the correct one might not be chosen at runtime.
> 
> The correct way to do this is to list it as a dependency, but link it into the app target.

I suggest to add this point to the [README's Adding frameworks to unit tests or a framework](https://github.com/Carthage/Carthage#adding-frameworks-to-unit-tests-or-a-framework).

mdiep:
Care to open a pull request? There's already a section on nested frameworks, and this would make a fine addition.

evgeniyd:
@mdiep could you please review the PR?

mdiep:
Fixed by #1441.

Issue #1401, opened 2016-07-22
Linking only embedded framework with other dynamic framework fails when build & run on device: "required code signature missing"


evgeniyd:
* `carthage version`: 0.17.2
* `xcodebuild -version`: Xcode 8.0; Build version 8S162m
* Are you using `--no-build`? YES, but just because I build the framework with Swift 2.3 toolchain
* Are you using `--no-use-binaries`? NO
* Are you using `--use-submodules`? NO

**Cartfile**
```
github "johnsundell/unbox" ~> 1.8 #this is my Other.framework
```

**Carthage Output**
```
TBA
```

## Setup:

My setup is pretty simple (Swift 2.3 & Xcode 8B2):
 
- Using Carthage I have build **Other.framework**
- **MyApp** has embeded **My.framework** framework.
- The app and framework projects are under one Xcode workspace.
- I hade linked **Other.framework** to **My.framework** ONLY (that means, **MyApp** is not linked to **Other.framework** at all). The point here is that, **MyApp** does not need to use **Other.framework** API.

## Problem:

Everything seems to work fine, until I **Build & Run the app on the device**. The app launched and than the process is aborted with the following Xcode error:

    dyld: Library not loaded: @rpath/Other.framework/Other  
      Referenced from: /private/var/containers/Bundle/Application/DCF0331F-FF23-43CF-AE79-B3857D5A6EE3/MyApp.app/Frameworks/My.framework/My  
      Reason: no suitable image found.  Did find:  
      /private/var/containers/Bundle/Application/DCF0331F-FF23-43CF-AE79-B3857D5A6EE3/MyApp.app/Frameworks/My.framework/Frameworks/Other.framework/Other: required code signature missing for '/private/var/containers/Bundle/Application/DCF0331F-FF23-43CF-AE79-B3857D5A6EE3/MyApp.app/Frameworks/My.framework/Frameworks/Other.framework/Other'  
 
I have checked the signature of **Other.framework** and it looked OK to me. Moreover,
 
## Workaround:
 
Link **MyApp** with **Other.framework**. Horrible... This feels broken.

Linking the very same binary **Other.framework** to **MyApp** and solving the issue this way, points out, the **Other.framework** is built OK and able to be re-signed correctly. Possibly, nothing to do with Carthage.

andreyz:
Seems related to #1396

mdiep:
> `MyApp.app/Frameworks/My.framework/Frameworks/Other.framework/Other`

[...]

PR #1441, opened 2016-08-16
#1427 README upd: link dependencies from embedded frameworks to the app target

evgeniyd:
Readme is updated with respective note about embedded framework.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ carthage update Box Result
 
 If the framework you want to add to your project has dependencies explicitly listed in a [Cartfile][], Carthage will automatically retrieve them for you. You will then have to **drag them yourself into your project** from the [Carthage/Build] folder.
 
+If the embedded framework in your project has dependencies to other frameworks you must list them in the [Cartfile][], and **link them to application target** (even if application target does not have dependency to that frameworks and never uses them).
+
 ### Using submodules for dependencies
 
 By default, Carthage will directly [check out][Carthage/Checkouts] dependencies’ source files into your project folder, leaving you to commit or ignore them as you choose. If you’d like to have dependencies available as Git submodules instead (perhaps so you can commit and push changes within them), you can run `carthage update` or `carthage checkout` with the `--use-submodules` flag.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,6 +116,8 @@ carthage update Box Result
 
 If the framework you want to add to your project has dependencies explicitly listed in a [Cartfile][], Carthage will automatically retrieve them for you. You will then have to **drag them yourself into your project** from the [Carthage/Build] folder.
 
+If the embedded framework in your project has dependencies to other frameworks you must list them in the [Cartfile][], and **link them to application target** (even if application target does not have dependency to that frameworks and never uses them).
```
NachoSoto: You don't need to add them to the `Cartfile`, they will get pulled into `Cartfile.private` automatically.

The rest of the sentence is correct though! I can merge once that's updated :) thanks for adding this note.

evgeniyd: But initially it has to be put to `Cartfile`, hasn't it? Otherwise, I don't understand, how it would be added to `Cartfile.resolved`.

For example, in this issue #1401 I described my **Setup**. In order to pull and build **Other.framework** I have to put it into `Cartfile` initially. After that, the record is added to `Cartfile.resolved`. 

Other developers then, could `bootstrap` it and get all dependancies ready to be used. 

What I am missing?

Note: the project I am talking about is here: https://github.com/evgeniyd/Giraffe

mdiep: You have an app. It depends on `A.framework.`. `A.framework` depends on `B.framework`. Only `A.framework` needs to be listed in the app's `Cartfile` (`B.framework` will be built automatically), but both frameworks will need to be linked to the application target.

NachoSoto: > If the embedded framework in your project has dependencies to other frameworks ~~you must list them in the Cartfile~~

Dependencies are brought in transitively. If your `Cartfile` depends on `A`, which depends on `a1` and `a2` you don't need to add `a1` and `a2` to `Cartfile`, just `A`.

evgeniyd: But I do not have a (separate) `Cartfile` for my embedded framework project. Should I create one, add my `Other.framework` to it and then, have my app project's `Cartfile` without `Other.framework` record?

mdiep:
Thanks for the pull request!

Issue #1438, opened 2016-08-14
Carthage preserves it's output if stdout is redirected


slabko:
When I redirect carthage's output, it preserves everything it usually prints to stdout until it completely finishes (or fails). A simple example, when `carthage bootstrap --platform iOS` immediately starts printing its progress on the screen, `carthage bootstrap --platform iOS | less` won't print anything until carthage has his job done completely. Another simple example would be `carthage bootstrap --platform iOS > output &` and then `less -f output`.

Although those particular cases don't take place in real life, it is sometimes really annoying, when build server, for example, gives no input for a long time. Travis kills the build completely based on the fact that the build process has not been producing any output for 10 minutes or so.

One thing to add, this problem doesn't exist if I add `--verbose` parameter, but in this case it becomes way too verbose.

I'm wondering if there is any way to fix that problem?

* `carthage version`: 
```
0.17.2 
```

* `xcodebuild -version`: 
```
Xcode 7.3.1
Build version 7D1014
```

* I'm not using `--no-build`?
* I'm not using `--no-use-binaries`?
* I'm not using `--use-submodules`?

**Cartfile**
```
#cat Cartfile
github "ReactiveCocoa/ReactiveCocoa"      "v4.1.0"
github "Alamofire/Alamofire"              "3.3.1"
github "thoughtbot/Argo"                  "v3.0.1"
github "thoughtbot/Curry"                 "v2.2.0"
github "thoughtbot/Runes"                 "v3.2.0"
github "AliSoftware/OHHTTPStubs"          "5.0.0"
github "Swinject/Swinject"                "1.1.1"
github "SnapKit/SnapKit"                  "0.21.0"
github "yeahdongcn/UIColor-Hex-Swift"     "2.1"
github "ashleymills/Reachability.swift"   "v2.3.3"
github "stephencelis/SQLite.swift"        "0.10.1"
github "MailOnline/ImageViewer"           "2.0.6"
```

**Carthage Output**
```
#carthage bootstrap --platform iOS
*** No Cartfile.resolved found, updating dependencies
*** Fetching Result
*** Checking out Alamofire at "3.3.1"
*** Checking out Result at "2.1.3"
*** Checking out Argo at "v3.0.1"
*** Checking out OHHTTPStubs at "5.0.0"
*** Checking out Curry at "v2.2.0"
*** Checking out ImageViewer at "2.0.6"
*** Checking out Reachability.swift at "v2.3.3"
[...]

PR #1444, opened 2016-08-17
Configured stdout to always flush the buffer at the end of each line

slabko:
Fixes [#1438](https://github.com/Carthage/Carthage/issues/1438)

The change makes `printf` flush it's buffer for every new line. This is standard behaviour when it runs in TTY, but when output is file or pipe, then the buffer mode changes to block buffering.

The solution tested and it works, but probably, calling `setlinebuf` in the beginning of `main.swift` is not the best idea. If you point out the right place for it, I can easily fix that.

Looking forward for any feedback.

### DIFF
--- a/Source/carthage/main.swift
+++ b/Source/carthage/main.swift
@@ -13,6 +13,8 @@ import ReactiveCocoa
 import ReactiveTask
 import Result
 
+setlinebuf(stdout)
+
 guard ensureGitVersion().first()?.value == true else {
 	fputs("Carthage requires git \(CarthageRequiredGitVersion) or later.\n", stderr)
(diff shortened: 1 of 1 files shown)


mdiep:
I think the location of the fix is fine. I'm not well-versed in this to know whether this is the behavior that we want. @Carthage/carthage Can anyone else speak to this?

jspahrsummers:
It really depends on the use case. Very verbose output can become a bottleneck with line buffering, but building a tool that consumes Carthage's output is much easier if it can read line-by-line as they occur.

Line buffering seems fine until/unless performance is proven to be an issue.

mdiep:
🚢

PR #1447, opened 2016-08-19
Added "clean" to xcodebuild -showBuildSettings to work around rdar://27052195

salling:
This is to work around an xcodebuild bug (rdar://27052195; introduced with xcode 8) that causes a complete hang if the target contains a core data model. Further discussion here:

https://forums.developer.apple.com/thread/50372

Credits for workaround to IngmarStein in dev forums

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -476,7 +476,11 @@ public struct BuildSettings {
 	/// Upon .success, sends one BuildSettings value for each target included in
 	/// the referenced scheme.
 	public static func loadWithArguments(arguments: BuildArguments) -> SignalProducer<BuildSettings, CarthageError> {
-		// Including build action "clean" here, because otherwise Xcode 8 (at least in betas) will hang
+		// xcodebuild (in Xcode 8) has a bug where xcodebuild -showBuildSettings
+		// can hang indefinitely on projects that contain core data models.
+		// rdar://27052195
+		// Including the action "clean" works around this issue, which is further
+		// discussed here: https://forums.developer.apple.com/thread/50372
 		let task = xcodebuildTask(["clean", "-showBuildSettings"], arguments)
 
 		return launchTask(task)
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -476,7 +476,8 @@ public struct BuildSettings {
 	/// Upon .success, sends one BuildSettings value for each target included in
 	/// the referenced scheme.
 	public static func loadWithArguments(arguments: BuildArguments) -> SignalProducer<BuildSettings, CarthageError> {
-		let task = xcodebuildTask("-showBuildSettings", arguments)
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -476,7 +476,8 @@ public struct BuildSettings {
 	/// Upon .success, sends one BuildSettings value for each target included in
 	/// the referenced scheme.
 	public static func loadWithArguments(arguments: BuildArguments) -> SignalProducer<BuildSettings, CarthageError> {
-		let task = xcodebuildTask("-showBuildSettings", arguments)
+		// Including build action "clean" here, because otherwise Xcode 8 (at least in betas) will hang
+		let task = xcodebuildTask(["clean", "-showBuildSettings"], arguments)
```
mdiep: Can you include more detail in the comment? It'd be nice if it mentioned (1) that the hang occurs when the project contains a core data model and (2) the devforums post.

mdiep:
Thanks for opening a pull request! ✨

Source/CarthageKit/Xcode.swift
salling: Done!

NachoSoto:
We had just removed this behavior by default, right?

mdiep:
I think you're thinking of #1419? I was just reviewing that.

salling:
I don't think it's related. My change is exclusively about -showBuildSettings and working around xcodebuild bug that hangs if target contains a core data model.

The other PR you are referring to, would it prevent Carthage from invoking xcodebuild -showBuildSettings? Or would it also add "clean" action to this xcodbuild invokation?

Source/CarthageKit/Xcode.swift
```diff
@@ -476,7 +476,12 @@ public struct BuildSettings {
 	/// Upon .success, sends one BuildSettings value for each target included in
 	/// the referenced scheme.
 	public static func loadWithArguments(arguments: BuildArguments) -> SignalProducer<BuildSettings, CarthageError> {
-		let task = xcodebuildTask("-showBuildSettings", arguments)
+		// xcodebuild (in Xcode 8) has a bug where xcodebuild -showBuildSettings
+		// can hang indefinitely on projects that contain core data models.
+		// rdar://27052195
+		// Including the action "clean" works around this issue, which is further
+		// discussed here: https://forums.developer.apple.com/thread/50372
+		let task = xcodebuildTask(["clean", "-showBuildSettings"], arguments)
```
mdiep: Does this work if you specify the `build` action instead?

The documentation for `-showBuildSettings` explicitly mentions that it won't start a build:

> Lists the build settings in a project or workspace and scheme. Does not initiate a build. Use with -project or -workspace and -scheme.

It doesn't mention `clean` explicitly, although it makes sense for it to not clean. But if `build` works, then I think it'd be preferable to use that.

salling: It does not, unfortunately.

mdiep:
🚢

mdiep:
Thanks for the PR!

PR #1459, opened 2016-09-03
Use `NSURL.URLByStandardizingPath` in `NSURL.hasSubdirectory` extension

ikesyo:
This should be a simpler solution for the issue described in #1431.

https://developer.apple.com/reference/foundation/nsurl/1414302-standardizingpath

> - Expand an initial tilde expression using stringByExpandingTildeInPath.
> - Reduce empty components and references to the current directory (that is, the sequences “//” and “/./”) to single path separators.
> - In absolute paths only, resolve references to the parent directory (that is, the component “..”) to the real parent directory if possible using stringByResolvingSymlinksInPath, which consults the file system to resolve each potential symbolic link.
>     In relative paths, because symbolic links can’t be resolved, references to the parent directory are left in place.
> - **Remove an initial component of “/private” from the path if the result still indicates an existing file or directory (checked by consulting the file system).**

### DIFF
--- a/Source/CarthageKit/FrameworkExtensions.swift
+++ b/Source/CarthageKit/FrameworkExtensions.swift
@@ -295,7 +295,15 @@ extension NSURL {
 	}
 
 	public func hasSubdirectory(possibleSubdirectory: NSURL) -> Bool {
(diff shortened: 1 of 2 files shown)


mdiep:
✨

PR #1474, opened 2016-09-14
Update shell completion

nafu:
Update completion for
- Bash
- Zsh
- Fish

because carthage has new commands and options that isn't included in current completions.


NachoSoto:
Thanks!

PR #1472, opened 2016-09-14
Replace quotation with apostrophe

nafu:
When executing `carthage help`, it outputs four kinds of quotation.
It would be nice to use only apostrophe (\u0027) for this.

| Before | After |
| :---: | :---: |
| screen shot 2016-09-14 at 1 28 06 pm | screen shot 2016-09-14 at 1 27 05 pm |
| Full output (carthage help build) | Full output (carthage help build) |
| screen shot 2016-09-14 at 12 54 42 pm | screen shot 2016-09-14 at 12 47 16 pm |

PR #1476, opened 2016-09-15
Replace full-width apostrophe (\uff07) with apostrophe (\u0027)

nafu:
PR for https://github.com/Carthage/Carthage/pull/1472#issuecomment-247046457

Run:
```
ag -l --nocolor "’" --ignore-dir Carthage | xargs sed -i '' "s/\’/'/g"
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,18 +1,18 @@
-We love that you’re interested in contributing to Carthage!
+We love that you're interested in contributing to Carthage!
 
 ## Carthage is simple
 
-Please file issues or submit pull requests for anything you’d like to see! However, we make no promises that they’ll be accepted—many suggestions will be rejected to preserve simplicity.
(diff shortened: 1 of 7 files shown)


NachoSoto:
`'` is fine for strings in user output, but there's no reason to update the docs. `‘` is correct.

nafu:
@NachoSoto 
Thank you for your response! https://github.com/Carthage/Carthage/pull/1476#issuecomment-247226391

So, what is the difference between [`you're`](https://github.com/Carthage/Carthage/pull/1476/files#diff-04c6e90faac2675aa89e2176d2eec7d8R47) and the other [`you’re`](https://github.com/Carthage/Carthage/pull/1476/files#diff-6a3371457528722a734f3c51d9238c13R40)?
Let me know which part is inappropriate, I'll update this.

mdiep:
@nafu Sorry for the confusion! We should limit this change to strings that are reported to the user in the terminal.

`’` is a nicer rendering of `'`, which is preferable typographically, but doesn't seem to render well in some terminals. As a result, `'` is preferable for terminal output.

Issue #1222, opened 2016-03-28
Validate Carthage/Build content is up to date with Carthage.resolved


baek-jinoo:
Context: I prefer to not include Carthage directory in the git repository (I cache the .framework files for my CI).

My goal is to allow our team members to be able to run the app while not having to worry about running `carthage bootstrap` if someone updates the dependency. For example, suppose we introduce a flag **--check-cache**. So, a command such as `carthage bootstrap --check-cache` could be added in the 'Build Phases' of the Xcode project and if the content of the 'Carthage/Build' directory is up to date with Carthage.resolved, it would result in a no-op. But, if the directory's content is stale when compared to Carthage.resolved, it would execute `carthage bootstrap`.

With this, I can foresee our development workflow improving in three ways:
- If the version difference doesn't cause build errors, then developers will be doing development without the latest dependencies and find out if there is any issues when CI catches it. This is prevented.
- If the build breaks when a dev fetches a change that causes any errors due to their stale Carthage folder, it will not break their initial build and keep working.
- One less step to setup the project to run when setting up the project for the first time.

This approach is standard in other toolchains in other languages, where the build and run step's dependency step(s) checks the project dependencies as part of the flow. I believe this approach will be beneficial in this tool-chain as well.

I am willing to implement this change if you think it complies with Carthage's values. Then, I would like to get some input on how to check the staleness of the Build folder content.

mdiep:
> So, a command such as carthage bootstrap --check-cache could be added in the 'Build Phases' of the Xcode project and if the content of the 'Carthage/Build' directory is up to date with Carthage.resolved, it would result in a no-op. But, if the directory's content is stale when compared to Carthage.resolved, it would execute carthage bootstrap.

How do you propose validating that it's up-to-date?

gfontenot:
I've been doing this in our projects by copying `Cartfile.resolved` into `Carthage` and then running `cmp Cartfile.resolved ./Carthage/Cartfile.resolved`. It's super naive, but it works pretty well.

mdiep:
That will verify the _checkout_, but not the built products.

baek-jinoo:
I posted this issue to pick your (@mdiep) brain on the approach to check the staleness of the frameworks in the Build folder. The source of truth is the .framework directories, but I don't see anything in there that's obvious to me to interpret a consistent means of determining the validity of the version of the Framework, especially when there might be slight differences in OS version, Xcode version and carthage version.

[...]

PR #1489, opened 2016-09-19
Avoid rebuilding frameworks unnecessarily

jasonboyle:
Proposed solution to https://github.com/Carthage/Carthage/issues/1222.

Carthage currently rebuilds all dependencies on every update/bootstrap/build command, even if the resolved version of a given dependency has already been built and can be found in the Carthage/Build directory.  These changes create a version file for each project that associates a commitish with the SHA1s of each built framework for each platform.  When a dependency would be built, if its version file and the version files for all of its dependencies match, it is skipped instead (unless the --no-cache-builds option is provided).


jasonboyle:
It looks like the Xcode 8 tests may be failing because of https://github.com/Homebrew/brew/issues/987: "brew install bats" is failing because ruby can’t be found.

Carthage/Checkouts/Argo
```diff
@@ -1 +1 @@
-Subproject commit ecd43f7245780dfd275bc044392afef66e5786a6
+Subproject commit 327ead1fc16f13bc4a37489f5ab81c6266c44470
```
mdiep: This looks like it may have been erroneously changed.

Source/CarthageKit/BuildOptions.swift
```diff
@@ -16,11 +16,14 @@ public struct BuildOptions {
 	public let toolchain: String?
 	/// The path to the custom derived data folder.
 	public let derivedDataPath: String?
+	/// Rebuild even if cached builds exist.
+	public let ignoreCachedBuilds: Bool
```
mdiep: I'd call this `cacheBuilds`.

Source/CarthageKit/NSData+SHA1.swift
```diff
@@ -0,0 +1,230 @@
...
+	}
+}
+
+private struct BytesSequence: SequenceType {
+	let chunkSize: Int
+	let data: Array<UInt8>
+	
+	func generate() -> AnyGenerator<ArraySlice<UInt8>> {
+		
+		var offset:Int = 0
```
mdiep: Style here and else where: Please put a space between the `:` and the type.

Source/CarthageKit/NSData+SHA1.swift
```diff
@@ -0,0 +1,230 @@
...
+	private class func withBytes(bytes: Array<UInt8>) -> NSData {
+		return NSData(bytes: bytes, length: bytes.count)
+	}
+}
+
+private extension _ArrayType where Generator.Element == UInt8 {
+	func toHexString() -> String {
+		return self.lazy.reduce("") { $0 + String(format:"%02x", $1) }
+	}
+}
```
mdiep: It might be worth pulling in a dependency to do this for us.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,102 @@
...
+import Foundation
+
+private let commitishKeyName = "commitish"
+private let cachedFrameworksKeyName = "cachedFrameworks"
+private let frameworkNameKeyName = "frameworkName"
+private let frameworkSha1KeyName = "frameworkSha1"
+
+private typealias VersionFile = [String: [String: AnyObject]]
+
+public func createVersionFileForDependency(dependency: Dependency<PinnedVersion>, forPlatforms platforms: Set<Platform>, buildProductURLs: [NSURL], rootDirectoryURL: NSURL) -> Bool {
```
mdiep: All the public interfaces should get documentation.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,102 @@
...
+public func versionFileMatchesDependency(dependency: Dependency<PinnedVersion>, forPlatforms platforms: Set<Platform>, rootDirectoryURL: NSURL) -> Bool {
+	let rootBinariesURL = rootDirectoryURL.appendingPathComponent(CarthageBinariesFolderPath, isDirectory: true).URLByResolvingSymlinksInPath!
+	let versionFileURL = rootBinariesURL.appendingPathComponent(".\(dependency.project.name).version")
+	guard let versionInfo = readVersionFileAtURL(versionFileURL) else { return false }
+	
+	let platformNames: [String] = !platforms.isEmpty ? platforms.map { $0.rawValue } : Array(versionInfo.keys)
+	guard platformNames.count > 0 else { return false }
+	
+	for platformName in platformNames {
+		guard let platformInfo: [String: AnyObject] = versionInfo[platformName] else { return false }
```
mdiep: This versioning information seems like it deserves a proper model type, rather than just dictionary lookups.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,102 @@
...
+	do {
+		let jsonData = try NSJSONSerialization.dataWithJSONObject(versionInfo, options: .PrettyPrinted)
+		try jsonData.writeToURL(versionFileURL, options: .DataWritingAtomic)
+	}
+	catch { return false }
+	
+	return true
+}
+
+public func versionFileMatchesDependency(dependency: Dependency<PinnedVersion>, forPlatforms platforms: Set<Platform>, rootDirectoryURL: NSURL) -> Bool {
```
mdiep: It doesn't make sense to me to have a function that lookups up and compares both version files.

It makes more sense to have an `Equatable` type that can be computed from each separately and then compared with `==` (or another method).

Source/CarthageKit/Xcode.swift
```diff
@@ -1180,6 +1240,8 @@ public func buildInDirectory(directoryURL: NSURL, withOptions options: BuildOpti
 		let locator = buildableSchemesInDirectory(directoryURL, withConfiguration: options.configuration, forPlatforms: options.platforms)
 			.replayLazily(Int.max)
 
+		var urls: [NSURL] = []
```
mdiep: It'd be preferable to not use local state to pass information outside of the signal chain.

Source/CarthageKit/Xcode.swift
```diff
@@ -1242,6 +1309,13 @@ public func buildInDirectory(directoryURL: NSURL, withOptions options: BuildOpti
 				return BuildSchemeProducer(value: .Success(initialValue))
 					.concat(buildProgress)
 			}
+			.on(completed: {
+				if let dependency = dependency, rootDirectoryURL = rootDirectoryURL {
+					if !createVersionFileForDependency(dependency, forPlatforms: options.platforms, buildProductURLs: urls, rootDirectoryURL: rootDirectoryURL) {
+						observer.sendFailed(CarthageError.CanNotBuildVersionFile(dependency.project))
```
mdiep: This doesn't seem like it should be an error. If the version file can't be created, should the build really fail?

Source/carthage/Build.swift
```diff
@@ -30,6 +30,7 @@ extension BuildOptions: OptionsType {
 			<*> m <| Option(key: "platform", defaultValue: .All, usage: "the platforms to build for (one of ‘all’, ‘Mac’, ‘iOS’, ‘watchOS’, 'tvOS', or comma-separated values of the formers except for ‘all’)" + addendum)
 			<*> m <| Option<String?>(key: "toolchain", defaultValue: nil, usage: "the toolchain to build with")
 			<*> m <| Option<String?>(key: "derived-data", defaultValue: nil, usage: "path to the custom derived data folder")
+			<*> m <| Option(key: "ignore-cached-builds", defaultValue: false, usage: "ignore cached build products")
```
mdiep: I'd like to see this be `cache-builds` with a default value of `true`.

Source/CarthageKit/VersionFile.swift
jasonboyle: There are no comparisons between version files in the current code; a version file matches if the commitish matches and the recorded SHA1s match the SHA1s of the corresponding frameworks in the Build directory for a given platform.

jasonboyle:
@mdiep  Thanks for your feedback.  I believe I've addressed most of your comments, with the exception of adding more granular unit tests and pulling in a dependency for the SHA1 algorithm (which I can look into later).  I would grateful if you could take another look over the code when you have time.

JonB:
@jasonboyle @baek-jinoo: This looks great and can't wait for it to be merged in for the benefits!

I just had one quick query about the use of NSData+SHA1.swift...

Roughly what size chunks of NSData end up having their SHA1 has calculated and have you tried profiling the speed of this SHA1 computation?

In some work projects we've been computing the SHA1 hash of 4Mb NSData chunks and found the use of CryptoSwift to be an order of magnitude slower, seconds vs milliseconds, than IDZCommonCrypto (which is a swift wrapper of the C lib CommonCrypto, as opposed to a reimplementation like CryptoSwift).

For our work projects the speed difference is important (and we switched) but is this something that Carthage needs to be worried about? I'm pretty certain if the NSData chunks are small then the speed difference wouldn't be noticeable but thought it worth flagging

baek-jinoo:
@JonB If what you're saying is relevant to what we pulled in from CryptoSwift, we should definitely do some testing to see how long it takes to get the SHA1 from a decently big binary file (it took about 500+ ms for the `Madness` 620KB  binary blob in the spec). For example, realm framework.zip has a 50+ MB binary file for iOS. If it takes seconds for 4MB NSData blobs, that would be a terrible experience especially when we expect this to be a "no-op".

```
start sha1 creation: 1475135246.4666
sha1 created: 1475135247.00945
```

BobElDevil:
In our preliminary checks when working with @jasonboyle we did notice that CryptoSwift seemed to be taking a huge amount of time (seconds) for the debug build but was near instant for the Release build. Didn't look any further but maybe it's whole module optimization or something similar where it can optimize away a ton of dynamic dispatch during the computation loop.

brentleyjones:
This seems to have stalled out. Is the SHA1 calculation speed holding this up, or is there more that needs to be changed from @mdiep's review?

jasonboyle:
@brentleyjones I believe @mdiep's comments have been addressed, and the code is ready for a more in-depth review.  I don't think the SHA1 calculation should be a hold up as it can be swapped out for a more efficient implementation if necessary.

baek-jinoo:
Especially being pretty sure that an inefficient SHA1 algorithm is probably way faster than rebuilding, we should not let that block this.

Jin

mdiep:
This is on my list, but I've been very busy. I'm starting to catch back up on things this week, and hope to get to this soon.

brentleyjones:
Awesome. I'm so looking forward to this finally making it in.

akyri:
Has there been any progress on this? It will be excellent to finally have this functionality in Carthage.

mdiep:
Still working on it. My time is unfortunately quite finite. ☺️

akyri:
Of course. Please let me know if there's anything I can do to help :)

Dschee:
It's been over a month again now, if we can be of any help @mdiep just tell us, I think there are quite a lot of people who want this to be finished sooner rather than later. What steps need to be made? Maybe we can help with some?

Source/CarthageKit/Project.swift
```diff
@@ -313,15 +316,20 @@ public final class Project {
...
 	}
 	
-	/// Loads the dependencies for the given dependency, at the given version.
-	private func dependencies(for dependency: Dependency<PinnedVersion>) -> SignalProducer<Dependency<VersionSpecifier>, CarthageError> {
+	/// Loads the Cartfile for the given dependency, at the given version.
+	private func cartfile(for dependency: Dependency<PinnedVersion>) -> SignalProducer<Cartfile, CarthageError> {
 		let revision = dependency.version.commitish
 		return self.cloneOrFetchDependency(dependency.project, commitish: revision)
-			.flatMap(.concat) { repositoryURL in
+			.flatMap(.Concat) { repositoryURL in
```
ikesyo: Please use `.concat` shim: #1584

Source/CarthageKit/Project.swift
```diff
@@ -313,15 +316,20 @@ public final class Project {
 			}
 	}
 	
-	/// Loads the dependencies for the given dependency, at the given version.
-	private func dependencies(for dependency: Dependency<PinnedVersion>) -> SignalProducer<Dependency<VersionSpecifier>, CarthageError> {
+	/// Loads the Cartfile for the given dependency, at the given version.
+	private func cartfile(for dependency: Dependency<PinnedVersion>) -> SignalProducer<Cartfile, CarthageError> {
```
ikesyo: This was changed in #1674.

Source/CarthageKit/Xcode.swift
```diff
@@ -668,7 +668,7 @@ public func buildInDirectory(directoryURL: URL, withOptions options: BuildOption
 					.take(first: 1)
 					.map { project, _ in (scheme, project) }
 			}
-			.map { (scheme: String, project: ProjectLocator) -> BuildSchemeProducer in
+			.flatMap(.Concat) { (scheme: String, project: ProjectLocator) -> SignalProducer<TaskEvent<URL>, CarthageError> in
```
ikesyo: Ditto (`.concat` shim)

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,221 @@
...
+
+private typealias VersionFileData = [String: [String: AnyObject]]
+
+/// Creates a version file for the current dependency in the
+/// Carthage/Build directory which associates its commitish with
+/// the SHA1s of the built frameworks for each platform in order
+/// to allow those frameworks to be skipped in future builds.
+///
+/// Returns true if the version file was successfully created.
+public func createVersionFileForDependency(dependency: Dependency<PinnedVersion>, forPlatforms platforms: Set<Platform>, buildProductURLs: [NSURL], rootDirectoryURL: NSURL) -> Bool {
```
mdiep: 1. This seems like it should, at the very least, use a `write(to url: URL)` method on `VersionFile`.

2. It should return a `SignalProducer` so that it can be a non-blocking call.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,221 @@
...
+		var frameworks = frameworksByPlatform[platformName] ?? []
+		frameworks.append(cachedFramework)
+		frameworksByPlatform[platformName] = frameworks
+	}
+	
+	var cachedPlatforms: [String: CachedPlatform] = frameworksByPlatform.reduce([:]) { (cachedPlatforms, entry: (platformName: String, cachedFrameworks: [CachedFramework])) in
+		var cachedPlatforms = cachedPlatforms
+		cachedPlatforms[entry.platformName] = CachedPlatform(commitish: dependency.version.commitish, cachedFrameworks: entry.cachedFrameworks)
+		return cachedPlatforms
+	}
```
mdiep: The above work seems like it should be done in an initializer on `VersionFile`.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,221 @@
...
+	let rootBinariesURL = rootDirectoryURL.appendingPathComponent(CarthageBinariesFolderPath, isDirectory: true).URLByResolvingSymlinksInPath!
+	let versionFileURL = rootBinariesURL.appendingPathComponent(".\(dependency.project.name).version")
+	
+	if let oldVersionFile: VersionFile = readVersionFileAtURL(versionFileURL) {
+		for platform in Platform.supportedPlatforms {
+			if let oldCachedPlatform = oldVersionFile.cacheForPlatform(platform) where cachedPlatforms[platform.rawValue] == nil {
+				cachedPlatforms[platform.rawValue] = oldCachedPlatform
+			}
+		}
+	}
```
mdiep: I think we should rip this bit out. I don't think it's wise to cache on a per-platform basis.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,221 @@
...
+	guard let versionFile: VersionFile = readVersionFileAtURL(versionFileURL) else { return false }
+	let commitish = dependency.version.commitish
+	
+	let cachedPlatforms = versionFile.cachedPlatforms()
+	let platformsToCheck = platforms.isEmpty ? cachedPlatforms : platforms
+	for platform in platformsToCheck {
+		if !checkPlatform(platform, versionFile: versionFile, commitish: commitish, rootDirectoryURL: rootDirectoryURL) {
+			return false
+		}
+	}
```
mdiep: The work of checking whether a platform can be skipped should be in a method on `VersionFile`.

Source/CarthageKit/Xcode.swift
```diff
@@ -1084,26 +1083,88 @@ public typealias BuildSchemeProducer = SignalProducer<TaskEvent<(ProjectLocator,
...
+					return .NoSharedFrameworkSchemes(dependency.project, platforms)
+
+				case let (.GitHub(repo), .NoSharedSchemes(project, _)):
+					return .NoSharedSchemes(project, repo)
+
+				default:
+					return error
+				}
+			}
+
```
mdiep: Why did all of this change? Is this fallout from a bad merge?

Source/CarthageKit/Xcode.swift
```diff
@@ -1225,26 +1285,50 @@ public func buildInDirectory(directoryURL: NSURL, withOptions options: BuildOpti
...
+							}
+						})
+						.map { (urls: [NSURL]) -> TaskEvent<(ProjectLocator, String)> in
+							return .Success(ignoredValue)
+						}
+						.filter { taskEvent in
+							return false
+						}
+				
+				Signal.merge([eventSignal, versionFileSignal]).observe(observer)
```
mdiep: This work (creating signals and merging them) shouldn't be done inside `startWithSignal`. Most likely you want to use `flatMap` instead.

Source/carthage/Build.swift
```diff
@@ -30,6 +30,7 @@ extension BuildOptions: OptionsType {
 			<*> m <| Option(key: "platform", defaultValue: .All, usage: "the platforms to build for (one of 'all', 'macOS', 'iOS', 'watchOS', 'tvOS', or comma-separated values of the formers except for 'all')" + addendum)
 			<*> m <| Option<String?>(key: "toolchain", defaultValue: nil, usage: "the toolchain to build with")
 			<*> m <| Option<String?>(key: "derived-data", defaultValue: nil, usage: "path to the custom derived data folder")
+			<*> m <| Option(key: "cache-builds", defaultValue: true, usage: "use cached builds when possible")
```
mdiep: This should default to `false` for the initial release, I think. Then people can try it out and use it—but if there are bugs, then it won't bite everybody.

Source/carthage/Build.swift
```diff
@@ -30,6 +30,7 @@ extension BuildOptions: OptionsType {
 			<*> m <| Option(key: "platform", defaultValue: .all, usage: "the platforms to build for (one of 'all', 'macOS', 'iOS', 'watchOS', 'tvOS', or comma-separated values of the formers except for 'all')" + addendum)
 			<*> m <| Option<String?>(key: "toolchain", defaultValue: nil, usage: "the toolchain to build with")
 			<*> m <| Option<String?>(key: "derived-data", defaultValue: nil, usage: "path to the custom derived data folder")
+			<*> m <| Option(key: "cache-builds", defaultValue: true, usage: "use cached builds when possible")
```
mdiep: I think this should default to `false` for the initial release. Then interested users can try it out, but the majority won't hit any bugs that are found.

Source/CarthageKit/Xcode.swift
```diff
@@ -311,18 +311,17 @@ private func mergeBuildProductsIntoDirectory(firstProductSettings: BuildSettings
...
 /// A callback function used to determine whether or not an SDK should be built
 public typealias SDKFilterCallback = (sdks: [SDK], scheme: String, configuration: String, project: ProjectLocator) -> Result<[SDK], CarthageError>
 
 /// Builds one scheme of the given project, for all supported SDKs.
 ///
 /// Returns a signal of all standard output from `xcodebuild`, and a signal
 /// which will send the URL to each product successfully built.
-public func buildScheme(scheme: String, withConfiguration configuration: String, inProject project: ProjectLocator, workingDirectoryURL: URL, derivedDataPath: String?, toolchain: String?, sdkFilter: SDKFilterCallback = { .success($0.0) }) -> SignalProducer<TaskEvent<URL>, CarthageError> {
-	precondition(workingDirectoryURL.isFileURL)
+public func buildScheme(scheme: String, withOptions options: BuildOptions, inProject project: ProjectLocator, workingDirectoryURL: URL, dependency: Dependency<PinnedVersion>? = nil, sdkFilter: SDKFilterCallback = { .success($0.0) }) -> SignalProducer<TaskEvent<URL>, CarthageError> {
```
mdiep: Why was `dependency` added here. It doesn't seem to be used?

Source/CarthageKit/Xcode.swift
BobElDevil: I believe we did it this way to have the "Building Scheme" events get sent out right away (so users get immediate feedback for xcode output and the like), while still allowing the version file signal to 'wait' until all the builds are done. If we used flatMap, we'd also require a 'collect' or something similar, which would stall out the xcode output

mdiep: In that case, you should be able to use `flatMapTaskEvents`? (Or whatever that's called.)

Source/CarthageKit/VersionFile.swift
BobElDevil: If you don't cache per platform you blow away state when using the '--platform' flag. For example if you first build with '--platform ios' then later '--platform mac', then build with '--platform ios' again, it won't be cached. 

Additionally you would still the platform variance somewhere, as the SHA's will be different / there may be different counts of frameworks in the different platforms (e.g., they have an ios only target and a mac os only target).

mdiep: > If you don't cache per platform you blow away state when using the '--platform' flag. For example if you first build with '--platform ios' then later '--platform mac', then build with '--platform ios' again, it won't be cached.

Yup. I think that's the right thing to do here. If you're using the cache, there's no reason to build on a per-platform basis like that. (You may want to exclude platforms you're not using at all, but you may as well build all of the platforms you _are_ using at once.)

> Additionally you would still the platform variance somewhere, as the SHA's will be different / there may be different counts of frameworks in the different platforms (e.g., they have an ios only target and a mac os only target).

Sorry, yeah. I just meant that old ones shouldn't be preserved.

Source/CarthageKit/Xcode.swift
BobElDevil: I'm still not sure flatMap can solve our problem here. The issue is that one stream (the `versionFileSignal`) has to wait until the end to perform some action: mainly writing out the version file. Notice the call to 'collect' midway through the definition for it.

The `eventSignal` on the other hand needs to be sending it's events upstream in real time, as the Task events contain the xcode stdout. Basically we want to fork the signal, so we can collect on it, but also let it continue.  

If we just used flatMap (or flatMapTaskEvents), we can't collect on the results, we still only get one url at a time. methods like 'scan' or other "collect but also send events" didn't work because we couldn't know when it was done, and injecting a 'takeLast' just halts the whole stream again. 

The other approach we had was to subscribe to the signal, so we could get notified of when it was done, but that either requires a collect again (which is essentially this solution here), or a local array, which is what we had done before, but it created local state which was undesired. 

We may be missing a solution to our problem, and I'm all for making it more streamlined without needing to do this merge call.  In all my searching around about 'forking' or 'splitting' signals/producers, it's either something that would need to be done higher up (at the subscribe level), or taking this approach. (e.g., (ReactiveCocoa/ReactiveCocoa#2243)

mdiep: I'll take a closer look after the other feedback is addressed. 👍

mz2:
I've tested the development version of this on two machines now for an app with a heavy use of Carthage based dependencies, some of them internal (= there is a big need for this). Thanks a lot for working on this – the feature appears to work, except at least with the `build` command it appears necessary to pass `--cache-builds` (i.e. it's not on by default) or the skipping behaviour will never happen. I noticed a commit above which I interpreted as enabling `--cache-builds` by default so just reporting this in case it's not expected behaviour and not a known issue.

mz2:
(Oops, looks like I just failed to spot https://github.com/Carthage/Carthage/pull/1489#discussion_r95500035 – feel free to delete these comments)

mz2:
I noticed a subtle issue, reproducible with the following minimal `Cartfile` contents:

```
github "erikdoe/ocmock" == 3.2.2
```

- If I now do repeated rebuilds with `carthage bootstrap --cache-builds --use-ssh --platform ios,watchOS` I observe that skipping does *not* happen.
- If I use `--platform ios` instead of `--platform ios,watchos` the skipping *does* happen.
- If I use `--platform watchos`, then I see why this probably is not working: `Dependency "ocmock" has no shared framework schemes for any of the platforms: watchOS`

I am guessing some comparison is made to the platforms provided as arguments, instead of the ones actually available, and then elsewhere the comparison is made against what actually is availalbe. You could treat it as an issue in OCMock I suppose, but the context is that you may want to use OCMock in tests on iOS only and some other dependencies you need at deploy or testing time, expressed in that same Cartfile, will build on watchOS (… and not necessarily for iOS of course). That is, you may have other Carthage built dependencies that are genuinely platform dependent and you use --platform to limit yourself to the union of those platforms.

jeeftor:
I'm assuming this is almost done?

BobElDevil:
@mdiep, @ikesyo this is ready for another look over

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,270 @@
...
+	
+	let macOS: [CachedFramework]?
+	let iOS: [CachedFramework]?
+	let watchOS: [CachedFramework]?
+	let tvOS: [CachedFramework]?
+	
+	static let commitishKey = "commitish"
+	static let xcodeVersionKey = "xcodeVersion"
+	
+	func cachesForPlatform(_ platform: Platform) -> [CachedFramework]? {
```
mdiep: Maybe make this a `subscript`?

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,270 @@
...
+		case .iOS:
+			return iOS
+		case .watchOS:
+			return watchOS
+		case .tvOS:
+			return tvOS
+		}
+	}
+	
+	func cachedPlatforms() -> Set<Platform> {
```
mdiep: This can be a `var` instead of a `func`.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,270 @@
...
+		}
+		self = versionFile
+	}
+	
+	func check(_ platform: Platform, commitish: String, xcodeVersion: String, rootDirectoryURL: URL) -> SignalProducer<Bool, CarthageError> {
+		guard commitish == self.commitish && xcodeVersion == self.xcodeVersion else {
+			return .init(value: false)
+		}
+		
+		let rootBinariesURL = rootDirectoryURL.appendingPathComponent(CarthageBinariesFolderPath, isDirectory: true).resolvingSymlinksInPath()
```
mdiep: The URL of the binaries folder should be passed in directly—instead of putting the logic here.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,270 @@
...
+		guard FileManager.default.fileExists(atPath: url.path),
+			let jsonData = try? Data(contentsOf: url),
+			let json = try? JSONSerialization.jsonObject(with: jsonData, options: .allowFragments),
+			let versionFile: VersionFile = Argo.decode(json) else {
+			return nil
+		}
+		self = versionFile
+	}
+	
+	func check(_ platform: Platform, commitish: String, xcodeVersion: String, rootDirectoryURL: URL) -> SignalProducer<Bool, CarthageError> {
```
mdiep: I think this name could use improvement. `satisfies(platform:commitish:xcodeVersion:rootDirectoryURL)` or `isValidFor(...)` or something.

mdiep: I'd like to see a pure version of this function that doesn't read from disk—along with tests. (Everything should be passed in as a value.) Then the side-effecting version can use the pure version.

Source/CarthageKit/Errors.swift
```diff
@@ -89,6 +89,9 @@ public enum CarthageError: Error, Equatable {
 
 	/// An error occurred while shelling out.
 	case taskError(TaskError)
+	
+	/// An error occurred while generating the version file.
+	case versionFileError(description: String)
```
mdiep: `CarthageError` should really only be for user-visible failures.

Since caching is entirely optional, this error case should be removed. (If you need an error internally, you can add an `internal struct`.)

Source/CarthageKit/Project.swift
```diff
@@ -120,6 +120,9 @@ public enum ProjectEvent {
 	/// Building the project is being skipped, since the project is not sharing
 	/// any framework schemes.
 	case skippedBuilding(ProjectIdentifier, String)
+	
+	/// Building the project is being skipped because it is cached.
+	case skippedBuildingCached(ProjectIdentifier)
```
mdiep: Maybe we should add a case for `rebuildingCachedProject`. 🤔

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,270 @@
...
+				}
+				.collect()
+				.flatMap(.concat) { _ -> SignalProducer<(), CarthageError> in
+					return writeVersionFile
+				}
+		}
+	} else {
+		// Write out an empty version file for dependencies with no built frameworks, so cache builds can differentiate between
+		// no cache and a dependency that has no frameworks
+		return writeVersionFile
```
mdiep: Why is this important? Either way, no work should be done, right?

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,270 @@
...
+				.flatMap(.merge) { platform in
+					return versionFile.check(platform, commitish: commitish, xcodeVersion: xcodeVersion, rootDirectoryURL: rootDirectoryURL)
+				}
+				.reduce(true) { current, result in
+					return current && result
+				}
+		}
+}
+
+private func currentXcodeVersion() -> SignalProducer<String, CarthageError> {
```
mdiep: Can you move this to `Xcode.swift`? This is a good function to have generally.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,270 @@
...
+	let task = Task("/usr/bin/xcrun", arguments: ["xcodebuild", "-version"])
+	return task.launch()
+		.mapError(CarthageError.taskError)
+		.ignoreTaskData()
+		.attemptMap { data in
+			guard let versionString = String(data: data, encoding: .utf8) else {
+				return .failure(.versionFileError(description: "Could not get xcode version"))
```
mdiep: When you move this, it can become a `readFailed` for now.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,270 @@
...
+
+			return .success(versionString.trimmingCharacters(in: .whitespacesAndNewlines))
+	}
+}
+
+private func md5ForFileAtURL(_ frameworkFileURL: URL) -> SignalProducer<String, CarthageError> {
+	guard FileManager.default.fileExists(atPath: frameworkFileURL.path) else {
+		return .init(error: .versionFileError(description: "File does not exist for md5 generation \(frameworkFileURL)"))
+	}
+	let task = Task("/usr/bin/env", arguments: ["md5", "-q", frameworkFileURL.path])
```
mdiep: Is there a reason you settled on MD5 instead of SHA?

BobElDevil: No technical reason, it's just since we are using the command line version now, the `md5` tool has a `-q` option which prevents it from spitting out anything except the checksum, while the `shasum` tool also always spit out the filename. So it was a matter of convenience for the parsing.

We can always switch it back to sha1 if there's some reason we didn't think of to prefer it.

Cartfile.resolved
```diff
@@ -1,12 +1,12 @@
...
+github "jdhealy/PrettyColors" "6f54504f8715c5d51114bcff821c48fd87d8a8b7"
 github "jspahrsummers/xcconfigs" "0.9"
-github "thoughtbot/Argo" "v4.1.2"
-github "Carthage/Commandant" "0.11.3"
-github "ReactiveCocoa/ReactiveSwift" "1.0.0"
-github "Carthage/ReactiveTask" "96e1013e0a6782bf0faba27485fe2d0a1153b531"
 github "mdiep/Tentacle" "0.5.0"
+github "thoughtbot/Argo" "v4.1.2"
+github "thoughtbot/Curry" "v3.0.0"
+github "thoughtbot/Runes" "v4.0.1"
```
BobElDevil: Sure, this was just the output from running a carthage update to double check the versions were good. Since the versions are the same i'll revert.

Documentation/VersionFile.md
```diff
@@ -0,0 +1,27 @@
...
+
+    {
+      "commitish" : "1.6.0",
+      "Mac" : [
+       {
+           "sha1" : "ad4c4bfe83c546d18418825b8c481df11b3910cc",
+           "name" : "Prelude"
+       }
+      ]
+    }
```
BobElDevil: Will update

Source/carthage/Extensions.swift
```diff
@@ -116,6 +116,9 @@ internal struct ProjectEventSink {
 
 		case let .skippedBuilding(project, message):
 			carthage.println(formatting.bullets + "Skipped building " + formatting.projectName(project.name) + " due to the error:\n" + message)
+			
+		case let .skippedBuildingCached(project):
+			carthage.println(formatting.bullets + "Skipped building " + formatting.projectName(project.name))
```
BobElDevil: 👍

Source/CarthageKit/VersionFile.swift
BobElDevil: 👍

Source/CarthageKit/VersionFile.swift
BobElDevil: Turns out this wasn't needed at all. The only call site was actually a bug. We need to use the set of all platforms rather than cachedPlatforms when provided with an empty platform set (basically no `--platform` option specified), so it's going to be removed.

Source/CarthageKit/VersionFile.swift
BobElDevil: ok, will rename and create a purely functional version as well

Source/CarthageKit/VersionFile.swift
BobElDevil: It's important because the commitish still needs to be validated to determine if downstream dependencies should be rebuilt. For example, if you have an xcodeConfigs repo that changes it's version that dependency B relies on, then you'll want to recompile B even if B is still the same version that was last compiled. 

Writing out an empty file distinguishes between "We ran a build, and determined that this dependency had no frameworks" and "This dependency didn't exist / wasn't considered during the last build, so rebuild"

Source/CarthageKit/VersionFile.swift
BobElDevil: 👍

Source/CarthageKit/VersionFile.swift
BobElDevil: Hm, so readFailed takes a non-optional URL, and we're not reading a url really, just text from a command output. There doesn't seem to be an appropriate existing error to use here other than possibly 'parseError', though that's commented to be specific to Cartfile parsing.

Source/CarthageKit/Project.swift
BobElDevil: we could do that, though it would require changing the logic a bit. currently if `--cache-builds` isn't provided we completely bypass the version file check. If we want to detect this case we'll have to always check the version file, and then spit out this case when we build anyway.

Unless you're only referring to the case where we rebuild because there was a version file mismatch, explicitly pointing out that there has been an update to account for

mdiep: Yeah, I was only referring to the latter case.

Source/CarthageKit/VersionFile.swift
BobElDevil: Ok, looked around, and it appears 'parseError' is used in other similar situations (like when a git task fails to read from ls-tree), so it's just the comment that's wrong. Using parse error here as well, and updating the comment

Source/CarthageKit/Project.swift
jasonboyle: @mdiep Added logging to report when a valid/invalid cache or no cache is found.  I believe this is PR ready for another review.

BobElDevil:
Also reading through the comments, @mz2 I believe the latest version should handle what you're talking about. We now write out a version file with an empty array of frameworks for a platform if nothing was generated there (like ocmock), so that on the next pass it knows that ocmock explicitly generated nothing, and can be counted as 'cached' for watchOS

Source/CarthageKit/Project.swift
```diff
@@ -733,20 +742,46 @@ public final class Project {
...
 	/// optionally they are limited by the given list of dependency names.
+	/// Cached dependencies whose dependency trees are also cached will not
+	/// be rebuilt unless otherwise specified via build options.
 	///
 	/// Returns a producer-of-producers representing each scheme being built.
 	public func buildCheckedOutDependenciesWithOptions(_ options: BuildOptions, dependenciesToBuild: [String]? = nil, sdkFilter: @escaping SDKFilterCallback = { .success($0.0) }) -> SignalProducer<BuildSchemeProducer, CarthageError> {
 		return loadResolvedCartfile()
 			.flatMap(.merge) { resolvedCartfile in
 				return self.buildOrderForResolvedCartfile(resolvedCartfile, dependenciesToInclude: dependenciesToBuild)
+					.flatMap(.concat) { dependency -> SignalProducer<(Dependency<PinnedVersion>, Set<ProjectIdentifier>), CarthageError> in
```
mdiep: All of this can be outdented a level.

```swift
return loadResolvedCartfile()
    .flatMap(.merge) { resolvedCartfile in
        return self.buildOrderForResolvedCartfile(resolvedCartfile, dependenciesToInclude: dependenciesToBuild)
    }
    .flatMap(.concat) { dependency -> SignalProducer<(Dependency<PinnedVersion>, Set<ProjectIdentifier>), CarthageError> in
        return self
            .dependencyProjects(for: dependency)
            .map { projects in (dependency, projects) }
     }
     .reduce([]) { ... }
```

Source/CarthageKit/Project.swift
```diff
@@ -733,20 +742,46 @@ public final class Project {
...
+						self.dependencyProjects(for: dependency)
+							.map { projects in (dependency, projects) }
```
mdiep: Please use `return` if a closure spans multiple lines.

Also, please follow the general indentation pattern we use.

```swift
        return self
            .dependencyProjects(for: dependency)
            .map { projects in (dependency, projects) }
```

Source/CarthageKit/Project.swift
```diff
@@ -733,20 +742,46 @@ public final class Project {
...
+						}
+
+						var dependenciesToBuild = includedDependencies
+						_ = self.shouldSkipBuildForDependency(nextDependency, dependencyProjects: projects, dependenciesToBeBuilt: includedDependencies, platforms: options.platforms)
+							.on(value: { shouldSkip in
+								if !shouldSkip {
+									dependenciesToBuild.append(nextDependency)
+								}
+							})
+							.wait() // We must wait since subsequent calls to reduce rely on the results of the previous project skipping or not
```
mdiep: I think it should be possible to do this without a `wait` if you move this check:

```swift
		let projectsToBeBuilt = Set(dependenciesToBeBuilt.map { $0.project })
		guard dependencyProjects.intersection(projectsToBeBuilt).isEmpty else {
			return .init(value: false)
		}
```

1. Get build order
2. Get dependencies for project
3. Get whether project has cached version
4. Reduce `(Dependency, Set<ProjectIdentifier>, Bool)` into a set of projects to be built

README.md
```diff
@@ -136,6 +136,10 @@ If you want to work on your dependencies during development, and want them to be
 
 Note that you should be [using submodules](#using-submodules-for-dependencies) before doing this, because plain checkouts [should not be modified][Carthage/Checkouts] directly.
 
+### Caching builds
+
+By default Carthage will avoid rebuilding a dependency if its resolved version and the hash of its built frameworks are unchanged for a given platform.  Caching information is stored in invisible version files in the Build directory, which are described in more detail in Carthage’s documentation.  If the --no-cache-builds option is provided, both cached and non-cached frameworks will be built.
```
BobElDevil: 👍

Source/CarthageKit/Project.swift
BobElDevil: 👍

Source/CarthageKit/Project.swift
BobElDevil: 👍

Source/CarthageKit/Project.swift
BobElDevil: Looks like that worked. Done

Source/CarthageKit/Project.swift
```diff
@@ -736,20 +739,46 @@ public final class Project {
...
+
+						return dependenciesToBuild
+					}
+					.flatMap(.concat) { dependencies -> SignalProducer<Dependency<PinnedVersion>, CarthageError> in
+						SignalProducer<Dependency<PinnedVersion>, CarthageError>(dependencies)
+					}
```
BobElDevil: Tried using flatten instead but unfortunately the swift compiler claims there's ambiguity when trying to make the flatten call. And since flatten doesn't take a closure that we can add a type to to nudge the compiler in the right direction, I think we have to stick with the manual flatMap.

BobElDevil:
Latest batch of review fixes pushed

Source/CarthageKit/Project.swift
```diff
@@ -733,20 +742,57 @@ public final class Project {
...
+				return self
+					.dependencyProjects(for: dependency)
+					.map { projects in (dependency, projects) }
+			}
+			.flatMap(.concat) { nextGroup -> SignalProducer<(Dependency<PinnedVersion>, Set<ProjectIdentifier>, Bool?), CarthageError> in
+				let (dependency, nestedDependencies) = nextGroup
+				return versionFileMatchesDependency(dependency, forPlatforms: options.platforms, rootDirectoryURL: self.directoryURL)
+					.map { matches in
+						return (dependency, nestedDependencies, matches)
+					}
```
mdiep: Now that we've got the structure cleaned up a bit, we can do this:

```swift
return loadResolvedCartfile()
    .flatMap(.concat) { resolvedCartfile -> SignalProducer<Dependency<PinnedVersion>, CarthageError> in
        return self.buildOrderForResolvedCartfile(resolvedCartfile, depndenciesToInclude: dependenciesToBuild)
    }
    .flatMap(.concat) { dependency in
        return SignalProducer.combineLatest(
            SignalProducer(value: dependency),
            dependencyProjects(for: dependency),
            versionFileMatches(dependency, for: options.platforms, rootDirectory: self.directoryURL)
        )
    }
    .reduce([]) { ... }
```

Source/CarthageKit/Project.swift
```diff
@@ -733,20 +742,57 @@ public final class Project {
...
+			}.reduce([]) { (includedDependencies, nextGroup) -> [Dependency<PinnedVersion>] in
```
mdiep: This `reduce` should start on its own line.

mdiep: You can also get rid of `nextGroup`:

```swift
.reduce([]) { (includedDependencies, (nextDependency, projects, matches)) -> [Dependency<PinnedVersion>] in
```

Source/CarthageKit/Project.swift
```diff
@@ -762,6 +808,37 @@ public final class Project {
 					}
 			}
 	}
+	
+	/// Checks the version files for the dependency and each of the dependencies
+	/// listed in its Cartfile.
+	///
+	/// Returns true if the dependency does not need to be rebuilt.
+	private func shouldSkipBuildForDependency(_ dependency: Dependency<PinnedVersion>, dependencyProjects: Set<ProjectIdentifier>, dependenciesToBeBuilt: [Dependency<PinnedVersion>], platforms: Set<Platform>) -> SignalProducer<Bool, CarthageError> {
```
mdiep: This could use a Swift 3 name.

```swift
private func shouldSkipBuild(
    for dependency: Dependency<PinnedVersion>,
    dependingOn dependencyProjects: Set<ProjectIdentifier>,
    whenBuilding dependenciesToBeBuilt: [Dependency<PinnedVersion>],
    on platforms: Set<Platform>
) -> SignalProducer<Bool, CarthageError> {

File: Source/CarthageKit/VersionFile.swift [hunk 44]
```diff
@@ -0,0 +1,272 @@
...
+		guard let cachedFrameworks = self[platform] else {
+			return .init(value: false)
+		}
+
+		return SignalProducer<CachedFramework, CarthageError>(cachedFrameworks)
+			.flatMap(.concat) { cachedFramework -> SignalProducer<(String, String), CarthageError> in
+				let frameworkName = cachedFramework.name
+				let platformURL = binariesDirectoryURL.appendingPathComponent(platform.rawValue, isDirectory: true).resolvingSymlinksInPath()
+				let frameworkURL = platformURL.appendingPathComponent("\(frameworkName).framework", isDirectory: true)
+				let frameworkBinaryURL = frameworkURL.appendingPathComponent("\(frameworkName)", isDirectory: false)
```
mdiep: May as well combine these.

```swift
let url = binariesDirectoryURL
    .appendingPathComponent(platform.rawValue, isDirectory: true)
    .resolvingSymlinksInPath()
    .appendingPathComponent("\(frameworkName).framework", isDirectory: true)
    .appendingPathComponent(frameworkName, isDirectory: false)
```

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,272 @@
...
+		guard FileManager.default.fileExists(atPath: url.path),
+			let jsonData = try? Data(contentsOf: url),
+			let json = try? JSONSerialization.jsonObject(with: jsonData, options: .allowFragments),
+			let versionFile: VersionFile = Argo.decode(json) else {
+			return nil
+		}
+		self = versionFile
+	}
+
+	func satisfies(platform: Platform, commitish: String, xcodeVersion: String, binariesDirectoryURL: URL) -> SignalProducer<Bool, CarthageError> {
```
mdiep: I'm wondering if we should split this into 2 methods:

1. Check whether the cache is valid for a platform, commitish, and Xcode version
2. Check whether the frameworks on disk match the cache

That might simplify some of the logic?

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,272 @@
...
+
+/// Determines whether a dependency can be skipped because it is
+/// already cached.
+///
+/// If a set of platforms is not provided, all platforms are checked.
+///
+/// Returns an optional bool which is nil if no version file exists,
+/// otherwise true if the version file matches and the build can be
+/// skipped or false if there is a mismatch of some kind.
+public func versionFileMatchesDependency(_ dependency: Dependency<PinnedVersion>, forPlatforms platforms: Set<Platform>, rootDirectoryURL: URL) -> SignalProducer<Bool?, CarthageError> {
```
mdiep: This could use a Swift 3 name.

```swift
public func versionFileMatches(_ dependency: Dependency<PinnedVersion>, for platforms: Set<Platform>, rootDirectory: URL) -> SignalProducer<Bool?, CarthageError> {
```

jasonboyle:
Thanks for the review @mdiep!  Please take another look when you get the chance.

adomanico:
Super excited for this feature. Thanks for making this happen.

Will save a lot of time generating our daily builds

scottrhoyt:
Thanks for implementing this @jasonboyle @BobElDevil !

README.md
```diff
@@ -136,6 +136,10 @@ If you want to work on your dependencies during development, and want them to be
 
 Note that you should be [using submodules](#using-submodules-for-dependencies) before doing this, because plain checkouts [should not be modified][Carthage/Checkouts] directly.
 
+### Caching builds
+
+By default Carthage will rebuild a dependency regardless of whether it's the same resolved version as before. Passing the `--cache-builds` will cause carthage to avoid rebuilding a dependency if it can. See information on [version files][VersionFile] for details on how carthage performs this caching.
```
mdiep: This isn't going to be by default for now. This needs to be updated to reflect that it's off by default.

Source/CarthageKit/VersionFile.swift
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	func satisfies(platform: Platform, commitish: String, xcodeVersion: String, binariesDirectoryURL: URL) -> SignalProducer<Bool, CarthageError> {
+		guard let cachedFrameworks = self[platform] else {
+			return SignalProducer(value: false)
+		}
+		
+		let md5s = self.md5s(for: cachedFrameworks, platform: platform, binariesDirectoryURL: binariesDirectoryURL)
+		return self.satisfies(platform: platform, commitish: commitish, xcodeVersion: xcodeVersion, md5s: md5s)
+	}
```
mdiep: Let's make the `satisfies` right below this take a `md5s: [String?]` instead of a `SignalProducer`.

```swift
return self
    .md5s(for: cachedFrameworks, platform: platform, binariesDirectoryURL: binariesDirectoryURL)
    .collect()
    .map { md5s in
        return self.satisfies(platform: platform, commitish: commitish, xcodeVersion: xcodeVersion: md5s: md5s)
    }

func satisfies(platform: Platform, commitish: String, xcodeVersion: String, md5s: [String?]) -> Bool {
    guard let cachedFrameworks = self[platform], commitish == self.commitish, xcodeVersion == self.xcodeVersion else {
        return SignalProducer(value: false)
    }
    
     return md5s == cachedFrameworks.map { $0.md5 }
}
```

---

Although this really makes we think we should do this instead:

1. Add a function that creates a `VersionFile` from the state on disk
2. Add a method that asks whether one `VersionFile` is satisfied by another
3. Here, load the `VersionFile` from disk (1) and ask whether it satisfies (2)
4. Later, use (1) to persist the new version file

Essentially, `createVersionFile` below would return a `SignalProducer<VersionFile, CarthageError>` and _not_ write out the file to disk. It could then be reused for both writing the version file and checking whether it's up to date.

README.md
BobElDevil: That's what it says. "By default Carthage will _rebuild_ a dependency". It's explaining the default behavior is to not performing caching.

mdiep: 🤦🏼‍♂️

Source/CarthageKit/VersionFile.swift
BobElDevil: I don't think this would work. Each dependency doesn't have a separate folder for the build products it produced, so the disk is not a source of truth for what was produced by this specific dependency. 

We currently get this list by collecting on the urls spit out by the build scheme signal, so the current createVersionFile that takes a list of urls would still need to exist in some form.

The closest I think we could come is create a method that creates a comparison only version file based on the disk state, given a reference version file (putting any frameworks matching the reference version file into their own), so we could encapsulate the comparison logic into the struct itself. But it wouldn't be able to be reused for writing out the 'real' version file

Source/CarthageKit/Project.swift
```diff
@@ -733,20 +742,52 @@ public final class Project {
...
+			.flatMap(.concat) { resolvedCartfile -> SignalProducer<Dependency<PinnedVersion>, CarthageError> in
 				return self.buildOrderForResolvedCartfile(resolvedCartfile, dependenciesToInclude: dependenciesToBuild)
 			}
+			.flatMap(.concat) { dependency -> SignalProducer<(Dependency<PinnedVersion>, Set<ProjectIdentifier>, Bool?), CarthageError> in
+				return SignalProducer.combineLatest(
+					SignalProducer(value: dependency),
+					self.dependencyProjects(for: dependency),
+					versionFileMatches(dependency, platforms: options.platforms, rootDirectoryURL: self.directoryURL)
+				)
+			}.reduce([]) { (includedDependencies, nextGroup) -> [Dependency<PinnedVersion>] in
```
jasonboyle: I tried replacing `nextGroup` with a tuple as you suggest but I couldn't get it to compile.

Source/CarthageKit/Xcode.swift
```diff
@@ -680,33 +679,56 @@ public func buildInDirectory(_ directoryURL: URL, withOptions options: BuildOpti
...
+								.flatMapError { _ in .empty }
+						}
+						.map { _ -> TaskEvent<(ProjectLocator, String)> in
+							return .success(ignoredValue)
+						}
+						.filter { taskEvent in
+							return false
+						}
+				
+				Signal.merge([eventSignal, versionFileSignal]).observe(observer)
```
mdiep: I think the best solution here is to write a new version of `collect`—similar to `flatMapTaskEvents`.

```swift
extension SignalProducer where Value: TaskEventType {
    func collect() -> SignalProducer<TaskEvent<[Value.T]>, Error> {
        ...
    }
}
```

That should let you maintain a saner structure here, where you don't need to use `observer` directly—it can remain in its own `startWithSignal` block at the end.

The existing code here is definitely a mess. And `TaskEvent`s make it difficult. So is the best solution I see giving the constraints.

BobElDevil: just to clarify a bit, this new collect extension would 'collect' on the 'success' task events, but let the rest through.

So instead of a stream of stdin, success(val1), stdin, succes(val2) we'd get 'stdin, stdin, success([val1, val2]) ?

should this extension just live in Xcode.swift for now?

mdiep: That's correct! `Xcode.swift` is fine for now.

BobElDevil:
@mdiep made the suggested changes (and finally added that newline for `.reduce` ;) )

For the extension, I wrote it for both Signal and SignalProducer (via lift), so that if in the future it wanted to be included in ReactiveTask, the two functions could be dropped in there without much effort.

mdiep:
The ReactiveTask and PrettyColors submodules were modified at some point. Can you revert those changes?

Source/CarthageKit/Xcode.swift
```diff
@@ -680,33 +679,43 @@ public func buildInDirectory(_ directoryURL: URL, withOptions options: BuildOpti
...
-					.map { taskEvent in
-						return taskEvent.map { _ in initialValue }
-					}
-					.filter { taskEvent in taskEvent.value == nil }
-
-				return BuildSchemeProducer(value: .success(initialValue))
-					.concat(buildProgress)
+					.on(started: {
+						observer.send(value: .success(initialValue))
+					})
```
mdiep: Can we avoid using `observer` here?

Source/CarthageKit/Xcode.swift
```diff
@@ -813,6 +822,42 @@ extension SignalProducerProtocol where Value == URL, Error == CarthageError {
...
+					observer.send(value: .success(taskValues))
+					observer.sendCompleted()
+				case let .failed(error):
+					observer.send(error: error)
+				case .interrupted:
+					observer.sendInterrupted()
+				}
+			}
+		}
+	}
```
mdiep: This looks great. 👍

mdiep:
How would you feel about rigging up an integration tests that tests caching and invalidation?

Source/CarthageKit/Xcode.swift
BobElDevil: this runs into the old "Need to collect the task events but also want to send them through" issue (the stdout/stderr portion was resolved by `collectTaskEvents`)

So here, the original code basically inverted the 'usual' task output. It forced the 'initialValue' to be sent at the beginning of the task instead of the end, and then used map/filter to throw away the 'real' task event. In the current codebase it was ok to throw away the real task value, it was never actually used. The problem is that real task value is the `URL` values we want to collect on. 

So if I tack it onto the beginning, it'll get caught up in the collect (since it's not a stdout event). If I map to it after the collect, it'll wait til the build is done. I think the real solution would be to get rid of the 'invert the task event' workaround altogether, and maybe somehow send out an event to the project events observer, or something higher up. But as it is now, there's no real way to "tell" anyone the build started while still waiting for the build to finish in the same producer.

jasonboyle:
@mdiep I believe we already have such tests in ProjectSpec.  The tests build a project with two dependent frameworks (Prelude > Either > Madness) with various invalid cache states.  Is there something more you have in mind?

mdiep:
> The tests build a project with two dependent frameworks (Prelude > Either > Madness) with various invalid cache states. Is there something more you have in mind?

Ah, okay. That'll be sufficient.

This will need a merge from master now that #1760 has been merged.

BobElDevil:
Pushed a merge with master, so hopefully this is all set to go now :)

BobElDevil:
@mdiep Is there anything else needed to be done with this one?

BobElDevil:
Also looked at the travis failure, looks like both jobs were having connectivity issues. Both failed with RPC SSLRead failures. Not sure how to rerun a travis build

mdiep:
This needs another merge. 😩 

This is at the top of my Carthage queue. I'll give it another look tomorrow morning.

Source/CarthageKitTests/ProjectSpec.swift
```diff
@@ -12,9 +12,113 @@ import Nimble
...
 	override func spec() {
+		describe("createAndCheckVersionFiles") {
+			let directoryURL = Bundle(for: type(of: self)).url(forResource: "DependencyTest", withExtension: nil)!
+			let buildDirectoryURL = directoryURL.appendingPathComponent(CarthageBinariesFolderPath)
+			
+			func buildDependencyTest(platforms: Set<Platform> = [], cacheBuilds: Bool = true) -> Set<String> {
+				var builtSchemes: [String] = []
+				
+				let project = Project(directoryURL: directoryURL)
+				let result = project.buildCheckedOutDependenciesWithOptions(BuildOptions(configuration: "Debug", platforms: platforms, toolchain: "com.apple.dt.toolchain.Swift_2_3", cacheBuilds: cacheBuilds))
```
mdiep: Does this need to specify the toolchain? This test won't work for long if it requires Swift 2.3.

mdiep:
This is looking good. After the conflicts are resolved, I think it'll be ready to go.

Source/CarthageKitTests/ProjectSpec.swift
BobElDevil: At the moment, yes it does. These tests were written before the swift 3 stuff. They were chosen because they were simple projects (i.e., don't take forever to compile), but also had nested dependencies. Unfortunately they don't have branches that compile with swift3. 

What we can do is construct a better, toolchain safe set of dependencies in the fixture and replace this test. I'll start work on getting that in place, but we're hoping it can be a follow-up PR to avoid any more conflicts / delays?

jasonboyle:
@mdiep  Merge conflicts have been resolved.  I believe this is ready to go.

Source/CarthageKitTests/ProjectSpec.swift
mdiep: Sounds good. 👍

mdiep:
Thank you @baek-jinoo, @jasonboyle, and @BobElDevil!!! 🎉 

I know this was a long, difficult process. Thank you for all your patience and hard work. ✨ This took much longer than I would have liked because of limitations on my time.

mdiep:
Also, I'll try to release this by the end of the week. There's one more long-running PR that I'd like to get merged first.

mz2:
Sorry to be a party pooper but on testing this feature again, now off the current master, I found still something I'd certainly personally consider a bug: https://github.com/Carthage/Carthage/issues/1785 – making local commits to a dependency pulled in as a submodule, then `carthage build --cache-builds` for that dependency does *not* lead to realising that the cache should be invalidated (that the cache is invalid in that scenario is suggested by `carthage build` and `carthage build --cache-builds` outputting a different artefact in that situation). I'm guessing the issue is that there is no handling here for submodule based dependencies here, whereas at the build stage the commitish comparison for a submodule based dependency should be between what's in the version file and the current `rev-parse HEAD` of the submodule, not against what's in Cartfile.resolved.

mdiep:
That's why this feature wasn't enabled by default—to give people time to opt in to test it. ☺️

ffittschen:
I don't understand the move from SHA1 to MD5 🤔 I can see the point that cached builds don't really require _cryptographic_ integrity, but why move to something that is [explicitly recommended to not be used anymore][1]?

Using an XCTestCase measure block to test the time it takes to calculate a MD5 Hash vs. a SHA256 Hash shows, that the difference is not that big (in absolute values) and way below the mentioned "seconds for 4MB chunks". I used the Realm.framework (the binary has 56.4MB) as reference, because it was mentioned earlier in this thread:
MD5 Measure Block results: 0.153 sec (3% STDEV)
SHA256 Measure Block results: 0.347 sec (6% STDEV)

To test the SHA256 calculation, I replaced `/usr/bin/env md5 -q` with `/usr/bin/openssl dgst -sha256` in the `md5ForFileAtURL(_:)` method.

[1]: https://www.kb.cert.org/vuls/id/836068

BobElDevil:
@ffittschen There was no technical / performance reason for the decision. It was purely a convenience situation when trying to convert to a command line usage. Since the sample space is so small, and as you said cryptographic integrity isn't a concern, it didn't really stand out.

However, you have a valid point that it really should just be avoided anyway. Filed #1807 to not lose track of this. If you want to open a PR for it that would of course be awesome

ffittschen:
Okay, great 👍🏻 I'll see what I can do on the weekend.

PR #1507, opened 2016-09-27
Update Tentacle

### LINKED ISSUE #1505: `fetch` fails when using HTTP access on Github Enterprise
* `carthage version`: 0.18
* `xcodebuild -version`: Xcode 8.0 Build version 8A218a

**Cartfile**
```
github "http://ghe.foobar.local/anytime/anywhere"
```

**Carthage Output**
```

Illegal instruction: 4
```

I am not prepare a SSL certificate for Github Enterprise.
The above error occurs when execute `carthage fetch`.
It seems to implement that `preferHTTPS` always becomes `true` in `Fetch.swift`.
Is it intended?

// That does not matter because it is no probrem to use SSH.

thanks

ikesyo:
Should fix #1505.


mdiep:
Thanks!

PR #1512, opened 2016-10-04
Eliminate redundant builds

### LINKED ISSUE #1360: Carthage is potentially 50% slower than it needs to be
(This is a bit of a meta-ticket - sorry)

Carthage has developed a bit of a reputation (inside my company and among other users I know who use it) for being slow. Too slow. Slow enough that it can often cause CI jobs to fail (https://github.com/Carthage/Carthage/issues/1310). Slow enough that someone decided a solution was to make carthage_cache to prevent Carthage from running at all on CI (https://github.com/guidomb/carthage_cache)

I wanted to know just what Carthage was doing so I took a build of Carthage and added logging to ReactiveTask so that all NSTasks would get logged to disk.

For my current day job project of 14 dependencies a Carthage build would take 1631 seconds (~27 minutes) (on a 3.1Ghz MBP with eSATA SSD).

After collecting the log file and removing everything in it that wasn't a `xcodebuild clean build` or  `dwarfdump`, `lipo` or `dsymutil`. I "ran" the log and the same build took 813 seconds (~14 minutes).

So this experiment shows that Carthage might be almost twice as slow as it could be.

Furthermore by watching Activity Monitor during the build you could plainly see that the CPU on the machine were mostly at 100% during the process. Whereas watching the monitor during normal Carthage runs the CPUs spend a significant portion of their time idle (It's my understanding that a lot of this idle time is spend in timeouts waiting for `xcodebuild -list`/`xcodebuild -showBuildSettings` to return.)

Is there any work being done to reduce the total build time of Carthage? Is this something the maintainers agree is a problem and would like to address?

If much of the slow down is due to polling `xcodebuild -list` and/or `xcodebuild -showBuildSettings` could this be solved by not calling xcodebuild info so often, possibly caching that data for known versions of repositories, or perhaps by requiring the user to specify the info in the Cartfile?

mdiep:
Fixes #1360.

For schemes that built multiple targets, Carthage was building the scheme multiple times. This is because we’d read the build settings for the scheme, separate the settings by target, and build any buildable targets by building the scheme. Since building the scheme builds all the targets inside, this was redundant.

This fixes the issue by `collect`ing the settings and then flattening them after the scheme is built.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -932,8 +932,10 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 							return false
 						}
 					}
+					.collect()
 					.flatMap(.Concat) { settings -> SignalProducer<TaskEvent<BuildSettings>, CarthageError> in
-						if settings.bitcodeEnabled.value == true {
+						let bitcodeEnabled = settings.reduce(true) { $0 && ($1.bitcodeEnabled.value ?? false) }
+						if bitcodeEnabled {
 							argsForBuilding.bitcodeGenerationMode = .Bitcode
 						}
 
@@ -941,9 +943,7 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 						buildScheme.workingDirectoryPath = workingDirectoryURL.path!
 
 						return launchTask(buildScheme)
-							.map { taskEvent in
-								taskEvent.map { _ in settings }
-							}
+							.flatMapTaskEvents(.Concat) { _ in SignalProducer(values: settings) }
 							.mapError(CarthageError.TaskError)
 					}
 			}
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -932,18 +932,18 @@ public func buildScheme(scheme: String, withConfiguration configuration: String,
 							return false
 						}
 					}
+					.collect()
 					.flatMap(.Concat) { settings -> SignalProducer<TaskEvent<BuildSettings>, CarthageError> in
-						if settings.bitcodeEnabled.value == true {
+						let bitcodeEnabled = settings.reduce(true) { $0 && ($1.bitcodeEnabled.value ?? false) }
+						if bitcodeEnabled {
```
mdiep: I'm a little uncertain about this part, but it seemed like the best option given my very limited knowledge of bitcode.

ikesyo: This looks correct, it should be sure that all the targets are bitcode-enabled to enable bitcode for the scheme.

PR #1530, opened 2016-10-13
Fix a failing integration test

ikesyo:
The failure is seen here: https://travis-ci.org/Carthage/Carthage/jobs/167228879. That is because Argo v2.3.0 did not specify version constraint for Curry in its `Cartfile.private`. So when running `carthage update` in the checked out repository, the latest Curry v3.0.0 for Swift 3, which can't be built by Xcode 7.3, will be used. ~~Using Argo v3.1.0 should fix the problem.~~

### DIFF
--- a/Source/CarthageIntegration/update.bats
+++ b/Source/CarthageIntegration/update.bats
@@ -2,9 +2,10 @@
 
 setup() {
     cd $BATS_TMPDIR
-    rm -rf Argo
-    git clone -b v2.3.0 https://github.com/thoughtbot/Argo.git
-    cd Argo
+    rm -rf UpdateTest
+    mkdir UpdateTest && cd UpdateTest
+    echo 'github "antitypical/Result" == 2.1.3' > Cartfile
+    echo 'github "Quick/Nimble" == 4.1.0' > Cartfile.private
 }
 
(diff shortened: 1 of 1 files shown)


ikesyo:
> github "thoughtbot/Runes" >= 3.0.0

Argo v3.1.0 also has the issue: https://github.com/thoughtbot/Argo/blob/v3.1.0/Cartfile.private#L2. 😞

ikesyo:
Switched to using [Commandant](https://github.com/Carthage/Commandant).

Source/CarthageIntegration/update.bats
```diff
@@ -2,9 +2,10 @@
 
 setup() {
     cd $BATS_TMPDIR
-    rm -rf Argo
-    git clone -b v2.3.0 https://github.com/thoughtbot/Argo.git
-    cd Argo
+    rm -rf UpdateTest
+    mkdir UpdateTest && cd UpdateTest
+    echo 'github "antitypical/Result" == 2.1.3' > Cartfile
+    echo 'github "Quick/Nimble" == 4.1.0' > Cartfile.private
```
mdiep: This is a way better approach. 👍

PR #1535, opened 2016-10-17
Code simplification using map.

Nirma:
I thought using map here might be a little more brief.

How does it look?

Thanks!

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -27,13 +27,7 @@ private func gitHubUserAgent() -> String {
 extension Repository {
 	/// The URL that should be used for cloning this repository over HTTPS.
 	public var HTTPSURL: GitURL {
-		let auth: String
-		if let token = tokenFromEnvironment(forServer: server) {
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/GitHub.swift
```diff
@@ -27,12 +27,7 @@ private func gitHubUserAgent() -> String {
 extension Repository {
 	/// The URL that should be used for cloning this repository over HTTPS.
 	public var HTTPSURL: GitURL {
-		let auth: String
-		if let token = tokenFromEnvironment(forServer: server) {
-			auth = "\(token)@"
-		} else {
-			auth = ""
-		}
+		let auth: String = tokenFromEnvironment(forServer: server).map { "\(token)@" } ?? ""
```
NachoSoto: This should be `$0` instead of `token`.
Also, you can drop the explicit type.

Nirma:
@NachoSoto So sorry!!! I pushed the wrong commit, here I fixed it now!

How does it look?

NachoSoto:
Thank you!

PR #1540, opened 2016-10-19
Clear up installation instructions

JetForMe:
This is my attempt at clearing up the installation instructions a bit. The issue has some discussion indicating the Xcode requirement is only for building. If that's verified, I'll change this.


mdiep:
Thanks for the pull request!

PR #1553, opened 2016-10-28
Add support for pre-release and build metadata, as per SemVer specs

marcoconti83:
# Issue addressed

Our `Cartfile` reads:

```
github "wireapp/wire-ios-data-model" ~> 25.0
```

The `wire-ios-data-model` repository currently has the following tags:

```
...
23.0.0
23.0.0-callkit
23.1.0
23.2.0
23.2.0-callkit
23.3.0
23.3.1
24.0.0
25.0.0
25.0.1
...
```

When I run `carthage update --platform iOS`, I would expect to fetch `wire-ios-data-model` at version `25.0.1`. Instead, Carthage will pick `23.2.0-callkit`. After running, my `Cartfile.resolved` is now:

```
...
github "wireapp/wire-ios-data-model" "23.2.0-callkit"
...
```
# Investigation

The pre-release part of the version `23.2.0-callkit` (which follows the [semver specs](http://semver.org/#spec-item-9)) cause the version to fail to parse, because CarthageKit does not yet support pre-release and build metadata (see [here](https://github.com/Carthage/Carthage/blob/279f4b88a75c409f6b586cd8630740d87ab7713c/Source/CarthageKit/Version.swift#L66)).

Additionaly, CarthageKit assumes that a version that can not be parsed will satisfy any requirements (see [here](https://github.com/Carthage/Carthage/blob/279f4b88a75c409f6b586cd8630740d87ab7713c/Source/CarthageKit/Version.swift#L178)). This leads to the first non-parsable version to be picked as a match even if it is a pre-release of an previous incompatible version.
# Changes

Modified `Version.swift` so that now CarthageKit correctly parses pre-release and build metadata informations as per [semver specs](http://semver.org).
# Testing

Versions with pre-release and build metadata have been added to the unit tests of the `SemanticVersion` implementation.
Additionally I manually tested the locally-compiled Carthage version using the example listed earlier and it now behaves correctly, picking `25.0.1` instead of `23.2.0-callkit`.

### DIFF
--- a/Source/CarthageKit/Version.swift
+++ b/Source/CarthageKit/Version.swift
@@ -14,6 +14,7 @@ import ReactiveCocoa
 public protocol VersionType: Equatable {}
 
 /// A semantic version.
+/// - Note: See <http://semver.org/>
 public struct SemanticVersion: VersionType, Comparable {
 	/// The major version.
 	///
@@ -30,7 +31,17 @@ public struct SemanticVersion: VersionType, Comparable {
 	///
 	/// Increments to this component represent backwards-compatible bug fixes.
 	public let patch: Int
-
+	
+	/// The pre-release identifier
+	///
+	/// Indicates that the version is unstable
+	public let preRelease : String?
+	
+	/// The build metadata
+	///
+	/// Build metadata is ignored when comparing versions
+	public let buildMetadata : String?
+	
 	/// The pin from which this semantic version was derived.
 	public var pinnedVersion: PinnedVersion?
 
@@ -40,10 +51,12 @@ public struct SemanticVersion: VersionType, Comparable {
 		return [ major, minor, patch ]
 	}
 
-	public init(major: Int, minor: Int, patch: Int) {
+	public init(major: Int, minor: Int, patch: Int, preRelease: String? = nil, buildMetadata: String? = nil) {
 		self.major = major
 		self.minor = minor
 		self.patch = patch
+		self.preRelease = preRelease
+		self.buildMetadata = buildMetadata
 	}
 
 	/// The set of all characters present in valid semantic versions.
@@ -63,8 +76,6 @@ public struct SemanticVersion: VersionType, Comparable {
 				version.pinnedVersion = pinnedVersion
 				return .Success(version)
 			} else {
-				// Disallow versions like "1.0a5", because we only support
-				// SemVer right now.
 				return .Failure(CarthageError.ParseError(description: "syntax of version \"\(version)\" is unsupported"))
 			}
 		}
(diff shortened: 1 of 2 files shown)


marcoconti83:
Probably ties in to https://github.com/Carthage/Carthage/issues/469

mdiep:
First off, thanks for the PR! It's really great when people jump in to fix things. 🤘 

Unfortunately, this fix is incomplete and will cause some regressions. This is a tricky area, which is why it hasn't been fixed yet. This does get us partway there. There are 3 important requirements:

1. Pre-release versions should not be picked unless explicitly requested
2. Pre-release should override release version if explicitly specified somewhere in the chain
3. Requirements must always be respected—if you specify `>= 4.0`, you should never get anything less.

This fixes half of (1), but it also causes some regressions. e.g., if you add `github "ReactiveCocoa/ReactiveCocoa"`, which currently has an alpha release of 5.0.0, but not a stable release, it will download the alpha version instead of the latest stable (4.x) release.

marcoconti83:
Thanks for the feedback @mdiep!

Actually I wrote this PR to address exactly point (3), as with current Carthage I get `23.2.0-callkit` even if I specify `>= 25.0`. But your point is still valid when considering that I could get `4.0-pre` when specifying `4.0`.

I will try translate your comments into unit tests and see if I can come up with a fix that will address my original issue and the regressions you mentioned. Thanks for the input!

mdiep:
Thanks @marcoconti83! That'd be great! Please not that these apply to other non-semantic-version tags as well.

marcoconti83:
I added comparison between pre-release versions as per semantic versioning specifications, and a strategy to decide whether a pre-release version satisfies a carthage requirement. 

However I have the feeling I might have missed something.
I'm not sure what is the expected behavior for `VersionSpecifier.GitReference` when comparing it with a semantic version. It used to be satisfied by anything, and I left it so, but I feel like this is wrong, unless there is also a comparison somewhere else that did not manage to find. See https://github.com/Carthage/Carthage/pull/1553/commits/6c2629a03249b43d76f95eaa8d16d3f97d365e2f#diff-aa0796654f69d3d5ad0a58b156377953R327.

marcoconti83:
It would be great if you could review my last commit to see if it covers the three points you mentioned. I am also not sure what you meant by:

> Please not that these apply to other non-semantic-version tags as well.

I implemented it as discarding any non-semantic versioning compatible version when matching a `Any` requirement.

Source/CarthageKit/Version.swift
```diff
@@ -30,7 +31,17 @@ public struct SemanticVersion: VersionType, Comparable {
 	///
 	/// Increments to this component represent backwards-compatible bug fixes.
 	public let patch: Int
-
+	
+	/// The pre-release identifier
+	///
+	/// Indicates that the version is unstable
+	public let preRelease : String?
```
mdiep: I'm wondering if there should be a inner `PreRelease` struct to use for this.

```swift
public struct SemanticVersion: VersionType, Comparable {
    public struct PreRelease: Comparable {
        public let identifiers: [String]
    }
}
```

Then a bunch of the code below (like `isPreReleaseLesser`) would have a more natural home.

Source/CarthageKit/Version.swift
```diff
@@ -75,33 +91,68 @@ extension SemanticVersion: Scannable {
...
+			return stringWithPrefix.substringFromIndex(stringWithPrefix.startIndex.advancedBy(prefix.characters.count))
+		} else {
+			return nil
+		}
 	}
 }
 
 public func <(lhs: SemanticVersion, rhs: SemanticVersion) -> Bool {
+	if lhs.components == rhs.components {
+		return lhs.isPreReleaseLesser(rhs.preRelease)
```
mdiep: I'd switch here to handle the `nil`s right here:

```swift
if lhs.components == rhs.components {
    switch (lhs.prerelease, rhs.prerelease) {
        case let (lhs?, rhs?):
            return lhs < rhs
        case (.some, .none):
            return true
        case (.none, _):
            return false
    }
}
```

Source/CarthageKit/Version.swift
```diff
@@ -175,34 +314,54 @@ public enum VersionSpecifier: VersionType {
...
 
 		case let .AtLeast(requirement):
-			return withSemanticVersion { $0 >= requirement }
-
+			return withSemanticVersion { version in
+				let versionIsNewer = version >= requirement
+				
+				// Only pick a pre-release version if the requirement is also
+				// a pre-release of the same version
+				let notPreReleaseOrSameComponents =	!version.isPreRelease
```
mdiep: Weird indentation here after the `=`.

Source/CarthageKit/Version.swift
```diff
@@ -175,34 +314,54 @@ public enum VersionSpecifier: VersionType {
...
+					|| (requirement.isPreRelease && version.hasSameNumericComponents(requirement))
+				return notPreReleaseOrSameComponents && versionIsNewer
+			}
 		case let .CompatibleWith(requirement):
 			return withSemanticVersion { version in
+				
+				let versionIsNewer = version >= requirement
+				let notPreReleaseOrSameComponents =	!version.isPreRelease
```
mdiep: Same here.

mdiep:
@marcoconti83 This is a tricky PR to review since the potential for breakage is rather high. How would you feel about breaking this into 3 separate PRs?

1. Add exhaustive tests for `VersionSpecifier.satisifiedBy`. That would make it much easier to see which cases changed in this PR, making it easier to catch any errors.

2. Add parsing for pre-release versions without changing any of the tests from (1).

3. Building on (1) and (2), change satisfiability.

(1) would be the biggest help, I think.

That would give me a lot more confidence that we weren't making undesired changes.

mdiep:
Actually, I don't know if (2) can be done separately. But doing (1) separately should still be a big benefit.

marcoconti83:
I'll give it a try!

blender:
ping

marcoconti83:
Sorry for not getting back on this in such a long time, this fell off my radar. I will give it a  shoot over the weekend.

marcoconti83:
Will close this one and reopen as separate PRs.

marcoconti83:
Created first one:
- #2392

marcoconti83:
> [...] How would you feel about breaking this into 3 separate PRs?
>
> 1. Add exhaustive tests for VersionSpecifier.satisifiedBy. That would make it much easier to see which cases changed in this PR, making it easier to catch any errors.
> 2. Add parsing for pre-release versions without changing any of the tests from (1).

@mdiep could you please clarify what you would like to see in (1), as discussed above?

I was about to port the tests I added in the current PR to a new separate PR but I realized that all the tests I added are related to pre-releases satisfying requirements so I can not port them on a "clean" PR, I can only port them on top of #2392.

On the other hand, `VersionSpecifier.satisifiedBy` tests in `VersionSpec.swift` are already quite comprehensive, so I would not know what more to test unless we throw pre-release versions in the mix.

Is there still any part of the satisfaction rules that is lacking testing?

mdiep:
I think I was thinking that you could add the tests for pre-releases but have them just return that they're not satisfied for now. Then the PR that turns on that behavior can just turn `false`s into `true`s.

marcoconti83:
Created: https://github.com/Carthage/Carthage/pull/2394

Issue #1558, opened 2016-11-04
Upgrade to Swift 3


mdiep:
Xcode 8.2 won't include Swift 2, so now is the time to start upgrading to Swift 3.

ikesyo:
Xcode 8.2 still includes Swift 2.3 (that is the final release supporting 2.3) but switching to Swift 3 now should be good anyway.

ikesyo:
Dependencies status:

- [ ] ReactiveTask `"master"`
- [x] Commandant `~> 0.11.2`
- [ ] PrettyColors `"development"`
- [x] ReactiveSwift `"1.0.0-alpha.3"`
- [ ] Tentacle
- [x] Quick `~> 0.10`
- [x] Nimble `~> 5.0`

ikesyo:
I'm planning have some pre-migration tasks for easier review, those makes diff volumes of an actual migration pull request fewer:

- Adopt API Design Guidelines (mainly functions, methods)
- Make enum cases lowercased
- Rename `URL` to `url`
- Drop `NS` prefixes using `typealias`
- `dynamictype` -> `typeof()`
- ...

mdiep:
What's the best way for me to help with this?

It seems that Xcode 8.3 is just around the corner, so we should try to finish this up.

ikesyo:
Thank you @mdiep. I'd like to submit an actual migration pull request once #1734 and #1735 are merged.

ikesyo:
This should have been done (except for entirely following API Design Guidelines).

mdiep:
💖

PR #1571, opened 2016-11-11
[swift3-prep] Make enum cases lowerCamelCase

ikesyo:
https://github.com/Carthage/Carthage/issues/1558#issuecomment-259115655

- `Platform`
- `SDK`
- `BuildPlatform`

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -222,20 +222,20 @@ public func schemesInProjects(projects: [(ProjectLocator, [String])]) -> SignalP
 
 /// Represents a platform to build for.
 public enum Platform: String {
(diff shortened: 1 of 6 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -222,20 +222,20 @@ public func schemesInProjects(projects: [(ProjectLocator, [String])]) -> SignalP
 
 /// Represents a platform to build for.
 public enum Platform: String {
-	/// Mac OS X.
-	case Mac
+	/// macOS.
+	case macOS = "Mac"
```
ikesyo: The raw value should not be changed for backward compatibility since that is used in `Carthage/Build/Mac` dir.

Issue #1541, opened 2016-10-19
Carthage auto-selecting clean action for Cocoapods target rather than the framework's target


maxkramerbcgdv:
Hi there,

I'm switching over to Carthage from Cocoapods given the need for significantly faster build times (that don't require re-building 6/7 dependencies from scratch frequently). Everything has worked great and it has been a very simple process thus far. One of our dependencies `ACSPinKit`: https://github.com/arconsis/ACSPinKit does not support Carthage outright, but has no reason not to. 

I followed the Supporting Carthage guide (https://github.com/Carthage/Carthage#supporting-carthage-for-your-framework) and the project already adheres to all of the requirements, but when attempting to build it from within our project, and within the ACSPinKit project itself (fresh clone), we are getting the error below.
- `carthage version`:  0.18.1
- `xcodebuild -version`: 
  Xcode 8.0
  Build version 8A218a
- Are you using `--no-build`? no
- Are you using `--no-use-binaries`? no
- Are you using `--use-submodules`? no

**Cartfile**

```
none
```

**Carthage Output**
`carthage build --no-skip-current --verbose`

```
A shell task (/usr/bin/xcrun xcodebuild -workspace /private/tmp/ACSPinKit/Example/ACSPinKit.xcworkspace -scheme Pods-ACSPinKit-ACSPinKit -configuration Release CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean -showBuildSettings) failed with exit code 66:
xcodebuild: error: Scheme Pods-ACSPinKit-ACSPinKit is not currently configured for the clean action.
```

What appears to be the problem is that for some reason the auto-selected scheme is incorrect, and it should be selecting `ACSPinKit` instead. Having tried to execute the same command with the correct scheme, the project builds as expected.

`/usr/bin/xcrun xcodebuild -workspace /private/tmp/ACSPinKit/Example/ACSPinKit.xcworkspace -scheme Pods-ACSPinKit-ACSPinKit -configuration Release CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean`

From what I've seen so far, there doesn't seem to be a way of specifying the scheme/target that the workspace should build with, so I was wondering what you recommend to fix the error.

Cheers!
Max.

mdiep:
[...]

PR #1572, opened 2016-11-11
Adds the -skipUnavailableActions flag when retrieving build settings

zachwaugh:
`xcodebuild clean -showBuildSettings` can fail if the scheme doesn't support the clean action, same as issue https://github.com/Carthage/Carthage/issues/1541. This adds a flag to skip unavailable actions instead of failing.

An example with PINRemoteImage:

```shell
$ carthage update
*** Checking out PINRemoteImage at "3.0.0-beta.6"
*** xcodebuild output can be found in /var/folders/4x/k22fcvxj09q33bnxlpc1ngv00000gn/T/carthage-xcodebuild.D68Rup.log
A shell task (/usr/bin/xcrun xcodebuild -workspace /Users/zach/code/tmp/CarthageTest/Carthage/Checkouts/PINRemoteImage/Example-Mac/PINRemoteImage.xcworkspace -scheme PINRemoteImage -configuration Release CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean -showBuildSettings) failed with exit code 66:
xcodebuild: error: Scheme PINRemoteImage is not currently configured for the clean action.
```

Running directly:
```shell
$ xcodebuild -workspace PINRemoteImage.xcworkspace -scheme PINRemoteImage -configuration Release CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean -showBuildSettings
Build settings from command line:
    CARTHAGE = YES
    CODE_SIGN_IDENTITY = 
    CODE_SIGNING_REQUIRED = NO

xcodebuild: error: Scheme PINRemoteImage is not currently configured for the clean action.
```

xcodebuild has a `-skipUnavailableActions` option:
> -skipUnavailableActions
           Skip actions that cannot be performed instead of failing. This option is only honored if -scheme is passed.

With that option, we get the settings and no error:

```shell
$ xcodebuild -workspace PINRemoteImage.xcworkspace -scheme PINRemoteImage -configuration Release CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean -showBuildSettings -skipUnavailableActions
Build settings from command line:
    CARTHAGE = YES
    CODE_SIGN_IDENTITY = 
    CODE_SIGNING_REQUIRED = NO
```

I don't know if there are any side-effects from that flag, but seems pretty safe since we're just retrieving the build settings here. I'm not very familiar with the codebase, so let me know if this the right place to put that.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -481,7 +481,7 @@ public struct BuildSettings {
 		// rdar://27052195
 		// Including the action "clean" works around this issue, which is further
 		// discussed here: https://forums.developer.apple.com/thread/50372
-		let task = xcodebuildTask(["clean", "-showBuildSettings"], arguments)
+		let task = xcodebuildTask(["clean", "-showBuildSettings", "-skipUnavailableActions"], arguments)
 
 		return launchTask(task)
 			.ignoreTaskData()


ikesyo:
Thanks for the pull request! Using the flag seems reasonable for me. :ok_man: 

@mdiep Another thought?

mdiep:
This is great! ✨ 

> So I think the flag should be added here too

I think it's probably not required there? If I understand correctly, we primarily hit this issue because Carthage inquires about every scheme in every project. Dynamic framework schemes should always have a clean action?

ikesyo:
Okay, let's merge this as is then. 🚢

sirvon:
How can i use this version of carthage instead of the one supplied by ```brew install```?

mdiep:
You can build from source.

We will be releasing a new version sometime in the next couple weeks.

PR #1595, opened 2016-11-18
[swift3-prep] Add Foundation.Scanner shims

ikesyo:
#1558

### DIFF
--- a/Source/CarthageKit/Swift3Shims.swift
+++ b/Source/CarthageKit/Swift3Shims.swift
@@ -134,6 +134,27 @@
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


Source/CarthageKit/Swift3Shims.swift
```diff
@@ -134,6 +134,27 @@
 		}
 	}
 
+	public typealias Scanner = NSScanner
```
ikesyo: This must be `public` because of this: https://github.com/Carthage/Carthage/pull/1595/files#diff-9a71f3c0c991f37b53338e11f4d93de2R18

PR #1611, opened 2016-11-23
Allow the user to cache build artefacts for a particular revision of a dependency 

nickoto:
This patch caches the build products or a given dependency at a given commitish. Subsequent builds for that specific dependency+commitish will then use the cached artefacts instead of building again. The cache is only used when `--use-build-products-cache` is specified on the command line.

I had not taken a look at the pending PRs when I started this and it turns out that the goal is is similar to #1489 however it works across projects. It is more or less a work in progress and I would love some feedback on this.

### DIFF
--- a/Source/CarthageKit/BuildOptions.swift
+++ b/Source/CarthageKit/BuildOptions.swift
@@ -16,11 +16,14 @@ public struct BuildOptions {
 	public let toolchain: String?
 	/// The path to the custom derived data folder.
 	public let derivedDataPath: String?
+	/// The path to the cached binaries folder.
+	public let useBuildProductsCache: Bool
 
-	public init(configuration: String, platforms: Set<Platform> = [], toolchain: String? = nil, derivedDataPath: String? = nil) {
+	public init(configuration: String, platforms: Set<Platform> = [], toolchain: String? = nil, derivedDataPath: String? = nil, useBuildProductsCache: Bool = false) {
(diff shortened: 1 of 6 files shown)


freesuraj:
@nickoto can u fix the conflict and resubmit this PR?

leoneparise:
Any thoughts on this? It will improve update times a lot.

mdiep:
This and #1489 are on my list, but I've been very busy with other OSS (ReactiveCocoa, ReactiveSwift), so I haven't had time to review this yet.

freesuraj:
Folks, lets make Carthage great again !! Let's merge this PR.

soheilbm:
If anyone looking for alternate solution until they merge this PR, try this one: https://github.com/soheilbm/carthage-cache

mdiep:
It doesn't seem like cached products are ever invalidated. Nor does it look like the centralized cache is ever pruned. Those are definitely both dealbreakers. (#1489 does the first, but doesn't implement a centralized cache.)

nickoto:
What should be the deciding factor for both pruning and invalidating the cached products?

mdiep:
> for both pruning

This I'm not sure about. I'm open to suggestions. ☺️ 

> invalidating the cached products

If the version changes or the compiler version changes. Note that the cached product should be validated somehow by a SHA or something.

mdiep:
#1489 has been merged.

Issue #1227, opened 2016-03-30
Skip schemes


mstfy:
When I build RxSwift with carthage it builds RxSwift, RxCocoa, RxBlocking and RxTests. Is there a way to tell carthage to build only RxSwift and RxCocoa and skip other schemes?

mdiep:
No, there's not. :relaxed:

robertjpayne:
@mstfy I have a similar need and I've simply scripted Carthage via a Rakefile and the `subprocess` gem. Luckily schemes are located inside `*.xcodeproj/xcshareddata/xcschemes/**`and are pretty easy to find and remove.

jeeftor:
so you do a `carthage update --no-build` -- delete schemes by a ``carthage build` ?

Antondomashnev:
@robertjpayne can you please elaborate on your solution, would be helpful, because I have similar problem.

lukescott:
I have a similar problem with https://github.com/aws/aws-sdk-ios

They have one project file, but multiple build targets. It builds all the frameworks:
```
*** Building scheme "AWSDynamoDB" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSMachineLearning" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSCognitoIdentityProvider" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSAutoScaling" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSCognito" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSAPIGateway" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSSQS" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSSNS" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSKinesis" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSCloudWatch" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSSES" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSElasticLoadBalancing" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSLambda" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSIoT" in AWSiOSSDKv2.xcodeproj
*** Building scheme "AWSCore" in AWSiOSSDKv2.xcodeproj
[...]

PR #1616, opened 2016-11-29
allow frameworks to specify which schemes Carthage should build

jpsim:
By specifying an optional `.carthage_schemes` file at the root of a framework repo, Carthage can skip much of the work involved in evaluating which schemes to build in a directory.

If present, the `.carthage_schemes` file should be a newline-separated list of scheme names that Carthage should build.

The work involved in evaluating which schemes to build can be substantial due to requiring `xcodebuild` invocations. In the case of the Realm frameworks, for example, using this approach shaves minutes off the total update time.

Since frameworks are sometimes useful for purposes internal to a repo and not meant to be consumed externally, specifying a subset of schemes to build in `.carthage_schemes` can also great speed up build times.

This can also be useful to work around certain Xcode bugs, such as when Xcode considers unit test bundle products to actually be framework products [rdar://29407087](http://www.openradar.me/29407087).

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -164,14 +164,26 @@ public func schemesInProject(project: ProjectLocator) -> SignalProducer<String, 
 		.map { (line: String) -> String in line.stringByTrimmingCharactersInSet(.whitespaces) }
 }
 
+
+/// Returns schemes specified by the `.carthage_schemes` file at the directory specified.
+/// Returns `nil` if it doesn't exist.
+private func carthageSchemesFileSchemes(directoryURL: NSURL) -> [String]? {
+	let carthageSchemesFileURL = directoryURL.appendingPathComponent(".carthage_schemes")
+	let contents = try? String(contentsOfURL: carthageSchemesFileURL, encoding: NSUTF8StringEncoding)
+	return contents?.componentsSeparatedByCharactersInSet(.newlines).filter { !$0.isEmpty }
+}
+
 /// Finds schemes of projects or workspaces, which Carthage should build, found
 /// within the given directory.
 public func buildableSchemesInDirectory(directoryURL: NSURL, withConfiguration configuration: String, forPlatforms platforms: Set<Platform> = []) -> SignalProducer<(ProjectLocator, [String]), CarthageError> {
 	precondition(directoryURL.fileURL)
 
+	let filteredSchemes = carthageSchemesFileSchemes(directoryURL)
+
 	return locateProjectsInDirectory(directoryURL)
 		.flatMap(.concat) { project -> SignalProducer<(ProjectLocator, [String]), CarthageError> in
 			return schemesInProject(project)
+				.filter { scheme in return filteredSchemes?.contains(scheme) ?? true }
 				.flatMap(.merge) { scheme -> SignalProducer<String, CarthageError> in
 					let buildArguments = BuildArguments(project: project, scheme: scheme, configuration: configuration)
 


jpsim:
Building both `Realm.framework` and `RealmSwift.framework` is ~40% faster with this PR (499s vs 839s), since it works around [rdar://29407087](http://www.openradar.me/29407087).

---

Current `master`:

```shell
$ time carthage update --platform osx
*** Fetching realm-cocoa
*** Checking out realm-cocoa at "a772de1626c9e6b3279c508375a539c29c77244c"
*** xcodebuild output can be found in /var/folders/np/5y4sgt8x3r13sk7x0n31z_3m0000gn/T/carthage-xcodebuild.JBCzD7.log
*** Building scheme "Realm" in Realm.xcworkspace
*** Building scheme "Object Server Tests" in Realm.xcworkspace
*** Building scheme "RealmSwift" in Realm.xcworkspace
carthage update --platform osx  839.58s user 86.48s system 262% cpu 5:52.89 total
```

This PR:

```shell
$ time carthage update --platform osx
*** Fetching realm-cocoa
*** Checking out realm-cocoa at "a772de1626c9e6b3279c508375a539c29c77244c"
*** xcodebuild output can be found in /var/folders/np/5y4sgt8x3r13sk7x0n31z_3m0000gn/T/carthage-xcodebuild.VXjlfY.log
*** Building scheme "Realm" in Realm.xcworkspace
*** Building scheme "RealmSwift" in Realm.xcworkspace
carthage update --platform osx  499.41s user 45.25s system 238% cpu 3:48.07 total
```

---

Unfortunately, Carthage still builds `Realm.framework` once too many, since `RealmSwift.framework` uses it as a dependency, and Carthage always cleans before building, but that's a different battle.

younata:
That looks like a pretty good idea, and has been something I've been wanting to do myself.

Would it be possible to add some test coverage for this?

Pr0Ger:
I think it's better to give this kind of control to project instead of framework. For example project can build only 'Realm' scheme, or only 'RealmSwift' dependes on what it is required.

Relevant issue: #1227

jpsim:
> I think it's better to give this kind of control to project instead of framework.

That's a nice idea, but orthogonal to this feature. It doesn't address preventing internal-only frameworks from being exposed externally.

For example, Realm uses a framework called `PlaygroundFrameworkWrapper` to enable using a prebuilt framework in a playground (typically only allowed for framework _targets_ defined in an Xcode workspace), which is only used for packaging purposes, and should never be exposed externally. We've had to unshare this framework scheme to cater to Carthage's behavior (realm/realm-cocoa@9c38192), which isn't ideal and shouldn't be required of any project maintainer since it adversely affects non-Carthage use cases.

jpsim:
> Would it be possible to add some test coverage for this?

I'm happy to add tests and documentation, but before I invest time into that, I'd like to get a sense from Carthage maintainers whether or not this is an idea they're generally comfortable with.

mdiep:
Thanks for the pull request. I always really appreciate when people open a PR instead of expecting maintainers to do all the work. ☺️ 

I do feel a little uncomfortable with this because it opens the door to a lot of additional configuration and complexity—and Carthage has eschewed all configuration. I feel this both on an implementation level (where the codebase needs to be refactored to handle the existing complexity) and conception level (where we'd probably want a larger vision for what configuration we want to allow and how it works).

But I also respect that this is a very pragmatic choice. That's part of why I've let this sit for a couple days before responding. Idealism vs. pragmatism is a hard line to manage.

I noticed that your radar had a response. Have you tried out the solution they gave? Knowing whether that solves the issue at hand seems like the next step.

jpsim:
> I always really appreciate when people open a PR instead of expecting maintainers to do all the work.

Same! And this is a better way to spawn a more concrete conversation than by having a hand-wavy feature request issue.

> I do feel a little uncomfortable with this because it opens the door to a lot of additional configuration and complexity—and Carthage has eschewed all configuration. I feel this both on an implementation level (where the codebase needs to be refactored to handle the existing complexity) and conception level (where we'd probably want a larger vision for what configuration we want to allow and how it works).

I can appreciate that. However in practice, as a library maintainer with a fairly complex repo structure, I can safely say that conforming to Carthage's expectations has been more of a burden, and needed more refactoring than this simple configuration option involves.

For example, there's no way currently to have internal-only frameworks (like the `PlaygroundFrameworkWrapper` example above) without also making compromises, which in some cases aren't viable, such as unsharing the framework scheme in that specific example.

If we explicitly controlled which schemes we wanted Carthage to consider building, there would be less of a burden on library maintainers to consider the Carthage implications of adding another framework in their repo, sometimes deeply nested in an `examples/` subdirectory.

Please keep in mind as well that this file is entirely optional and won't affect the majority of library maintainers with very simple project structures.

As for wanting to explore a more general strategy for configuration, that's a fair argument and I respect if that's the outcome of this. However, it is disappointing. I'd much rather see this implemented with a warning saying that it's a temporary solution until a more comprehensive configuration approach can be designed and implemented. As you can tell, it's also not a feature that will require a high maintenance burden (it's dead simple for this reason), so it could conceivably be preserved even after a different configuration strategy is introduced.

> I noticed that your radar had a response. Have you tried out the solution they gave? Knowing whether that solves the issue at hand seems like the next step.

Yes, but that's only a partial solution to one of the motivating factors for this feature (that Xcode behavior). It doesn't address the "internal-only framework" scenario presented above.

guidomb:
This could be a solution to the problem of working with dependencies like https://github.com/aws/aws-sdk-ios. They have lots of scheme and the entire project takes around 30 minutes to compile. Even if as a library consumer I only want to use AWS S3 I need to be build all their services frameworks. Users of AWS iOS SDK that include it via CocoaPods don't have this issue because they can use subspec but this is not an option with Carthage.

Although we can discuss if the way AWS SDK is structure is adequate or not, having the solution that @jpsim propose will make our life easier and will lower the barrier for library maintainers to adopt Carthage.

guidomb:
@jpsim Is this an option only for the library maintainer or for libraries consumers too? Because if this is only for maintainers then what I said before about AWS SDK does not apply and maybe a solution like #1227 would be better.

jpsim:
As-is, this would only be for library maintainers.

mdiep:
> Since frameworks are sometimes useful for purposes internal to a repo and not meant to be consumed externally

Do you have any examples of this? Internal frameworks that are part of shared schemes? (The `Object Server Tests` case above doesn't fit since it was an Xcode/configuration issue.)

_(Sorry for the delay. But now that ReactiveSwift 1.0 has shipped, I should be able to wrap up this discussion.)_

jpsim:
An example is the `PlaygroundFrameworkWrapper` that I mentioned in a previous issue.

jpsim:
*previous comment

mdiep:
Unfortunately, I don't think that example warrants adding this to Carthage.

This feature brings a fair bit of risk to Carthage. Although the feature itself isn't complicated, the configuration issue could quickly become a large breaking change in the community. If authors start relying on it to hide broken schemes, then we can't break that support.

Given that this is a minor case, has a workaround, and should really be a enhancement request for Xcode, I don't think it's a good idea to accept it. Carthage has to walk a very fine line here. Since the impact and reach are both small, I don't think it's a good idea.

jpsim:
Understood.

arkdan:
@mdiep i deeply regret you judged the question this way.

1. since the change is optional, there is a long way before the authors start checking in "broken schemes" because of it.
2. can we really consider carthage stopping authors from "hiding broken schemes"?
3. on one side we have real value to developers ("pragmatism") vs your opinion ("idealism"), which, i'm sorry to say, can not be enough to stop this discussion. Rather, you could suggest another approach around the problem, or maybe keep the discussion open for the community and other maintainers.
4. i would definitely benefit if carthage supported specifying a scheme.
5. `defer {}` is not needed most of the time - but when it is, it brings real value. I sure wasn't sad when i got `defer {}` at my disposal.

Dschee:
Anybody coming across this issue nowadays, please checkout my comment [here](https://github.com/Carthage/Carthage/pull/1990#issuecomment-479442640).

takasurazeem:
Swift Package Manager lets you select schemes you want to build and so should Carthage allow now, there's a huge array of dependencies. The project I work on rely on Carthage and there's is one dependency called SkyFloatingLabelTextField that has an old Objective-C scheme that arbitrarily would be selected sometimes and sometimes not really depends if Carthage wants to be helpful with me and if it does select it, it would end with an error 65.

PR #1631, opened 2016-12-08
Reduce NSString API usages

ikesyo:


Source/CarthageKit/GitHub.swift
```diff
@@ -51,7 +51,7 @@ extension Repository {
 	/// Enterprise instances.
 	public static func fromIdentifier(identifier: String) -> Result<Repository, CarthageError> {
 		// GitHub.com
-		let range = NSRange(location: 0, length: (identifier as NSString).length)
+		let range = NSRange(location: 0, length: identifier.utf16.count)
```
mdiep: Is `utf16` correct here? Not `characters`?

ikesyo: That should be correct. The documentation would be a help: https://github.com/apple/swift/blob/swift-3.0.1-RELEASE/stdlib/public/core/String.swift#L137-L159.

PR #1638, opened 2016-12-12
[swift3-prep] Make scan's `into` default to `nil`

BasThomas:
#1558


ikesyo:
This does not looks correct to me, there are no default arguments for `Scanner` APIs: https://developer.apple.com/reference/foundation/scanner.

BasThomas:
You're right, let's not change this then.

PR #1668, opened 2016-12-28
Detect duplicates early

mdiep:
This is the first step in some cleanup I want to do—changing Cartfiles to have `Set`s of dependencies instead of `Array`s, changing `Resolver` to not know about `Cartfile`s, etc.

### DIFF
--- a/Source/CarthageKit/Cartfile.swift
+++ b/Source/CarthageKit/Cartfile.swift
@@ -90,18 +90,18 @@ public struct Cartfile {
 			return Cartfile
 				.from(string: cartfileContents)
 				.mapError { error in
-					if case let .duplicateDependencies(dupes) = error {
-						let dupes = dupes
-							.map { dependency in
-								return DuplicateDependency(
-									project: dependency.project,
-									locations: [ cartfileURL.carthage_path ]
-								)
-							}
-						return .duplicateDependencies(dupes)
-					} else {
+					guard case let .duplicateDependencies(dupes) = error else {
 						return error
 					}
+					
+					let dependencies = dupes
+						.map { dependency in
+							return DuplicateDependency(
+								project: dependency.project,
+								locations: [ cartfileURL.carthage_path ]
+							)
+						}
+					return .duplicateDependencies(dependencies)
 				}
 		} catch let error as NSError {
 			return .failure(CarthageError.readFailed(cartfileURL, error))
--- a/Source/CarthageKit/Cartfile.swift
+++ b/Source/CarthageKit/Cartfile.swift
@@ -69,14 +69,40 @@ public struct Cartfile {
 			}
 		}
 
-		return result.map { _ in cartfile }
+		return result.flatMap { _ in
+			let dupes = cartfile
+				.dependencyCountedSet
+				.filter { $0.1 > 1 }
+				.map { $0.0 }
+				.map { DuplicateDependency(project: $0, locations: []) }
+			
+			if !dupes.isEmpty {
+				return .failure(.duplicateDependencies(dupes))
+			}
+			return .success(cartfile)
+		}
 	}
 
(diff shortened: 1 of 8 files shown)


Source/CarthageKitTests/CartfileSpec.swift
```diff
@@ -74,23 +74,26 @@ class CartfileSpec: QuickSpec {
...
+					.sort { $0.description < $1.description }
+				expect(dupes.count) == 2
+				
+				let self2Dupe = dupes[0]
+				expect(self2Dupe) == ProjectIdentifier.gitHub(Repository(owner: "self2", name: "self2"))
+				
+				let self3Dupe = dupes[1]
+				expect(self3Dupe) == ProjectIdentifier.gitHub(Repository(owner: "self3", name: "self3"))
+			} else {
+				fatalError("should error")
```
mdiep: I wasn't sure if there was a better way to do this with Quick/Nimble. If we were using XCTest directly, I'd use `XCFail`. But this certainly works. 😕

ikesyo: FYI Nimble has `fail` functions: https://github.com/Quick/Nimble/blob/v5.1.1/Sources/Nimble/DSL.swift#L21-L35.

mdiep: Thanks! Xcode's code completion wasn't working. 🙃

ikesyo:
`let mainLocation = ...` in `ProjectSpec.swift` could be removed.

Source/CarthageKit/Cartfile.swift
```diff
@@ -69,14 +69,40 @@ public struct Cartfile {
...
 
 	/// Attempts to parse a Cartfile from a file at a given URL.
 	public static func from(file cartfileURL: URL) -> Result<Cartfile, CarthageError> {
 		do {
 			let cartfileContents = try String(contentsOf: cartfileURL, encoding: .utf8)
-			return Cartfile.from(string: cartfileContents)
+			return Cartfile
+				.from(string: cartfileContents)
+				.mapError { error in
+					if case let .duplicateDependencies(dupes) = error {
```
ikesyo: `guard` would be preferable:

```swift
guard case let .duplicateDependencies(dupes) = error else {
	return error
}

let dupes = dupes
	.map { dependency in
		return DuplicateDependency(
			project: dependency.project,
			locations: [ cartfileURL.carthage_path ]
		)
	}
return .duplicateDependencies(dupes)
```

Source/CarthageKitTests/CartfileSpec.swift
```diff
@@ -74,23 +74,26 @@ class CartfileSpec: QuickSpec {
...
-			expect(dupes.count) == 2
-
-			let self2Dupe = dupes[0]
-			expect(self2Dupe) == ProjectIdentifier.gitHub(Repository(owner: "self2", name: "self2"))
-
-			let self3Dupe = dupes[1]
-			expect(self3Dupe) == ProjectIdentifier.gitHub(Repository(owner: "self3", name: "self3"))
+			expect(result.error).notTo(beNil())
+
+			if case let .duplicateDependencies(dupes)? = result.error {
```
ikesyo: Ditto:

```swift
guard case let .duplicateDependencies(dupes)? = result.error else {
	fail("Cartfile should error with duplicate dependencies")
}

let dupes = dupes
	.map { $0.project }
	.sort { $0.description < $1.description }
expect(dupes.count) == 2

let self2Dupe = dupes[0]
expect(self2Dupe) == ProjectIdentifier.gitHub(Repository(owner: "self2", name: "self2"))

let self3Dupe = dupes[1]
expect(self3Dupe) == ProjectIdentifier.gitHub(Repository(owner: "self3", name: "self3"))
```

PR #1680, opened 2016-12-31
Remove SemanticVersion.pinnedVersion

mdiep:


Source/CarthageKitTests/VersionSpec.swift
```diff
@@ -64,83 +55,100 @@ class VersionSpecifierSpec: QuickSpec {
...
 			}
 
 			it("should allow only greater patch versions to satisfy 0.x") {
-				let specifier = VersionSpecifier.compatibleWith(versionZeroOne)
-				expect(specifier.isSatisfied(by: versionZeroOneOne.pinnedVersion!)) == true
-				expect(specifier.isSatisfied(by: versionZeroTwo.pinnedVersion!)) == false
+				let specifier = VersionSpecifier.compatibleWith(SemanticVersion.fromPinnedVersion(v0_1_1).value!)
+				expect(specifier.isSatisfied(by: v0_1_0)) == false
+				expect(specifier.isSatisfied(by: v0_1_1)) == true
+				expect(specifier.isSatisfied(by: v0_2_0)) == false
```
ikesyo: Looks like those do test the same patch version and lower patch versions but does not test `greater patch versions` now.

PR #1681, opened 2016-12-31
Sort Cartfile.resolved

### LINKED ISSUE #717: Sort Cartfile.resolved

mdiep:
Fixes #717.

### DIFF
--- a/Source/CarthageKit/Cartfile.swift
+++ b/Source/CarthageKit/Cartfile.swift
@@ -176,6 +176,7 @@ public struct ResolvedCartfile {
 extension ResolvedCartfile: CustomStringConvertible {
 	public var description: String {
 		return dependencies
+			.sort { $0.project.description < $1.project.description }
 			.map { $0.description + "\n" }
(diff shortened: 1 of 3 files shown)


Source/CarthageKit/Cartfile.swift
```diff
@@ -131,7 +131,7 @@ public func duplicateProjectsIn(_ cartfile1: Cartfile, _ cartfile2: Cartfile) ->
 /// checked out for each dependency.
 public struct ResolvedCartfile {
 	/// The dependencies listed in the Cartfile.resolved.
-	public var dependencies: [Dependency<PinnedVersion>]
+	public var dependencies: Set<Dependency<PinnedVersion>>
```
mdiep: I want to get rid of this entirely and use the dictionary below (with a similar change in `Cartfile`), but this is a good first step.

PR #1689, opened 2017-01-04
Rewrite ResolverSpec to contain all its data

mdiep:
This makes it _much_ easier IMO to understand what it’s testing and to add new tests. It removes the use of all files (test cartfiles, git repos) in favor of a `DB` struct.


Source/CarthageKitTests/ResolverSpec.swift
```diff
@@ -14,229 +14,306 @@ import ReactiveCocoa
...
+				updating: [ github2 ]
+			)
+			expect(resolved.value!) == [
+				(github3, .v1_2_0),
+				(github2, .v1_1_0),
+				(github1, .v1_0_0),
+			]
+		}
+		
+		pending("should resolve a subset when given specific dependencies that have constraints") {
```
mdiep: Because it doesn't pass. ☺️ 

I'm hoping that this simplification of these tests will let me expand the resolver and fix a number of lingering issues.

mdiep:
> Does it make sense to test the following condition?

Done. 👍

PR #1742, opened 2017-01-31
Start a framework for using xcodebuild: XCDBLD

mdiep:
I've been thinking for a while that we should have a separate framework for controlling `xcodebuild`. The logic of how that actually works obstructs the logic of CarthageKit. Ideally, all the specifics of how to serialize arguments and deserialize output would be hidden somewhere else.

This starts that framework: `XCDBLD.framework`.

I've only begun to move things into the framework—starting with the basic types. It's hard to move the code over because its very poorly factored—`xcodebuild` logic and `carthage` logic are intertwined. IMO this only strengthens the argument that we should have a separate framework.

I'd like to merge this and consider the work of extracting XCDBLD to be a work in progress. Trying to do it all at once would be a bit of a nightmare—and would definitely be a very long lived PR.

Once we're happy with the design of the framework, I think we can extract it to a separate repository.


Source/CarthageKit/XCDBLDExtensions.swift
```diff
@@ -1,33 +1,25 @@
-import Foundation
 import ReactiveSwift
-
-/// Describes how to locate the actual project or workspace that Xcode should
-/// build.
-public enum ProjectLocator: Comparable {
```
mdiep: Some parts of these types didn't fit will in XCDBLD, so they now live in extensions here. Some of them can probably move eventually, but it will require some work to do so.

ikesyo:
I'm not sure why but Xcode immediately crashes if I open `Carthage.xcworkspace`.

mdiep:
🙌

PR #1760, opened 2017-02-13
Implementing binary origin

michaelmcguire:
Implements #1308 

- Defines a new `binary` origin that must point to an `https` endpoint that returns the JSON-formatted definition of the binary-only vended frameworks.
- Documentation updated to describe format of JSON
- Current limitation: this JSON file is not cached even though the actual vended frameworks are


mdiep:
Thanks @michaelmcguire! 🎉

steipete:
Wohoo! That means we can now start offering [PSPDFKit](https://pspdfkit.com/) via Carthage. Thanks! 💙

mdiep:
@steipete Definitely open an issue (or better yet—a PR 😉) if you encounter any problems.

PR #1718, opened 2017-01-19
Don't pass Xcode errors through

mdiep:
Direct users to the xcodebuild log to help them self-diagnose errors.

Fixes #1626.

## Before
```
$ carthage update --no-use-binaries
*** Fetching Result
*** Checking out Result at "1.0.2"
*** xcodebuild output can be found in /var/folders/xj/sydx1wkd0p147x0k_7v667940000gn/T/carthage-xcodebuild.yIaqhL.log
*** Building scheme "Result-Mac" in Result.xcodeproj
** CLEAN FAILED **


The following build commands failed:
	Check dependencies
(1 failure)
** BUILD FAILED **


The following build commands failed:
	Check dependencies
(1 failure)
A shell task (/usr/bin/xcrun xcodebuild -project /Users/mdiep/Downloads/tmp/Carthage/Checkouts/Result/Result.xcodeproj -scheme Result-Mac -configuration Release ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean build) failed with exit code 65:
** CLEAN FAILED **


The following build commands failed:
	Check dependencies
(1 failure)
** BUILD FAILED **


The following build commands failed:
	Check dependencies
(1 failure)
```

## After
```
$ carthage-dev update --no-use-binaries
*** Fetching Result
*** Checking out Result at "1.0.2"
*** xcodebuild output can be found in /var/folders/xj/sydx1wkd0p147x0k_7v667940000gn/T/carthage-xcodebuild.NCkGIT.log
*** Building scheme "Result-Mac" in Result.xcodeproj
Build Failed
	Task failed with exit code 65:
	/usr/bin/xcrun xcodebuild -project /Users/mdiep/Downloads/tmp/Carthage/Checkouts/Result/Result.xcodeproj -scheme Result-Mac -configuration Release ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES clean build

This usually indicates that project itself failed to compile.Please check the xcodebuild log for more details: /var/folders/xj/sydx1wkd0p147x0k_7v667940000gn/T/carthage-xcodebuild.NCkGIT.log
```

PR #1761, opened 2017-02-14
Don't filter output so it still makes it to the log file

BobElDevil:
This fixes an issue I found, caused by the changes for Carthage/Carthage#1718

1718 filtered the output when !verbose, unfortunately this filtered it out from writing to the log file either, so when in non-verbose mode, you get no xcode output at all.

This instead just routes errors -> stdout when not verbose, thus sending all messages to the log file.


mdiep:
🤦🏼‍♂️

Thanks @BobElDevil!

PR #1764, opened 2017-02-15
Fix printing of newIssueURL

dcaunt:
By making this a `URL`, `absoluteString` is no longer optional.

Output was:

> *** Skipped building libextobjc due to the error:
> Dependency "libextobjc" has no shared framework schemes for any of the platforms: iOS
> 
> If you believe this to be an error, please file an issue with the maintainers at Optional("https://github.com/jspahrsummers/libextobjc/issues/new")
> 

Now:

> *** Skipped building libextobjc due to the error:
> Dependency "libextobjc" has no shared framework schemes for any of the platforms: iOS
> 
> If you believe this to be an error, please file an issue with the maintainers at https://github.com/jspahrsummers/libextobjc/issues/new

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -39,8 +39,8 @@ extension Repository {
 	}
 
 	/// The URL for filing a new GitHub issue for this repository.
-	public var newIssueURL: NSURL {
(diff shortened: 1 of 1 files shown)


mdiep:
Thanks!

PR #1765, opened 2017-02-15
Allow installation to any Volume

### LINKED ISSUE #1701: Installing to any disk other that the boot disk fails
The .pkg runs, and claims success. However....

1. a binary file has arbitrarily been placed on that disk in a fake /usr/bin/ folder
1. ...but no paths have been modified, so it's not actually "installed" it's merely "randomly deposited on the filesystem", and obviously doesn't work from command line
1. Bizarrely, that file cannot then be moved; some lock is placed on it and even the OS X root user cannot "mv" it.

Fortunately, I was able to delete the entire fake /usr/ folder in Finder, and start again.

Suggestions:

1. Remove the dialogue options in installer for choosing a drive, since they don't work
1. Eventually, if someone makes this work correctly (including modifying the path, etc), re-add them.

dcaunt:
This should fix #1701 – tested by installing to a new `dmg`.

### DIFF
--- a/Carthage.xcodeproj/project.pbxproj
+++ b/Carthage.xcodeproj/project.pbxproj
@@ -1030,7 +1030,7 @@
 			baseConfigurationReference = D0D1213419E878CC005E4BAA /* Mac-Application.xcconfig */;
 			buildSettings = {
 				INFOPLIST_FILE = Source/carthage/Info.plist;
-				LD_RUNPATH_SEARCH_PATHS = "@executable_path/. @executable_path/CarthageKit.framework/Versions/Current/Frameworks /Library/Frameworks /Library/Frameworks/CarthageKit.framework/Versions/Current/Frameworks $(inherited)";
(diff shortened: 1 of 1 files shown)


mdiep:
This seems like a weird thing to do. And even if it can find `CarthageKit.framework`, it still won't be in the path. 😕 

Unless this is something you'd actually want, I think we're better off disabling the drive selection in the installer.

dcaunt:
This isn't something that I need personally, no. Are you aware of a mechanism to restrict drive selection? I couldn't find anything when I looked at this.

mdiep:
I'm not too familiar with how this works, but it does look to be possible. [This](https://developer.apple.com/library/content/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html) documentation page for `pkgbuild` specifies `customLocation` and `customLocationAllowAlternateVolumes` can control this. But I'm not sure whether we currently have a distribution XML file.

dcaunt:
Got it. We don't have the distribution file but it [looks straightforward to implement](https://thegreyblog.blogspot.co.uk/2014/06/os-x-creating-packages-from-command_2.html). It seems that `domains` is actually the key we need.  I'll close this in favour of a PR which restricts installation to a volume with Mac OS installed.

PR #1772, opened 2017-02-20
Restrict installation to volumes with macOS

### LINKED ISSUE #1701: Installing to any disk other that the boot disk fails

dcaunt:
Fixes #1701 

If you check this out and run `Carthage.pkg` you can see there's a different disk selection UI. We _could_ restrict installation to the local system volume, but this mechanism already limits installation to Volumes containing macOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ PREFIX?=/usr/local
 
 XCODEFLAGS=-workspace 'Carthage.xcworkspace' -scheme 'carthage' DSTROOT=$(TEMPORARY_FOLDER)
 
+INTERNAL_PACKAGE=CarthageApp.pkg
 OUTPUT_PACKAGE=Carthage.pkg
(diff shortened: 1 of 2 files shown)


mdiep:
This is great. Thanks!

PR #1779, opened 2017-02-22
Use checked out cartfile when appropriate

BobElDevil:
I'm not sure if this is desired behavior or not, so opening this to start the discussion. 

Noticed something while trying to create a test case that doesn't rely on real repositories (for [this comment](https://github.com/Carthage/Carthage/pull/1489#discussion_r102023679))

When fetching Cartfile information to get the list of nested dependencies, we always fetch the data from the git repository, even if we've already performed a checkout. This seemed a little odd, because I would expect something like `carthage build` to just build what's there, without needing to look in the git repo with a possible clone/fetch operation. 

So I took a stab at having the dependency list be fetched from the checked out dependency directory instead of the git repository, when appropriate.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -296,8 +296,8 @@ public final class Project {
 	}
 
 	/// Produces the sub dependencies of the given dependency
-	func dependencyProjects(for dependency: Dependency<PinnedVersion>) -> SignalProducer<Set<ProjectIdentifier>, CarthageError> {
-		return self.dependencies(for: dependency)
+	func dependencyProjects(for dependency: Dependency<PinnedVersion>, usingCheckout: Bool) -> SignalProducer<Set<ProjectIdentifier>, CarthageError> {
+		return self.dependencies(for: dependency, usingCheckout: usingCheckout)
 			.map { $0.project }
 			.collect()
 			.map { Set($0) }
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -398,20 +398,33 @@ public final class Project {
...
-			return self.cloneOrFetchDependency(dependency.project, commitish: revision)
-				.flatMap(.concat) { repositoryURL in
-					return contentsOfFileInRepository(repositoryURL, CarthageProjectCartfilePath, revision: revision)
-				}
-				.flatMapError { _ in .empty }
-				.attemptMap(Cartfile.from(string:))
+			let cartfileProducer: SignalProducer<Cartfile, CarthageError>
+			let dependencyCheckoutURL = self.directoryURL.appendingPathComponent(dependency.project.relativePath)
+			 // If the checkout directory doesn't exist, it may be a binary download, so fall back to fetching the cartfile
+			if usingCheckout && FileManager.default.fileExists(atPath: dependencyCheckoutURL.path) {
```
jdhealy: With https://github.com/BobElDevil/Carthage/commit/c9d6c6fb059206590b1cf1ba0d532e2deae7d3af, there‘s a race condition between this `fileExists` call and when the SignalProducer is started into the file being read.

To minimize this time difference, we‘d want the `fileExists` call as part of the SignalProducer chain.

Source/CarthageKit/Project.swift
```diff
@@ -398,20 +398,33 @@ public final class Project {
...
+				let cartfileURL = dependencyCheckoutURL.appendingPathComponent(CarthageProjectCartfilePath)
+				cartfileProducer = SignalProducer<Cartfile, CarthageError>
+					.attempt {
+						// Ignore errors since the cartfile may not exist if it's a leaf in the dependency graph.
+						return Cartfile.from(file: cartfileURL).recover(with: Cartfile.from(string: ""))
```
jdhealy: Won‘t this `recover` also preempt [parsing errors (for example)](https://github.com/Carthage/Carthage/blob/9f1cc9e/Source/CarthageKit/Cartfile.swift#L73)?

Lesser point: just semantically (not sure it has meaningful effect one way or the other), I‘d rather avoid the `Cartfile.from(string: "")` in favor of having the SignalProducer not produce a value (`flatMap`ing `.empty` or the like).

BobElDevil: Ah yes it will. I can do another file check separately that's pulled into the chain like the one above. That way it mimics the other code path and will error on parsing, but not on missing files.

BobElDevil:
So I wasn't thinking of this as a performance gain. (it might be a little faster if only because we're not invoking git operations) but that wasn't the original motivation.

The issue I ran into is that we have a test case that uses real github projects to test some build caching. The test fixture that's zipped up already has everything setup, as if you had run `carthage checkout` in the root of it. So no additional fetching/cloning is needed, this is purely a building test. So the issue with the test is the projects only built with swift2.3. When I went to fix it, I created some completely fake projects that build really fast, and won't be reliant on swift version (it's basically an empty framework), and mimic'd the same structure. (3 dependencies: A depends on B depends on C). When I went to run a `carthage build` / the unit test, I was surprised to see it fail because it wanted to fetch / clone the dependency repository to look at the Cartfile.

From a flow perspective this seemed odd because it already has all that information in the Checkouts folder. It also is a correctness thing for a (maybe not even officially supported) edge case, where if you're working with submodules and modify a dependency Cartfile and commit it, the `carthage build` command will still check the old Cartfile for build ordering, while it will build the new version of the code. So to me it seems like if we have the source sitting there, and we're asked to build it, we should just reference the source for our build needs.

Note that if it's decided we should still always look at git, the test isn't blocked. I can just dump these fake frameworks into some dummy repos somewhere, or force the test to 'clone' all 3 into the shared cache prior to running the test, so it doesn't actually need to find a real endpoint. Making my test writing easier wasn't the _only_ motivator though =)

jdhealy:
Of the three building commands (`update`, `bootstrap`, `build`), all three resolve what to build based on the combined Cartfiles.

Because `build` doesn’t de facto entail writing (and potentially overwriting) in the `Carthage/Checkouts` directory, maybe one intuits that, for a `build` command, an on-disk Cartfile in a `Carthage/Checkouts` subdirectory might trump part of a `Cartfile.resolved` resolution.

In my opinion, all three command should continue to use the same resolution basis (combined Cartfiles) rather than have one differ.

jdhealy:
There might be an opportunity for better user-facing warnings. You brought up a considerable point of confusion in the case: 

- of a submodule
  - especially if that submodule is a branch-pointing dependency
    - in which, that on-disk checkout has had commits added to that branch
    - and, the `Cartfile.resolved` still points to the previous branch `HEAD`

If you'd like to open an issue or another exploratory PR, it would be greatly appreciated (thanks again!)

PR #1780, opened 2017-02-22
Readme: Added leading slash for shell example

louisdh:
Turns out `bin/sh` does not work, and will make you tear your hair out for 15 minutes until you notice the missing leading slash. This readme change will hopefully spare some people their sanity 🙂


mdiep:
Oof. Thanks!

PR #1786, opened 2017-02-23
Add support for .tar.gz binary archives

dcaunt:
Some organisations provide binary releases in .tar.gz archives rather than zips. With support for extracting these, it becomes easier for them or a third-party to provide the JSON files for binaries. 

This could be expanded to support BZIP etc.

### DIFF
--- a/Source/CarthageKit/Archive.swift
+++ b/Source/CarthageKit/Archive.swift
@@ -36,9 +36,41 @@ public func unzip(archive fileURL: URL, to destinationDirectoryURL: URL) -> Sign
 		.then(SignalProducer<(), CarthageError>.empty)
 }
 
+/// Untars the gzipped archive at the given file URL, extracting into the given
+/// directory URL (which must already exist).
+public func untargz(archive fileURL: URL, to destinationDirectoryURL: URL) -> SignalProducer<(), CarthageError> {
+	precondition(fileURL.isFileURL)
+	precondition(destinationDirectoryURL.isFileURL)
+
+	let task = Task("/usr/bin/env", arguments: [ "tar", "-xzf", fileURL.carthage_path, "-C", destinationDirectoryURL.carthage_path ])
+	return task.launch()
+		.mapError(CarthageError.taskError)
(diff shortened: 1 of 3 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -523,7 +523,13 @@ public final class Project {
 	/// Sends the temporary URL of the unzipped directory
 	private func unzipAndCopyBinaryFrameworks(zipFile: URL) -> SignalProducer<URL, CarthageError> {
 		return SignalProducer<URL, CarthageError>(value: zipFile)
-			.flatMap(.concat, transform: unzip(archive:))
+			.flatMap(.concat) { fileURL -> SignalProducer<URL, CarthageError> in
+				if fileURL.pathExtension == "gz" {
+					return untargz(archive: fileURL)
+				} else {
+					return unzip(archive: fileURL)
+				}
```
mdiep: Would you mind adding an `unarchive` function that contains this logic? And rename this function to `unarchiveAndCopyBinaryFrameworks`?

Source/CarthageKit/Archive.swift
```diff
@@ -36,9 +36,41 @@ public func unzip(archive fileURL: URL, to destinationDirectoryURL: URL) -> Sign
...
+		.flatMap(.merge) { directoryURL in
+			return unzip(archive: fileURL, to: directoryURL)
+				.then(SignalProducer<URL, CarthageError>(value: directoryURL))
+		}
+}
+
+
+/// Untars the gzipped archive at the given file URL into a temporary directory, 
+/// then sends the file URL to that directory.
+public func untargz(archive fileURL: URL) -> SignalProducer<URL, CarthageError> {
```
mdiep: I think we can make `untargz` and `unzip` private if we add a public `unarchive`.

mdiep:
It seems like a test may have broken… but I'm rerunning CI just in case.

dcaunt:
Ah, this is my fault - tests in ArchiveSpec will need updating.

mdiep:
Thanks!

Issue #1785, opened 2017-02-23
--cache-builds will skip building even if a commit different to what's been built is locally available


mz2:
The new --cache-builds behaviour is not behaving at least to my expectation if I use it together with `--use-submodules` for a local package editing workflow. Here's a demonstration:

1.  Build your dependent package X: `carthage build --cache-builds X` where X is a package that's been pulled in with `--use-submodules` to `Carthage/Checkouts/X`.
2. Make local changes to X and commit them to the submodule at `Carthage/Checkouts/X`.
3. Repeat building X: `carthage build --cache-builds X` – observe that X is skipped building and the commit from the local checkout was not read, presumably because the comparison is only done against the commitish in Cartfile.resolved.

I would argue this is a bug, because `carthage build` without --cache-builds will in fact build the current revision from `Carthage/Checkouts/X`. Requiring the editing of Cartfile.resolved as part of a local package editing workflow would be both illogical (leaks the implementation detail of what keys are used to invalidate the cache) and actually kind of dangerous because you may end up committing by accident for example changes to `Cartfile.resolved` that have intentionally not been pushed out yet.

* `carthage version`: current head of master (9f1cc9ef73c40eb83f3c251dad66b2a3f958e2dd).
* `xcodebuild -version`: Xcode 8.2.1 (8C1002)
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? yes

BobElDevil:
I think it would make sense to compare against the current 'subproject commit' value. That way we could also detect 'dirty' working trees as well and rebuild. (though with 'dirty' trees we would have to make sure that commit-dirty != commit-dirty).

Non submodule usage still won't be able to detect changes to the working copy, not sure if we want to get into that game. It might do to just say "if you're not using submodules you shouldn't be touching the Checkouts folder"

mz2:
> Non submodule usage still won't be able to detect changes to the working copy, not sure if we want to get into that game. It might do to just say "if you're not using submodules you shouldn't be touching the Checkouts folder"

With or without this particular feature, I think that's a pretty reasonable guidance to give.

scottrhoyt:
I think `--use-cache` should be incompatible with `--use-submodules` for the time being. The guidance should remain that modifying Checkouts without `--use-submodules` results in undefined behavior. What do you think about that?

mz2:
Why exactly should those be incompatible?

scottrhoyt:
[...]

PR #1790, opened 2017-02-25
Update README for cache and submodule incompatibility.

scottrhoyt:
Choosing the easy route from #1785 for starters.


mdiep:
Thanks!

PR #1795, opened 2017-02-28
Ignore draft releases for the new Carthage release check

### LINKED ISSUE #1791: The check for new Carthage release should ignore draft releases
```
*** Please update to the latest Carthage version: 0.20.0. You currently are on 0.19.1
```

This shouldn't happen until the release has been marked as a non-draft.

ikesyo:
Resolves #1791.

### DIFF
--- a/Source/carthage/BuildVersion.swift
+++ b/Source/carthage/BuildVersion.swift
@@ -20,7 +20,7 @@ public func localVersion() -> SemanticVersion {
 
 public func remoteVersion() -> SemanticVersion? {
 	let latestRemoteVersion = Client(.dotCom)
-		.releases(in: Repository(owner: "Carthage", name: "Carthage"), perPage: 1)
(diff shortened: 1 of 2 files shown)


Source/carthage/BuildVersion.swift
```diff
@@ -21,7 +21,9 @@ public func localVersion() -> SemanticVersion {
 public func remoteVersion() -> SemanticVersion? {
 	let latestRemoteVersion = Client(.dotCom)
 		.releases(in: Repository(owner: "Carthage", name: "Carthage"), perPage: 1)
```
mdiep: We should probably update this to be 2 per page. Otherwise this will require 2 network requests if the first release is a draft.

ikesyo: Nice catch! Addressed in 6d40232.

mdiep:
💖

Issue #1794, opened 2017-02-27
Getting error in Carthage Update


grangej:
* `carthage version`: .20
* `xcodebuild -version`: Xcode 8.2.1
* Are you using `--no-build`? yes
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? no

**Carthage Output**
Since updating to .20 I am getting the following error on one of my dependencies: 

`Failed to write to /*****/Carthage/Checkouts/*****/Carthage/Checkouts/logger: Error Domain=NSCocoaErrorDomain Code=516 "The file “logger” couldn’t be saved in the folder “Checkouts” because a file with the same name already exists." UserInfo={NSFilePath=/****/Carthage/Checkouts/******/Carthage/Checkouts/logger, NSUnderlyingError=0x7fb6b2b7cfb0 {Error Domain=NSPOSIXErrorDomain Code=17 "File exists"}}
`

I have tried removing this Checkout directory but the error simply comes back again sometimes in a different framework.

ikesyo:
This may be related to #1715.

jakeadams:
I am seeing the same issue. A clean bootstrap works fine, but subsequent bootstraps, even with `--cache-builds` is failing on the symlink file

mdiep:
Can you share a `Cartfile` that reproduces the issues? That will be a big help in fixing it.

vincentisambart:
I'm not sure it's the same problem, but it looks like it.
Cartfile
```
github "ikesyo/Himotoki" ~> 3.0
github "ishkawa/APIKit" ~> 3.0
```

```
$ carthage bootstrap --platform iOS
*** No Cartfile.resolved found, updating dependencies
*** Fetching Himotoki
*** Fetching APIKit
*** Fetching Result
*** Checking out Himotoki at "3.0.1"
*** Downloading Result.framework binary at "3.1.0"
*** Checking out APIKit at "3.1.2"
Failed to write to /private/tmp/test/Carthage/Checkouts/APIKit/Carthage/Checkouts/Result: Error Domain=NSCocoaErrorDomain Code=516 "The file “Result” couldn’t be saved in the folder “Checkouts” because a file with the same name already exists." UserInfo={NSFilePath=/private/tmp/test/Carthage/Checkouts/APIKit/Carthage/Checkouts/Result, NSUnderlyingError=0x7f9a1b60e580 {Error Domain=NSPOSIXErrorDomain Code=17 "File exists"}}

[...]

PR #1715, opened 2017-01-17
Before symlinking, filter submodules in `CarthageProjectCheckoutsPath` which would conflict with dependencies.

jdhealy:
*Supersedes #1493.*

- - - 

Previously, no symlinks would be created in projects handled as submodules through the `--use-submodules` flag. See #1484, thoughtbot/Argo#412.

`checkoutRepositoryToDirectory` no longer handles submodules of inputted repository's working tree. Remove currently never-overridden `shouldCloneSubmodule: Submodule -> Bool = { _ in true }` in `checkoutRepositoryToDirectory`.
`cloneSubmodulesForRepository` removed, `cloneOrFetchDependency` subsumes behavior for all pre-existing code paths.

In `symlinkCheckoutPathsForDependencyProject`, input dependency of `Dependency<PinnedVersion>` and no longer input  subDependencies of `Set<ProjectIdentifier>`.

PR #1805, opened 2017-03-01
Execute `git ls-tree` to list just the directory contents, without recursion

jdhealy:
*Hopefully, addresses most of #1794*

`git ls-tree -r` would list checked-in submodules, symlinks, the files inside recursed-through directories, but (problematically) not the directories themselves.

Instead, without recursion and with a trailing path separator, `ls-tree` lists just the directory contents, which is what should have been implemented to begin with.

Remove now unnecessary path shallowness check as the task now outputs only the desired depth.

Major apologies for introducing these errors in #1715.

### ~~TODO~~

- More Integration tests
  - [x] [Test in `/private` sub-directories](https://github.com/jdhealy/Carthage/commit/dcfd2444176dd42da3f4877faf95f28e8488d268#diff-45b1265a2399de422a4be9bc0dbcd748R6)
  - [x] [Test with conflicting `Carthage/Checkouts`-house directories, directories checked-in to `git`](https://github.com/jdhealy/Carthage/blob/dcfd2444176dd42da3f4877faf95f28e8488d268/Source/CarthageIntegration/checked-in-conflicting-differing-directory.bats)
  - [x] Test with conflicting `Carthage/Checkouts`-housed symlinks
   - conflicting in both the cases where 
      - [the symlink points to the same directory our symlink would](https://github.com/jdhealy/Carthage/commit/dcfd2444176dd42da3f4877faf95f28e8488d268#diff-469c149f42a15fb39e27c78049674004R68)
        - [x] Determine whether this errors or not
          - *It does, same write error.*
      - [the symlink points somewhere else entirely](https://github.com/jdhealy/Carthage/blob/dcfd2444176dd42da3f4877faf95f28e8488d268/Source/CarthageIntegration/checked-in-conflicting-differing-symlink.bats#L26-L28)
   - with these symlinks both coming from 
      - [a git repository checkout](https://github.com/jdhealy/Carthage/blob/dcfd2444176dd42da3f4877faf95f28e8488d268/Source/CarthageIntegration/checked-in-conflicting-differing-symlink.bats)
      - alternatively, from a post-`bootstrap` file-system write

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -918,9 +918,9 @@ public final class Project {
 						.isDirectoryKey
 					])
 
-					if dependencyCheckoutURLResource?.isSymbolicLink == .some(true) {
+					if dependencyCheckoutURLResource?.isSymbolicLink == true {
 						_ = dependencyCheckoutURL.carthage_path.withCString(Darwin.unlink)
-					} else if dependencyCheckoutURLResource?.isDirectory == .some(true) {
+					} else if dependencyCheckoutURLResource?.isDirectory == true {
 						// older version of carthage wrote this directory?
 						// user wrote this directory, unaware of the precedent not to circumvent carthage’s management?
 						// directory exists as the result of rogue process or gamma ray?
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -879,20 +879,7 @@ public final class Project {
 		return self.dependencyProjects(for: dependency)
 			.zip(with: // file system objects which might conflict with symlinks
 				list(treeish: dependency.version.commitish, atPath: CarthageProjectCheckoutsPath, inRepository: repositoryURL)
-					.flatMap(.merge) { (path: String) -> SignalProducer<String, CarthageError> in
-						let componentsRelativeToDirectoryURL = {
-							return URL(string: $0, relativeTo: self.directoryURL)?.standardizedFileURL.pathComponents
-						}
-						if
-							let components = componentsRelativeToDirectoryURL(path),
-							let comparator = componentsRelativeToDirectoryURL(CarthageProjectCheckoutsPath),
-							Array(components.dropLast(1)) == comparator // file system object is contained (shallowly) by `CarthageProjectCheckoutsPath`
-						{
-							return .init(value: components.last!)
-						} else {
-							return .empty
-						}
-					}
+					.map { (path: String) in (path as NSString).lastPathComponent }
 					.collect()
 			)
 			.attemptMap { (dependencies: Set<ProjectIdentifier>, components: [String]) -> Result<(), CarthageError> in
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -913,6 +913,24 @@ public final class Project {
 					let subdirectoryPath = (CarthageProjectCheckoutsPath as NSString).appendingPathComponent(name)
 					let linkDestinationPath = relativeLinkDestinationForDependencyProject(dependency.project, subdirectory: subdirectoryPath)
 
+					let dependencyCheckoutURLResource = try? dependencyCheckoutURL.resourceValues(forKeys: [
+						.isSymbolicLinkKey,
+						.isDirectoryKey
+					])
+
+					if dependencyCheckoutURLResource?.isSymbolicLink == .some(true) {
+						_ = dependencyCheckoutURL.carthage_path.withCString(Darwin.unlink)
+					} else if dependencyCheckoutURLResource?.isDirectory == .some(true) {
+						// older version of carthage wrote this directory?
+						// user wrote this directory, unaware of the precedent not to circumvent carthage’s management?
+						// directory exists as the result of rogue process or gamma ray?
+
+						// TODO: explore possibility of messaging user, informing that deleting said directory will result
+						// in symlink creation with carthage versions greater than 0.20.0, maybe with more broad advice on
+						// “from scratch” reproducability.
+						continue
+					}
+
 					do {
 						try fileManager.createSymbolicLink(atPath: dependencyCheckoutURL.carthage_path, withDestinationPath: linkDestinationPath)
 					} catch let error as NSError {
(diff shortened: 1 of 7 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -878,21 +878,8 @@ public final class Project {
...
-							let comparator = componentsRelativeToDirectoryURL(CarthageProjectCheckoutsPath),
-							Array(components.dropLast(1)) == comparator // file system object is contained (shallowly) by `CarthageProjectCheckoutsPath`
-						{
-							return .init(value: components.last!)
-						} else {
-							return .empty
-						}
-					}
+				list(treeish: dependency.version.commitish, atPath: CarthageProjectCheckoutsPath + "/", inRepository: repositoryURL)
+					.map { (path: String) in (path as NSString).pathComponents.last! }
```
ikesyo: You can use `lastPathComponent` here.

jdhealy:
```swift
} else if dependencyCheckoutURLResource?.isDirectory == .some(true) {
	// older version of carthage wrote this directory?
	// user wrote this directory, unaware of the precedent not to circumvent carthage’s management?
	// directory exists as the result of rogue process or gamma ray?

	// TODO: explore possibility of messaging user, informing that deleting said directory will result
	// in symlink creation with carthage versions greater than 0.20.0, maybe with more broad advice on
	// “from scratch” reproducibility.
	continue
}
```

Probably out-of-scope for this PR, but wanted to kick off discussion? Thoughts?

BobElDevil:
Well, I looked at the documentation we have [here](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#carthagecheckouts)

> You are not required to commit this folder to your repository, but you may wish to, if you want to guarantee that the source checkouts of each dependency will always be accessible at a later date.
>Unless you are using submodules, the contents of this directory **should not be modified**, as they may be overwritten by a future carthage checkout command.

It seems while we explicitly mention not to _modify_ the checkout folder, having it exist in the repo is supported. Messaging the user about the situation I think would be undesired as it's likely out of their control, since the dependency may be purposely versioning the Checkouts folder. Unless we want to adopt a more firm "Only use submodules or gitignore the Checkouts" in the documentation moving forward, and then put up a message about older versions, and recommending to maintainers that they stop tracking it.

jdhealy:
https://travis-ci.org/Carthage/Carthage/jobs/209034858 should be demonstrating a few failing test — not sure why all are reporting as “ok”…

Looking into this…

jdhealy:
Both a local VM and [Bitrise have failing tests as expected](https://www.bitrise.io/build/1531dd9ac3cd3944) for https://github.com/jdhealy/Carthage/commit/dcfd2444176dd42da3f4877faf95f28e8488d268, which is just the integration tests (precedes the fixes).

[Travis-CI has reports of issues with `set -e`](https://github.com/travis-ci/travis-ci/issues///3586#issuecomment-274585174) (not sure if `test -e` there is a typo) and [bats relies fundamentally on it (⌘F for `set -e`)](https://github.com/sstephenson/bats#bats-bash-automated-testing-).

This would call into question not just the tests I added, but also the `setup` and `teardown` of all the pre-existing tests.

https://github.com/jdhealy/Carthage/commit/dcfd2444176dd42da3f4877faf95f28e8488d268, in both a local VM and [on Bitrise](https://www.bitrise.io/build/1035a3ac824aedff), passes all integration tests, so I’m stripping the `[WIP]` from this PR and it’s ready for review.

Source/CarthageKit/Project.swift
```diff
@@ -926,6 +913,24 @@ public final class Project {
 					let subdirectoryPath = (CarthageProjectCheckoutsPath as NSString).appendingPathComponent(name)
 					let linkDestinationPath = relativeLinkDestinationForDependencyProject(dependency.project, subdirectory: subdirectoryPath)
 
+					let dependencyCheckoutURLResource = try? dependencyCheckoutURL.resourceValues(forKeys: [
+						.isSymbolicLinkKey,
+						.isDirectoryKey
+					])
+
+					if dependencyCheckoutURLResource?.isSymbolicLink == .some(true) {
+						_ = dependencyCheckoutURL.carthage_path.withCString(Darwin.unlink)
```
mdiep: Does `NSFileManager.removeItem(at:)` not work here?

Source/CarthageKit/Project.swift
```diff
@@ -926,6 +913,24 @@ public final class Project {
 					let subdirectoryPath = (CarthageProjectCheckoutsPath as NSString).appendingPathComponent(name)
 					let linkDestinationPath = relativeLinkDestinationForDependencyProject(dependency.project, subdirectory: subdirectoryPath)
 
+					let dependencyCheckoutURLResource = try? dependencyCheckoutURL.resourceValues(forKeys: [
+						.isSymbolicLinkKey,
+						.isDirectoryKey
+					])
+
+					if dependencyCheckoutURLResource?.isSymbolicLink == .some(true) {
```
mdiep: The `.some` isn't needed here or below.

mdiep:
There's a lot of code to that integration test. I wish there was an easier way to test this. 😕

Source/CarthageKit/Project.swift
jdhealy: `Darwin.unlink` differs in not removing directories.

But, since we have the check beforehand, if `removeItem` makes more sense to you?

jdhealy:
> There's a lot of code to that integration test. I wish there was an easier way to test this. 😕

Yeah, if `@test` could be inside a bash for-loop, those three added bats files would be one, but `bats` has a preprocessor which doesn’t accommodate that.

Source/CarthageKit/Project.swift
mdiep: Yeah, since we're checking let's use `removeItem`. I feel more comfortable with the `NSFileManager` APIs.

jdhealy: Less comfortable with `NSFileManager.removeItem(at:)` re-reading the docs, actually: “If the URL specifies a directory, the contents of that directory are recursively removed.”

Another thought: the slim possibility of a race condition between the check, swapping in a directory, and a recursive directory deletion.

Thinking I have a stronger opinion towards `Darwin.unlink` than earlier today.

Issue #1807, opened 2017-03-01
--cache-builds should use sha256 instead of md5


BobElDevil:
* `carthage version`:  0.20
* `xcodebuild -version`: 8.2.1
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? yes

As brought up [here](https://github.com/Carthage/Carthage/pull/1489#issuecomment-283006776), we should use sha256 instead of md5 for our cache checksums. 

Creating this issue to not lose track of this. When implemented, it'll incur a one-time cost of rebuilding (since all caches would be invalid as they would lack a sha256 value), but otherwise should be pretty straightforward to implement.

PR #1811, opened 2017-03-03
Change md5 to sha256

ffittschen:
This PR resolves issue #1807.

Other than previously stated [here](https://github.com/Carthage/Carthage/pull/1489#issuecomment-283006776), I used `/usr/bin/shasum -a 256` instead of `/usr/bin/openssl dgst -sha256`. It returns the hash before the path, so it is easier to parse, in case the path contains spaces.

I also renamed occurances of `md5` to `hash` (instead of `sha256`) in case the hashing function has to be replaced again some time in the future. Since this also included renaming the key in the VersionFile, I don't know if this causes trouble. @BobElDevil said in the issue above, that in case the hash doesn't match anymore, carthage will just rebuild the dependency, but what happens if the key doesn't match? I guess it will also just rebuild the dependencies, because it will fall through [this guard](https://github.com/Carthage/Carthage/blob/d21adfa7333764d257ead1513dff31aa99ac0f5c/Source/CarthageKit/VersionFile.swift#L235)

### DIFF
--- a/Source/CarthageKit/VersionFile.swift
+++ b/Source/CarthageKit/VersionFile.swift
@@ -258,14 +258,15 @@ private func md5ForFileAtURL(_ frameworkFileURL: URL) -> SignalProducer<String, 
 	guard FileManager.default.fileExists(atPath: frameworkFileURL.path) else {
 		return SignalProducer(error: .readFailed(frameworkFileURL, nil))
 	}
-	let task = Task("/usr/bin/env", arguments: ["md5", "-q", frameworkFileURL.path])
+	let task = Task("/usr/bin/shasum", arguments: ["-a", "256", frameworkFileURL.path])
 	return task.launch()
 		.mapError(CarthageError.taskError)
 		.ignoreTaskData()
 		.attemptMap { data in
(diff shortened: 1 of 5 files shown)


Source/CarthageKit/VersionFile.swift
```diff
@@ -120,24 +120,24 @@ struct VersionFile {
...
-		return SignalProducer<(String?, CachedFramework), CarthageError>(Swift.zip(md5s, cachedFrameworks))
-			.map { (md5, cachedFramework) -> Bool in
-				guard let md5 = md5 else {
+		return SignalProducer<(String?, CachedFramework), CarthageError>(Swift.zip(hashes, cachedFrameworks))
+			.map { (hash, cachedFramework) -> Bool in
+				guard let hash = hash else {
 					return false
 				}
-				return md5 == cachedFramework.md5
+				return hash == cachedFramework.hash
```
NachoSoto: This could be just `return hash == cachedFramework.hash`.

mdiep:
Thanks for the PR!

Issue #1810, opened 2017-03-03
Feature request: make copy-frameworks optionally strip architectures


NachoSoto:
In any project with several Carthage dependencies, calling `/usr/local/bin/carthage copy-frameworks` starts becoming a very large part of the compilation time. AFAIK most of that time is due to stripping architectures.

However, I just realized that if we made that optional, we could make it only do that when archiving for submission, and in that way make iteration times much lower.

mdiep:
Maybe we should just not do it for debug builds?

NachoSoto:
Some people might want to send debug builds to TestFlight. But I think that could be a simpler way, too, if we don't want to add a parameter, everyone would get the optimization for free

NachoSoto:
Or better yet, look at `ONLY_ACTIVE_ARCH`. If `YES`, don't try to strip it (it means we're debugging to just one architecture, either simulator or device), and that would also not be valid for production. I think that'd be the most accurate flag to use.

erichoracek:
After taking a shot at implementing this, as far as I can tell the (non-optional) code signing step vastly outweighs the stripping step in terms of time expense:

- Baseline in our project with ~30 frameworks: ~41s
- Without stripping architectures/headers ~41s
- Without code signing ~10s

Measured with Carthage `0.20.1-11-gf9f5575` when building for a device (`ONLY_ACTIVE_ARCH=YES`).

Perhaps I've missed something—the logic was as follows:

```swift
private func copyFramework(_ source: URL, target: URL, validArchitectures: [String]) -> SignalProducer<(), CarthageError> {
  return SignalProducer.combineLatest(copyProduct(source, target), codeSigningIdentity())
    .flatMap(.merge) { (url, codesigningIdentity) -> SignalProducer<(), CarthageError> in
      let strip: SignalProducer<(), CarthageError>

      // If only one architecture is active, we're building for speed and
      // thus can get away with not stripping architectures nor headers.
      if onlyActiveArchitecture() {
        strip = SignalProducer<(), CarthageError>.empty
      } else {
        strip = stripFramework(url, keepingArchitectures: validArchitectures)
      }

      let sign = signFramework(url, codesigningIdentity: codesigningIdentity)

      if buildActionIsArchiveOrInstall() {
        return strip
          .concat(sign)
          .then(copyBCSymbolMapsForFramework(url, fromDirectory: source.deletingLastPathComponent()))
          .then(SignalProducer<(), CarthageError>.empty)
      } else {
[...]

PR #1841, opened 2017-03-20
Copy frameworks concurrently (~4x speedup)

erichoracek:
From a quick test on a project with 30 frameworks, this speeds up the copy frameworks step for the device target in our application by nearly 4x on a quad-core machine. Before this change, the copy frameworks step takes ~40s, and after this change it takes ~10s.

### DIFF
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -41,7 +41,7 @@ public struct CopyFrameworksCommand: CommandProtocol {
 						}
 					}
 					// Copy as many frameworks as possible in parallel.
-					.observe(on: QueueScheduler(name: "org.carthage.CarthageKit.CopyFrameworks.copy"))
+					.start(on: QueueScheduler(name: "org.carthage.CarthageKit.CopyFrameworks.copy"))
 			}
 			.waitOnCommand()
 	}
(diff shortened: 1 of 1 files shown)


erichoracek:
@NachoSoto Thanks for the review! I've taken a shot at implementing #1810 but was unable to get a speed improvement from it. It seems code signing dramatically outweighs architecture stripping in terms of time expense.

Source/carthage/CopyFrameworks.swift
```diff
@@ -39,7 +39,9 @@ public struct CopyFrameworksCommand: CommandProtocol {
 										.then(SignalProducer<(), CarthageError>.empty)
 								}
 						}
-				}
+					}
+					// Copy as many frameworks as possible in parallel.
+					.observe(on: QueueScheduler(name: "org.carthage.CarthageKit.CopyFrameworks.copy"))
```
mdiep: It seems like this should be `start(on:)`, not `observe(on:)`?

mdiep:
It's a pity that ReactiveSwift doesn't have a `FlattenStrategy` to limit concurrency. That seems like it'd be a nice addition here.

Source/carthage/CopyFrameworks.swift
erichoracek: After measuring this again, it looks like `.merge` (with no scheduler) results in the same speedup as both observing or starting on a background scheduler. I'd assume that this is due to the fact that `Task` is non-blocking, and most of the work in `copy-framework` spent running `Task`s.

However, it does seem to be more correct to give each copied framework its own scheduler to ensure that if there ever is a blocking operation it doesn't result in another slowdown. Up to you!

mdiep: I wondered… since `observe(on:)` shouldn't have had an effect.

mdiep:
Thanks!

Issue #937, opened 2015-11-23
Would be useful to have carthage & xcode versions inside Cartfile.resolved


christopherstott:
At buddybuild, when we build iOS apps in the cloud, we infer the dependency managers apps/repos use, and make sure the environment is setup correctly for the app to build.

Sometimes builds rely on a particular version of Carthage being installed, so we'd like to make sure that it is.

With cocoapods, the Podfile.lock contains the cocoapod version - so we can make sure the right version is installed. Cartfile.resolved does not currently contain equivalent info. 

Also useful would be the xcode version that was used.

We'd be happy to put together a pull request to add these - I'm not sure what the Cartfile.resolve should look like with this info in, or maybe there's a better way to achieve this?

Chris

NachoSoto:
I think this makes a lot of sense, specially the Xcode version. Given the fact that Swift's ABI is not stable so compiled frameworks aren't backwards compatible, this would make things a lot better.
For example, it would mean that Carthage can automatically detect the version is old and not download a binary even though `--no-use-binaries` was not specified.

christopherstott:
@NachoSoto - any thoughts on what the format should be?

Something like this?

```
carthage "0.10.0"
xcode "7.1"
github "Quick/Nimble" "v0.2.0"
github "Quick/Quick" "v0.2.2"
github "ReactiveCocoa/ReactiveCocoa" "v2.4.3"
github "jspahrsummers/xcconfigs" "0.7.1"
```

mdiep:
I'd rather find a way to get the Swift version of (1) the currently selected Xcode and (2) the project in question. Adding the Xcode version specifically seems fragile, as there's not necessarily any relationship between Xcode and Swift versions.

NachoSoto:
> as there's not necessarily any relationship between Xcode and Swift versions.

In practice there is though, but I hear you. The reason why I think Xcode version is probably more appropriate is because it's general for `Obj-C` + `Swift` frameworks.

christopherstott:
For our particular use case, just knowing the xcode version they last ran carthage on would be useful.

Looks like you can get the swift version from 
```xcrun swift --version
Apple Swift version 2.1 (swiftlang-700.1.101.6 clang-700.1.76)
Target: x86_64-apple-darwin14.5.0```

rasharab:
We just hit an issue where having at least the carthage version in cartfile.resolved would've been helpful, as using the latest version of carthage broke the customer's build.

-Rashin

[...]

PR #1842, opened 2017-03-20
Print Carthage version into resolved file

Palleas:
This is a first attempt at addressing #937.

> TL;DR:
Sometimes builds rely on a particular version of Carthage being installed, so we'd like to make sure that it is.

We are still getting build errors due to the wrong version of carthage being used. After trying a few things, including treating the Carthage version as any other dependency (git/gitlab/binary) during parsing, this seemed like the best approach. 

Let me know what you think!


mdiep:
> We are still getting build errors due to the wrong version of carthage being used.

What kind of errors are you getting?

I don't think it makes sense to tie building to a particular Carthage version. And putting it `Cartfile.resolved`, while very pragmatic, doesn't make sense to me conceptually.

Palleas:
A good example would be #1794. When people end up having to downgrade waiting for a fix, they also have to switch the version they use in our dashboard.

> I don't think it makes sense to tie building to a particular Carthage version. And putting it Cartfile.resolved, while very pragmatic, doesn't make sense to me conceptually.

It's not so much about tying as it is about providing additional informations. Knowing the version they used locally makes it easier to help them when the build fails.

What you be your philosophy in that case? In #937, you mentioned adding other informations like Xcode and Swift toolchain. Would you rather have something like this?

@NachoSoto you mentioned in #937 that it made sense, do you have any input on this?

mdiep:
I definitely don't want to add anything that would complain if the local carthage version was _higher_ than the version in the resolved Cartfile.

I'd much rather focus on getting everyone on the latest version and refactoring to make everything more testable.

PR #1850, opened 2017-03-28
Create version files for prebuilt binaries

### LINKED ISSUE #1839: Verify that the Carthage build folder is in sync with the Cartfile.resolved
In a project that is frequently undergoing updates, a common pain point of development is accidentally building with a `Carthage/Build` folder that does not match the contents of your `Cartfile.resolved`. For example, if a bug was fixed in a dependency and a `carthage update` was performed by developer A, developer B must first make sure to run `carthage bootstrap` before they build and run the project or else they will encounter the same bug in their local environment.

As of the addition of `--cache-builds` command, `.Project.version` files are now written to the `Carthage/Build` directory. As such, I propose that these `.Project.version` files could be used in a new command to validate that the contents of the `Carthage/Build` folder is in sync with the `Cartfile.resolved`. This command could be added to a run script build phase with a `bash -e` shell to fail the build with a non-zero exit status and explanation of the dependencies that are out of date.

As far as I can tell, the majority of the work in implementing this command would be:
- Archiving `.Project.version` files into the archive when used with the `carthage archive` command
- Unpacking `.Project.version` files into the `Carthage/Build` from prebuilt binaries directory on checkout/bootstrap
- Iterating through all .`Project.version` files when this new command is invoked to verify that the `commitish` field matches the same field in the `Cartfile.resolved`

For this command, I propose the following syntax:

```
carthage verify
```

Invoking `carthage verify` with a build folder that is in sync with the `Cartfile.resolved` would produce the following output with a 0 exit status:

```
All builds are in sync with their resolved version.
```

Invoking `carthage verify` with a build folder that is _not_ in sync with the `Cartfile.resolved` would produce the following output with a non-zero exit status:

```
The following builds are out of sync with their resolved version:
*** ReactiveCocoa at "4.0.0", expected "5.0.0"
You must bootstrap to ensure they are in sync.
```

Finally, since many prebuilt binaries will not be compatible at first, I propose writing an abbreviated `.Project.version` file on checkout of a prebuilt binary (only if not included in the archive itself). This file would only include the `commitish ` field from the release that the prebuilt binary was attached to, e.g.:

```
{
  "commitish" : "5.0.0",
}
```

Please let me know if I've missed an exiting proposal for this. If not, let me know if this seems like something that would be a welcome addition to Carthage, and if so, whether it would be something that I could take a shot at implementing. Thanks!

erichoracek:
This change adds the following behavior:
- All `.Project.version` files matching the frameworks provided to the archive command are archived into the resulting zip.
- When the `build` command is performed with the `--no-skip-current` argument, a `.CurrentBuild.version` file is written to the `Carthage/Build` directory with no commitish field (it is not known at the time of build).
- When an archive is extracted, any `.Project.verion` files found within are extracted alongside the prebuilt frameworks and symbols.
  - If a `.CurrentBuild.version` file is found during extraction, it is rewritten as `.RepositoryName.version` with the `commitish` field populated with the `commitish` field of the dependency.

The general goal of this change is to ensure that every framework that was built with a version of Carthage after this change was introduced has a corresponding `.Project.version` file in the `Carthage/Build` directory, even if it is prebuilt.

Fixes #1839


mdiep:
I'm not sure that we should rely on a `.Project.version` file being included in the zip. We haven't required zips to be created with `carthage archive`, so this could break the workflow for some teams.

I think we have all the information we need to generate the file from the zips in the archive. It'd be preferable to do that if we can.

erichoracek:
I'm hesitant to _always_ attempt to generate the `.version` file for the following reasons:
- As far as I can tell, it would not be possible to populate the `xcodeVersion` field of the `.version` file if it was not archived into prebuilt binaries without Xcode.
- Additionally, this would also add the unnecessary overhead of computing the hash for a framework to every checkout from a release.
- Finally, the `commitsh` of frameworks that are archived alongside the primary framework in a repository could not be computed. For example, if archiving the ReactiveCocoa project and additionally embedding ReactiveSwift alongside it (via `carthage archive ReactiveCocoa ReactiveSwift` in the ReactiveCocoa repository), it would be impossible for consumers of the archive to know what version of `ReactiveSwift` was included without an embedded `.version` file.

To me, it seems like we should make a best effort to package these `.version` files into archives that are built using Carthage (since it will save time for all consumers and provide data that would otherwise be impossible to know), and then have a fallback abbreviated version file as discussed in #1839 if the archive does not include a `.version` file.

erichoracek:
I've updated this PR to now create an abbreviated `.Project.version` file if one is not present in the zip when downloading prebuilt binaries. This will occur if there are any prebuilt frameworks found that do are not included in a bundled `.Project.version`. file. If a `.Project.version` is found for every framework, no abbreviated `.Project.version` is generated.

erichoracek:
To clarify further, I believe that this addresses the following use cases:
- When downloading a prebuilt framework that was not packaged with `carthage archive,` a `.Project.version` file will still be created
- When download a prebuilt framework that was packaged prior to the version of Carthage that contains this change, a `.Project.version` file will still be created

This means that from this change forwards all dependencies will have a corresponding `.Project.version` file, regardless of whether they were built locally, attached to a github release, or downloaded from a `binary` json map.

erichoracek:
@mdiep is this change something you'd be open to? If you're not convinced of its usefulness, I can revert the behavior that includes `.Project.version` files in archives. Just let me know.

mdiep:
I think it should be fine to include. :+1 This PR is on my list, but I haven't quite gotten to it yet. Sorry!

mdiep:
Okay, sorry for the slow response on this one.

I've given this some more thought to this and looked at the code this weekend. And, unfortunately, I think we should always generate the version file.

The primary reason for this it that it will be easier to maintain. With two separate behaviors, we always risk breaking one without noticing. This is particularly true as most of the projects that Carthage contributors are likely to test with in the real world will be created with `carthage version`.

> I'm hesitant to always attempt to generate the .version file for the following reasons:
> 
> As far as I can tell, it would not be possible to populate the xcodeVersion field of the .version file if it is not present the archive alongside the prebuilt binaries.

This is true. However, the `xcodeVersion` is really just a poor proxy for the swift version. Since we now have code to read the Swift version—both of a compiled framework and from the local swift version—we should switch the cache to use this instead. Objective-C frameworks don't actually need to be rebuilt within the Xcode project changes.

Switching to the Swift version will let us generate the information we need.

> Additionally, this would also add the unnecessary overhead of computing the hash for a framework to every checkout from a release.

Since we're already doing this for non-binary releases, I don't think it's a big deal.

> Finally, the commitsh of frameworks that are archived alongside the primary framework in a repository could not be computed. For example, if archiving the ReactiveCocoa project and additionally embedding ReactiveSwift alongside it (via carthage archive ReactiveCocoa ReactiveSwift in the ReactiveCocoa repository), it would be impossible for consumers of the archive to know what version of ReactiveSwift was included without an embedded .version file.

That's fine. The commitish is for the repository that's responsible for the binary download. We don't really support embedding other frameworks anyway. If ReactiveCocoa embeds ReactiveSwift, ReactiveSwift will still get built on its own.

erichoracek:
All of your points sound reasonable. I've updated this PR to only create version files for `binary` and `github` references with prebuilt frameworks.

mdiep:
I'm traveling this week, so there may be a bit of a delay on my review. I'll try to get to this soon!

Source/CarthageKit/VersionFile.swift
```diff
@@ -40,7 +40,7 @@ extension CachedFramework: Decodable {
 
 struct VersionFile {
 	let commitish: String
-	let xcodeVersion: String
+	let xcodeVersion: String?
```
mdiep: I think we need to switch this to be the Swift version. Then we can store the Swift version for binary versions of Swift frameworks. (Failure to do so would basically be removing our detection that the Swift version changed.)

There are already methods in CarthageKit to get the Swift version from a compiled framework and from the currently selected Swift version, so I don't think this will be too hard.

erichoracek: @mdiep sounds reasonable. I've gone ahead and made a separate PR to do this, #1896

erichoracek:
@mdiep this now has the changes from #1896 integrated—it would be great if you could take another pass when you get a chance. Thanks for reviewing as always!

mdiep:
I'm very pleased with where this ended up! Thanks for all your effort on it!

PR #1862, opened 2017-03-30
Small change to README.md to save development time

dustinrwh:
After using Carthage for awhile, I realized that my builds during development were taking a little bit longer than expected. I then released that I really didn't need to copy-frameworks during the development because I wasn't submitting to the App Store. I made a small change to the script in the README to include a check to see if we are building for Release. This way, developers will save time as they do development builds. Every second counts 😃

Thanks for all the hard work. Carthage is great!


dustinrwh:
It appears that this may introduce issues when NOT in Release. Closing for now.

PR #1863, opened 2017-03-31
Add support for custom log paths.

### LINKED ISSUE #1860: Support for custom Log paths
* `carthage version`:  0.20.1
* `xcodebuild -version`:  Xcode 8.3 Build version 8E162

We are looking to capture build logs as part of our Artifacts. It would be nice if carthage supported a `--log-path` option on commands that generate a log.
As you can see in the Screenshot it's tricky to reconcile which builds belong to what build on the box. 

screen shot 2017-03-30 at 2 19 22 pm

A workaround for us at the moment is to nuke all the old logs. Or we could capture the last line of `carthage build` and grab it from there.

aaroncrespo:
closes #1860
depends on https://github.com/mdiep/Tentacle/issues/68

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -42,10 +42,10 @@ public struct BuildCommand: CommandProtocol {
 		public let colorOptions: ColorOptions
 		public let isVerbose: Bool
 		public let directoryPath: String
-		public let logPath: String
+		public let logPath: String?
 		public let dependenciesToBuild: [String]?
 
-		public static func create(_ buildOptions: BuildOptions) -> (Bool) -> (ColorOptions) -> (Bool) -> (String) -> (String) -> ([String]) -> Options {
+		public static func create(_ buildOptions: BuildOptions) -> (Bool) -> (ColorOptions) -> (Bool) -> (String) -> (String?) -> ([String]) -> Options {
 			return { skipCurrent in { colorOptions in { isVerbose in { directoryPath in { logPath in { dependenciesToBuild in
 				let dependenciesToBuild: [String]? = dependenciesToBuild.isEmpty ? nil : dependenciesToBuild
 				return self.init(buildOptions: buildOptions, skipCurrent: skipCurrent, colorOptions: colorOptions, isVerbose: isVerbose, directoryPath: directoryPath, logPath: logPath, dependenciesToBuild: dependenciesToBuild)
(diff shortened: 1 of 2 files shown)


aaroncrespo:
adds a `log-path` option which is a string path to a custom log output dir.

aaroncrespo:
https://github.com/Carthage/Carthage/pull/1866

Source/carthage/Build.swift
```diff
@@ -58,6 +59,7 @@ public struct BuildCommand: CommandProtocol {
 				<*> ColorOptions.evaluate(m)
 				<*> m <| Option(key: "verbose", defaultValue: false, usage: "print xcodebuild output inline")
 				<*> m <| Option(key: "project-directory", defaultValue: FileManager.default.currentDirectoryPath, usage: "the directory containing the Carthage project")
+				<*> m <| Option(key: "log-path", defaultValue: BuildCommand.defaultLogPath, usage: "path to the xcode build output. A temporary file is used by default")
```
mdiep: It'd be preferable to use a `nil` default value here. Then (1) the temporary value won't show up in the help description and (2) `openLogFile` won't need to check `path == BuildCommand.defaultLogPath`. I think you could get rid of `defaultLogPath` entirely and move the logic inside the method.

mdiep:
Thanks for the PR! I have one request about the handling of the option. Once that's fixed, this should be ready to go.

Source/carthage/Build.swift
aaroncrespo: I initially had that approach but then used this one for the PR because: 

1. It makes passing the value into Build from Update operations easier without duplicating the default string value
2. avoiding `mktemps` or documenting that `mktemps` will happen on input. When a custom log is used. A user would be surprised if they passed `custom-log-for-project-X` and the logs appeared in `custom-log-for-project-6`

I can move it back to optional to cleanup the help text, but I still think the equality check is handy. If it goes away we need to make the help text longer and explain the special behavior around `X` in paths.

BobElDevil: I don't think he's suggesting that we apply mktemps to provided log paths. Just that if you make it an optional `String?`, then at the point where you decide whether to use mktemps, it can become:
`if path == nil {`
instead of
`if path == BuildCommand.defaultLogPath {`

aaroncrespo: gotcha

mdiep:
This is great. Thanks!

PR #1866, opened 2017-03-31
Update Tentacle to avoid Swift 3.1 compiler crash.

jdhealy:
Temporarily remove Tentacle’s `create(file:path:repository:branch:)` — unused by carthage — due to Swift 3.1 compiler crash when attempting to specialize generics in this function.

See: <https://github.com/mdiep/Tentacle/issues/68>.

PR #1873, opened 2017-04-04
Remove hardcoding of swift version in tests.

scottrhoyt:
As mentioned in #1755 
Fixes tests in #1866 

@jdhealy, thanks!

### DIFF
--- a/Source/CarthageKitTests/XcodeSpec.swift
+++ b/Source/CarthageKitTests/XcodeSpec.swift
@@ -33,7 +33,7 @@ class XcodeSpec: QuickSpec {
 			_ = try? FileManager.default.removeItem(at: targetFolderURL)
 		}
 
-		describe("determingSwiftInformation") {
+		describe("determineSwiftInformation:") {
 			let currentSwiftVersion = swiftVersion.single()?.value
 			let testSwiftFramework = "Quick.framework"
 			let currentDirectory = URL(fileURLWithPath: FileManager.default.currentDirectoryPath, isDirectory: true)
--- a/Source/CarthageKitTests/XcodeSpec.swift
+++ b/Source/CarthageKitTests/XcodeSpec.swift
@@ -59,6 +59,7 @@ class XcodeSpec: QuickSpec {
 			it("should determine a framework's Swift version") {
 				let result = frameworkSwiftVersion(testSwiftFrameworkURL).single()
 
+				expect(FileManager.default.fileExists(atPath: testSwiftFrameworkURL.path)) == true
 				expect(result?.value) == currentSwiftVersion
 			}
 
--- a/Source/CarthageKitTests/XcodeSpec.swift
+++ b/Source/CarthageKitTests/XcodeSpec.swift
@@ -17,8 +17,6 @@ import Tentacle
 import XCDBLD
 
 class XcodeSpec: QuickSpec {
-	static let currentSwiftVersion = "3.0.2"
-
 	override func spec() {
 		// The fixture is maintained at https://github.com/ikesyo/carthage-fixtures-ReactiveCocoaLayout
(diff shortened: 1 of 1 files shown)


Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -36,6 +34,7 @@ class XcodeSpec: QuickSpec {
 		}
 
 		describe("determingSwiftInformation") {
+			let currentSwiftVersion = swiftVersion.single()?.value
```
mdiep: `currentSwiftVersion` should probably be non-optional. Otherwise if the command failed for some reason, it could mask a failure when we compare below.

scottrhoyt: This expectation:

```swift
expect(currentSwiftVersion?.isEmpty) == false
```

should fail if `currentSwiftVersion` is `nil`. I thought it might be better to have that test fail as an indicator that things aren't working as we would expect rather than having the tests crash. What  do you think @mdiep?

Source/CarthageKitTests/XcodeSpec.swift
```diff
@@ -35,7 +33,8 @@ class XcodeSpec: QuickSpec {
 			_ = try? FileManager.default.removeItem(at: targetFolderURL)
 		}
 
-		describe("determingSwiftInformation") {
+		describe("determineSwiftInformation:") {
+			let currentSwiftVersion = swiftVersion.single()?.value
 			let testSwiftFramework = "Quick.framework"
 			let currentDirectory = URL(fileURLWithPath: FileManager.default.currentDirectoryPath, isDirectory: true)
```
jdhealy: Is this current directory assumed (or, maybe guaranteed) to be the enclosing directory of `CarthageKitTests.xctest`?

Is `Quick.framework` assumed to be there as a built product of a (not executed by `CarthageKitTests.xctest`) Xcode build?

scottrhoyt: I think the current directory is guaranteed to be the directory that the test target is being run from. But I could be wrong on that.

Since the test is written as a quick spec, it is assumed that there will be a `Quick.framework` present. This assumption could be broken if Quick were integrated manually, integrated as a static framework, or if the current configuration were modified to load it from a different location.

In both cases, I think a failing test will give a good indication of where to sort out the problem. But perhaps a check to make sure that the test framework does in fact exist at the location it is assumed to be at would help someone disambiguate a test that failed because the test framework wasn't there versus one that failed because we couldn't determine it's Swift version. I went ahead and added that.

As far as writing a test that doesn't have these assumptions, I'm open to other suggestions on how to locate a framework that we are sure was compiled by the version of Swift used to compile the tests.

scottrhoyt:
Looks like the build failed for other reasons. Could someone with permissions please restart it? Thanks!

Source/CarthageKitTests/XcodeSpec.swift
mdiep: Ah, I missed that. Not sure how I feel about that approach, but it should be good enough. 👍

mdiep:
The build failed because of changes to the Travis CI image. Can you merge in master to pick up #1875? That should fix the build.

Source/CarthageKitTests/XcodeSpec.swift
jdhealy: Just noticed we build a fixtured Swift framework (`WorkspaceWithDependency`) later in `XcodeSpec`.

I'm gonna explore the possibility of sharing the side effects of this build with multiple `describe` and `it` calls…

And sorry about the Travis CI image thing, that was unfortunate timing, apologies.

mdiep:
Thanks!

PR #1534, opened 2016-10-16
Update XCode Scheme Share link.

orarbel:

PR #1879, opened 2017-04-09
Revert "Update XCode Scheme Share link."

cokochiaki:
Reverts Carthage/Carthage#1534


mdiep:
Why would we want to revert this? That link no longer works.

PR #1881, opened 2017-04-11
Extract temporary directory creation into reactive FileManager extension

erichoracek:
Now that `FileManager` has a `temporaryDirectory` property as of macOS 10.12, it makes sense to couple temporary directory creation logic to it.


mdiep:
Thanks!

PR #1884, opened 2017-04-14
[Minor Bugfix] Unwrap  currentLine  in ScannableError's description method

Nirma:
`currentLine` is being printed as an optional as is, so when this description function is called 
the contents of `currentLine` are being printed as `Optional("Nirma/Attributed")` instead of just `"Nirma/Attributed"`.

I discovered this while making a mistake myself after editing my `Cartfile` so I don't think users see it often but it bugged (knee slap) me enough to send over a PR.

How does it look?

Thanks!

### DIFF
--- a/Source/CarthageKit/ScannableError.swift
+++ b/Source/CarthageKit/ScannableError.swift
@@ -12,7 +12,7 @@ public struct ScannableError: Error {
 
 extension ScannableError: CustomStringConvertible {
 	public var description: String {
-		return "\(message) in line: \(currentLine)"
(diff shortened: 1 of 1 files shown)


mdiep:
It looks like you've mistakenly committed `Source/CarthageKit/.ScannableError.swift.swp`. Can you remove that?

Other than that, this looks great!

Nirma:
@mdiep Wow! Sorry for that slip up! I pushed again, how does it look now?

mdiep:
That's great! Thanks!

PR #1896, opened 2017-04-23
Replace Xcode version in version files with inferring Swift version at build time

erichoracek:
Now that we have a way to read Swift versions from frameworks directly, the Xcode version value can be removed from version files. In its place, the Swift version is now read directly from the cached framework at build-time rather than writing it directly to the version file, since Frameworks contain their Swift version.

This will be backwards-compatible with existing version files as they are only losing a key, not gaining a new one.

This was proposed as part of #1850, but it seemed to make sense to split it out into a new PR.


Source/CarthageKit/VersionFile.swift
```diff
@@ -115,29 +121,66 @@ struct VersionFile {
...
 			return SignalProducer(value: false)
 		}
 		
-		return self.hashes(for: cachedFrameworks, platform: platform, binariesDirectoryURL: binariesDirectoryURL)
+		return SignalProducer
+			.zip(
+				self.hashes(for: cachedFrameworks, platform: platform, binariesDirectoryURL: binariesDirectoryURL),
+				self.swiftVersionMatches(for: cachedFrameworks, platform: platform, binariesDirectoryURL: binariesDirectoryURL, localSwiftVersion: localSwiftVersion)
+			)
 			.collect()
```
mdiep: This `collect` could go inside the `zip`—then you wouldn't need to `map` below.

```swift
let hashes = self
    .hashes(for: cachedFrameworks, platform: platform, binariesDirectoryURL: binariesDirectoryURL)
    .collect()
let swiftVersionMatches = self
    .swiftVersionMatches(for: cachedFrameworks, platform: platform, binariesDirectoryURL: binariesDirectoryURL, localSwiftVersion: localSwiftVersion)
    .collect()
return SignalProducer
    .zip(hashes, swiftVersionMatches)
```

Source/CarthageKit/VersionFile.swift
```diff
@@ -239,11 +279,12 @@ public func versionFileMatches(_ dependency: Dependency<PinnedVersion>, platform
 
 	let platformsToCheck = platforms.isEmpty ? Set<Platform>(Platform.supportedPlatforms) : platforms
 
-	return currentXcodeVersion()
```
mdiep: I think `currentXcodeVersion` can be removed now.

erichoracek:
@mdiep thanks for the review! Let me know if you have any other requested changes.

Issue #1905, opened 2017-04-27
`--cache-builds` incorrectly detects some libraries as having an invalid cache in 0.21


klundberg:
This just recently started after I upgraded to 0.21. If I run the same carthage command multiple times in a row with --cache-builds specified, later times still try to rebuild dependencies even though they shouldn't be rebuilt. Here is the full command I'm running (building with Xcode 8.2.1):

`carthage bootstrap --platform iOS --toolchain com.apple.dt.toolchain.Swift_2_3 --no-use-binaries --cache-builds --configuration Debug`

In the carthage output below the libraries that are detected as invalid will always rebuild. My hunch as to the cause is that these libraries all use swift, while the ones that are treated as valid are objc-only (or at least they don't contain swift code).

* `carthage version`: 0.21
* `xcodebuild -version`: 8.2.1
* Are you using `--no-build`? No 
* Are you using `--no-use-binaries`? Yes
* Are you using `--use-submodules`? No
* Are you using `--cache-builds`? Yes

**Cartfile**
```
github "paypal/PayPal-iOS-SDK" == 2.14.5
github "iziz/libPhoneNumber-iOS" == 0.8.13
github "johnsundell/unbox" == 1.9
github "johnsundell/wrap" == 1.1.1
github "FutureKit/FutureKit" == 2.3
github "jdg/MBProgressHUD" == 0.9.2
github "wayfair/brickkit-ios" == 1.4.0
github "facebook/Shimmer" "bfeedb7"

github "seanscal/SDWebImage" "master"
github "klopez89/PagedArray" "release4.17"
github "seanscal/card.io-iOS-source" == 5.4.2

github "akashivskyy/GoogleAnalytics-Framework"
github "ekurutepe/CLTokenInputView" "master"

github "jonreid/OCMockito" == 3.0.2
github "hamcrest/OCHamcrest" == 5.3.0
github "Quick/Quick" == 0.9.3
github "Quick/Nimble" == 4.1.0
github "seanscal/Mockingjay" "MockingjayMinusDependencies"
```

**Carthage Output**
```
*** Checking out brickkit-ios at "1.4.0"
*** Checking out card.io-iOS-source at "5.4.2"
*** Checking out CLTokenInputView at "bb3064786da864ae68ef2d39c891cd630ca9eb48"
*** Checking out FutureKit at "2.3.0"
*** Checking out libPhoneNumber-iOS at "0.8.13"
*** Checking out MBProgressHUD at "0.9.2"
[...]

PR #1906, opened 2017-04-28
Implemented toolchain-aware swift version check

klundberg:
Potential fix for #1905.

I was able to manually test and verify that the cache was respected for non-default toolchains by using this cartfile:

```
github "Quick/Quick" == 0.9.1
```

and this command with Xcode 8.2.1 selected (using my locally built carthage with the implemented fix):

```
.../carthage update --toolchain com.apple.dt.toolchain.Swift_2_3 --platform osx --cache-builds
```

where the stock 0.21 version of carthage did not respect the cache.

I'm not 100% certain how this can be unit tested, though. This requires either the unit tests to be run on a machine that has Xcode 8-8.2.1, or perhaps that the machine in question has a custom toolchain installed from swift.org that can be used. Is there a reasonable way to unit test this I'm not thinking of?

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -514,12 +514,12 @@ public final class Project {
 	/// Updates the dependencies of the project to the latest version. The
 	/// changes will be reflected in Cartfile.resolved, and also in the working
 	/// directory checkouts if the given parameter is true.
-	public func updateDependencies(shouldCheckout: Bool = true, dependenciesToUpdate: [String]? = nil) -> SignalProducer<(), CarthageError> {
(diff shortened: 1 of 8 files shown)


mdiep:
> I'm not 100% certain how this can be unit tested, though. This requires either the unit tests to be run on a machine that has Xcode 8-8.2.1, or perhaps that the machine in question has a custom toolchain installed from swift.org that can be used. Is there a reasonable way to unit test this I'm not thinking of?

Yeah, this is an ongoing challenge for us. Making unit tests possible/easy is my focus when I have time to actually write code for Carthage.

For now, this is good enough.

mdiep:
I've released 0.22.0 and submitted the release to homebrew: https://github.com/Homebrew/homebrew-core/pull/12988

PR #1910, opened 2017-04-29
Checkout concurrently (~2.5x speedup)

erichoracek:
With our test project with 26 checkouts, before:

```
$ time carthage checkout
carthage checkout  11.63s user 5.81s system 97% cpu 17.848 total
```

after:

```
$ time carthage checkout
carthage checkout  13.68s user 6.97s system 270% cpu 7.618 total
```


erichoracek:
Per the discussion in #1839

erichoracek:
After taking a look at bootstrapping with a cold checkouts cache (removing `~Library/Caches/org.carthage.CarthageKit` before bootstrap), this should also be sped up by 2x with the latest commit.

before:

```
$ carthage checkout
carthage checkout  38.90s user 18.98s system 99% cpu 58.040 total
```

after:

```
$ carthage checkout
carthage checkout  36.84s user 16.64s system 40% cpu 2:12.18 total
``` 

As far as I can tell, there's no reason that cloning to the same directory would need to be on the same scheduler (although I do understand why adding submodules would be).

mdiep:
I feel a little uncomfortable with this until we can upgrade ReactiveSwift and get the new `.concurrent` merge strategy to limit the number of concurrent operations. 😕

andersio:
`OperationQueue` can be used as a transitional solution.

erichoracek:
I've updated this logic to checkout concurrently with a limit of the active number of processors. The speed seems unaffected, but it should hopefully prevent unbounded thread growth with a large dependency graph.

andersio:
Is there a problem with using an `OperationQueue` and `maxConcurrentOperationCount`? While both dispatch queues are at the same QoS classes in https://github.com/Carthage/Carthage/pull/1910/commits/58237f004848254b0a38a3842e11962a6ba80806, I'd still rather avoid semaphores, or at least the concurrent queue should get higher priority.

erichoracek:
The primary issue in my mind was that migrating to use `OperationQueue` would require a decent amount of code around managing an `Operation` subclass that wrapped a `SignalProducer` (since `BlockOperation` would be insufficient by itself as far as I could tell. It seemed like this may not be worth implementing when we'll soon have a first-class `.concurrent(...)` flatten strategy.

Can you briefly explain why we'd want to ensure that the concurrent queue has a higher priority?

andersio:
ICYMI you can use `OperationQueue` like a `DispatchQueue` with `BlockOperation`.

> Can you briefly explain why we'd want to ensure that the concurrent queue has a higher priority?

Semaphore is unsafe due to priority inversion. After a second thought, it appears that marking the concurrent queue is not enough, if the `producer` starts the work that could yield terminal event asynchronously in lower QoS classes.

erichoracek:
Makes sense—I understand why this would have priority inversion issues from the perspective of the queues having different priorities, but what is the issue if they share the same priority?

erichoracek:
With respect to using `BlockOperation` and `OperationQueue`, are you recommending that concurrent signal producers are implemented by blocking the `BlockOperation` using a construct like `SignalProducer.wait()` within the `BlockOperation`'s block?

andersio:
Hmm, Sorry about overlooking the needs here. [An `Operation` subclass is probably necessary](https://gist.github.com/andersio/33faf2fb9eddefee9bab80c1ea34ccda). Blocking operators use semaphore, so they won't help here.

> but what is the issue if they share the same priority?

Within the scope of this PR, there isn't an issue. But unless we are sure that every producer started by `ProducerQueue` performs work and sends events on the same QoS class as the `ProducerQueue`, it cannot be declared priority inversion safe.

erichoracek:
I've migrated to `OperationQueue` using your suggested `BlockOperation` subclass @andersio—please take a look when you have a chance. Thanks!

Source/CarthageKit/ProducerQueue.swift
```diff
@@ -8,50 +8,70 @@
...
+		return _isFinished && super.isFinished
+	}
+
+	var _isFinished: Bool = false {
+		willSet { willChangeValue(forKey: "isFinished") }
+		didSet { didChangeValue(forKey: "isFinished") }
+	}
+
+	init(_ block: @escaping (ManuallyFinishingOperation) -> Void) {
+		super.init()
```
andersio: Could you please make this an unowned reference? Other than this LGTM.

erichoracek: Sure—done.

Source/CarthageKit/ProducerQueue.swift
```diff
@@ -8,50 +8,73 @@
...
+/// Serializes or parallelizes (up to a limit) the execution of many
+/// SignalProducers, like flatten(.concat) or flatten(.merge), but without all
+/// needing to be enqueued in the same context.
 ///
 /// This allows you to manually enqueue producers from any code that has access
 /// to the queue object, instead of being required to funnel all producers
 /// through a single producer-of-producers.
 internal final class ProducerQueue {
-	private let queue: DispatchQueue
+	private let operationQueue: OperationQueue
```
mdiep: Would you mind creating a new queue class for this?

When we upgrade ReactiveSwift and get `.concurrent()`, we'll want to revert these changes. It'll be easier to do that if it's a separate class.

erichoracek:
@mdiep should be addressed—thanks!

mdiep:
This is really great. ✨

PR #1915, opened 2017-05-03
Rename ProjectIdentifier to Dependency

mdiep:
We still refer to `Dependency`s as _projects_ in many places, but I think this is a good start.

### DIFF
--- a/Source/CarthageKit/Errors.swift
+++ b/Source/CarthageKit/Errors.swift
@@ -237,14 +237,14 @@ extension CarthageError: CustomStringConvertible {
 		case let .missingEnvironmentVariable(variable):
 			return "Environment variable not set: \(variable)"
 
-		case let .noSharedFrameworkSchemes(Dependency, platforms):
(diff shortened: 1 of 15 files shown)


Source/CarthageKit/Errors.swift
```diff
@@ -237,14 +237,14 @@ extension CarthageError: CustomStringConvertible {
 		case let .missingEnvironmentVariable(variable):
 			return "Environment variable not set: \(variable)"
 
-		case let .noSharedFrameworkSchemes(projectIdentifier, platforms):
-			var description = "Dependency \"\(projectIdentifier.name)\" has no shared framework schemes"
+		case let .noSharedFrameworkSchemes(Dependency, platforms):
```
ikesyo: [nit] The variable should be lowercase.

mdiep: Oops! Search and replace gone wrong! 🙈

PR #1918, opened 2017-05-05
Log the version warning to stderr

BobElDevil:
We have some scripts that do different things based on the version of carthage they have. This allows us to roll out carthage updates across our build machines without any downtime when we uptake changes in carthage. 

Recently noticed a reversion in behavior, and it was because our usage of `carthage version` didn't work for parsing the version anymore as the warning about being out of date got in the way. It made sense to me that this sort of warning belongs in stderr, as it's not part of the 'normal' output of the command.

### DIFF
--- a/Source/carthage/main.swift
+++ b/Source/carthage/main.swift
@@ -21,8 +21,7 @@ guard ensureGitVersion().first()?.value == true else {
 }
 
 if let remoteVersion = remoteVersion(), localVersion() < remoteVersion {
-	let formatting = ColorOptions.Formatting(true)
-	fputs(formatting.bullets + "Please update to the latest Carthage version: \(remoteVersion). You currently are on \(localVersion())" + "\n", stderr)
+	fputs("Please update to the latest Carthage version: \(remoteVersion). You currently are on \(localVersion())" + "\n", stderr)
 }
 
 if let carthagePath = Bundle.main.executablePath {
--- a/Source/carthage/main.swift
+++ b/Source/carthage/main.swift
@@ -22,7 +22,7 @@ guard ensureGitVersion().first()?.value == true else {
 
 if let remoteVersion = remoteVersion(), localVersion() < remoteVersion {
 	let formatting = ColorOptions.Formatting(true)
-	carthage.println(formatting.bullets + "Please update to the latest Carthage version: \(remoteVersion). You currently are on \(localVersion())")
+	fputs(formatting.bullets + "Please update to the latest Carthage version: \(remoteVersion). You currently are on \(localVersion())" + "\n", stderr)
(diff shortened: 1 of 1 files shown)


mdiep:
My original thought was that lines with `***` could be ignored. But I understand that this is a pain in this particular case—we had to make some adjustments to our script as well.

Do you what the standard practice is for notices like this? I'm struggling to think of an example right now, but I know we're not the only ones who have update messages like this.

BobElDevil:
I know that homebrew routes these type of messages to stderr. Anything that is a warning is stderr even if it didn't actually break the command you ran. I can't say how common it is in general though.

Source/carthage/main.swift
```diff
@@ -22,7 +22,7 @@ guard ensureGitVersion().first()?.value == true else {
 
 if let remoteVersion = remoteVersion(), localVersion() < remoteVersion {
 	let formatting = ColorOptions.Formatting(true)
```
jdhealy: I’d lean against styling `stderr` output — for `stdout` we’ve followed the convention of `git` and other that when piping or when `env[TERM] == 'DUMB'` [(see checks here)]( https://github.com/Carthage/Carthage/blob/5f0d04aa95e2946cd3ee454c21ee6d29058177d2/Source/carthage/Environment.swift#L23-L39), we don’t apply styling.

Additionally, we provide the `--color never` flag+argument, we should avoid violating that.

Probably sensible to not style, as we have with other `stderr` output.

Definitely makes sense to message this on `stderr` though, PR's good for me with this change, thanks!

BobElDevil: Done
👍

jdhealy:
Seeming unrelated failure in `"should rebuild cached frameworks (and dependencies) whose version does not match the version file"` — attaching the Travis log and restarting the Xcode 8.2 Build+Test job.

[travis_build-230479163_job-230479165_log.txt](https://github.com/Carthage/Carthage/files/987867/travis_build-230479163_job-230479165_log.txt)

jdhealy:
[The restarted job](https://travis-ci.org/Carthage/Carthage/jobs/230479165) (see job worker information below) passed all tests, so `insert shrugging emoticon here`, LGTM

<details>
<summary>Restarted Job Worker information</summary>

```
hostname: wjb-2:cffe5885-829b-4106-aff6-ec50d3bd48b6
version: v2.6.2 https://github.com/travis-ci/worker/tree/fdccca4efd347ebc889baae641ccbf55bb871d19
instance: ef4bef6f-d8d9-4432-8729-eab45dc9ab6c:travis-ci-macos10.12-xcode8.2-1490755335
```
</details>

PR #1924, opened 2017-05-11
Reduce unnecessary calls of `gitRootDirectoryForRepository` in `submodulesInRepository`

ikesyo:
There should be no need to call it for each entry of `.gitmodules`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ matrix:
       language: objective-c
       osx_image: xcode8.2
       script:
-        - brew update
         - brew uninstall carthage
         - brew install bats
         - make install
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
       osx_image: xcode8.2
       script:
         - brew uninstall carthage
-        - brew install bats
+        - HOMEBREW_NO_AUTO_UPDATE=1 brew install bats
         - make install
         - bats Source/CarthageIntegration
       env:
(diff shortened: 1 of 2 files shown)


ikesyo:
I'm not sure why but this didn't pass the integration tests unfortunately...

mdiep:
Have you tried running those tests (or a manual recreation of them) locally?

jdhealy:
I ran them on Bitrise and saw the same `fatal` git results in the integration tests.

<em>Can comment again with those logs when I'm not on mobile</em>

.travis.yml
```diff
@@ -25,7 +25,6 @@ matrix:
       language: objective-c
       osx_image: xcode8.2
       script:
-        - brew update
```
ikesyo: https://travis-ci.org/Carthage/Carthage/jobs/231640984#L354

jdhealy: This preceded the `brew install bats`, but [bats hasn't had a release since 2014](https://github.com/sstephenson/bats/releases), so in a pragmatic sense, I agree that it's unnecessary.

ikesyo: > Update: …and homebrew updates on install commands too, right, right…

😞

jdhealy: ah, https://github.com/Homebrew/brew/issues/1670#issuecomment-267096602

DonkeyAfterDark:
I always want to bypass the "Updating Homebrew..." bit, because on my Network it hangs and never completes.

Issue #1379, opened 2016-06-30
Static *framework* support


davbeck:
I know that Carthage only support dynamic frameworks at the moment. I realize it's significantly more difficult to support plain static libraries because you need to keep track of both the binary library and the headers, but has there been much thought about support static _frameworks_?

Xcode doesn't provide a template for this out of the box, but if you create a dynamic framework target and then change the Mach-O type to static library, the compiled product is then a .framework, with proper modulemap and headers. I've been playing around with this today and it seems to work pretty well when just dropping the framework into a project. Here is a test framework that I made and used lipo to create a universal iOS binary: http://cl.ly/3s332V3g0S08. You can drop it into any project (both framework and app) and import like any other framework. Even in a Swift project. (`AThing.coolThing()` if you want to try and use the code in the framework.)

Note that there is one big downside to static frameworks in that they can not contain assets like images, but I would bet a large majority of libraries don't use assets.

I've been thinking about this quit a bit after attending WWDC 2016 Session 406 "Optimizing App Startup Time". In that talk they discuss how dynamic frameworks can slow down the launch of your app because of the work needed to link the libraries. According to the session, static libraries don't have this issue because the work is done at compile time.

mdiep:
We haven't supported static frameworks because AFAIK static frameworks are explicitly supported by Apple. But if they are, I'd certainly be willing to look at a PR that added support for them. I can't make any guarantees that I'll accept it though—it depends on the details and I'd need to play around with them.

NachoSoto:
> Xcode doesn't provide a template for this out of the box, but if you create a dynamic framework target and then change the Mach-O type to static library, the compiled product is then a .framework, with proper modulemap and headers. 

But the framework doesn't get linked **statically**, does it?

davbeck:
It does get linked statically as far as I can tell. `file` reports the framework is static and I can link to it in an app but not embed it and it will run on a device. Examining the ipa, the framework isn't included in the Frameworks directory.

Also, it looks like Fabric/Crashltyics is using a static framework, although I'm not sure what their build process looks like.

```bash
$ file Crashlytics
Crashlytics: Mach-O universal binary with 5 architectures
Crashlytics (for architecture armv7):	current ar archive random library
Crashlytics (for architecture armv7s):	current ar archive random library
Crashlytics (for architecture i386):	current ar archive random library
Crashlytics (for architecture x86_64):	current ar archive random library
Crashlytics (for architecture arm64):	current ar archive random library
```

[...]

PR #1932, opened 2017-05-15
Gracefully handle frameworks with no debug symbols

erichoracek:
This is a prerequisite to static framework support. See #1379.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -587,11 +587,24 @@ public func buildScheme(_ scheme: String, withOptions options: BuildOptions, inP
 			}
 		}
 		.flatMapTaskEvents(.concat) { builtProductURL -> SignalProducer<URL, CarthageError> in
-			return createDebugInformation(builtProductURL)
-				.then(SignalProducer<URL, CarthageError>(value: builtProductURL))
-		}
+			return UUIDsForFramework(builtProductURL)
+				.collect()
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -587,11 +587,24 @@ public func buildScheme(_ scheme: String, withOptions options: BuildOptions, inP
...
+					// one dSYM architecture UUID in the framework. This can 
+					// occur if the framework is a static framework packaged 
+					// like a dynamic framework.
+					if uuids.isEmpty {
+						return SignalProducer(value: builtProductURL)
+					}
+
+					return createDebugInformation(builtProductURL)
+						.then(SignalProducer<URL, CarthageError>(value: builtProductURL))
+				}
```
mdiep: Would you mind altering this slightly to move the `.then` outside?

```swift
			return UUIDsForFramework(builtProductURL)
				.collect()
				.flatMap(.concat) { uuids -> SignalProducer<URL, CarthageError> in
					// Only attempt to create debug info if there is at least 
					// one dSYM architecture UUID in the framework. This can 
					// occur if the framework is a static framework packaged 
					// like a dynamic framework.
					if uuids.isEmpty {
						return .empty
					}

					return createDebugInformation(builtProductURL)
				}
				.then(SignalProducer<URL, CarthageError>(value: builtProductURL))
```

(The return type in that `flatMap` is most likely not right now.)

PR #1933, opened 2017-05-15
Fix regression causing sorted build order to be lost

erichoracek:
This regressed in #1914


mdiep:
Good catch! 😅

A test would definitely be great here if we can add one.

erichoracek:
I've updated the `DependencyTest` frameworks to be dependent on one other, and updated the spec to verify their relative ordering—please take a look!

BobElDevil:
Thanks for the test! looks good to me

mdiep:
🤘

ikesyo:
🙈 Thanks for the fix!

PR #1935, opened 2017-05-17
[WIP] Support building static frameworks

### LINKED ISSUE #1379: Static *framework* support

ikesyo:
With this and #1932 we should be able to fully support static frameworks.

Should resolve #1379.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -308,7 +308,7 @@ private func mergeModuleIntoModule(_ sourceModuleDirectoryURL: URL, _ destinatio
 
 /// Determines whether the specified framework type should be built automatically.
 private func shouldBuildFrameworkType(_ frameworkType: FrameworkType?) -> Bool {
-	return frameworkType == .dynamic
+	return frameworkType != nil
 }
 
 /// Determines whether the given scheme should be built automatically.
@@ -330,7 +330,7 @@ private func shouldBuildScheme(_ buildArguments: BuildArguments, _ forPlatforms:
 			}
 		}
 		.filter(shouldBuildFrameworkType)
-		// If we find any dynamic framework target, we should indeed build this scheme.
+		// If we find any framework target, we should indeed build this scheme.
 		.map { _ in true }
 		// Otherwise, nope.
 		.concat(value: false)
@@ -644,7 +644,7 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
 			let xcodebuildAction: BuildArguments.Action = sdk.isDevice ? .archive : .build
 			return BuildSettings.load(with: argsForLoading, for: xcodebuildAction)
 				.filter { settings in
-					// Only copy build products that are dynamic frameworks
+					// Only copy build products that are frameworks
 					guard let frameworkType = settings.frameworkType.value, shouldBuildFrameworkType(frameworkType), let projectPath = settings.projectPath.value else {
 						return false
 					}
--- a/Tests/CarthageKitTests/XcodeSpec.swift
+++ b/Tests/CarthageKitTests/XcodeSpec.swift
@@ -243,7 +243,7 @@ class XcodeSpec: QuickSpec {
 			}
 		}
 
-		it("should skip projects without shared dynamic framework schems") {
+		it("should skip projects without shared framework schems") {
 			let dependency = "SchemeDiscoverySampleForCarthage"
 			let _directoryURL = Bundle(for: type(of: self)).url(forResource: "\(dependency)-0.2", withExtension: nil)!
 			let _buildFolderURL = _directoryURL.appendingPathComponent(Constants.binariesFolderPath)


ikesyo:
This might be annoying for some repositores who have both static and dynamic framework schemes (e.g https://github.com/BoltsFramework/Bolts-ObjC/tree/1.8.4/Bolts.xcodeproj/xcshareddata/xcschemes)

erichoracek:
Will this build static frameworks that have an output type of a `.a` file, or will it only build "fake" dynamic frameworks, e.g. a static framework wrapped within a `.framework` package?

At first glance, it feels like this may be something that should go behind a flag. This change could cause Carthage to start building targets that consumers may not want to build (especially if we're supporting overriding the `xcconfig`s to cause Xcode to build a dynamic framework as a static one). Perhaps something like `--dynamic` or `--static`, with `--dynamic` as the inferred default?

What do you think?

ikesyo:
@erichoracek 

> Will this build static frameworks that have an output type of a `.a` file, or will it only build "fake" dynamic frameworks, e.g. a static framework wrapped within a `.framework` package?

The former is just a static library, not a static framework I think. So I mean the latter here. An example is [BoltsFramework/Bolts-ObjC](https://github.com/BoltsFramework/Bolts-ObjC) backed by [the config](https://github.com/parse-community/xctoolchain/blob/master/Configurations/Product/StaticFramework.xcconfig) (the important part is `MACH_O_TYPE = staticlib`, which is `mh_dylib` for dynamic frameworks).

> At first glance, it feels like this may be something that should go behind a flag. This change could cause Carthage to start building targets that consumers may not want to build (especially if we're supporting overriding the `xcconfig`s to cause Xcode to build a dynamic framework as a static one). Perhaps something like `--dynamic` or `--static`, with `--dynamic` as the inferred default?

Adding a flag will be good, but `--dynamic` or `--static` seems bit weird to me––those imply complete switching. Even if the flag is specified, Carthage will not build dynamic frameworks as static frameworks and vice-versa.

The flag should let Carthage accept static framework schemes to be built. So I prefer something like `--allow-static-frameworks`.

ikesyo:
> Perhaps something like `--dynamic` or `--static`, with `--dynamic` as the inferred default?

Oh I may misunderstand that. Do you mean that only building dynamic ones or static ones?

erichoracek:
That make sense—thanks. Apologies for conflating library and framework—these are often used interchangeably in terms of code but can mean quite different things in Xcode land when discussing static vs. dynamic.

In terms of flags, I think we would want something that would allow you to build _either_ static or dynamic libraries. My thinking is that this could cause collisions in the `Carthage/Build/<Platform>` directory. As an example, Bolts can produce two frameworks both named `Bolts.framework` for iOS: one static and one dynamic.

drkibitz:
@ikesyo How's this coming? Really looking forward to carthage potentially supporting this through and through. My team is really on the fence of using Carthage at this point, and having the "full" functionality to use it without a Xcode + Carthage + Merge-Workaround would be great (reduced to just Xcode + Carthage). It becomes an easier sell as it is easier to manage in a larger team.

mdiep:
I'm okay with this change, but I think we need to figure out the naming collision problem first. I don't want to break support for any repos, and I think a flag isn't a great option.

If we see that there are multiple schemes that produce a product with the same name, maybe we should nest them in a folder named after the scheme? `Carthage/Build/iOS/Bolts-iOS-Dynamic/Bolts.framework`?

Or we could sequester static frameworks somewhere and leave dynamic frameworks where they are. `Carthage/Build/iOS/static/Bolts.framework`

ikesyo:
> If we see that there are multiple schemes that produce a product with the same name, maybe we should nest them in a folder named after the scheme? `Carthage/Build/iOS/Bolts-iOS-Dynamic/Bolts.framework`?
>
> Or we could sequester static frameworks somewhere and leave dynamic frameworks where they are. `Carthage/Build/iOS/static/Bolts.framework`

I prefer the latter.

erichoracek:
As of Xcode 9 beta 4, building a dynamic Swift framework target statically is [officially supported](https://twitter.com/daniel_dunbar/status/889546788633321472) and as simple as overriding the `MACH_O_TYPE` build setting with a value of `staticlib` (overriding `LD` as discussed [here](https://github.com/Carthage/Carthage/blob/master/Documentation/StaticFrameworks.md) is no longer necessary). Would it make sense for Carthage to add first-party support for building dynamic framework targets statically with this build setting overridden, rather than expecting all existing Carthage-compatible targets to add a new static framework target?

Perhaps this is a mutually exclusive idea—I could see Carthage having support for building static framework targets and building dynamic framework targets statically with the `MACH_O_TYPE` setting overridden. However, as far as I can tell, the build setting override approach would make adoption of static frameworks much easier than attempting to get all of the Carthage compatible projects to create a new static framework target, and could potentially trim complexity as well.

mdiep:
> Would it make sense for Carthage to add first-party support for building dynamic framework targets statically with this build setting overridden, rather than expecting all existing Carthage-compatible targets to add a new static framework target?

I think that could make sense. It's definitely an interesting idea. I'm curious what @ikesyo thinks.

I'd love to see a PR that explores it and includes documentation for both contributors and consumers of Carthage. I don't have a great sense of what this would involve.

drkibitz:
How's this work going?

Aranoledur:
Hey guys. What is holding this fix? Cocoa pods already supporting this, but a lot of frameworks can't support static linking. Maybe some help needed? As I think, names collision problem is not on carthage side, but a framework. Like, if you want to support both linking, just name your targets correctly, e.g. `Bolts-static.framework, Bolts-dynamic.framework`

wedkarz:
@Aranoledur agree! What is holding this off?

blender:
@wedkarz feel free to volunteer is taking this further.

wedkarz:
@blender Sure! What can I do?

blender:
@wedkarz I think the main problem is the naming collision. I.e. statically built Alamofire will clash with dynamically built Alamofire.

Suggestion so far has build separating the the name spaces in a `static` dir or otherwise ignore the issue.

You can advocate for one of the two and suggest a way forward. Or maybe take inspiration from what Cocoapods has done.

All ideas welcome at this stage imho.

Frizlab:
I’d personally vote for two folders, `dynamic` and `static` (naming does not matter) and compile within those two.

drkibitz:
+1 @Frizlab except naming does matter to me, would very much like it to correspond to Mach-O-Type values

Aranoledur:
@blender Hey! Can you suggest where to implement this logic? Or how? E.g. 
```swift
if target.matchOType == .static {
    createFolder("osType/Static")
} else {
    createFolder("osType/Dynamic")
}
```
Thats correct?

r-peck:
Using a directory for both seems like it would be a painful breaking change for anyone expecting the dynamic frameworks to be in a specific directory. Maybe just putting the static ones inside a directory while leaving the dynamic ones as is for now would make sense?

r-peck:
@Aranoledur in Source/CarthageKit/Xcode.swift there's a function `buildScheme` that does the building. In a closure inside that function, there's a call to `mergeBuildProducts` passing the output folder URL. It passes that to `copyBuildProductIntoDirectory`, which determines the final product URL. Changes to output directory or file name likely belong somewhere along that path. The `BuildSettings` struct has a `frameworkType` property on it, and that struct is available all along that chain of calls, so wherever the right place to adjust the URL is should have access to the info needed to branch based on dynamic vs static.

There is probably some more thought that needs to go into the UX around this. Given the same project might output both a dynamic framework and a static one, consuming projects probably don't want both in their framework search path. Maybe that's just on the user to sort out afterward for now (sort of like any other need to filter unneeded frameworks built by Carthage)?

mdiep:
If Carthage could reliably override the settings to turn dynamic framework targets into static framework targets, then we could conceivably add a flag for this instead of worrying about naming collisions.

r-peck:
@mdiep I'm not sure that would always be feasible. Any framework that needs resources will have to use a resource bundle when built static, but could bundle in the resources when dynamic. It also seems plausible that someone may want to use a mix of static and dynamic depending on where they need each dependency. Transitive dependencies could be problematic as well. If they are built statically because of the settings change but still linked into other frameworks Carthage builds, we start heading toward duplicate symbols problems.

The resources issue brings up a separate concern regarding when the resource bundle is built and where it goes. Should that be copied to the build directory too, or should there be a convention to put it in the static framework and let the consumer copy it correctly?

Aranoledur:
@r-peck I've solved name collision for result product by creating `Static` folder, but there is still this problem for `DerivedData`. `xcodebuild ... build` (not `archive`) command can't handle cases when you trying to build static target and `DerivedDataPath/Build/Products` has some binary package with the same name. E.g. `xcodebuild SocketRocket build` when there is `DerivedDataPath/Build/Products/.../SocketRocket` binary package. It won't compare mach-o type of the binary with the target's mach-o type and will pass this old package to `lipo`. And lipo will create a package with static binaries for device architectures and dynamic binaries for simulator architectures, `dsymutil` will be called and fail. Revers situation will be if there are static binaries in DerivedData and dynamic target. So the possible solutions are

1. Delete `DerivedDataPath/Build` folder before `xcodebuild` call
2. Change DerivedData path for static targets

r-peck:
Changing the derived data path seems more appropriate, but I think both approaches may have some impact on the cache builds behavior. I'm not familiar enough with how derived data is used in the cache, but it seems like there are probably broader implications.

Aranoledur:
Is it possible to merge my changes here, or I have to create another PR?

Aranoledur:
@r-peck I still can't figure out how to properly copy test framework from fixtures. Please help) Also decided to dwell on the solution that deletes Product folder, because derived data can be nil. Other things are done, was able to submit bundle to ITC.

r-peck:
> copy test framework from fixtures

Could you clarify what you mean by this? Is this referring to copying the built product from derived data to the final output location?

As far as deleting the product folder within derived data, my comments here have just been thoughts about possible approaches and pitfalls. I think this really needs input from someone with more authority on the project.

r-peck:
I'm honestly not familiar enough with the project to know whether or not that's the right place for that.

Issue #1936, opened 2017-05-17
Statically link Carthage frameworks into Carthage?


erichoracek:
One of the more confusing parts around how Carthage installation works currently is that it requires a dylib (itself with many more embedded dylibs) to be installed alongside the `carthage` binary (`CarthageKit.framework`). If we have support for #1379, would it be possible to statically link Carthage's dependencies into the primary `carthage` binary? If so, it could mean that there would be only one file to install.

The primary unanswered question in my mind is whether it's possible to statically link against the swift core dylibs. Does anyone know if there's an exposed way to do this? They seem to be present at this path `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift_static/macosx`, so perhaps it is as simple as linking against these `.a` files.

It would be a little strange for `carthage` to use the non-default flow for embedding built frameworks, but it is a CLI (rather than an app) so perhaps this may be a good choice for this scenario. Additionally, this would likely resolve issues where another version of `CarthageKit.framework` steaks into a user's `@rpath` before the one that they just downloaded, causing a new version of the `carthage` binary to use the wrong version of `CarthageKit.framework`.

Thoughts? Thanks for reading.

mdiep:
My preference would be to build Carthage itself with SwiftPM (#1559). That would accomplish the same thing.

@ikesyo was looking into this, but I don't know how far he made it.

erichoracek:
That's fair—although it may be a bit perplexing to observers that Carthage would be built using another dependency manager with a similar feature set. It seems to me to be a good goal for Carthage to support building itself—unless perhaps if Carthage is only considered suitable for building applications, not CLIs.

mdiep:
I think Carthage's current setup to build a CLI tool is too complicated. If we can simplify that in a supported way, then I'm all for it. But I think SwiftPM might be the only supported way to do so.

ikesyo:
I'm still working on SwiftPM build and that works fine basically. There is still a little problem about `Bundle`-related codes (we don't have bundled Info.plist in this case) and making the test suites work.

We can build a statically-linked binary using SwiftPM with the following:

```
swift build -c release -Xswiftc --static-stdlib
```

ikesyo:
I'm not sure about the difference or the internal, but using SwiftPM-built binary would be appropriate I think.

[...]

Issue #1559, opened 2016-11-04
Build carthage with SwiftPM


mdiep:
I think we should be able to build `carthage` with SwiftPM. I think that will let us distribute a single binary without the need to ship `.framework`s alongside it.

mdiep:
This should be doable now that we're on Swift 3.*



_*May require changes to our dependencies_

mdiep:
Looks like this doesn't work right now because:

1. We require macOS 10.11
2. SwiftPM builds for 10.10+
3. SwiftPM doesn't currently have the ability to specify a deployment target (but it will in Swift 5)

mdiep:
Resolved with #2665

PR #1938, opened 2017-05-18
Build carthage with SwiftPM

ikesyo:
Refs: #1559 and #1936 

We are not ready to run the test suites on SwiftPM nor distribute a binary built with it yet. But this should be a good first step.


ikesyo:
Thanks! :shipit:

PR #1943, opened 2017-05-19
Switch to Xcode 8.3

ikesyo:
SwiftPM job still uses Xcode 8.2 for a while.

Once this is merged, Xcode version in the homebrew formula should be updated as well for a next release.


ikesyo:
Thanks for the 0.23 release! Let's mege this now.

PR #1945, opened 2017-05-20
Add support for Swift Package Manager only projects

### LINKED ISSUE #1226: Supporting SwiftPM Generated Xcode Projects
:warning: *I wanted to open a discussion, I know it's obviously too soon to actually start supporting this but I wanted to gather feedback and create a discussion around it.* :warning:

In newer versions of SwiftPM you can generate Xcode projects (via `swift build -X`) which means that Carthage could potentially detect `Package.swift` (SwiftPM) and the lack of an Xcode project and then generate the project to use. This would make it far easier for package authors to support Carthage.

Another option could be adding a dependency of Carthage onto SwiftPM as a library, and calling the internals directly to build the the package as a framework for use by Carthage.

devxoul:
## Background

Swift Package Manager doesn't have Xcode dependency so the library authors can distribute their code without Xcode project files. Already there are many projects that don't have Xcode project files in their Git repository (e.g. [ReactorKit](https://github.com/ReactorKit/ReactorKit)). One of the most important design goal of Carthage is to reduce the responsibility of Xcode:

> Ultimately, we created Carthage because we wanted the simplest tool possible—a dependency manager that gets the job done without taking over the responsibility of Xcode, and without creating extra work for framework authors.

This PR makes Carthage to have less dependency of Xcode by supporting SPM-only libraries.

## Summary

* Run `swift package generate-xcodeproj` before each build processes if the dependency doesn't have both `.xcodeproject` and `.xcworkspace` but`Package.swift`.
* Add a fixture for testing SPM-only library.

## Related Issues

* Resolves #1226


devxoul:
@ikesyo, could you please review this PR when you're available? Thanks!

mdiep:
Last time I talked to the SwiftPM folks, it was suggested that we not rely on this functionality. I'll try to find someone at WWDC to talk to about this.

devxoul:
@mdiep, I'm curious why we should not rely on that function. I think unstable solution is better than no solution. Even we can offer this functionality as a beta (or with a explicit flag) and change the implementation future. I hope I could distribute my code-only libs to Carthage without Xcode :)

mdiep:
> I think unstable solution is better than no solution.

I don't think that's true.

An unstable solution would likely:

1. Greatly increase the support load for Carthage. Which would decrease the time available for development.

2. Remove the incentive for project creators to add proper Carthage support—which will hurt everyone when/if this function breaks

devxoul:
You might be true if the `swift package generate-xcodeproj` is *definitely* unstable, but I can't understand why we cannot use it :)

Anyway, do you have any suggestion for adding Carthage support to existing SPM-only project?

mdiep:
I think generating the xcodeproj is a fine way to start Carthage support. Generate, share the scheme, and check it in, and you should be good to go.

devxoul:
@mdiep You mean the Xcode project from `Xcode.app` not by `swift package generate-xcodeproj`?

mdiep:
No, I mean generating with `swift package generate-xcodeproj`, sharing the scheme, and then checking it in.

devxoul:
Hmm, I'm confused because you just said that `swift package generate-xcodeproj` is unstable so we should not rely on it. 🤔🤔🤔 Is there any difference between doing that on Carthage and each lib? Maybe the schemes?

mdiep:
It's not guaranteed to work for all projects. But if it's a fine starting place. If it works out of the box for your project, great! This acts as a QA step for projects—they have to confirm that the generated project works.

devxoul:
Thanks for your kind reply. I hope we could find a better solution 👍 Feel free to close or keep open this PR.

devxoul:
@mdiep, was there any updates on WWDC? :)

mdiep:
After talking to someone from the SwiftPM team, I'm tentatively on board with adding support for this. It will require some testing. I'll try to play with this soon, but it will need to be a lower priority than some of my other OSS maintenance.

Source/CarthageKit/Project.swift
```diff
@@ -960,6 +960,22 @@ public final class Project {
 			}
 	}
 
+	public func generateSwiftPackageManagerXcodeProjectIfAvailable(forDependencyAt dependencyPath: String) -> SignalProducer<(), NoError> {
+		let files = (try? FileManager.default.contentsOfDirectory(atPath: dependencyPath)) ?? []
+
+		let hasXcodeProjectFile = files.lazy.filter { $0.hasSuffix(".xcodeproj") || $0.hasSuffix(".xcworkspace") }.first != nil
```
mdiep: Instead of doing `foo.lazy.filter { ... }.first`, you should be able to do `foo.first { ... }`.

Source/CarthageKit/Project.swift
```diff
@@ -960,6 +960,22 @@ public final class Project {
...
+		guard !hasXcodeProjectFile else { return .init(value: ()) }
+
+		let hasPackageFile = files.lazy.filter { $0 == "Package.swift" }.first != nil
+		guard hasPackageFile else { return .init(value: ()) }
+
+		let task = Task("/usr/bin/swift", arguments: [ "package", "generate-xcodeproj" ], workingDirectoryPath: dependencyPath)
```
mdiep: This should definitely have a project event that gets printed to the console.

Tests/CarthageKitTests/ProjectSpec.swift
```diff
@@ -171,7 +171,42 @@ class ProjectSpec: QuickSpec {
...
+			let directoryURL = Bundle(for: type(of: self)).url(forResource: "SwiftPackageManagerTest", withExtension: nil)!
+			let buildDirectoryURL = directoryURL.appendingPathComponent(CarthageBinariesFolderPath)
+
+			func buildFramework() -> [String] {
+				let project = Project(directoryURL: directoryURL)
+				let result = project.buildCheckedOutDependenciesWithOptions(BuildOptions(configuration: "Debug"))
+					.flatten(.concat)
+					.ignoreTaskData()
+					.on(value: { (project, scheme) in
+						NSLog("Building scheme \"\(scheme)\" in \(project)")
```
mdiep: Can you remove this log from the test?

devxoul:
Sounds cool. I found that SPM generated project can cause a problem if the library has dependencies. I'll update this PR soon with more specific tests. Thanks!

Source/CarthageKit/Project.swift
ikesyo: This will not be sufficient––Carthage allows having Xcode projects or workspaces in any subdirectories so you should need to check the dependency directory recursively.

Source/CarthageKit/Project.swift
ikesyo: Generating an Xcode project to the dependency directory makes it dirty so that should be avoided. You should generate a project to a temporary directory.

ikesyo:
Another viewpoint: Deployment target settings of a generated Xcode project are not set for iOS, tvOS, and watchOS ([macOS is set to `10.10`](https://github.com/apple/swift-package-manager/blob/swift-3.1-branch/Sources/Xcodeproj/pbxproj().swift#L88-L93)). That means that the actual deployment target determined for each platforms will be its latest SDK (iOS 10.3, tvOS 10.2, watchOS 3.2 in Xcode 8.3.x). So actually the generated project should be useless without any modification (are your application targeting iOS 10.3 only? 🙂).

I'm still reluctant to add this feature.

mdiep:
> Another viewpoint: Deployment target settings of a generated Xcode project are not set for iOS, tvOS, and watchOS (macOS is set to 10.10). That means that the actual deployment target determined for each platforms will be its latest SDK (iOS 10.3, tvOS 10.2, watchOS 3.2 in Xcode 8.3.x). So actually the generated project should be useless without any modification (are your application targeting iOS 10.3 only? 🙂).

When talking to the SwiftPM folks, this did seem like a potential issue. I'm not super knowledgable about it, but they mentioned the idea of passing an `.xcconfig` file when building to set those flags.

I guess any SwiftPM project wouldn't depend on the SDK, so Carthage could hardcode the oldest viable option?

> I'm still reluctant to add this feature.

I think you're more knowledgable about this than I am, so definitely speak up if you think something is a bad idea.

ikesyo:
> When talking to the SwiftPM folks, this did seem like a potential issue. I'm not super knowledgable about it, but they mentioned the idea of passing an .xcconfig file when building to set those flags.
>
> I guess any SwiftPM project wouldn't depend on the SDK, so Carthage could hardcode the oldest viable option?

Sounds reasonable to me. We can pass `--xcconfig-overrides` option to `swift package generate-xcodeproj` command which will be better than overriding on building.

devxoul:
We can modify the xcconfig and Info.plist. Currently I'm archiving SPM generated xcodeproj as following:

* https://github.com/ReactorKit/ReactorKit/blob/master/.travis.yml#L38-L46
* https://github.com/ReactorKit/ReactorKit/blob/master/scripts/carthage_helper.rb

I think I can do similar with Swift but I'm not sure this would be the best solution 🤔

olbrichj:
Is there any update on this one?

mdiep:
There's some unaddressed feedback above.

devxoul:
I'm having busy days and don't have enough time for this issue at time time. I'll update this PR when I'm available :)

regexident:
Any progress on this, @devxoul?

devxoul:
@regexident, no actually.

kiliankoe:
Would love to see something like this included in Carthage! ❤️ 

Just out of curiosity, how are dependencies handled here? I've run into the issue before that dependencies declared in my Package.swift will break the build for a project that includes the SPM project as a Carthage dependency using the SPM-generated xcodeproj.

SebastianThiebaud:
Hey,

I'd love to see this released and can help to finish what has been started. What is missing before this can get approved?

Thanks

kiliankoe:
The entire swiftpm ecosystem has gotten a lot more stable since this PR was first opened. I'm having no issues generating xcodeprojects and using those as carthage dependencies in other projects. Is there any chance to get this merged or restart the discussion? 😊

fabb:
How are subdependencies handled? What about subdependencies declared in both a Package.swift + in a Cartfile of another dependency?

fabb:
Having SwiftPM support could bridge the gap until SwiftPM can be used on its own, what do you think?
https://forums.swift.org/t/roadmap-of-swiftpm-integration-into-xcode/13466/18

blender:
I'm going to reopen this one in light of https://github.com/Carthage/Carthage/issues/2529 

This feature is up for grabs if someone wants to push it further

yhkaplan:
@devxoul I made a PR to this one which resolves the merge conflicts and some of the comments. 

I just need to resolve the issues below, which might be easier in a separate PR:

1. Generating projects to a temporary directory
2. The implementation of `hasXcodeProjectFile` (making it check the dependency directory recursively)

Source/CarthageKit/Project.swift
yhkaplan: @ikesyo For a temporary directory, would `$TMPDIR` be preferred?

Source/CarthageKit/Project.swift
yhkaplan: Hmm, something like below might work...
```diff
  public func generateSwiftPackageManagerXcodeProjectIfAvailable(forDependencyAt dependencyPath: String) -> SignalProducer<(), NoError> {
-   let files = (try? FileManager.default.contentsOfDirectory(atPath: dependencyPath)) ?? []
-
-   let hasXcodeProjectFile = files.lazy.first { $0.hasSuffix(".xcodeproj") || $0.hasSuffix(".xcworkspace") } != nil
+   // Search for each sub-dir recursively for Xcode Project file
+   while let fileOrFolder = FileManager.default.enumerator(atPath: dependencyPath)?.nextObject() as? String {
+     let hasXcodeProjectFile = fileOrFolder.hasSuffix(".xcodeproj") || fileOrFolder.hasSuffix(".xcworkspace")
      if hasXcodeProjectFile { return .init(value: ()) }
+   }

+   let files = (try? FileManager.default.contentsOfDirectory(atPath: dependencyPath)) ?? []
    let hasPackageFile = files.lazy.first { $0 == "Package.swift" } != nil
    guard hasPackageFile else { return .init(value: ()) }
```

yhkaplan:
My understanding is that generating projects to a temporary directory is the only remaining issue, but are there any more concerns?

Source/CarthageKit/Project.swift
```diff
@@ -1107,6 +1126,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 				options.derivedDataPath = derivedDataVersioned.resolvingSymlinksInPath().path
 
 				return self.symlinkBuildPathIfNeeded(for: dependency, version: version)
+					.then(self.generateSwiftPackageManagerXcodeProjectIfAvailable(forDependencyAt: dependencyPath))
 					.then(build(dependency: dependency, version: version, self.directoryURL, withOptions: options, sdkFilter: sdkFilter))
```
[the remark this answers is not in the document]
mdiep: It would need to by passing in an xcconfig with the settings required for an iOS target.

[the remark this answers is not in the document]
ikesyo: The generated Xcode target supports multiplatform (iOS, tvOS, watchOS as well) by default. An additional requirement is overriding deployment target as I said multiple times.

- https://github.com/Carthage/Carthage/pull/1945#issuecomment-309082053
- https://github.com/Carthage/Carthage/pull/1945#issuecomment-433048983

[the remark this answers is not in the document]
ollitapa: Ah, sorry, should've read more closely 😅

yhkaplan:
I’ll try to make another PR to this one and resolve the remaining issues when I have time (probably in a couple weeks)

mxcl:
I'm interested in finishing this. Though it’s not completely clear to me what is left to do, just getting deployment targets set?

mdiep:
> I'm interested in finishing this

Awesome!

> Though it’s not completely clear to me what is left to do, just getting deployment targets set?

As far as I know, yes.

yhkaplan:
If we wait for Swift 5.0, then minimum deployment target settings can be written directly in the Package.swift file: https://github.com/apple/swift-evolution/blob/master/proposals/0236-package-manager-platform-deployment-settings.md

This should remove the need to read in xcconfig files, making things a whole lot easier!

mxcl:
How feasible is only allowing this feature if you have Xcode 10.2? Like, I think it’s fine.

yhkaplan:
Good point. I think on Carthage’s side, it would involve checking the Package.swift file’s version (if that property is available), otherwise spitting out a warning about targets not being supported in Swift versions lower than 5.0, but still building.

The reason I say warning and not error is that some people who only need the latest minimum target version supported use Carthage and SPM today, so it might be good to not break this behavior.

mxcl:
The version can be checked. The top of `Package.swift` specifies the version it supports, it would specify: `swift-tools-version: 5.0`, eg: https://github.com/mxcl/Path.swift/blob/fixes/Package%40swift-5.0.swift#L1

Notably, packages can have multiple `Package.swift`s like in the above example it is named `Package@swift-5.0.swift`.

I think these are reasonable criterions though. Any package author that wants to support Carthage with only a swift-package would have no reason to object to providing a swift-5-tools SwiftPM manifest.

mdiep:
> If we wait for Swift 5.0, then minimum deployment target settings can be written directly in the Package.swift file

If you specify an iOS deployment target and use `swift package genarate-xcodeproj`, does the generated dynamic framework target have all the right settings for use on iOS?

> Any package author that wants to support Carthage with only a swift-package would have no reason to object to providing a swift-5-tools SwiftPM manifest

Agreed

yhkaplan:
> If you specify an iOS deployment target and use swift package genarate-xcodeproj, does the generated dynamic framework target have all the right settings for use on iOS?

Let me download Xcode 10.2 beta and find out 😉

mxcl:
It can be built for iOS but it won’t have anything special set. It won’t have app-extension only specified for example.

mxcl:
The generated xcodeproj has all platforms available as build targets, I spoke to @aciidb0mb3r about it and he said that the platform specification in `Package.swift` files doesn't actually restrict the platforms it merely allows deployment targets to be set. So this idea is bunk, we still need `xcconfig` files otherwise the generated project will almost certainly fail to build since it will pretend that it can build for eg. tvOS and then fail if it (likely) doesn't support tvOS actually.

mxcl:
Hey everyone, so I've been thinking about this a lot and here is my current plan:

* Only support this if the toolchain is a Swift 5 toolchain
* Only support this if the package provides a Swift 5.0 or higher tool-version manifest
* Generate the xcodeproj
* Only build for platforms explicitly declared in the manifest

I will begin work on this, this weekend, so if there are concerns with my approach, please speak up!

Thanks for the help 🙏

mxcl:
As an update here I have submitted a PR to SwiftPM to get the platform information from the `dump-package` command.

mxcl:
Update: the dump-package changes are finally in Xcode 10.2-beta4 so I will attempt to finish the patch.

rynecheow:
May I ask how are we doing with this PR?

mxcl:
I am unlikely to find time for this in near future, here’s my progress though:

https://github.com/mxcl/Carthage/tree/spm

JonasVautherin:
@mxcl: Would you mind elaborating a bit on what would need to be done on your PR?

mxcl:
I believe the remaining part is parsing dump-package and setting deployment targets, nothing else.

ollieatkinson:
From what I can see these are the following settings which would need to be added to the generated Xcodeproj:

```swift
IPHONEOS_DEPLOYMENT_TARGET = 8.0
SUPPORTED_PLATFORMS = iphonesimulator iphoneos
SDKROOT = iphoneos
DEFINES_MODULE = YES
MODULEMAP_FILE = $(PACKAGE_NAME).xcodeproj/GeneratedModuleMap/$(TARGET_NAME)/module.modulemap
```

Steps:
1. Run `swift package dump-package`, grab the following information:
```
"platforms" : [
    {
      "platformName" : "ios",
      "version" : "8.0"
    }
],
```
2. Create an `xcconfig` file inside of the temporary with the values above, transformed appropriately 
3. When generating the xcodeproj pass `--xcconfig-overrides path/to/tmp/config.xcconfig`

If the project does not specify platform information, I'm guessing we should just fail out - but some steer on that from the core contributors would be good.

If the above sounds reasonable, I can give it a go - but I will hang fire until I get some feedback.

thedavidharris:
> _Edit_: Have I found a further complication?
> 
> Unless I'm missing something, I think we will also need to generate an xcodeproj for each platform we want to build for because only one target is generated by SwiftPM.

I don't think so, `SUPPORTED PLATFORMS` and the OS flags should be enough. https://davedelong.com/blog/2018/11/15/building-a-crossplatform-framework/ describes a way of doing this, then when the platforms are targeted it'll pick up the right version of everything. 

@ollieatkinson @mxcl any continued work on this? I may be able to help pick it up, as we currently are writing a bunch of SPM modules internally that need Carthage integration for some of our current apps and support both is a bit of a pain for the interim, so would be interested in helping out on this

thedavidharris:
So, looks like there's an edge case that won't make this workable for us, and that's when an SPM project has dependencies that are also Carthage dependencies by the consuming project. Probably won't take this on anymore myself, as I don't think there's any solution to that problem :/

Sajjon:
@devxoul Hey! What is the status of this PR? It is open since 2017. It would be great if Carthage could build Swift Packages :)

JonasVautherin:
@thedavidharris: I understand the issue of having common dependencies between SPM and Carthage, and that's an issue because Carthage is not designed to deal with that manually.

However, it is currently a problem that libraries that only support SPM cannot be used in a Carthage project.

Do you think there could be a way to deal with that problem by leveraging xcodegen? I'm thinking of a way to generate an xcodeproj that would embed SPM packages, while allowing to ignore SPM dependencies that also come from Carthage. Would be fine if it requires some manual work to get the setup correct, as it would allow Carthage users to benefit from SPM-only libraries.

Or do you think that it's not worth it and SPM will quickly take over now that it is embedded in Xcode?

scosman:
Anyone trying to accomplish this, I got it running fairly easily with another tool in combination with Carthage. No need to manually maintain xcodeproj files, as this can generate them from your `Package.swift`.

 - Install XcodeGen tool: https://github.com/yonaskolb/XcodeGen
 - Add a project.yml file for the framework. Can be quite simple, but need to point to your Sources directory, include `info` section (just a path to generate Info.plist to), and include `schemes: {}` sections. Sample: file https://github.com/CriticalMoments/CriticalMoments/blob/ebfe2ead8783dd43b94d89b135225a0b2e12b80e/ios/project.yml
 - Generate the project files Carthage needs with `xcodegen generate --spec ./project.yml`
 - Then Carthage works: `carthage build --no-skip-current --use-xcframeworks`

PR #1952, opened 2017-05-24
Change test directories layout for SwiftPM

ikesyo:


Package.swift
```diff
@@ -19,9 +19,8 @@ let package = Package(
     ],
     exclude: [
         "Source/CarthageIntegration",
-        "Source/CarthageKitTests",
         "Source/Scripts",
-        "Source/XCDBLDTests",
         "Source/carthage/swift-is-crashy.c",
+        "Tests/CarthageKitTests/FakeOldObjc.framework",
```
ikesyo: This is needed to suppress the error below:

```
error: the package has an unsupported layout, modulemap (Carthage/Tests/CarthageKitTests/FakeOldObjc.framework/Modules/module.modulemap) is not allowed to be mixed with sources
fix: move the modulemap inside include directory
```

ikesyo:
Thanks for the re-review!

PR #1958, opened 2017-05-27
Fix --log-path option

### LINKED ISSUE #1928: --log-path option appears to be broken
* `carthage version`: 0.22.0
* `xcodebuild -version`: 8.3.2 8E2002

**Cartfile**
```
```

**Carthage Output**
```
$ carthage bootstrap --log-path mylog.log
*** Cloning ReactiveCocoa
*** Fetching ReactiveSwift
*** Fetching Result
*** Checking out Result at "3.2.1"
*** Checking out ReactiveSwift at "1.1.2"
*** Downloading ReactiveCocoa.framework binary at "5.0.3"
Failed to write to /var/folders/q4/xrj10h0n20zg_b2jwt5f84t4wlq244/T: Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"
```

Using the `--log-path` option seems to break Carthage.  I've tried specifying a specific file, a specific directory with relative and absolute paths, but I always get this error.  My guess is that something further down the line is assuming the temporary directory is being created but it doesn't when you use this option.

michaelmcguire:
Resolves #1928 

There were two issues:

* The existing implementation 1) doesn't create the log file specified if it doesn't exist and 2) even if it does exist, it fails because the `FileHandle` used is immediately deallocated upon being created (so you get a Bad Descriptor exception later)
* The error handling incorrectly reported back the problem as involving the temporary directory

This PR uses the original `FileHandle` used to open the file so the resource management issue is fixed and updates the method to return a `CarthageError` so the correct error and URL can be reported.

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -163,34 +163,43 @@ public struct BuildCommand: CommandProtocol {
 		}
 	}
 
-	/// Opens a temporary file on disk, returning a handle and the URL to the
+	/// Opens an existing file, if provided, or creates a temporary file if not, returning a handle and the URL to the
 	/// file.
-	private func openLogFile(_ path: String?) -> SignalProducer<(FileHandle, URL), NSError> {
+	private func openLogFile(_ path: String?) -> SignalProducer<(FileHandle, URL), CarthageError> {
 		return SignalProducer.attempt {
-			let logFD: Int32
-			let logPath: String
-			
 			if let path = path {
-				logFD = FileHandle(forUpdatingAtPath: path)?.fileDescriptor ?? -1
-				logPath = path
+				if !FileManager.default.fileExists(atPath: path) {
+					FileManager.default.createFile(atPath: path, contents: nil, attributes: nil)
+				}
+				
+				let fileURL = URL(fileURLWithPath: path, isDirectory: false)
+				
+				guard let handle = FileHandle(forUpdatingAtPath: path) else {
+					let error = NSError(domain: CarthageKitBundleIdentifier,
+					                    code: 1,
+					                    userInfo: [NSLocalizedDescriptionKey: "Unable to open file handle for file at \(path)"])
+					return .failure(.writeFailed(fileURL, error))
+				}
+				
+				return .success((handle, fileURL))
 			} else {
 				var temporaryDirectoryTemplate: ContiguousArray<CChar>
 				temporaryDirectoryTemplate = (NSTemporaryDirectory() as NSString).appendingPathComponent("carthage-xcodebuild.XXXXXX.log").utf8CString
-				logFD = temporaryDirectoryTemplate.withUnsafeMutableBufferPointer { (template: inout UnsafeMutableBufferPointer<CChar>) -> Int32 in
+				let logFD = temporaryDirectoryTemplate.withUnsafeMutableBufferPointer { (template: inout UnsafeMutableBufferPointer<CChar>) -> Int32 in
 					return mkstemps(template.baseAddress, 4)
 				}
-				logPath = temporaryDirectoryTemplate.withUnsafeBufferPointer { (ptr: UnsafeBufferPointer<CChar>) -> String in
+				let logPath = temporaryDirectoryTemplate.withUnsafeBufferPointer { (ptr: UnsafeBufferPointer<CChar>) -> String in
 					return String(validatingUTF8: ptr.baseAddress!)!
 				}
+				if logFD < 0 {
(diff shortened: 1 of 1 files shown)


Source/carthage/Build.swift
```diff
@@ -163,34 +163,43 @@ public struct BuildCommand: CommandProtocol {
...
 			if let path = path {
-				logFD = FileHandle(forUpdatingAtPath: path)?.fileDescriptor ?? -1
-				logPath = path
+				if !FileManager.default.fileExists(atPath: path) {
+					FileManager.default.createFile(atPath: path, contents: nil, attributes: nil)
+				}
+				
+				let fileURL = URL(fileURLWithPath: path, isDirectory: false)
+				
+				guard let handle = FileHandle(forUpdatingAtPath: path) else {
```
mdiep: Instead of doing the `fileExists` check above, it looks like we should use `FileHandle(forWritingTo:)` here.

jdhealy: @mdiep the `fileExists` check guards not-so-much the handle, but more the `createFile(atPath:…)` call, which [can overwrite: “If a file already exists at path, this method overwrites the contents of that file if the current process has the appropriate privileges to do so.”](https://developer.apple.com/reference/foundation/filemanager/1410695-createfile)

[`FileHandle.init(forUpdatingAtPath:)`](https://developer.apple.com/reference/foundation/filehandle/1411131-init) (from the description at least) is a superset of the behavior of [`FileHandle.init(forWritingAtPath:)`](https://developer.apple.com/reference/foundation/filehandle/1414405-init).

mdiep: Oops. I was thinking that `FileHandle.init(forWritingAtPath:)` didn't require the file to exist. But that's not the case.

jdhealy: The documentation for [`FileHandle.init(forUpdatingAtPath:)`](https://developer.apple.com/reference/foundation/filehandle/1411131-init) tasks us with closing the file descriptor.

My inclination (with no existing `FileHandle.init(forUpdatingAtPath:closeOnDealloc:)` to call) is to do some sort of transfer from `FileHandle.init(forUpdatingAtPath:)` `flatMap`-ing `FileHandle(fileDescriptor: $0.fileDescriptor, closeOnDealloc: true)`.

Do you think (knowing the “immediately deallocated” issue you described), the following would work?
```swift
guard
	!FileManager.default.fileExists(atPath: path) &&
	FileManager.default.createFile(atPath: path, contents: nil, attributes: nil),
	case let fileURL = URL(fileURLWithPath: path, isDirectory: false),
	let handle = FileHandle(forUpdatingAtPath: path)
		.flatMap({ FileHandle(fileDescriptor: $0.fileDescriptor, closeOnDealloc: true) })
		.flatMap({ $0.fileDescriptor >= 0 ? $0 : nil })
else {
	let error = NSError(
		domain: CarthageKitBundleIdentifier,
		code: 1,
		userInfo: [NSLocalizedDescriptionKey: "Unable to open file handle for file at \(path)"]
	)
	return .failure(.writeFailed(fileURL, error))
}

return .success((handle, fileURL))
```

mdiep: As I read it, the `FileHandle` will close the file descriptor—the owner of the `FileHandle` isn't responsible.

> When using this method to create a file handle object, the file handle owns its associated file descriptor and is responsible for closing it.

mdiep: And actually, the original bug shows that that's the case. (Because the `FileHandle` wasn't retained, and the file descriptor was returned, but was no longer valid because the `FileHandle` had deallocated.)

jdhealy:
Ack, sorry @mdiep — I spent too long rewording a change request, still have questions with this.

I can commit a fixup, with @michaelmcguire providing guidance on https://github.com/Carthage/Carthage/pull/1958#pullrequestreview-41008815

@michaelmcguire Nice work on this, sorry to post-merge review.

Source/carthage/Build.swift
mdiep: But I'm not at all impressed by the `FileHandle` documentation. It's not very thorough or clear.

jdhealy: Second reading I see it @mdiep's way, and [the analogous `swift-corelibs-foundation` code seems to corroborate](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/FileHandle.swift#L165-L172).

Very ambiguous documentation language on whether file description closing by handle “is responsible”  through user-appended code or by baked in code-path — now I see, it's baked-in…

jdhealy:
Nice work, @michaelmcguire 🎉

Source/carthage/Build.swift
michaelmcguire: Not much to say here other than @mdiep's understanding was mine.  And yes, I agree that the documentation is not great.

PR #1965, opened 2017-05-30
[CI] Run `swift test`

ikesyo:
Since SwiftPM creates a test bundle on macOS,

- Let SwiftPM just build the test bundle without running it
- Copy test resources into the bundle
- Finally run the test bundle

### DIFF
--- a/Tests/CarthageKitTests/XcodeSpec.swift
+++ b/Tests/CarthageKitTests/XcodeSpec.swift
@@ -35,15 +35,19 @@ class XcodeSpec: QuickSpec {
 
 		describe("determineSwiftInformation:") {
 			let currentSwiftVersion = swiftVersion().single()?.value
+			#if !SWIFT_PACKAGE
 			let testSwiftFramework = "Quick.framework"
 			let currentDirectory = URL(fileURLWithPath: FileManager.default.currentDirectoryPath, isDirectory: true)
 			let testSwiftFrameworkURL = currentDirectory.appendingPathComponent(testSwiftFramework)
+			#endif
 
+			#if !SWIFT_PACKAGE
 			it("should determine that a Swift framework is a Swift framework") {
 				let result = isSwiftFramework(testSwiftFrameworkURL).single()
 
 				expect(result?.value) == true
 			}
+			#endif
 
 			it("should determine that an ObjC framework is not a Swift framework") {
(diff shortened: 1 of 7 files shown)


script/copy-fixtures
```diff
@@ -4,10 +4,10 @@
...
 # "subprojects" that are actually just originating from a fixture.
 
-destination="${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
+destination="$2"
 
 for fixture in "carthage-fixtures-ReactiveCocoaLayout-master" "CartfileOnly" "CartfilePrivateOnly" "NoCartfile" "SchemeDiscoverySampleForCarthage-0.2" "Swell-0.5.0" "SampleMultipleSubprojects" "SampleGitSubmodule" "DependencyTest" "WorkspaceWithDependency"
 do
 	rm -Rf "$destination/$fixture"
-	unzip -q -d "$destination" "${SRCROOT}/Tests/CarthageKitTests/fixtures/$fixture"
+	unzip -q -d "$destination" "$1/$fixture"
```
jdhealy: We should probably add existence checks to all these variables with `${foo:?}`.

Based on a quick reading of `man unzip`, I don't think our execution of it will overwrite, but redundant safety doesn't hurt…

Tests/CarthageKitTests/XcodeSpec.swift
```diff
@@ -56,18 +60,22 @@ class XcodeSpec: QuickSpec {
 				expect(currentSwiftVersion?.isEmpty) == false
 			}
 
+			#if !SWIFT_PACKAGE
```
mdiep: Why are these disabled?

ikesyo: Because SwiftPM does not build frameworks (`.framework`) so we can't test this using the existence of `Quick.framework`.

Tests/CarthageKitTests/XcodeSpec.swift
```diff
@@ -35,15 +35,19 @@ class XcodeSpec: QuickSpec {
 
 		describe("determineSwiftInformation:") {
 			let currentSwiftVersion = swiftVersion().single()?.value
+			#if !SWIFT_PACKAGE
 			let testSwiftFramework = "Quick.framework"
```
ikesyo: https://github.com/Carthage/Carthage/pull/1965/files#r119255259

PR #1976, opened 2017-06-05
Pre-build Scripts

nidegen:
Hi there,

what I would like to see (or missed to find so far) is the possibility to execute a script before Carthage builds something. For instance, i would like to generate an xcodeproj with cmake before Carthage builds the framework, in order not to have the .xcodeproj in the repo..

Or is this already possible?

Before I start work on it, I am happy for any comments where to implement. I expect it in the CarthageKit, right?


mdiep:
I haven't tried it, but I think you could do this like so:

1. Create a project to commit in your project
2. Generate your project locally
3. Add the generated project to the committed project
4. Add the target from the generated project to your scheme (after the target from the committed project)
5. Delete the generated project on disk
6. In the committed project's target, add a run script build phase that generates the Xcode project

nidegen:
yes i have actually thought about this possibility:) it def is an option for me.

however, executing a script would be nicer. so if it would be acceptable, i would like to implement that right away.

or might there be any security concerns about it?

mdiep:
I think this is Carthage should stay away from.

Arbitrary scripts make it much harder for build systems to reason about dependencies. Tools like llbuild and SwiftPM are trying to explicitly capture the structure and semantics of a build so they can be more intelligent.

How were you hoping to generate the Xcode project? With `swift package generate-xcodeproj`?

nidegen:
Ok sure. Agree on the issues for public Carthage repos, but i guess it could be cool for internal dependency management:) I have a Cmake-generated XCode project for a Obj-C-Wrapped c++ library where i wanted to do sth like it.

Issue #1644, opened 2016-12-14
Normalize GitHub repository names specified with git keyword


CodeEagle:
* `carthage version`: 0.18.1
* `xcodebuild -version`:  8b62
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`?no 
* Are you using `--use-submodules`? no

**Cartfile**
```
some other framework with some common frameworks
```
**Cartfile.resolved**
```
...
github "Alamofire/Alamofire" "4.2.0"
git "git@github.com:Alamofire/Alamofire.git" "4.2.0"
github "steipete/Aspects" "1.4.2"
git "git@git1.luoo.net:iOS/AudioProtocol.git" "0.1.5"
github "CodeEagle/CoverFlowPlainLayout" "0.0.6"
git "git@github.com:krzyzanowskim/CryptoSwift.git" "0.6.6"
github "krzyzanowskim/CryptoSwift" "0.6.6"
github "CodeEagle/DGElasticPullToRefresh" "2.0.0"
github "CodeEagle/FDFullscreenPopGesture" "1.2"
github "hkellaway/Gloss" "1.1.1"
git "git@github.com:hkellaway/Gloss.git" "1.1.1"
git "git@github.com:CodeEagle/HMSegmentedControl_CodeEagle.git" "0.2.4"
github "CodeEagle/HMSegmentedControl_CodeEagle" "0.2.4"
github "CodeEagle/KVOBlock" "0.1.3"
git "git@git1.luoo.net:iOS/LoIconFont.git" "0.0.7"
git "git@github.com:jdg/MBProgressHUD.git" "1.0.0"
github "nixzhu/MonkeyKing" "1.1.0"
git "git@git1.luoo.net:iOS/NetworkImageExtension.git" "0.0.7"
github "CodeEagle/OpaqueTextLayer" "0.0.6"
git "git@github.com:SnapKit/SnapKit.git" "3.1.1"
github "SnapKit/SnapKit" "3.1.1"
git "git@github.com:radex/SwiftyTimer.git" "2.0.0"
git "git@git1.luoo.net:iOS/SystemEnhancement.git" "0.2.9"
github "CodeEagle/TimeAgo" "0.1.5"
github "CodeEagle/YYCache" "1.0.6"
git "git@github.com:ibireme/YYCache.git" "1.0.4"
github "ibireme/YYText" "1.0.7"
github "CodeEagle/ZLSwipeableViewSwift" "0.0.9"
[...]

PR #1995, opened 2017-06-17
Git dependencies containing github.com

blender:
Interpret git dependencies containing github.com as github dependencies instead.

This fixes: https://github.com/Carthage/Carthage/issues/1644

```
git "git@github.com:SnapKit/SnapKit" ~> 3.2.0 
git "ssh://git@github.com:SnapKit/SnapKit" ~> 3.2.0 
git "https://github.com/SnapKit/SnapKit" ~> 3.2.0  
```
will all resolve to 
```
github "SnapKit/SnapKit" "v3.2.0"
```

Does not affect current Cartfile.resolved. Upon update resolution will be unified.

### DIFF
--- a/Source/CarthageKit/Cartfile.swift
+++ b/Source/CarthageKit/Cartfile.swift
@@ -81,6 +81,7 @@ public struct Cartfile {
 		}
 
 		return result.flatMap { _ in
+			print(duplicates)
 			if !duplicates.isEmpty {
 				return .failure(.duplicateDependencies(duplicates.map { DuplicateDependency(dependency: $0, locations: []) }))
 			}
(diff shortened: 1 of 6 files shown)


Source/CarthageKit/Cartfile.swift
```diff
@@ -254,7 +255,33 @@ extension Dependency: Scannable {
...
+					let separatorFound = (gitbubHostScanner.scanString("/", into: nil) || gitbubHostScanner.scanString(":", into: nil)) && !gitbubHostScanner.isAtEnd
+					
+					let startOfOwnerAndNameSubstring = gitbubHostScanner.scanLocation
+					
+					if separatorFound && startOfOwnerAndNameSubstring <= urlString.utf16.count {
+						let ownerAndNameSubstring = urlString[urlString.index(urlString.startIndex, offsetBy: startOfOwnerAndNameSubstring)..<urlString.endIndex]
+						return Repository.fromIdentifier(ownerAndNameSubstring as String).map { self.gitHub($0, $1) }
+					}
+					else {
+						return .failure(ScannableError(message: "invalid git URL found", currentLine: scanner.currentLine))
```
mdiep: This seems like maybe it should just fallback to a normal `.git(gitURL)`.

Source/CarthageKit/Cartfile.swift
```diff
@@ -254,7 +255,33 @@ extension Dependency: Scannable {
 			}
 		} else if scanner.scanString("git", into: nil) {
 			parser = { urlString in
-				return .success(self.git(GitURL(urlString)))
+				
+				// This is a dependency expressed as git "..."
+				let gitURL = GitURL(urlString)
+				
+				if gitURL.containsGithubDomain {
```
mdiep: These logic seems like it should be broken out into a separate function. It's quite a bit of code to put in this method and it's also a bit more general.

Maybe it could be in an initializer on `Depndency`?

```swift
extension Dependency {
  init(gitURL: GitURL) {
  }
}
```

Source/CarthageKit/Cartfile.swift
```diff
@@ -81,6 +81,7 @@ public struct Cartfile {
 		}
 
 		return result.flatMap { _ in
+			print(duplicates)
```
blender: Sorry, this one slipped through. Removed

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -170,6 +170,42 @@ class DependencySpec: QuickSpec {
...
+				it("should read a git dependency as github") {
+					let scanner = Scanner(string: "git \"ssh://git@github.com:owner/name\"")
+					
+					let dependency = Dependency.from(scanner).value
+					
+					let expectedRepo = Repository(
+						owner: "owner",
+						name: "name")
+					
+					expect(dependency).to(equal(Dependency.gitHub(.dotCom, expectedRepo)))
```
ikesyo: Actually `expect(...) ==` is more popular in our test suites. I will address it.

ikesyo: #1999

Source/CarthageKit/Cartfile.swift
blender: I have done so but make it fileprivate. Otherwise there would be globally accessible

```
Dependency(gitURL: GitURL) 
Dependency.git(GitURL)
```

Source/CarthageKit/Cartfile.swift
blender: Done.

That would be the case when there is a URL like so `http://some.host/github.com/repo.git`

mdiep:
Thanks! This is a great fix.

PR #1996, opened 2017-06-17
Try to support standard `http_proxy` environment variable

### LINKED ISSUE #1682: [Enhancement] Support http_proxy
* `carthage version`:  0.18.1
* `xcodebuild -version`:  8C1002

Carthage can't use proxy server to download prebuilt binaries.
It's a much needed feature for developers in China.

ikesyo:
Would resolve #1682.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -145,14 +145,46 @@ extension Client {
 			Client.userAgent = gitHubUserAgent()
 		}
 
+		let urlSession = URLSession.makeProxyConfiguredSession()
+
 		if !isAuthenticated {
-			self.init(server)
+			self.init(server, urlSession: urlSession)
 		} else if let token = tokenFromEnvironment(forServer: server) {
-			self.init(server, token: token)
+			self.init(server, token: token, urlSession: urlSession)
 		} else if let (username, password) = credentialsFromGit(forServer: server) {
-			self.init(server, username: username, password: password)
+			self.init(server, username: username, password: password, urlSession: urlSession)
 		} else {
-			self.init(server)
+			self.init(server, urlSession: urlSession)
 		}
+	}
+}
+
+extension URLSession {
+	public static func makeProxyConfiguredSession() -> URLSession {
+		let configuration = URLSessionConfiguration.default
+		let env = ProcessInfo.processInfo.environment
+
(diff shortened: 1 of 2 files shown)


mdiep:
Any suggestion for how to test this? I might be able to check this at work this week, but I'm not sure how to check from my personal machine.

tspike:
any chance we can get this merged? currently a blocker for us

bartoszj:
Hi. Any progress with this task? A proxy setting could be tested with [Charles Proxy](https://www.charlesproxy.com/). Also, it would be nice to add the proxy setting to binary framework downloads.

pnull:
👍

Source/CarthageKit/GitHub.swift
```diff
@@ -145,14 +145,46 @@ extension Client {
...
+
+			if let port = components.port {
+				dictionary[kCFNetworkProxiesHTTPPort as AnyHashable] = port
+				dictionary[kCFNetworkProxiesHTTPSPort as AnyHashable] = port
+				components.port = nil
+			}
+			dictionary[kCFNetworkProxiesHTTPEnable as AnyHashable] = true
+			dictionary[kCFNetworkProxiesHTTPSEnable as AnyHashable] = true
+			dictionary[kCFNetworkProxiesHTTPProxy as AnyHashable] = components.url
+			dictionary[kCFNetworkProxiesHTTPSProxy as AnyHashable] = components.url
```
mdiep: I don't think we should be configuring HTTP and HTTPS with the same proxy. We should set them from the respective environment variables.

mdiep:
Finally wired up Charles to test this out. Sorry it took me so long!

I think we should treat HTTP/HTTPS separately, but other than that this looks good!

xiongmao86:
@ikesyo Could you support socks5 proxy as well?

xiongmao86:
Also has a question, The team replied that they don’t support http way of downloading git repo, why setting http proxy?

Source/CarthageKit/GitHub.swift
jakeadams: I think a `Proxy` type would be good to have so it can also be testable.

blender:
This has been improved and superseded by #2371 . Closing. 

Thanks everyone for the effort and the valuable contributions.

PR #2001, opened 2017-06-19
Improve Code Quality by using SwiftLint

Dschee:
I have setup **SwiftLint** to work with this project and run automatically on build. Note that you need to have the latest SwiftLint installed for this (`brew install swiftlint` if you don't have it yet, `brew upgrade swiftlint` for updating it).

I've added a `.swiftlint.yml` where I **configured the rules** and even added **many custom rules** to ensure the code is streamlined and consistently the same. I've also done the hard work and have **already fixed most of the issues** (there were 800+) – leaving only around 17 which can be solved over time with other PRs.

Note that all of the rules are Swift Community rules or extracted from popular projects such as Alamofire and Moya where they are implicitly used.

PR #2006, opened 2017-06-21
Use NSRange initializer

ikesyo:


Dschee:
Wow, looks like I'm gonna have to rebase many times, seeing you making the same changes I've already made in #2001 in smaller chunks ...

ikesyo:
Because looks like you are not going to do that 🙂

mdiep:
These will make the diff in #2001 much smaller, which is nice for reviewing!

Dschee:
Well, thtat‘s true, go for it. No problem with this on my side, was just a little surprised at first sight. ^^

PR #2025, opened 2017-06-22
copy-frameworks Output Files

mdiep:
One of the things I learned at WWDC this year is that Xcode is smart about _Run Script_ build phases if you specify both input and output files. If the input files haven't changed and the output files exist, then Xcode doesn't need to run the script again—saving you time on dirty builds.

Projects using `carthage copy-frameworks` can use this to prevent unnecessary copying of frameworks during dirty builds. This should be a nice speed boost for some projects.

A couple caveats apply:

1. Although we copy `.framework`s, `.dsym`s, and `.bcsymbolmap`s, I've only specified the `.framework`s here. I think that should be enough—it's unlikely that the other files would be removed while the `.framework` still existed.

2. This path isn't the correct path for Archive builds—they're copied to a different directory. But I think that's acceptable—archive builds are rare and no worse off than before (because I think it should copy every time).

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,24 +64,24 @@ Additionally, you'll need to copy debug symbols for debugging and crash reportin
 1. On your application targets’ “Build Phases” settings tab, click the “+” icon and choose “New Run Script Phase”. Create a Run Script in which you specify your shell (ex: `/bin/sh`), add the following contents to the script area below the shell:
 
     ```sh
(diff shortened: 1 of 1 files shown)


mdiep:
> But the rendering looks broken.

🙈

This must work differently with CommonMark. Fixed!

NachoSoto:
Oh awesome! This speeds up my build quite a lot :D

PR #2027, opened 2017-06-23
Fix checkout removal

### LINKED ISSUE #1969: Checkouts are moved to trash when using --cache-builds and valid cache is found
* `carthage version`: 0.23.0
* `xcodebuild -version`: Xcode 8.3.2
Build version 8E2002

**Cartfile**
```
github "card-io/card.io-iOS-source" == 5.4.1
github "CocoaLumberjack/CocoaLumberjack" == 3.1.0
github "airbnb/lottie-ios" == 1.5.1
...
```

**Carthage Output**
```
[10:53:03]: ▸ *** Checking out lottie-ios at "1.5.1"
[10:53:03]: ▸ *** Checking out CocoaLumberjack at "3.1.0"
[10:53:04]: ▸ *** Checking out card.io-iOS-source at "iOS_5.4.1"
"aec802ba2c2f524f8a47708aa6f714a48be73f49"
[10:53:12]: ▸ *** xcodebuild output can be found in /var/folders/bn/y5hxkvm9625gwfdtdl4f527h0000gp/T/carthage-xcodebuild.8OuaPI.log
[10:53:13]: ▸ *** Valid cache found for card.io-iOS-source, skipping build
[10:53:13]: ▸ *** Valid cache found for CocoaLumberjack, skipping build
[10:53:13]: ▸ *** Valid cache found for lottie-ios, skipping build
...
```

Bootstrapping carthage with --cache-builds on a CI machine fills up disk space pretty quickly, since all checkouts, for which a valid cache is found, are moved to trash.

BobElDevil:
Fixes #1969

Does a couple things
  *  Move the event listener for removing the checkout folder to within the binary unarchive flatMap. This ensures it only occurs after something is actually unarchived, and not for every checkout, regardless of binary status
  *  Listen for 'terminated' instead of 'completed', since the signal at this point in the operator chain will be interrupted by the later 'take' call. 
  *  Use `removeItem` instead of `trashItem`. This can easily be reverted, but I saw no reason for using the trash like this. It's generally expected that command line tools will be a little less forgiving on removal of things anyway, and this appeared to be the only place we used trash

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -586,9 +586,9 @@ public final class Project {
 						}
 						.flatMap(.concat) {
 							return self.unarchiveAndCopyBinaryFrameworks(zipFile: $0, projectName: dependency.name, commitish: revision, toolchain: toolchain)
-								.on(terminated: {
-									// We can receive the 'interrupted' event due to the 'take(first: 1)' below, so listen to 'terminated' to ensure we catch it
-									_ = try? FileManager.default.removeItem(at: checkoutDirectoryURL)
+								.on(value: { _ in
+									// Trash the checkouts folder when we've successfully copied binaries, to avoid building unnecessarily
+									_ = try? FileManager.default.trashItem(at: checkoutDirectoryURL, resultingItemURL: nil)
 								})
 						}
 						.flatMap(.concat) { self.removeItem(at: $0) }
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -588,7 +588,7 @@ public final class Project {
 							return self.unarchiveAndCopyBinaryFrameworks(zipFile: $0, projectName: dependency.name, commitish: revision, toolchain: toolchain)
 								.on(terminated: {
 									// We can receive the 'interrupted' event due to the 'take(first: 1)' below, so listen to 'terminated' to ensure we catch it
-									_ = try? FileManager.default.trashItem(at: checkoutDirectoryURL, resultingItemURL: nil)
+									_ = try? FileManager.default.removeItem(at: checkoutDirectoryURL)
 								})
 						}
 						.flatMap(.concat) { self.removeItem(at: $0) }
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -584,10 +584,13 @@ public final class Project {
 							}
 							return self.downloadMatchingBinaries(for: dependency, atRevision: revision, fromRepository: repository, client: Client(server: server, isAuthenticated: false))
 						}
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -584,10 +584,13 @@ public final class Project {
...
 						}
-						.flatMap(.concat) { self.unarchiveAndCopyBinaryFrameworks(zipFile: $0, projectName: dependency.name, commitish: revision, toolchain: toolchain) }
-						.on(completed: {
-							_ = try? FileManager.default.trashItem(at: checkoutDirectoryURL, resultingItemURL: nil)
-						})
+						.flatMap(.concat) {
+							return self.unarchiveAndCopyBinaryFrameworks(zipFile: $0, projectName: dependency.name, commitish: revision, toolchain: toolchain)
+								.on(terminated: {
+									// We can receive the 'interrupted' event due to the 'take(first: 1)' below, so listen to 'terminated' to ensure we catch it
+									_ = try? FileManager.default.removeItem(at: checkoutDirectoryURL)
```
mdiep: I think this used `trashItem` because we're trashing the checkouts folder. If you make local modifications and then run `carthage update`, your changes will be gone. Trashing them lets people recover them. I think it's good to be a little forgiving in this scenario.

Source/CarthageKit/Project.swift
```diff
@@ -584,10 +584,13 @@ public final class Project {
...
 							return self.downloadMatchingBinaries(for: dependency, atRevision: revision, fromRepository: repository, client: Client(server: server, isAuthenticated: false))
 						}
-						.flatMap(.concat) { self.unarchiveAndCopyBinaryFrameworks(zipFile: $0, projectName: dependency.name, commitish: revision, toolchain: toolchain) }
-						.on(completed: {
-							_ = try? FileManager.default.trashItem(at: checkoutDirectoryURL, resultingItemURL: nil)
-						})
+						.flatMap(.concat) {
+							return self.unarchiveAndCopyBinaryFrameworks(zipFile: $0, projectName: dependency.name, commitish: revision, toolchain: toolchain)
+								.on(terminated: {
+									// We can receive the 'interrupted' event due to the 'take(first: 1)' below, so listen to 'terminated' to ensure we catch it
```
mdiep: Won't this remove the checkout if we skipped installing binaries?

I think maybe we should do this `on(value:)`, since that's the case where we've actually downloaded a binary successfully?

Source/CarthageKit/Project.swift
BobElDevil: ok, I'll put the trash item call back

Source/CarthageKit/Project.swift
BobElDevil: That sounds reasonable. I was worried that it was sending an event for each framework it found, and didn't want to delete checkouts until it finished, but looking again, it just runs a `.then` with the unzipped directory after successful copying, so listening for a single value will work.

mdiep:
Thanks for investigating and fixing this!

PR #2035, opened 2017-06-28
[WIP] Adding Git Output for Verbosity #630

### LINKED ISSUE #630: Should the --verbose flag affect Git output

ragboyjr:
This is a WIP because to fix #630 it will require a decent sized design change that I'd like to run by before actually starting any real work.

The main issue is the `launchGitTask` func in Git.swift of CarthageKit. It specifically calls the `ignoreTaskData` and only listens for the final success event fired from the ReactiveTask. In this MR, i was able to grab a hold of that data before we ignored and provided some rudimentary logging :p to just show that those events are the ones we'd want to end up logging.

I propose that we update launchGitTask to be a protocol/class (or a higher order func) so we can decorate the launching capabilities.

Then we'd have one launcher that actually creates the reactive task like the current implementation but we don't ignore the task data. One launcher that sits on top and will fire ProjectEvents from the project event observer, and another launcher on top of that that will ignoreTaskData and map the output to a string.

We'd then need to refactor the codebase to pass in the task launcher instead of just referring to it globally.

And then, once the project events are firing, we can easily just listen and output them if the verbose flag is set in the carthage project.

There other ways we could solve this, so @mdiep let me know what you think :).

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -167,6 +167,27 @@ public func launchGitTask(
 	let taskDescription = Task("/usr/bin/env", arguments: [ "git" ] + arguments, workingDirectoryPath: repositoryFileURL?.path, environment: updatedEnvironment)
 
 	return taskDescription.launch(standardInput: standardInput)
+		.filter { taskEvent in
+			switch taskEvent {
+			case .success(_):
+				return true
+			case .launch(_):
+				return true
+			case let .standardOutput(data):
+				return data.count > 0
+			case let .standardError(data):
+				return data.count > 0
+			}
+		}
+		.on(event: { event in
+			guard let taskEvent = event.value else {
+				return
+			}
+			if case .success(_) = taskEvent {
+				return
+			}
+			print(taskEvent)
+		})
 		.ignoreTaskData()
 		.mapError(CarthageError.taskError)
 		.map { data in
@@ -199,7 +220,7 @@ public func cloneRepository(_ cloneURL: GitURL, _ destinationURL: URL, isBare: B
 		arguments.append("--bare")
 	}
 
-	return launchGitTask(arguments + [ "--quiet", cloneURL.urlString, destinationURL.path ])
+	return launchGitTask(arguments + [ "-v", cloneURL.urlString, destinationURL.path ])
 		.on(completed: {
 			FetchCache.updateLastFetchTime(forURL: cloneURL)
 		})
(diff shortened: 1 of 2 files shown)


ragboyjr:
@abbeycode I saw your previous attempt at resolving this issue; let me know if you have any concerns.

abbeycode:
@ragboyjr I had gotten bogged down working with Reactive Cocoa, basically.

ragboyjr:
@abbeycode dude, totally understand. I maybe adjusted 5 lines of code and it took me like 6 hours :p. ReactiveSwift and ReactiveTask are very hard to wrap your head around for new comers (like me).

The abstractions that finally helped me "ish" grasp this is that ReactiveSwift signal = event emitter and a SignalProducer = a promise with some extra features.

Source/CarthageKit/Project.swift
```diff
@@ -47,6 +47,12 @@ public enum ProjectEvent {
 
 	/// Building an uncached project.
 	case buildingUncached(Dependency)
+
+	/// Launching a git task.
+	case launchingGitTask(Task)
+
+	/// Running a git task.
+	case runningGitTask(String)
```
mdiep: I'm not sure that we need these events. I don't think we want to report these to users.

Source/CarthageKit/Git.swift
```diff
@@ -167,6 +167,27 @@ public func launchGitTask(
...
+		.on(event: { event in
+			guard let taskEvent = event.value else {
+				return
+			}
+			if case .success(_) = taskEvent {
+				return
+			}
+			print(taskEvent)
+		})
 		.ignoreTaskData()
```
mdiep: This is the issue here. Instead of ignoring the task data, it needs to be sent through. (The `standardOutput` and `standardError` events should bubble up to be printed. You can search for where this is done for `xcodebuild` builds.)

ragboyjr: @mdiep OK, great, I'll do some more digging and report back!

ragboyjr: @mdiep OK, so few issues.

1. launchGitTask returns a `SignalProducer<String, CarthageError>`. So all of the places that expect a string of the success output would have to change.
2. And even if we did that, we still wouldn't be able to let *all* of the git output be logged. Because there are some instances like in `listTags` where they need to operate on the successful string data and map into something else.

What would you think about adding another optional parameter to the launchGitTask func that is an observer/event sink. Then if verbose, we push to stdout, else we ignore.

mdiep:
> launchGitTask returns a SignalProducer<String, CarthageError>. So all of the places that expect a string of the success output would have to change

Yeah, that's expected. A lot of work, but it could be done piecemeal.

> And even if we did that, we still wouldn't be able to let all of the git output be logged. Because there are some instances like in listTags where they need to operate on the successful string data and map into something else.

The way that ReactiveTask is designed is that you map over the data, but the output is always passed through unchanged. So that's not an issue.

ragboyjr:
@mdiep kk, thanks for the help man. As I mentioned earlier, I am new to ReactiveSwift, but I'll take another crack at it and let you know.

Issue #1466, opened 2016-09-11
Always checkout project—even if pre-built binary is available


mdiep:
Otherwise dependencies can't rely on the source being there—which they may want to do. (c.f. #1421)

jessesquires:
This would be great. Currently having this problem with [OCMock](https://github.com/erikdoe/ocmock).

ikesyo:
This may conflict with #1485.

kenji21:
Having some issue with cmd-clicking some method defined in a swift framework, we add xcodeproj to our project to being able to "jump to definition". But as some dependencies are not checked out... Exemple: realm-cocoa, but only when defined via 

    github "realm/realm-cocoa" ~> 1.1.0

But ok if added with full 'https' URL:

    git "https://github.com/realm/realm-cocoa" ~> 1.1.0


using 0.18.2 on Sierra

mdiep:
Carthage will currently build a project if it exists in `Carthage/Checkouts/`—even if a binary has been downloaded. So this will need to account for that.

PR #2041, opened 2017-06-30
Use checked out cartfile

### LINKED ISSUE #2034: Unexpected fetch on build command
* `carthage version`: 0.23.0
* `xcodebuild -version`: 8.3.3

We are committing the checkout folder with our app to use it with CI. The CI builds with 

`build --cache-builds --platform iOS`

Unexpected: Carthage performs a fetch for some dependencies before the build. That breaks the CI for dependencies that need credentials - since we don't supply them on CI at the moment.

We would not expect `build` to perform a fetch. 

Is the fetch expected - and why? How could we ensure that build does not perform a fetch?

BobElDevil:
Fixes #2034 

This attempts to read the Cartfile from the current source checkout first (except during 'update' resolution) before falling back to the current git cache read/fetch.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -337,16 +337,16 @@ public final class Project { // swiftlint:disable:this type_body_length
 				cartfileSource = SignalProducer<Bool, NoError> {
 					var isDirectory: ObjCBool = false
 					return FileManager.default.fileExists(atPath: dependencyURL.path, isDirectory: &isDirectory) && isDirectory.boolValue
+				}
+				.flatMap(.concat) { directoryExists -> SignalProducer<Cartfile, CarthageError> in
+					if directoryExists {
+						return SignalProducer(result: Cartfile.from(file: dependencyURL.appendingPathComponent(Constants.Project.cartfilePath)))
+							.flatMapError { _ in .empty }
+					} else {
+						return cartfileFetch
 					}
-					.flatMap(.concat) { directoryExists -> SignalProducer<Cartfile, CarthageError> in
-						if directoryExists {
-							return SignalProducer(result: Cartfile.from(file: dependencyURL.appendingPathComponent(Constants.Project.cartfilePath)))
-								.flatMapError { _ in .empty }
-						} else {
-							return cartfileFetch
-						}
-					}
-					.flatMapError { _ in return .empty }
+				}
+				.flatMapError { _ in return .empty }
 			} else {
 				cartfileSource = cartfileFetch
 			}
@@ -772,12 +772,8 @@ public final class Project { // swiftlint:disable:this type_body_length
 
 		return loadResolvedCartfile()
 			.map { resolvedCartfile -> [(Dependency, PinnedVersion)] in
-				let dependencies = resolvedCartfile.dependencies
-				return dependencies.keys
-					.filter { dependenciesToCheckout?.contains($0.name) ?? true }
-					.map { key in
-						(key, dependencies[key]!)
-					}
+				return resolvedCartfile.dependencies
+					.filter { dep, _ in dependenciesToCheckout?.contains(dep.name) ?? true }
 			}
 			.zip(with: submodulesSignal)
 			.flatMap(.merge) { dependencies, submodulesByPath -> SignalProducer<(), CarthageError> in
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -216,16 +216,6 @@ public final class Project { // swiftlint:disable:this type_body_length
 		}
 	}
 
-	/// Produces the sub dependencies of the given dependency
-	fileprivate func dependencies(for dependency: Dependency, version: PinnedVersion) -> SignalProducer<Set<Dependency>, CarthageError> {
-		return self.dependencies(for: dependency, version: version)
-			.map { $0.0 }
-			.collect()
-			.map { Set($0) }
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -745,10 +771,13 @@ public final class Project { // swiftlint:disable:this type_body_length
 			}
 
 		return loadResolvedCartfile()
-			.flatMap(.merge) { resolvedCartfile in
-				return self
-					.buildOrderForResolvedCartfile(resolvedCartfile, dependenciesToInclude: dependenciesToCheckout)
-					.collect()
+			.map { resolvedCartfile -> [(Dependency, PinnedVersion)] in
+				let dependencies = resolvedCartfile.dependencies
+				return dependencies.keys
```
BobElDevil: Since we were already doing checkouts concurrently, I realized we were wasting cycles figuring out the build order, and we really just want to checkout anything we find in the list.

Source/CarthageKit/Project.swift
```diff
@@ -314,17 +304,53 @@ public final class Project { // swiftlint:disable:this type_body_length
...
-			return self.cloneOrFetchDependency(dependency, commitish: revision)
+			let cartfileFetch: SignalProducer<Cartfile, CarthageError> = self.cloneOrFetchDependency(dependency, commitish: revision)
 				.flatMap(.concat) { repositoryURL in
 					return contentsOfFileInRepository(repositoryURL, Constants.Project.cartfilePath, revision: revision)
 				}
 				.flatMapError { _ in .empty }
 				.attemptMap(Cartfile.from(string:))
+
+			let cartfileSource: SignalProducer<Cartfile, CarthageError>
+			if tryCheckoutDirectory {
```
ikesyo: I'm not confident that if this is really correct / safe––looks like the revision is not taken into account. Please correct me if I am wrong.

Source/CarthageKit/Project.swift
```diff
@@ -314,17 +304,53 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+				let dependencyURL = self.directoryURL.appendingPathComponent(dependency.relativePath)
+				cartfileSource = SignalProducer<Bool, NoError> {
+					var isDirectory: ObjCBool = false
+					return FileManager.default.fileExists(atPath: dependencyURL.path, isDirectory: &isDirectory) && isDirectory.boolValue
```
ikesyo: Wrong indentations.

Source/CarthageKit/Project.swift
BobElDevil: It's not taking the revision explicitly into account. However this is the same amount of verification that any step of the build phase takes. Right now we check the Cartfile.resolved for build order but build whatever happens to be sitting in  the checkout folder. So if we're assuming that we're building the right revision at that point, then it is no worse to also read the Cartfile from there.

Source/CarthageKit/Project.swift
```diff
@@ -745,10 +771,13 @@ public final class Project { // swiftlint:disable:this type_body_length
...
-			.flatMap(.merge) { resolvedCartfile in
-				return self
-					.buildOrderForResolvedCartfile(resolvedCartfile, dependenciesToInclude: dependenciesToCheckout)
-					.collect()
+			.map { resolvedCartfile -> [(Dependency, PinnedVersion)] in
+				let dependencies = resolvedCartfile.dependencies
+				return dependencies.keys
+					.filter { dependenciesToCheckout?.contains($0.name) ?? true }
+					.map { key in
+						(key, dependencies[key]!)
```
mdiep: I think you could get rid of this `map` if you get rid of the `keys` above?

```swift
return resolvedCartfile
    .dependencies
    .filter { dependenciesToCheckout?.contains($0.name) ?? true }
```

BobElDevil: Well dependencies is a dictionary, and we want to process each item, but that can be worked down a level when it's zipped with the submodule dictionary, rather than doing it here to force it into an array.

Source/CarthageKit/Project.swift
BobElDevil: Will fix, got thrown off by where the braces go for an init with closure + additional operations, but found another example to base it off of

Source/CarthageKit/Project.swift
```diff
@@ -314,17 +304,53 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+				cartfileSource = SignalProducer<Bool, NoError> {
+					var isDirectory: ObjCBool = false
+					return FileManager.default.fileExists(atPath: dependencyURL.path, isDirectory: &isDirectory) && isDirectory.boolValue
+				}
+				.flatMap(.concat) { directoryExists -> SignalProducer<Cartfile, CarthageError> in
+					if directoryExists {
+						return SignalProducer(result: Cartfile.from(file: dependencyURL.appendingPathComponent(Constants.Project.cartfilePath)))
+							.flatMapError { _ in .empty }
+					} else {
+						return cartfileFetch
```
mdiep: I'm not sure we should fetch the `Cartfile` here. If the checked out version is going to be the source of truth, then we should probably make it the source of truth.

What do you think?

BobElDevil: This covers the case where the checkout isn't there, like when we have a binary release. 

If we resolved #1466 (and solved the blocking issue that we build everything in the folder without verifying there were binary releases), then we wouldn't need this backup behavior anymore.

PR #2045, opened 2017-07-02
[gardening] Simplify `volumeSupportsFileCloning()` implementation

ikesyo:


jdhealy:
Given this is (hopefully) a temporary fix for a bug, I'd say the `NSLocalizedFailureReasonError` (despite not actually being localized) are (temporarily) worth their weight in additional code lines, if there might be more to diagnose out of user reports.

`fileprivate` and `key` instead of `keys` are good changes though.

I’ll let @BobElDevil make the final call though…

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -275,11 +275,11 @@ extension URL {
 			throw error(failureReason: "Expected single resource value: «actual count: \(values.count)».")
```
ikesyo: Though I'm not sure how this would happen event if we only pass single key to `resourceValues(forKeys:)` (have you ever experienced such unexpected behavior?), let's keep it for now.

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -275,11 +275,11 @@ extension URL {
...
 		}
 
-		guard case (keys.first!, let volumeSupportsFileCloning as CFBoolean)? = values.first else {
-			throw error(failureReason: "Unable to extract a CFBoolean from «\(String(describing: values.first))».")
+		guard let volumeSupportsFileCloning = values[key] as? NSNumber else {
```
ikesyo: The document says the return value is `NSNumber`: https://developer.apple.com/documentation/foundation/urlresourcekey/1642798-volumesupportsfilecloningkey

PR #1984, opened 2017-06-09
Check swift compiler

BobElDevil:
Fixes #1983 

Instead of using the language version from the swift version string, use the compiler version. That way binary compatible but different SWIFT_VERSION builds will still properly cache / be considered compatible.

Issue #1983, opened 2017-06-07
--cache-builds does not work for Xcode 9 for carthage update or build


kashifshaikh:
* `carthage version`: 0.23.0
* `xcodebuild -version`: Xcode 9.0 Build version 9M136h
* Are you using `--no-build`?  no
* Are you using `--no-use-binaries`?  no
* Are you using `--use-submodules`? no 
* Are you using `--cache-builds`? yes

**Cartfile**
```
github "realm/realm-cocoa" "tg/xcode-9"
```

**Carthage Output**
```
KashifsWorkMBP:iOS.EsiMobile kshaikh$ carthage update --cache-builds --platform iOS
*** Fetching realm-cocoa
*** Checking out realm-cocoa at "0cc6f7a0e008476c21ea6d8fb06d90f57f4ef0e6"
*** xcodebuild output can be found in /var/folders/q7/pr0h22nd4cdgb2f6zrx7sg_w0000gn/T/carthage-xcodebuild.HOmVdT.log
*** No cache found for realm-cocoa, building with all downstream dependencies
*** Building scheme "Realm" in Realm.xcworkspace
*** Building scheme "RealmSwift" in Realm.xcworkspace

KashifsWorkMBP:iOS.EsiMobile kshaikh$ carthage update --cache-builds --platform iOS
*** Fetching realm-cocoa
*** Checking out realm-cocoa at "0cc6f7a0e008476c21ea6d8fb06d90f57f4ef0e6"
*** xcodebuild output can be found in /var/folders/q7/pr0h22nd4cdgb2f6zrx7sg_w0000gn/T/carthage-xcodebuild.lnX9Za.log
*** Invalid cache found for realm-cocoa, rebuilding with all downstream dependencies
*** Building scheme "Realm" in Realm.xcworkspace
```
It will always rebuild framework - doesn't matter which package I choose.

With Xcode 8, cached builds works.

BobElDevil:
It looks like it's due to the swift version detection in a framework. I saw this happen in my frameworks, and looked at the "*-swift.h" header and this was the first line:
`// Generated by Apple Swift version 4.0 effective-3.2 (swiftlang-900.0.43 clang-900.0.22.8)`

while `swift --version` prints out

`Apple Swift version 4.0 (swiftlang-900.0.43 clang-900.0.22.8)`

[...]

PR #2052, opened 2017-07-06
Include Swift compiler version in addition to swiftlang version and clang version

ikesyo:
That would be more comprehensible.

See also #1983 and #1984. ~~Originally I assumed this in [this comment](https://github.com/Carthage/Carthage/issues/1983#issuecomment-307257847).~~


mdiep:
I'm not sure which is the best approach here. This is certainly more comprehensible. The current implementation is probably more correct, but perhaps not meaningfully so. Outside of some very small potential edge cases, I think these should behave the same.

What do you think @BobElDevil?

ikesyo:
I'm concerned about Xcode patch version updates. For example (the below should not be actual values):

- Xcode 9.0
  - `4.0` and `swiftlang-900.0.43 clang-900.0.22.8`
- Xcode 9.0.1
  - `4.0` and `swiftlang-900.0.44 clang-900.0.22.9`

If we use the former a build cache will still be valid, but if we use the latter will not be valid anymore.

However, it may be safer to use the latter value if we take into account a situation that there is no binary compatibility between patch version updates.

BobElDevil:
Yeah, incompatibility between patch versions is what I'm worried about too. So I think we should stick with the swiftlang/clang version. However I think for readability, we could take this opportunity to pull in both parts of the version string, so we can make more readable messaging. So while we still compare compiler versions, we still change it from:
`"Incompatible Swift version - framework was built with swiftlang-100.0.00 clang-100.0.00.1 and the local version is swiftlang-200.0.00 clang-200.0.00.2."`
to
`"Incompatible Swift version - framework was built with 3.0 (swiftlang-100.0.00 clang-100.0.00.1) and the local version is 3.1 (swiftlang-200.0.00 clang-200.0.00.2.")`

mdiep:
> If we use the former a build cache will still be valid, but if we use the latter will not be valid anymore.

But _is_ the latter valid? I'm not sure that we can count on compatibility between patch versions.

ikesyo:
> compatibility between patch versions

That will not be guaranteed. Also, I'm not sure if swiftlang/clang versions are bumped between patch versions.

mdiep:
I think we should stick with the current implementation because it's safer. But I think it'd be fine to pull in the Swift version like @BobElDevil suggested.

PR #2054, opened 2017-07-08
cp command failed on copying hard links if prefix_install already exists

waffleboot:
make prefix_install PREFIX=<somedir>

cp -Rf "/tmp/Carthage.dst/Library/Frameworks/CarthageKit.framework" "MyCarthage/Frameworks/"
cp: cannot overwrite directory MyCarthage/Frameworks/CarthageKit.framework/Resources with non-directory /tmp/Carthage.dst/Library/Frameworks/CarthageKit.framework/Resources

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ installables: clean bootstrap
 
 prefix_install: installables
 	mkdir -p "$(PREFIX)/Frameworks" "$(PREFIX)/bin"
-	cp -Rf "$(TEMPORARY_FOLDER)$(FRAMEWORKS_FOLDER)/$(OUTPUT_FRAMEWORK)" "$(PREFIX)/Frameworks/"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,7 +54,7 @@ installables: clean bootstrap
 
 prefix_install: installables
 	mkdir -p "$(PREFIX)/Frameworks" "$(PREFIX)/bin"
-	cp -Rf "$(TEMPORARY_FOLDER)$(FRAMEWORKS_FOLDER)/$(OUTPUT_FRAMEWORK)" "$(PREFIX)/Frameworks/"
+	rsync -a "$(TEMPORARY_FOLDER)$(FRAMEWORKS_FOLDER)/$(OUTPUT_FRAMEWORK)" "$(PREFIX)/Frameworks/"
```
mdiep: We probably also want the `--delete` option here?

waffleboot: fixed. I tried to find correct 'cp' flags also (like -L) and tried to use 'ditto' but ditto doesn't remove exhausted files.

mdiep:
Thanks!

PR #2055, opened 2017-07-08
cli-prog used old dylib from /Library/Frameworks

waffleboot:
'add_rpath' doesn't work as it adds new rpath to the end of rpaths list and program will use old dylib from /Library/Frameworks instead of prefix-dir. I have production version of Carthage and developed version in prefix dir.

If I made incompatibility changes in my own Carthage version and install it to prefix dir then

$./carthage build

dyld: lazy symbol binding failed: Symbol not found: __TIFV11CarthageKit12BuildOptionscFT13configurationSS9platformsGVs3SetO6XCDBLD8Platform_9toolchainGSqSS_15derivedDataPathGSqSS_11cacheBuildsSb6activeSb_S0_A4_
  Referenced from: /Users/yangand/Downloads/MyCarthage/bin/./carthage
  Expected in: /Library/Frameworks/CarthageKit.framework/CarthageKit

dyld: Symbol not found: __TIFV11CarthageKit12BuildOptionscFT13configurationSS9platformsGVs3SetO6XCDBLD8Platform_9toolchainGSqSS_15derivedDataPathGSqSS_11cacheBuildsSb6activeSb_S0_A4_
  Referenced from: /Users/yangand/Downloads/MyCarthage/bin/./carthage
  Expected in: /Library/Frameworks/CarthageKit.framework/CarthageKit

Abort trap: 6

Because carthage still uses /Library/Frameworks dylib because -add_rpath addes prefix dir to the end of list

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ prefix_install: installables
 	mkdir -p "$(PREFIX)/Frameworks" "$(PREFIX)/bin"
 	cp -Rf "$(TEMPORARY_FOLDER)$(FRAMEWORKS_FOLDER)/$(OUTPUT_FRAMEWORK)" "$(PREFIX)/Frameworks/"
 	cp -f "$(TEMPORARY_FOLDER)$(BINARIES_FOLDER)/carthage" "$(PREFIX)/bin/"
-	install_name_tool -add_rpath "@executable_path/../Frameworks/$(OUTPUT_FRAMEWORK)/Versions/Current/Frameworks/"  "$(PREFIX)/bin/carthage"
+	install_name_tool -rpath "/Library/Frameworks" "@executable_path/../Frameworks/$(OUTPUT_FRAMEWORK)/Versions/Current/Frameworks/"  "$(PREFIX)/bin/carthage"
 
 package: installables
 	pkgbuild \


mdiep:
Makes sense! Thanks!

Makefile
```diff
@@ -56,7 +56,7 @@ prefix_install: installables
 	mkdir -p "$(PREFIX)/Frameworks" "$(PREFIX)/bin"
 	cp -Rf "$(TEMPORARY_FOLDER)$(FRAMEWORKS_FOLDER)/$(OUTPUT_FRAMEWORK)" "$(PREFIX)/Frameworks/"
 	cp -f "$(TEMPORARY_FOLDER)$(BINARIES_FOLDER)/carthage" "$(PREFIX)/bin/"
-	install_name_tool -add_rpath "@executable_path/../Frameworks/$(OUTPUT_FRAMEWORK)/Versions/Current/Frameworks/"  "$(PREFIX)/bin/carthage"
+	install_name_tool -rpath "/Library/Frameworks" "@executable_path/../Frameworks/$(OUTPUT_FRAMEWORK)/Versions/Current/Frameworks/"  "$(PREFIX)/bin/carthage"
```
ikesyo: This seems a bit odd to me. We should have `@executable_path/../Frameworks` in rpaths as well, right? So following how SwiftLint handles rpaths will be good: https://github.com/realm/SwiftLint/blob/0.20.1/Makefile#L62-L76.

ikesyo:
#495 is related to this.

Makefile
mdiep: I'm not sure. This seemed better than what we had before, but it may not be complete.

Issue #169, opened 2014-12-01
Use `archive` action by default, instead of `build`?


robrix:
Right now if you have this in your `Cartfile`:

```
github "robrix/Either" == 1.0.1
```

builds will fail with this error:

```
Resonance:foo rob$ carthage update
*** Fetching Either
*** Checking out Either at "1.0.1"
*** xcodebuild output can be found in /var/folders/9y/f7n4rvx901nctm_5d16c1wm40000gn/T/carthage-xcodebuild.rEyFdX.log
*** Building scheme "Either" in Either.xcworkspace
A shell task failed with exit code 65:
>>> /usr/bin/xcrun xcodebuild -workspace /Users/rob/Desktop/foo/Carthage.checkout/Either/Either.xcworkspace -scheme Either -configuration Release -sdk macosx build
2014-12-01 11:34:36.605 xcodebuild[9518:786882]  DVTAssertions: Warning in /SourceCache/IDEXcode3ProjectSupport/IDEXcode3ProjectSupport-6269/Xcode3Core/LegacyProjects/Frameworks/DevToolsCore/DevToolsCore/SpecificationTypes/BuiltInSpecifications/Compilers/XCGccMakefileDependencies.m:77
Details:  Failed to load dependencies output contents from ``/Users/rob/Library/Developer/Xcode/DerivedData/Either-gkclfdotewehumboutptpufldxma/Build/Intermediates/Either.build/Release/EitherTests.build/Objects-normal/x86_64/EitherTests.d''. Error: Error Domain=NSCocoaErrorDomain Code=260 "The file “EitherTests.d” couldn’t be opened because there is no such file." UserInfo=0x7fceb4a47da0 {NSFilePath=/Users/rob/Library/Developer/Xcode/DerivedData/Either-gkclfdotewehumboutptpufldxma/Build/Intermediates/Either.build/Release/EitherTests.build/Objects-normal/x86_64/EitherTests.d, NSUnderlyingError=0x7fceb4a57120 "The operation couldn’t be completed. No such file or directory"}. User info: {
    NSFilePath = "/Users/rob/Library/Developer/Xcode/DerivedData/Either-gkclfdotewehumboutptpufldxma/Build/Intermediates/Either.build/Release/EitherTests.build/Objects-normal/x86_64/EitherTests.d";
    NSUnderlyingError = "Error Domain=NSPOSIXErrorDomain Code=2 \"The operation couldn\U2019t be completed. No such file or directory\"";
}.
[...]

PR #2057, opened 2017-07-11
[xcodebuild] Disable Code Coverage explicitly

### LINKED ISSUE #2056: Targets built with coverage data in Xcode 9
One of my test builds I just submitted to iTunes Connect (built with Xcode 9) failed processing with the following message:

> **Invalid Bundle** - Disallowed LLVM instrumentation. Do not submit apps with LLVM profiling instrumentation or coverage collection enabled. Turn off LLVM profiling or code coverage, rebuild your app and resubmit the app.

This is caused by [this issue](https://forums.developer.apple.com/thread/81893). The new Xcode seems to build coverage data into all builds, not only for testing. From the changelog:

> Schemes with Code Coverage enabled (in the Test action) will now always build with Code Coverage, including for the Run and Launch actions. This reduces the need to build multiple versions of the application. Users who require code coverage to be off for their non-test actions can explicitly use a separate scheme with coverage disabled. (31848014)

When building frameworks during the "Archive" action Xcode seems to handle that. But since Carthage dependencies are built externally and copied into the bundle later, they still contain coverage data.

Somehow the Carthage build process needs to make sure that gathering coverage data is disabled during build, even though it's enabled in the project.

ikesyo:
Should fix #2056.

### DIFF
--- a/Source/XCDBLD/BuildArguments.swift
+++ b/Source/XCDBLD/BuildArguments.swift
@@ -116,6 +116,10 @@ public struct BuildArguments {
 			args += [ "BITCODE_GENERATION_MODE=\(bitcodeGenerationMode.rawValue)" ]
 		}
 
+		// Disable Code Coverage explicitly.
+		// See https://github.com/Carthage/Carthage/issues/2056 for the details.
+		args += [ "-enableCodeCoverage", "NO" ]
+
 		// Disable code signing requirement for all builds
 		// Frameworks get signed in the copy-frameworks action
 		args += [ "CODE_SIGNING_REQUIRED=NO", "CODE_SIGN_IDENTITY=" ]
--- a/Tests/XCDBLDTests/BuildArgumentsSpec.swift
+++ b/Tests/XCDBLDTests/BuildArgumentsSpec.swift
@@ -10,7 +10,9 @@ class BuildArgumentsSpec: QuickSpec {
 				let workspace = ProjectLocator.workspace(URL(string: "file:///Foo/Bar/workspace.xcworkspace")!)
 				let project = ProjectLocator.projectFile(URL(string: "file:///Foo/Bar/project.xcodeproj")!)
 
-				let codeSignArguments = [
+				let fixedArguments = [
+					"-enableCodeCoverage",
+					"NO",
 					"CODE_SIGNING_REQUIRED=NO",
 					"CODE_SIGN_IDENTITY=",
 					"CARTHAGE=YES",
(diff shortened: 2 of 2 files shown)


ikesyo:
Hmm `-enableCodeCoverage` option is only permitted on test action (at least with Xcode 8). :disappointed:

ikesyo:
So this may be impossible without modifying schemes but that should not be an option.

mdiep:
It's not clear to me that this is something that we should fix. (It should maybe be on the framework authors to fix their schemes.) I think that's probably the best solution.

But if we wanted to, we could probably check the Xcode version to conditionally pass `-enableCodeCoverage`?

ikesyo:
> But if we wanted to, we could probably check the Xcode version to conditionally pass `-enableCodeCoverage`?

This didn't work for Xcode 9 either.

frankschlegel:
Wow, that was dismissed a bit too fast, I think.
@mdiep You say _every_ framework author should disable code coverage generation in their project to be compatible with Carthage? Not sure if this is going to happen anytime soon…

It seems Xcode does not generate coverage data when using the "Archive" action. Can Carthage maybe somehow leverage that?

mdiep:
> You say every framework author should disable code coverage generation in their project to be compatible with Carthage?

That seems like it's more-or-less the intent of the Xcode team.

> Users who require code coverage to be off for their non-test actions can explicitly use a separate scheme with coverage disabled

frankschlegel:
I don't think so. They explicitly enabled it everywhere now because they think the users might want to get coverage data outside of the test target. But they didn't consider external build systems.

frankschlegel:
I think a good fix for Carthage to work around this is to actually disable that option in the scheme. Normally you're not supposed to use Carthage's checkouts aside from building your dependencies anyways, right?

ikesyo:
Using `archive` action is not an option, unfortunately: https://github.com/Carthage/Carthage/issues/169#issuecomment-68624884

ikesyo:
> I think a good fix for Carthage to work around this is to actually disable that option in the scheme. 

I believe modifying checkouts (making checkouts dirty) should be strongly avoided.

frankschlegel:
I see. Can you somehow "remove" the `-fprofile-instr-generate -fcoverage-mapping` parameters from the build?

ikesyo:
We are not sure. It would be great if you could detect the way to do that.

TofPlay:
Did you see this [Technical Q&A QA1964 - LLVM Instrumentation](https://developer.apple.com/library/content/qa/qa1964/_index.html)?

PR #2060, opened 2017-07-12
[gardening] Refactor `cloneSubmoduleInWorkingDirectory` with `attemptMap`.

jdhealy:
Mostly non-functional, straight refactoring — I'll point out the few functional changes in comments…

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -294,83 +294,52 @@ public func checkoutRepositoryToDirectory(_ repositoryFileURL: URL, _ workingDir
 /// repository, but without any Git metadata.
 public func cloneSubmoduleInWorkingDirectory(_ submodule: Submodule, _ workingDirectoryURL: URL) -> SignalProducer<(), CarthageError> {
 	let submoduleDirectoryURL = workingDirectoryURL.appendingPathComponent(submodule.path, isDirectory: true)
+
+	func repositoryCheck<T>(attempt closure: () throws -> T, reasonForFailure: String) -> Result<T, CarthageError> {
+		do {
+			return .success(try closure())
+		} catch let error as NSError {
+			return .failure(
+				.repositoryCheckoutFailed(workingDirectoryURL: submoduleDirectoryURL, reason: reasonForFailure, underlyingError: error)
+			)
+		}
+	}
+
+	// swiftlint:disable switch_case_on_newline
 	let purgeGitDirectories = FileManager.default.reactive
 		.enumerator(at: submoduleDirectoryURL, includingPropertiesForKeys: [ .isDirectoryKey, .nameKey ], catchErrors: true)
-		.flatMap(.merge) { enumerator, url -> SignalProducer<(), CarthageError> in
-			var name: String?
-			do {
-				name = try url.resourceValues(forKeys: [ .nameKey ]).name
-			} catch let error as NSError {
-				return SignalProducer(
-					error: CarthageError.repositoryCheckoutFailed(
-						workingDirectoryURL: submoduleDirectoryURL,
-						reason: "could not enumerate name of descendant at \(url.path)",
-						underlyingError: error
-					)
-				)
+		.attemptMap { enumerator, url -> Result<(), CarthageError> in
+			switch repositoryCheck(attempt: {
+				try url.resourceValues(forKeys: [ .nameKey ]).name
+			}, reasonForFailure: "could not enumerate name of descendant at \(url.path)") {
+			case .failure(let error): return .failure(error)
+			case .success(let name) where name != ".git": return .success(())
+			default: break // proceed to the below `repositoryCheck`
 			}
 
-			if name != ".git" {
-				return .empty
-			}
-
-			var isDirectory: Bool?
-			do {
-				isDirectory = try url.resourceValues(forKeys: [ .isDirectoryKey ]).isDirectory
-				if isDirectory == nil {
-					return SignalProducer(
-						error: CarthageError.repositoryCheckoutFailed(
-							workingDirectoryURL: submoduleDirectoryURL,
-							reason: "could not determine whether \(url.path) is a directory",
-							underlyingError: nil
-						)
-					)
+			return repositoryCheck(attempt: {
+				switch try url.resourceValues(forKeys: [ .isDirectoryKey ]).isDirectory {
+				case .none: throw NSError() // this bogus error will probably never be seen
+				case .some(let bool): return bool
 				}
-			} catch let error as NSError {
-				return SignalProducer(
-					error: CarthageError.repositoryCheckoutFailed(
-						workingDirectoryURL: submoduleDirectoryURL,
-						reason: "could not determine whether \(url.path) is a directory",
-						underlyingError: error
-					)
-				)
-			}
+			}, reasonForFailure: "could not determine whether \(url.path) is a directory")
+				.flatMap { (isDirectory: Bool) in
+					if isDirectory { enumerator.skipDescendants() }
 
-			if let directory = isDirectory, directory {
-				enumerator.skipDescendants()
-			}
-
-			do {
-				try FileManager.default.removeItem(at: url)
-				return .empty
-			} catch let error as NSError {
-				return SignalProducer(
-					error: CarthageError.repositoryCheckoutFailed(
-						workingDirectoryURL: submoduleDirectoryURL,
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Git.swift
```diff
@@ -294,83 +294,52 @@ public func checkoutRepositoryToDirectory(_ repositoryFileURL: URL, _ workingDir
...
+			switch repositoryCheck(attempt: {
+				try url.resourceValues(forKeys: [ .nameKey ]).name
+			}, reasonForFailure: "could not enumerate name of descendant at \(url.path)") {
+			case .failure(let error): return .failure(error)
+			case .success(let name) where name != ".git": return .success(())
+			default: break // proceed to the below `repositoryCheck`
 			}
 
-			if name != ".git" {
-				return .empty
```
jdhealy: [In the refactored code, a Void value is now emitted](https://github.com/Carthage/Carthage/pull/2060/files#diff-6e4701d73a8e7b2cf7f3cd4fe4e4e60bR316) instead of `flatMap`-ing an empty SignalProducer.

These values have never been explicitly consumed anywhere, so not really an issue.

Source/CarthageKit/Git.swift
```diff
@@ -294,83 +294,52 @@ public func checkoutRepositoryToDirectory(_ repositoryFileURL: URL, _ workingDir
...
-					return SignalProducer(
-						error: CarthageError.repositoryCheckoutFailed(
-							workingDirectoryURL: submoduleDirectoryURL,
-							reason: "could not determine whether \(url.path) is a directory",
-							underlyingError: nil
-						)
-					)
+			return repositoryCheck(attempt: {
+				switch try url.resourceValues(forKeys: [ .isDirectoryKey ]).isDirectory {
+				case .none: throw NSError() // this bogus error will probably never be seen
```
jdhealy: Previously, [there would be a nil value for `underlyingError`](https://github.com/Carthage/Carthage/pull/2060/files#diff-6e4701d73a8e7b2cf7f3cd4fe4e4e60bL325).

This bogus `NSError` replaces that nil value, saves some duplication of error messaging, probably will never be seen (`isDirectory` seems mostly trustworthy).

jdhealy: Also, this switch statement could be a `guard` or `if let`, but I like the way `bool` is unwrap-assigned and immediately returned on same line.

mdiep: I think we should just use `!` if we don't think this should happen.

Source/CarthageKit/Git.swift
```diff
@@ -294,83 +294,52 @@ public func checkoutRepositoryToDirectory(_ repositoryFileURL: URL, _ workingDir
...
-						workingDirectoryURL: submoduleDirectoryURL,
-						reason: "could not enumerate name of descendant at \(url.path)",
-						underlyingError: error
-					)
-				)
+		.attemptMap { enumerator, url -> Result<(), CarthageError> in
+			switch repositoryCheck(attempt: {
+				try url.resourceValues(forKeys: [ .nameKey ]).name
+			}, reasonForFailure: "could not enumerate name of descendant at \(url.path)") {
+			case .failure(let error): return .failure(error)
```
mdiep: I think this could be structured to use `Result.flatMap` instead of returning the error like this. That would move the `name != ".git"` logic into the block below.

Source/CarthageKit/Git.swift
```diff
@@ -294,83 +294,52 @@ public func checkoutRepositoryToDirectory(_ repositoryFileURL: URL, _ workingDir
 /// repository, but without any Git metadata.
 public func cloneSubmoduleInWorkingDirectory(_ submodule: Submodule, _ workingDirectoryURL: URL) -> SignalProducer<(), CarthageError> {
 	let submoduleDirectoryURL = workingDirectoryURL.appendingPathComponent(submodule.path, isDirectory: true)
+
+	func repositoryCheck<T>(attempt closure: () throws -> T, reasonForFailure: String) -> Result<T, CarthageError> {
```
mdiep: We might be better off extending `FileManager` and `URL`?

```swift
extension URL {
    func name() -> Result<String, CarthageError> { ... }
}
```

Source/CarthageKit/Git.swift
```diff
@@ -294,83 +294,52 @@ public func checkoutRepositoryToDirectory(_ repositoryFileURL: URL, _ workingDir
...
+		do {
+			return .success(try closure())
+		} catch let error as NSError {
+			return .failure(
+				.repositoryCheckoutFailed(workingDirectoryURL: submoduleDirectoryURL, reason: reasonForFailure, underlyingError: error)
+			)
+		}
+	}
+
+	// swiftlint:disable switch_case_on_newline
```
mdiep: I think we should stick to the rule instead of disabling it.

Source/CarthageKit/Git.swift
```diff
@@ -294,83 +294,52 @@ public func checkoutRepositoryToDirectory(_ repositoryFileURL: URL, _ workingDir
...
-			} catch let error as NSError {
-				return SignalProducer(
-					error: CarthageError.repositoryCheckoutFailed(
-						workingDirectoryURL: submoduleDirectoryURL,
-						reason: "could not enumerate name of descendant at \(url.path)",
-						underlyingError: error
-					)
-				)
+		.attemptMap { enumerator, url -> Result<(), CarthageError> in
+			switch repositoryCheck(attempt: {
```
mdiep: If we're going to switch on this, can we pull out the the result of `repositoryCheck` into a local variable? I think that'd be far more readable.

Source/CarthageKit/Git.swift
```diff
@@ -294,83 +294,52 @@ public func checkoutRepositoryToDirectory(_ repositoryFileURL: URL, _ workingDir
...
-				isDirectory = try url.resourceValues(forKeys: [ .isDirectoryKey ]).isDirectory
-				if isDirectory == nil {
-					return SignalProducer(
-						error: CarthageError.repositoryCheckoutFailed(
-							workingDirectoryURL: submoduleDirectoryURL,
-							reason: "could not determine whether \(url.path) is a directory",
-							underlyingError: nil
-						)
-					)
+			return repositoryCheck(attempt: {
```
mdiep: If we put the string first, we could use trailing closure syntax here, which I think would be more readable.

```swift
// the error could add the "could not " before the reason string here
return repositoryCheck("determine whether \(url.path) is a directory") {
        switch try url.resourceValues(forKeys: [ .isDirectoryKey ]).isDirectory {
        case .none:
            throw NSError() // this bogus error will probably never be seen
        case .some(let bool):
            return bool
        }
     }
     .flatMap { (isDirectory: Bool) in
         if isDirectory { enumerator.skipDescendants() }
             return repositoryCheck("remove \(url.path)") {
                 try FileManager.default.removeItem(at: url)
             }
         }
     }
```

mdiep:
I'm excited to see these refactoring PRs. 💖

Source/CarthageKit/Git.swift
jdhealy: @mdiep Good notes, all around.

Side Question regarding an initializer on the _actual_ Result type: is the below still the favored style? (Over trailing closure?)
```
Result(attempt: {
…
})
```

mdiep: I'm not sure that there's a style preference for that. I think it's more about type inference. I'm happy to accept either.

Source/CarthageKit/Git.swift
```diff
@@ -294,83 +294,47 @@ public func checkoutRepositoryToDirectory(_ repositoryFileURL: URL, _ workingDir
...
-					error: CarthageError.repositoryCheckoutFailed(
-						workingDirectoryURL: submoduleDirectoryURL,
-						reason: "could not enumerate name of descendant at \(url.path)",
-						underlyingError: error
-					)
-				)
+		.attemptMap { enumerator, url -> Result<(), CarthageError> in
+			return repositoryCheck("enumerate name of descendant at \(url.path)") {
+				try url.resourceValues(forKeys: [ .nameKey ]).name
 			}
```
mdiep: Can you indent these 2 lines one more level? Our style is to match the `}` to the same level as the `.` on the next line—which means that the `try` line is indented 2 levels from the `return` line.

Same thing below.

mdiep:
Just a couple indentation fixes. Otherwise this looks good. Great improvement. 👏

PR #2061, opened 2017-07-12
[gardening] Avoid labeled if `break` in `Repository.fromIdentifier`.

jdhealy:
No functional changes.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -44,7 +44,7 @@ extension Repository {
 	/// for the github.com, or the form "http(s)://hostname/owner/name" for
 	/// Enterprise instances.
 	public static func fromIdentifier(_ identifier: String) -> Result<(Server, Repository), ScannableError> {
-		// GitHub.com
+		// ‘owner/name’ → GitHub.com
 		let range = NSRange(location: 0, length: identifier.utf16.count)
 		if let match = nwoRegex.firstMatch(in: identifier, range: range) {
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/GitHub.swift
```diff
@@ -53,14 +53,13 @@ extension Repository {
...
 		// GitHub Enterprise
-		breakpoint: if let url = URL(string: identifier), let host = url.host {
-			var pathComponents = url.pathComponents.filter { $0 != "/" }
-			guard pathComponents.count >= 2 else { break breakpoint }
-
-			// Consider that the instance might be in subdirectories.
-			let name = pathComponents.removeLast()
-			let owner = pathComponents.removeLast()
-
+		if
```
andersio: `guard` seems a better fit in this case. 🤔

Source/CarthageKit/GitHub.swift
```diff
@@ -53,14 +53,13 @@ extension Repository {
...
+			let url = URL(string: identifier),
+			let host = url.host,
+			case var pathComponents = url.pathComponents.filter({ $0 != "/" }),
```
mdiep: *mind blown*

Source/CarthageKit/GitHub.swift
jdhealy: @andersio done in https://github.com/Carthage/Carthage/pull/2061/commits/2426c491b73e54b5e7dc7e40312986b47e165ec2

mdiep:
🤘

PR #2066, opened 2017-07-15
[gardening] Postpone unneeded optional until after `reduce`

jdhealy:
One more real simple no-functional-change refactor while I’m at it.

### DIFF
--- a/Source/CarthageKit/VersionFile.swift
+++ b/Source/CarthageKit/VersionFile.swift
@@ -327,13 +327,8 @@ public func versionFileMatches(
 				.flatMap(.merge) { platform in
 					return versionFile.satisfies(platform: platform, commitish: commitish, binariesDirectoryURL: rootBinariesURL, localSwiftVersion: localSwiftVersion)
 				}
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/VersionFile.swift
```diff
@@ -327,13 +327,8 @@ public func versionFileMatches(
...
 					return versionFile.satisfies(platform: platform, commitish: commitish, binariesDirectoryURL: rootBinariesURL, localSwiftVersion: localSwiftVersion)
 				}
-				.reduce(true) { current, result in
-					if let current = current {
-						return current && result
-					} else {
-						return false
-					}
-				}
+				.reduce(true) { $0 && $1 }
```
mdiep: This could even be `.reduce(true, &&)`. 😄

jdhealy: Tried that, the `autoclosure` on `rhs` prevented it. 😉

mdiep: Weird. 😕

PR #2070, opened 2017-07-18
[gardening] Refactor: extend Result attempting a throwing closure taking a URL, failing with `CarthageError`

jdhealy:
```swift
extension Result where Error == CarthageError {
	/// Constructs a result from a throwing closure taking a `URL`, failing with `CarthageError` if throw occurs.
	/// - parameter carthageError: Defaults to `CarthageError.writeFailed`.
	internal init(
		at url: URL,
		carthageError: (URL, NSError) -> CarthageError = CarthageError.writeFailed,
		attempt closure: (URL) throws -> Value
	) {
		do {
			self = .success(try closure(url))
			self = .failure(carthageError(url, error))
		}
	}
}
```

[No functional changes (after force push)](https://github.com/Carthage/Carthage/pull/2070#issuecomment-316130854).

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -208,12 +208,9 @@ public final class Project { // swiftlint:disable:this type_body_length
 
 	/// Writes the given Cartfile.resolved out to the project's directory.
 	public func writeResolvedCartfile(_ resolvedCartfile: ResolvedCartfile) -> Result<(), CarthageError> {
-		do {
-			try resolvedCartfile.description.write(to: resolvedCartfileURL, atomically: true, encoding: .utf8)
-			return .success(())
-		} catch let error as NSError {
-			return .failure(.writeFailed(resolvedCartfileURL, error))
-		}
+		return Result(at: resolvedCartfileURL, attempt: {
+			try resolvedCartfile.description.write(to: $0, atomically: true, encoding: .utf8)
+		})
 	}
 
 	/// Limits the number of concurrent clones/fetches to the number of active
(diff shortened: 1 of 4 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -911,34 +888,29 @@ public func copyProduct(_ from: URL, _ to: URL) -> SignalProducer<URL, CarthageE
...
-		} catch let error as NSError {
-			if error.code != NSFileNoSuchFileError {
-				return .failure(.writeFailed(to, error))
+		let createDirectory = { try manager.createDirectory(at: $0, withIntermediateDirectories: true) }
+		return allowForFileWriteFileExistsError(Result(at: to.deletingLastPathComponent(), attempt: createDirectory))
+			.flatMap { _ in
+				allowForFileWriteFileExistsError(Result(at: to, attempt: manager.removeItem(at:)))
+			}
+			.flatMap { _ in
+				Result(at: to, attempt: { destination /* to */ in
```
jdhealy: `swiftlint` required longer than the two-character identifier `to`

jdhealy:
Had to force push, missed a functional change which made the commit message incorrect.

Now (truly) no functional changes.

Source/CarthageKit/Project.swift
```diff
@@ -922,11 +913,9 @@ public final class Project { // swiftlint:disable:this type_body_length
...
 					}
 
-					do {
-						try fileManager.createSymbolicLink(atPath: dependencyCheckoutURL.path, withDestinationPath: linkDestinationPath)
-					} catch let error as NSError {
-						return .failure(.writeFailed(dependencyCheckoutURL, error))
-					}
+					return Result(at: dependencyCheckoutURL, attempt: {
+						try fileManager.createSymbolicLink(atPath: $0.path, withDestinationPath: linkDestinationPath)
+					})
```
jdhealy: ⚠️ Accidentally introduced a functional change (moreover, a bug) by `return`ing the result’s value from 
 a for-loop — breaking the loop and returning from [the surrounding closure](https://github.com/jdhealy/Carthage/blob/ec4d59a44351ee56d545cb0446115a0c151dc3fa/Source/CarthageKit/Project.swift#L867-L922). Previously, [only the result’s error would break the for-loop](https://github.com/Carthage/Carthage/pull/2070/files#diff-3bd7f6198aa606c5ec6931e77cb3c42bL928).

PR #2072, opened 2017-07-20
Avoid breaking for-loop on Result’s value, only break on error in `symlinkCheckoutPaths`

jdhealy:
Fixes bug introduced in <https://github.com/Carthage/Carthage/commit/ec4d59a#commitcomment-23208406>.

Message assuring “no functional changes” is incorrect in commit ec4d59a44351ee56d545cb0446115a0c151dc3fa.

```diff
-	do {
-		try FileManager.default.copyItem(at: url, to: destinationURL, avoiding·rdar·32984063: true)
-		return .success(destinationURL)
-	} catch let error as NSError {
-		return .failure(.writeFailed(destinationURL, error))
-	}
+	return Result(at: destinationURL, attempt: {
+		try FileManager.default.copyItem(at: url, to: $0, avoiding·rdar·32984063: true)
+		return $0
+	})
```

Accidentally introduced a functional change (moreover, a bug) by `return`ing the result’s value from a for-loop — breaking the loop and returning from the surrounding closure.

Return to previous behavior where only the result’s error breaks the for-loop.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -913,9 +913,11 @@ public final class Project { // swiftlint:disable:this type_body_length
 						continue
 					}
 
-					return Result(at: dependencyCheckoutURL, attempt: {
+					if let error = Result(at: dependencyCheckoutURL, attempt: {
 						try fileManager.createSymbolicLink(atPath: $0.path, withDestinationPath: linkDestinationPath)
-					})
+					}).error {
+						return .failure(error)
+					}
 				}
 
 				return .success()


Source/CarthageKit/Project.swift
```diff
@@ -913,9 +913,11 @@ public final class Project { // swiftlint:disable:this type_body_length
 						continue
 					}
 
-					return Result(at: dependencyCheckoutURL, attempt: {
+					if let error = Result(at: dependencyCheckoutURL, attempt: {
```
ikesyo: It seems to me that restoring the original structure will be more readable / understandable:

```swift
do {
	try fileManager.createSymbolicLink(atPath: dependencyCheckoutURL.path, withDestinationPath: linkDestinationPath)
	return .failure(.writeFailed(dependencyCheckoutURL, error))
}
```

jdhealy: Mostly, this is just a stopgap before a larger refactor, which would look more like the following:

##### Preview with indenting changes removed:

```diff
 		// (which could be outside `CarthageProjectCheckoutsPath`).
 		if names.isEmpty { return .success() } // swiftlint:disable:this single_line_return
 
+		return Result(at: dependencyCheckoutsURL, attempt: {
 		do {
-			try fileManager.createDirectory(at: dependencyCheckoutsURL, withIntermediateDirectories: true)
+			try fileManager.createDirectory(at: $0, withIntermediateDirectories: true)
 		} catch let error as NSError {
 			if !(error.domain == NSCocoaErrorDomain && error.code == NSFileWriteFileExistsError) {
-				return .failure(.writeFailed(dependencyCheckoutsURL, error))
+				throw error
 			}
 		}
 
 		for name in names {
-			let dependencyCheckoutURL = dependencyCheckoutsURL.appendingPathComponent(name)
+			let dependencyCheckoutURL = $0.appendingPathComponent(name)
 			let subdirectoryPath = (carthageProjectCheckoutsPath as NSString).appendingPathComponent(name)
 			let linkDestinationPath = relativeLinkDestination(for: dependency, subdirectory: subdirectoryPath)
 
-			let dependencyCheckoutURLResource = try? dependencyCheckoutURL.resourceValues(forKeys: [
-				.isSymbolicLinkKey,
-				.isDirectoryKey,
-			])
+			let dependencyCheckoutURLCheck = {
+				try dependencyCheckoutURL.copyResource(bridgingFromType: CFBoolean.self, forIdentifierKey: $0) == .some(true)
+			}
 
-			if dependencyCheckoutURLResource?.isSymbolicLink == true {
+			if try dependencyCheckoutURLCheck(kCFURLIsSymbolicLinkKey) {
 				_ = dependencyCheckoutURL.path.withCString(Darwin.unlink)
-			} else if dependencyCheckoutURLResource?.isDirectory == true {
+			} else if try dependencyCheckoutURLCheck(kCFURLIsDirectoryKey) {
 				// older version of carthage wrote this directory?
 				// user wrote this directory, unaware of the precedent not to circumvent carthage’s management?
 				// directory exists as the result of rogue process or gamma ray?
 
 				// TODO: explore possibility of messaging user, informing that deleting said directory will result
 				// in symlink creation with carthage versions greater than 0.20.0, maybe with more broad advice on
-				// “from scratch” reproducability.
+				// “from scratch” reproducibility.
 				continue
 			}
 
-			if let error = Result(at: dependencyCheckoutURL, attempt: {
-				try fileManager.createSymbolicLink(atPath: $0.path, withDestinationPath: linkDestinationPath)
-			}).error {
-				return .failure(error)
-			}
+			try fileManager.createSymbolicLink(atPath: dependencyCheckoutURL.path, withDestinationPath: linkDestinationPath)
 		}
-
-		return .success()
+		})
 	}
```

mdiep:
I think this is a sign that this function needs more work. So I'm happy to merge as-is. I look forward to another PR! 😄

PR #2082, opened 2017-07-27
Utilize `FlattenStrategy.concurrent(limit:)`

ikesyo:
Related to #1910.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -742,13 +742,6 @@ public final class Project { // swiftlint:disable:this type_body_length
 			}
 	}
 
-	/// Limits the number of concurrent checkouts to the number of active 
-	/// processors.
-	let checkoutQueue = ConcurrentProducerQueue(
-		name: "org.carthage.Constants.Project.checkoutResolvedDependencies",
-		limit: ProcessInfo.processInfo.activeProcessorCount
-	)
-
 	/// Checks out the dependencies listed in the project's Cartfile.resolved,
 	/// optionally they are limited by the given list of dependency names.
 	public func checkoutResolvedDependencies(_ dependenciesToCheckout: [String]? = nil, buildOptions: BuildOptions?) -> SignalProducer<(), CarthageError> {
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -768,8 +761,9 @@ public final class Project { // swiftlint:disable:this type_body_length
 			}
 			.zip(with: submodulesSignal)
 			.flatMap(.merge) { dependencies, submodulesByPath -> SignalProducer<(), CarthageError> in
+				let limit = UInt(ProcessInfo.processInfo.activeProcessorCount)
 				return SignalProducer<(Dependency, PinnedVersion), CarthageError>(dependencies)
-					.map { (dependency, version) -> SignalProducer<(), CarthageError> in
+					.flatMap(.concurrent(limit: limit)) { (dependency, version) -> SignalProducer<(), CarthageError> in
```
mdiep: I realize that this is functionally equivalent to the existing behavior… but I don't think this is really an appropriate limit since this is a disk- and network-intensive task and not one that's CPU-limited. (Just because you a 16-core processor doesn't mean that your network and disk can handle 16 simultaneous checkouts.)

Could we lower this? Maybe set it to 3 or 4? 😕

ikesyo: Makes sense, I'll choose 4.

Source/CarthageKit/Project.swift
```diff
@@ -769,7 +762,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 			.zip(with: submodulesSignal)
 			.flatMap(.merge) { dependencies, submodulesByPath -> SignalProducer<(), CarthageError> in
 				return SignalProducer<(Dependency, PinnedVersion), CarthageError>(dependencies)
-					.map { (dependency, version) -> SignalProducer<(), CarthageError> in
+					.flatMap(.concurrent(limit: 4)) { (dependency, version) -> SignalProducer<(), CarthageError> in
```
andersio: Shouldn't we use `ProcessInfo.activeProcessorCount`?

andersio: It seems the changes here also undo the parallelism. The `concurrent` strategy doesn't automatically schedule things to global concurrent queues.

jdhealy: @mdiep [suggested to limit the concurrency here](https://github.com/Carthage/Carthage/pull/2082#discussion_r129959730).

Unless, maybe I misinterpreted your comment — are you suggesting to `limit` with `min(ProcessInfo.processInfo.activeProcessorCount, 4)`?

andersio: Hmm, I've missed that comment. It seems the work is eventually piped to a concurrent queue too, in `cloneOrFetch`.

mdiep:
🎉

PR #2098, opened 2017-08-04
[Makefile] Use variables for specifying repeated commands

zhongwuzw:
I met a guy, he downloaded the carthage zip file from github, then double-clicked the Makefile file because Makefile was executable.  Leads to remove all non-synstem file. e.g. `rm -rf "$(FRAMEWORKS_FOLDER)/$(OUTPUT_FRAMEWORK)"` will interpret as `rm -rf /` .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,14 @@ VERSION_STRING=$(shell git describe --abbrev=0 --tags)
 COMPONENTS_PLIST=Source/carthage/Components.plist
 DISTRIBUTION_PLIST=Source/carthage/Distribution.plist
 
+RM=rm -f
+RMD=rm -rf
+MKDIR=mkdir -p
+SUDO=sudo
+MV=mv -f
+CP=cp
+RSYNC=rsync -a --delete
+
 .PHONY: all bootstrap clean install package test uninstall
 
 all: bootstrap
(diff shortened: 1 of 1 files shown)


jdhealy:
Wow, that’s terrible, massive, massive apologies to anyone affected.

@zhongwuzw your knowledge of makefiles is evidently greater than mine, could you explain how [extracting the commands (as you did in `cb81ba6`)](https://github.com/Carthage/Carthage/pull/2098/commits/cb81ba6aad6abf858957c32c8cf860786454cf65) adds safety?

My inclination was towards [`TEMPORARY_FOLDER` (user-settable because of `?=`)](https://github.com/Carthage/Carthage/blob/d530ecdf7a5a9814f4ea4ff37cf58fe5dd5e16b1/Makefile#L37), which could hopefully be fixed by something like:
```make
RM_SAFELY = ZDOTDIR='/var/empty' zsh -c '[[ ! $${1:?} =~ "^[[:space:]]+\$$" ]] && [[ $${1:A} != "/" ]] && (( $${\#} == "1" )) && noglob rm -rf $${1:A}' --
TEMPORARY_FOLDER?=/tmp/MyProject.dst

.PHONY: default

default:
    $(RM_SAFELY) "$(TEMPORARY_FOLDER)"
```

Additionally, might also want to rename `TEMPORARY_FOLDER` to `CARTHAGE_TEMPORARY_FOLDER`.

zhongwuzw:
@jdhealy , Makefile use parentheses to get variable, like `$(Foo)`, shell use braces or not , like `${Foo}`, `$Foo`, In other words, double-clicked Makefile to run the script , interpreter wouldn't figure out that `$(Foo)` is a variable , but subshell. 
e.g. 
`$(RM) "$(INTERNAL_PACKAGE)"` , `$(RM)` would cause a subshell to execute `rm` command, `rm` lack of parameters, so wouldn't work .

jdhealy:
@zhongwuzw do you know of other open-source repositories that have committed `Makefile` safety changes like this?

zhongwuzw:
@jdhealy , `openresty` uses a lot safety operation , like [lua-redis-parser](https://github.com/openresty/lua-redis-parser/blob/master/Makefile#L31)

mdiep:
`RM` and `TEMPORARY_FOLDER` are defined the same way here—so if one is empty, the other should be too.

Defining variables like these for functions is fairly standard practice, usually because it makes it easier to customize the Makefile. (Check out http://www.oreilly.com/openbook/make3/book/ch03.pdf)

I'm a little confused about how this would get executed if you double-click on it. But I don't see any problems with these changes.

zhongwuzw:
@mdiep , double-clicked my updated Makefile, command like `RM` `MKDIR` would not execute correctly because of lacking enough parameters.

jdhealy:
thanks @zhongwuzw 🎉

zhongwuzw:
@jdhealy :beer::beer::beer::beer:

Issue #2062, opened 2017-07-14
Carthage derived cache issue when switching xcode-select


schlingding:
* `carthage version`: 0.24.0
* `xcodebuild -version`: XCode 9 Beta 3: 9M174d
* Are you using `--no-build`? No
* Are you using `--no-use-binaries`? Yes
* Are you using `--use-submodules`? No
* Are you using `--cache-builds`? Yes
* Command: `carthage update --no-use-binaries --platform iOS --cache-builds `

#Highlight
When building using Carthage in an newer version of xcodebuild, Carthage builds the new framework but reaches this error (Full output below):
`error: module compiled with Swift 3.1 cannot be imported in Swift 3.2`
OR
`error: module compiled with Swift 3.1 cannot be imported in Swift 4.0`

The Carthage directory in the project is deleted before running Carthage:
`rm -rf Carthage`

The problem occurs with a framework building against a dependent framework. FrameworkA builds. FrameworkB builds and has an `import FrameworkA` in the code. xcodebuild claims Framework A has this issue: `error: module compiled with Swift 3.1 cannot be imported in Swift 3.2`

Other colleagues are also having this issue so it is not isolated to just myself.

The only current fix is to delete Carthage DerivedData:
`rm -rf ~/Library/Caches/org.carthage.CarthageKit/DerivedData`

# Carthage Output:
*** Fetching FrameworkA
*** Fetching FrameworkB
*** Checking out FrameworkA at "v0.2.0"
*** Checking out FrameworkB at "v2.2.2"
*** No cache found for FrameworkA, building with all downstream dependencies
*** No cache found for FrameworkB, building with all downstream dependencies
*** xcodebuild output can be found in /var/folders/fg/xxx/T/carthage-xcodebuild.N10dNu.log
*** Building scheme "FrameworkA" in FrameworkA.xcodeproj
*** Building scheme "FrameworkB" in FrameworkB.xcodeproj

Build Failed
	Task failed with exit code 65:
[...]

Issue #1469, opened 2016-09-12
--no-clean option for the build command


mz2:
I'd like to have some more control over the `carthage build` command whereby the xcodebuild tasks executed by it would not include cleaning the target being built prior to building (i.e. make the `["clean", "build"]` tasks optionally to just `["build"]` in the function `buildScheme` in Xcode.swift.

In the case where I don't need a public release grade definite clean builds, this would commonly fit my needs and speed up my work, especially for libraries that have multiple Carthage based dependencies itself.

I've made a prototype implementation of the idea which I'm happy to polish up into a pull request if this is a feature you'd like to have included. Hopefully I have not overseen some existing functionality that already provides this some other way?

mdiep:
Our hope is that instead of adding a `--no-clean` option, we can transparently skip unchanged dependencies: #1222.

mz2:
I figured you'd say that :-) Thing is, this is achievable as a tiny (i.e. safe) change and I'm proposing it as a non-default option. Arguably it would be made redundant indeed by the work needed to complete #1222 successfully, though you may still want to be more lenient even with that in.

falcon283:
Any news on `--no-clean`
It will tremendously speed up the Private Libraries development adding them as submodule and rebuild only the changed files

ikesyo:
Closed by #2102.

PR #2102, opened 2017-08-07
[PoC] Split derived data directory by Xcode version

ikesyo:
This would make it enable to remove `clean` action from `xcodebuild clean build`.

This would also address #1469 and #2062. See also #1419.


ikesyo:
Thanks for the feedback!

Issue #506, opened 2015-05-30
Prefer semver tags


dbworku:
Carthage uses the latest git tag when checking out. To prevent Carthage from grabbing tags that may be bug or issue related, I suggest filtering semver style tags.

![screen shot 2015-05-30 at 6 11 58 pm](https://cloud.githubusercontent.com/assets/2839631/7899820/4992052e-06f9-11e5-8ab5-4da6b43bb8d2.png)

jspahrsummers:
We do. But in this particular case, it's hard for an algorithm to decide if that's SemVer or not—i.e., maybe it's meant to be version `2.0.0`, and all that other text is just a prefix.

On the other hand, we always filter out version strings that have a _suffix_, since that usually indicates something outside of SemVer—or else a prerelease tag, which we don't currently support.

PR #2111, opened 2017-08-12
[BREAKING] Make git tags parsing stricter on `SemanticVersion.from(_:)`

ikesyo:
The new behavior matches the SwiftPM's one.

This revisits #428, #506 and #508.

### DIFF
--- a/Source/CarthageKit/Version.swift
+++ b/Source/CarthageKit/Version.swift
@@ -44,9 +44,8 @@ public struct SemanticVersion: VersionType {
 	public static func from(_ pinnedVersion: PinnedVersion) -> Result<SemanticVersion, ScannableError> {
 		let scanner = Scanner(string: pinnedVersion.commitish)
 
-		// Skip leading characters, like "v" or "version-" or anything like
-		// that.
-		scanner.scanUpToCharacters(from: versionCharacterSet, into: nil)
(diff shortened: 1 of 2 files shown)


mdiep:
I think I'm 👎 on this.

It would be nice to match SwiftPM… but I don't think it's a good idea to make a breaking change like this. (This is also a challenge of a decentralized system: there's no central database we can use to assess compatibility.)

BUT I think we could maybe restrict things further. Maybe we could choose some more popular schemes and allow all of them? (e.g. `4.0`, `v4.0`, `release-4.0`, etc.) That might require some work to see what schemes are in use.

Down the road, when SwiftPM inevitably becomes the de facto solution, I think we could match it. (If Carthage remains useful.)

ikesyo:
Hm I withdraw this then. Thanks for the feedback!

ikesyo:
Refusing tags which have more than 2 hyphens might be good (e.g. `foo-bar-3.0.0`)

mdiep:
> Refusing tags which have more than 2 hyphens might be good (e.g. foo-bar-3.0.0)

That sounds like a reasonable requirement to me. There's a big gap between the current behavior and the SwiftPM behavior. I think we can move away from the current behavior.

PR #2112, opened 2017-08-14
Private Cache for concurrency

zorture:
Adding a new feature of private cache for the builds to achieve concurrency.

Update:

Issue:
1. While running Carthage update concurrently then one build fails.
2. Since build share once common cache i.e. ~/Library/Caches/org.carthage.CarthageKit/.

Fix:
1. Adding a feature of private cache for the builds to achieve concurrency while running Carthage update.
2. With the fix user has to include additional build parameter i.e. --private-cache and the build will have it's own private cache.
e.g.:  carthage update --platform ios --private-cache

### DIFF
--- a/Source/CarthageKit/Constants.swift
+++ b/Source/CarthageKit/Constants.swift
@@ -9,6 +9,9 @@ public struct Constants {
 	/// The name of the folder into which Carthage puts binaries it builds (relative
 	/// to the working directory).
 	public static let binariesFolderPath = "Carthage/Build"
+	
+	// create prive cache for the build to achieve concurrency
+	public static var createPrivateCache = false
 
 	/// The fallback dependencies URL to be used in case
 	/// the intended ~/Library/Caches/org.carthage.CarthageKit cannot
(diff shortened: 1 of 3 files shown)


ikesyo:
I'm not convinced if this is a good direction. I think this breaks the assumption that shared / cached contents (checkouts or downloaded binaries) can be reused for efficiency. So when the option is used there will be a lot of additional fetching repositories or downloading binaries for each execution and that should be inefficient / slow.

mdiep:
Yeah, I don't think this is a good solution. 👎 

It seems like what you're actually trying to achieve is concurrency while building? We could maybe add a parameter to allow concurrency there BUT I think it's unlikely to bring a speeding since building a project is already concurrent. (You're more likely to have resource contention.) Any such change would definitely require benchmarking.

zorture:
I am aware that my solution will not work for cached-builds. My solution is for the problem because of shared cache.

I do want to achieve concurrency while building. I do't think adding concurrency parameter will help. The issue is due to shared cache.

We can archive multiple builds using xcodebuild simultaneously then i guess we should be able to build our dependents simultaneously using carthage as well.

ikesyo:
> I am aware that my solution will not work for cached-builds.

I assume we didn't say anything about cached-builds.

> I do want to achieve concurrency while building. I do't think adding concurrency parameter will help. The issue is due to shared cache.
>
> We can archive multiple builds using xcodebuild simultaneously then i guess we should be able to build our dependents simultaneously using carthage as well.

As @mdiep said above, that should be done on the carhgae's building phase.

ikesyo:
I wonder that what you should do here are:

1. Running `carthage update --no-build` first
2. Then running `carthage build` concurrently with separated dependency groups

zorture:
Thanks for the tip. I got few more ideas but it did't solve the issue. 
I have attached some screen shot to explain the issue more clearly.

carthage concurrency issue

I am running carthage update in two separate folders simultaneously and one fails. I tried your solution as well but it still fails.

carthage no build

Reason for the build fail in the logs:

build failed log

The reason i am looking for private cache is for CI where we have different builds with different tags.

mdiep:
You can specify a custom derived data if you'd like to do that. `carthage bootstrap --derived-data <path to the custom derived data folder>`

Note that you probably want to run `carthage bootstrap` on CI, not `carthage update`.

zorture:
I tried but it still gives the lock error.

boot strap


Do You think making --private-cache work with the --cache-builds will be acceptable solution?

ikesyo:
As I said above, I'd recommend you to run just only `carthage build` concurrently. Do you really need `update` command?

mdiep:
I wouldn't expect `carthage checkout` to fail when concurrently from separate projects. I think we should address this.

I'm a little surprised that git takes a lock in this case. 😕

mdiep:
Oh, but I don't think this PR is the solution. Would you mind opening an issue so we can discuss how to best solve this?

zorture:
Carthage build don't work either

Consider the below 2 scenarios.

I am trying to run below cart file from 2 separate folders simultaneously. Notice the tags are different. This is the most important part of the issue.

From Folder A :  github "SnapKit/SnapKit" = 3.2.0
From Folder B :  github "SnapKit/SnapKit" = 3.3.2

Results:
Carthage build: only one builds. other one don't do any thing.
Carthage update: one fails.

Expected Result
Both builds should be executed as they are different. 

Reason
Both have same derive data cache folder name i.e. SnapKit

Example 2: Running below cart file simultaneously.

From Folder A :  github "SnapKit/SnapKit" = 3.2.0
From Folder B :  github "danielgindi/Charts" == 3.0.

Result:
Both are success.

Reason
Bothe have different derive data folder as they are separate projects.


Below is the issue i opened.
Carthage Concurrent Build Fail. #2120

PR #2116, opened 2017-08-15
[ResolverSpec] Add some examples to a pending test

ikesyo:
#1708

### DIFF
--- a/Tests/CarthageKitTests/ResolverSpec.swift
+++ b/Tests/CarthageKitTests/ResolverSpec.swift
@@ -326,9 +326,23 @@ class ResolverSpec: QuickSpec {
 				],
 			]
 
-			let resolved = db.resolve([ github1: .compatibleWith(.v3_0_0) ])
-			expect(resolved.value).to(beNil())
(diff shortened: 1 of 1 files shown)


jdhealy:
Great follow-through, @ikesyo — really admirable care on this!

If you’d permit me to suggest an alternate approach:

```swift
for toResolveUpon in [
	[ github1: .atLeast(.v3_0_0) as VersionSpecifier ],
	[ github1: .compatibleWith(.v3_0_0) ],
	[ github1: .exactly(.v3_0_0) ],
] {
	expect({
		switch db.resolve(toResolveUpon) {
		case .success(_): return .succeeded
		case .failure(let error): return .failed(reason: "resolution of «\(toResolveUpon)» from «\(db)» errored «\(error)»")
		}
	}).to(succeed())
}
```

Do you see any disadvantages to this that I might not be aware of?

ikesyo:
@jdhealy That seems interesting! But what we would expect here is a resolution failure, so that will be a bit different I think.

PR #2117, opened 2017-08-15
[gardening] Remove obsolete macOS version check

ikesyo:
macOS deployment target was already bumped to 10.10 in #1197.


ikesyo:
🤘

PR #2122, opened 2017-08-16
Resolver performance

BobElDevil:
This PR aims to address some inefficiencies I noticed in the Resolver (after staring at it for hours being utterly confused of course 😛 ). Especially when there's incompatible version specifiers somewhere in the tree. I'm going to describe what the resolver did before, and the changes I made to it.

Previously the resolver would start at the root repo, and do the following:

	1. Grab the direct dependency version specifiers from the Cartfile. If there are no dependencies, return
	2. Generate a list of versions for each dependency, that fit the specifier in the cartfile. If there are none, throw an error about missing versions.
	3. 'permute' the lists generated in (2) to generate a stream of all possible combinations of direct dependencies that fit the specifiers
	4. For each combination generated in 3, loop through each dependency and grab the Cartfile of that specific version.
	5. Recursively run steps 1-4 with this nested cartfile contents with the following change: In step 4, if the combination generated does not fit with the versions as provided, it throws an error about incompatible versions.

The main issues I noticed were related to the brute force nature of this approach
     
    1. In the recursive invocation of step (3), it will generate all permutations fitting the current specifiers, _before_ looking at the parent specifiers. Only when it loops through the permutations does it throw the error for incompatible versions. For example if you have a tree like:
    A:
        B // any
        C == 3.0
    B 2.0:
        C ~> 2.0
    B 1.0:
        C ~> 3.0
    And C had many 2.* versions, it would run through (and reject) every permutation that involves B 2.0 with some variant of C 2.*.
    2. Also in the recursive invocation of step (3), it generates permutations of all dependencies, even those that showed up previously in the parent. This means that we could end up checking duplicate trees, since the common dependencies would be permuted during the parent invocation _and_ the permuted during the child invocation.
    3. Due to the scheduling of permutations, it's in concept only processing one possible permutation at a time, but there's a huge queue of growing permutations stacking up in the signal producer queue, at all level of the recursion. This leads large memory usage with deep dendency trees.

 I attempted to fix these issues by modifying the algorithm such that instead of rejecting every permutation of a subtree when it's found to be incompatible, build in validation in the permutation generation itself, skipping over any whole chunks of permutations at once if any combinations known to be incompatible show up. In addition when recursing to deeper levels, only generate permutations over dependencies that aren't already in the graph

Since each permutation generated relies on the state of the previous one (due to needing to skip found incompatibilities), this implementation is much more of a standard recursive implementation, using an error cache to track incompatibilities, and doing away with many of the signal producers, instead using Results. This ensures we're processing exactly one thing at a time and we don't enqueue permutations before we have a chance to vet whether they should have been generated at all. 

That aspect is probably going to be the controversial portion of this, since we are pretty strict about staying stateless. I attempted to mimic the old algorithm as much as possible, and any state maintained is local to the resolver, no global state is used. I tried to figure out a way to use signals and _guarantee_ that we were only processing one thing at a time, but even calls like `flatMap(.concat)` which are 'serial' are still not guaranteed that certain events won't queue up while others are getting processed, leading to unpredictable behavior with regards to the error cache. 

Since this is such a big change, I toggled it with a new flag `--new-resolver`, and have the unit tests run it against both resolvers (in addition to adding some extra use cases). From my tests so far, it seems to be working as expected.

My main use case which prompted this work I unfortunately can't share due to private repos, but the basic situation was we have 17 top level dependencies, with many of those dependencies also mentioned in nested dependencies. One of those libraries got a major bump, and only half of the users of that library made the bump in their Cartfile. There was a valid permutation to be had, but unfortunately it was using the lower version, and since the resolver processes it in order from highest number to lowest, it had a lot to wade through first.

With the old resolver and this use case, it get up to about 1.5G of memory usage in about 1 minutes, and I actually never saw it finish, but killed it after 10 minutes
With this new implementation, it topped out at 20MB, and finished in about 2 seconds (after fetching).


mdiep:
> Since each permutation generated relies on the state of the previous one (due to needing to skip found incompatibilities), this implementation is much more of a standard recursive implementation, using an error cache to track incompatibilities, and doing away with many of the signal producers, instead using Results. This ensures we're processing exactly one thing at a time and we don't enqueue permutations before we have a chance to vet whether they should have been generated at all.
> 
> That aspect is probably going to be the controversial portion of this, since we are pretty strict about staying stateless. I attempted to mimic the old algorithm as much as possible, and any state maintained is local to the resolver, no global state is used.

That actually sounds great to me. 👍

I'll need some time to dig into this, but I'm very excited to see some improvements here! The current resolver is definitely a big bag of hurt. It was meant to provide correctness, but not necessarily more than that. I'll try to review this next week.

BobElDevil:
Just pushed an update that fixed the unit test toggling I had implemented (it accidentally always ran for the new resolver :/). Also realized a small addition could be made to fix one of the pending tests, with regards to doing invalid updates when given a subset of dependencies to update.

joshhinman:
I have a similar situation with dozens of dependencies, and Carthage was often unable to complete an update before being killed by the system for excess memory/CPU usage. This PR solves my problem. 👍

mdiep:
Would you mind opening a separate PR with the changes to the resolver tests here? I'd like to merge those separately because (a) that makes this PR smaller and (b) then I have the option to add more tests as I review.

BobElDevil:
Sure, opened it up here #2122. I'll rebase this branch once it's merged in

BobElDevil:
Just pushed a rebase onto master

Tests/CarthageKitTests/ResolverSpec.swift
```diff
@@ -259,76 +259,78 @@ class ResolverSpec: QuickSpec {
...
 				}
 
-				// Only the new resolver passes the following tests. Will change to non-pending when checked in
-				pending("should resolve a subset when given specific dependencies that have constraints") {
-					let db: DB = [
-						github1: [
-							.v1_0_0: [
-								github2: .compatibleWith(.v1_0_0),
+				// Only the new resolver passes the following tests.
+				if resolverType == NewResolver.self {
```
BobElDevil: Recommend ignoring whitespace for viewing the diffs in this file, as it's mostly a large indentation

mdiep:
The changes outside of `NewResolver.swift` look good. I'll try to get through the actual new resolver this week.

Source/CarthageKit/NewResolver.swift
```diff
@@ -0,0 +1,552 @@
...
+
+				return versionProducer
+					.map { DependencyNode(dependency: dependency, proposedVersion: $0, versionSpecifier: specifier, parent: parentNode) }
+					.collect()
+					.attemptMap { nodes in
+						guard !nodes.isEmpty else {
+							return .failure(CarthageError.requiredVersionNotFound(dependency, specifier))
+						}
+						return .success(nodes.sorted())
+				}
```
mdiep: There are a bunch of code style violations like the indentation of this `}`. If you install SwiftLint and build, they'll show up as warnings in Xcode.

Source/CarthageKit/NewResolver.swift
```diff
@@ -0,0 +1,552 @@
...
+	/// Error cache for used for this instance of the resolver
+	private let errorCache = ErrorCache()
+
+	/// Produces a lazily computed sequence of valid permutations of `dependencies`
+	/// previously seen errors, in the order they should be tried.
+	///
+	/// In other words, each permutation is an array of DependencyNodes the same length as
+	/// `dependencies`. Each array represents one possible permutation of those
+	/// dependencies (chosen from among the versions that actually exist for
+	/// each).
```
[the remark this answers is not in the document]
BobElDevil: Ah I missed some comment updates, I'll recheck them.

Source/CarthageKit/NewResolver.swift
BobElDevil: Hm, that's odd, I have swiftlint installed, and it was running before... maybe my xcode got confused about when to run the script. I'll take a look at all my indentation when I do a merge for conflicts.

BobElDevil:
@mdiep pushed the requested changes

Source/CarthageKit/NewResolver.swift
```diff
@@ -0,0 +1,557 @@
...
+				}
+
+				var newVersions = graph.versions(for: dependenciesToUpdate)
+				for (dependency, version) in lastResolved {
+					// If we don't have a new version for it, and it's still in the graph, grab the last resolved version
+					guard newVersions[dependency] == nil && graph.node(for: dependency) != nil else { continue }
+					newVersions[dependency] = version
+				}
+
+				return newVersions
```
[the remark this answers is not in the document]
BobElDevil: Actually, since I later added the fix for skipping updates that don't fit the whitelist,  it's even easier. We only need to filter out any new dependencies that didn't exist before in 'validateFinalGraph', then all the whitelist logic can be encapsulated there

BobElDevil:
@mdiep pushed the requested changes

BobElDevil:
Found a regression when specifying an update list, so pushed a fix + unit test for it in the new resolver. Basically I made it so we go back to 'adding' all nodes instead of only new nodes again. Instead I just limit the permutation to the previously pinned version. This way it still is functionally equivalent to the short circuit logic before, but will properly construct the edge set for node whitelisting correctly.

@mdiep can you give this another pass?

dennisbirch:
I've been using the new resolver version for a few weeks on a project with dozens of dependencies, and it works great. In fact, it is the only way I can build many of the components successfully. The release version dies before completing the task of resolving the dependency graph on all but the simplest of our components. I would really like to see this code brought into the release version of Carthage as soon as possible.

mdiep:
This is on my shortlist. However, I just got back from 2 weeks in China where I adopted my son. It will probably be another week or two until I can realistically review this PR. ☺️

HasanEdain:
In a large corporate environment where we have multiple internal dependencies having a resolver that is able to complete (even if it fails) in reasonable time is tremendously important. I would urge you to accept this PR and head towards a more time sensitive carthage.

DCForever:
@mdiep Any chance you or someone else can look again at this?

Source/CarthageKit/NewResolver.swift
```diff
@@ -0,0 +1,555 @@
...
+	fileprivate static func < (_ lhs: DependencyNode, _ rhs: DependencyNode) -> Bool {
+		let leftSemantic = SemanticVersion.from(lhs.proposedVersion).value ?? SemanticVersion(major: 0, minor: 0, patch: 0)
+		let rightSemantic = SemanticVersion.from(rhs.proposedVersion).value ?? SemanticVersion(major: 0, minor: 0, patch: 0)
+
+		// Try higher versions first.
+		return leftSemantic > rightSemantic
+	}
+
+	fileprivate static func == (_ lhs: DependencyNode, _ rhs: DependencyNode) -> Bool {
+		return lhs.dependency == rhs.dependency
```
mdiep: I'm not comfortable with this definition of equality for `DependencyNode`. This should changed to true equality.

Source/CarthageKit/NewResolver.swift
```diff
@@ -0,0 +1,555 @@
...
+		return .success(newGraph)
+	}
+}
+
+/// Custom iterator to perform permutations. Used in tandem with the `ErrorCache', 
+/// allowing us to short circuit permutations before generating them, rather than
+/// filtering results as they come.
+private struct NodePermutations: Sequence, IteratorProtocol {
+	private let baseGraph: DependencyGraph
+	private var permutation: [Dimension]
```
mdiep: This property could use a better name.

Source/CarthageKit/NewResolver.swift
```diff
@@ -0,0 +1,555 @@
...
+				// If it doesn't have a previous version, and isn't in the whitelist, remove it from the returned graph
+				filteredGraph.allNodes.remove(node)
+				filteredGraph.edges.removeValue(forKey: node)
+				continue
+			}
+
+			// If it's not in the whitelist, and it has a previous version, they should match
+			if lastVersion != node.proposedVersion {
+				return .failure(.unsatisfiableDependencyList(whitelist))
+			}
```
mdiep: This check seems to be new in this resolver and I don't quite understand whether (a) this failure would indicate a logic error in Carthage or (b) an incompatibility based on the constraints. Can you elaborate about when this would occur?

BobElDevil: This actually fixes a bug that exists in the current resolver. It's the formerly 'pending' test (which is now wrapped in a resolver type check). (See the changes below in the test class)

The previous resolver would:
- Find a valid graph
- filter out any non-whitelisted nodes from the graph, assigning them back to their 'lastResolved' value

This actually had the potential to result in an invalid graph, that was never subsequently validated (that's actually what the pending test is for). Because the 'lastResolved' may be incompatible with the new values for the nodes that were updated.

Instead, the new resolver has a slight variant, it now:
- Find a valid graph
- filter out any non-whitelisted nodes from the graph, and *validates that their values == the 'lastResolved' value*

Instead of attempting to figure out what the 'right' graph is post-filter, I'm just relying on the permutations to eventually find a graph that only updates the whitelisted nodes.

mdiep: This seems like too strict a check. We should be validating that the nodes that aren't updated are _compatible_ with the version specifiers—not that they resolve to the same versions.

Source/CarthageKit/NewResolver.swift
BobElDevil: renaming to 'currentNodeValues'

Source/CarthageKit/NewResolver.swift
mdiep: That would make for a good additional test.

If one of the dependencies not in the whitelist has a new minor version, updating the whitelist should still succeed.

Source/CarthageKit/NewResolver.swift
BobElDevil: Ok, this is the original comparison in the other resolver. I can make it true equality, but not sure how it'll affect the algorithm. I'll double check everything

Source/CarthageKit/NewResolver.swift
BobElDevil: I'm not sure I get what you're saying. The point of the whitelist is that anything _not_ in it doesn't change at all. That's what this check is doing. It's saying: "If you're not in the whitelist, you should be what you were before"

Source/CarthageKit/NewResolver.swift
mdiep: 🙈

Source/CarthageKit/NewResolver.swift
mdiep: It seems like either:

1. `node.proposedVersion` can only equal `lastVersion` because it's set to that explicitly during resolution. In this case, the check is redundant and could be eliminated.

2. `node.proposedVersion` is a result of normal resolution, in which case it could be e.g. `3.2.4` when `lastVersion = 3.2.3`. If the version specifier for this dependency is `~> 3.2`, then  we're making the whitelist much less useful unnecessarily.

BobElDevil: if we're allowing things outside the whitelist to be updated if they are still compatible with their version specifiers... isn't that not a whitelist? Isn't that functionally equivalent to just running the resolver on everything?

mdiep: I'm not saying update them. I'm saying we should allow the update of the whitelist to complete if one of the items outside the whitelist doesn't exactly match the last resolved version—so long as the last resolved version is compatible.

mdiep: This is obviously a bit confusing. If it'd help, I can put together a test case myself and submit it as a separate PR.

BobElDevil: I think I get it, it's an optimization not a correctness thing. If we come across a permutation that works within the specifiers, we could adjust the graph without breaking the validity of it. 

Currently it would reject it, but eventually 'come around' to the same graph when the permutations catch up to the same result, with matching proposed versions. 

I'll look into it, however that may be a problem due to how the resolution occurs in the first place. I don't "merge" version specifiers anymore when coming across children, I just reject outright if the children version specifier excludes the current permutation. 

What this means is at the point where I'm validating the whitelist, it may appear that the 'lastResolved' is compatible with the version specifier, but in fact it had been rejected by a child

BobElDevil:
@mdiep Made the two other changes

I went and tried to implement the 'just check if it's compatible' change, and I don't think it works, because changing the non-whitelisted dependencies can modify version specifiers of other dependencies.

It actually causes one of the new tests I wrote to fail "should resolve a subset when given specific dependencies that have constraints"

In this test, `github1 v1.0` says it needs a `github2 ~> 1.0` version. `github1 v2.0` says `~> 2.0`. The constraint on github1 itself though is `.any`.

So the first permutation the resolver finds is `github1 v2.0, github2 v2.0`. This is a valid permutation, but we whitelisted github2 only, so we don't want github1 to update. If we then check "Well, was 'lastResolved' still compatible with the version specifier", in this case it is (since we said `.any`). But just setting it to `1.0` is now invalid, because by doing that, we re-constrained `github2` to be `~> 1.0`. So we broke the graph.

So to solve that, we would then have to re-check all other nodes in the graph for validity, which seems to defeat the purpose of this optimization.

mdiep:
I think the key is to treat any existing, non-whitelisted repositories as additional constraints.

If `github1` is pinned to `1.1.1` and we're updating `github2`, then there should be an implicit `github ~> 1.1.1` constraint. We could add that at the root level before resolution begins. That should fix the issue here as well as the other issue that you reported—while keeping the optimization that you've implemented.

I think adding a test for that on `master` would make a good prerequisite for this PR. It could be rigged so that there are new major versions of `github1` and `github2` but the implicit constraint on `github1` keeps resolution on a new minor version and off the new major versions.

What do you think? Do you want to work on a test like that? Or should I?

BobElDevil:
> If github1 is pinned to 1.1.1 and we're updating github2, then there should be an implicit github ~> 1.1.1 constraint. We could add that at the root level before resolution begins.

This actually doesn't work, because our 'full' whitelist includes nested dependencies. So we don't actually know whether a dependency is _not_ in the whitelist  for sure until the graph is complete, since whenever we process a node, it could be adding items to the whitelist.

> I think adding a test for that on master would make a good prerequisite for this PR. It could be rigged so that there are new major versions of github1 and github2 but the implicit constraint on github1 keeps resolution on a new minor version and off the new major versions.

That test already exists, it's the test I brought up in my previous comment. It can be found [here](https://github.com/Carthage/Carthage/blob/master/Tests/CarthageKitTests/ResolverSpec.swift#L270) in master. I enable it in my PR for the new resolver, as it currently passes with the new resolver, but not the old one.

I feel like there may be some confusion here, maybe I should rename this method. When I say 'validateFinalGraph', I really mean 'validatePotentialFinalGraph'. It has to be the last step because of my first point above, we don't know the whitelist until this point. However, if it fails, we continue the search. That's why the test still passes, because by rejecting one that doesn't exactly match the 'lastResolved', the search will continue until it finds one that _does_ exactly match.

mdiep:
> I feel like there may be some confusion here, maybe I should rename this method. When I say 'validateFinalGraph', I really mean 'validatePotentialFinalGraph'. It has to be the last step because of my first point above, we don't know the whitelist until this point. However, if it fails, we continue the search. That's why the test still passes, because by rejecting one that doesn't exactly match the 'lastResolved', the search will continue until it finds one that does exactly match.

Ah, okay! I see that now. Sorry for the confusion. 😓 

Let's go ahead and merge this. Thanks for all the hard work!

PR #2123, opened 2017-08-20
[GitURL] Fix relative path handling in `normalizedURLString`

ikesyo:
Formerly relative paths are parsed as the scp syntax.

### DIFF
--- a/Source/CarthageKit/GitURL.swift
+++ b/Source/CarthageKit/GitURL.swift
@@ -9,13 +9,15 @@ public struct GitURL {
 	/// information. This is mostly useful for comparison, and not for any
 	/// actual Git operations.
 	internal var normalizedURLString: String {
-		let parsedURL: URL? = URL(string: urlString)
-
-		if let parsedURL = parsedURL, let host = parsedURL.host {
+		if let parsedURL = URL(string: urlString), let host = parsedURL.host {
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/GitURL.swift
```diff
@@ -9,13 +9,15 @@ public struct GitURL {
...
-		if let parsedURL = parsedURL, let host = parsedURL.host {
+		if let parsedURL = URL(string: urlString), let host = parsedURL.host {
 			// Normal, valid URL.
 			let path = strippingGitSuffix(parsedURL.path)
 			return "\(host)\(path)"
-		} else if urlString.hasPrefix("/") {
+		} else if urlString.hasPrefix("/") // "/path/to/..."
+			|| urlString.hasPrefix(".") // "./path/to/...", "../path/to/..."
+			|| urlString.hasPrefix("~") // "~/path/to/..."
+			|| !urlString.contains(":") // "path/to/..." with avoiding "git@github.com:owner/name"
```
mdiep: Would it work to use `URL.init(fileURLWithPath:)` here?

ikesyo: `URL.init(fileURLWithPath:)` is not failable, so that will not be help (`git@github.com:owner/name` is valid input).

PR #2125, opened 2017-08-21
Fix fetching a local repository with relative path

### LINKED ISSUE #968: Carthage doesn't appear to "see" latest commit of local repository specified with a relative path
This is kind of an obscure corner case. I have a `Cartfile` that specifies a dependency on a local repo like this:

```
git "../ISUtils" 
```

If I run `carthage update` I get

```
$ carthage update
*** Fetching ISUtils
*** Checking out ISUtils at "1.0.0"
```

I can change the Cartfile to specify the same repo using an absolute path

```
git "file:///Users/bryan/Documents/ISUtils"
```

Now if I run `carthage update` I get

```
$ carthage update
*** Fetching ISUtils
*** Checking out ISUtils at "1.0.1"
```

The second checkout is correct, the latest commit is tagged "1.0.1". The commit before that ("`HEAD~1`") is tagged as "1.0.0".

ikesyo:
Specifically for the case of `../some-repo`. Formerly `cloneOrFetch` fetches a repository itself as its remote in that case because of the given `GitURL` of `../some-repo`.

In short, running the following command within `~/Library/Caches/org.carthage.CarthageKit/dependencies/some-repo` is the problem.

```bash
# ../some-repo is pointing to the repository itself
/usr/bin/env git fetch --prune --quiet ../some-repo refs/tags/*:refs/tags/* +refs/heads/*:refs/heads/*
```

We should be able to use the `origin` remote instead of `../some-repo` here.

Fixes #968.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -1214,11 +1214,6 @@ public func cloneOrFetch(
 
 							return SignalProducer(value: (.fetching(dependency), repositoryURL))
 								.concat(
-									// Don't need to pass `remoteURL` here since the repository should have 
-									// the `origin` remote.
-									//
-									// See https://github.com/Carthage/Carthage/issues/968
-									// and https://github.com/Carthage/Carthage/pull/2125.
 									fetchRepository(repositoryURL, refspec: "+refs/heads/*:refs/heads/*")
 										.then(SignalProducer<(ProjectEvent?, URL), CarthageError>.empty)
 								)
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -1216,6 +1216,9 @@ public func cloneOrFetch(
 								.concat(
 									// Don't need to pass `remoteURL` here since the repository should have 
 									// the `origin` remote.
+									//
+									// See https://github.com/Carthage/Carthage/issues/968
+									// and https://github.com/Carthage/Carthage/pull/2125.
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -1208,13 +1208,18 @@ public func cloneOrFetch(
...
 
 							return SignalProducer(value: (.fetching(dependency), repositoryURL))
 								.concat(
-									fetchRepository(repositoryURL, remoteURL: remoteURL, refspec: "+refs/heads/*:refs/heads/*")
+									// Don't need to pass `remoteURL` here since the repository should have 
+									// the `origin` remote.
+									//
+									// See https://github.com/Carthage/Carthage/issues/968
+									// and https://github.com/Carthage/Carthage/pull/2125.
+									fetchRepository(repositoryURL, refspec: "+refs/heads/*:refs/heads/*")
```
BobElDevil: The only other place we use this function is when we fetch submodules, which we have also set up with the 'origin' remote. It might it make sense to just remove the remoteURL argument and fix up the submodule usage as well.

ikesyo: Thanks @BobElDevil, that makes sense. ✨

mdiep:
Nicely done. :clap

sukeshmlr:
Thanks a lot. This is a much needed fix.

dunqi:
You have to delete the dependency in:
 /Users/username/Library/Caches/org.carthage.CarthageKit/dependencies/nameofframework

And then checkout again - works for me

PR #2126, opened 2017-08-21
[WIP] Validating Cartfile name correctness

### LINKED ISSUE #1336: cartfile name is caseignore
I found that when cartfile file named `cartfile` is also works on carthage update in current project, but when your project is referenced with other project, only `Cartfile` name works for identify dependency.

spookyd:
Fixes #1336.

Currently, validating the Cartfile and Cartfile.private files a case correct and returning a file not found error if they do not match. However, I think it should be discussed whether or not a more specific error should be returned explaining that a Cartfile was discovered but is not properly named. In other words, return an error with a suggestion.

Thoughts?


jdhealy:
Thanks, this is a tough problem — thanks for working on it, glad to see this.

To best message the user, I think we need to inform about all potential problematically-named Cartfiles in one sweep — the frustration level of multiple passes might be too much otherwise…

This means we need to check `git` repositories for misnamed Cartfiles as well as the file-system.

I wasn’t even sure at first how to guide you through writing that code, so I started prototyping, hit a lot of compiler weirdness early on, and figured I'd [develop the basis out upfront and publish the commits](https://github.com/jdhealy/Carthage/commit/6508c508670bcdbccdba323fb6aeec198b4b34a9).

Still a bunch of implementation required on top of that — tests, fixtures, messaging, deciding at what point to actually call the `git`-checking function. Hoping you might be wiling to help with that and also some high-level code review?

@spookyd I've added you as a collaborator on jdhealy/Carthage so you can push to [that branch](https://github.com/jdhealy/Carthage/tree/message-user-misnamed-Cartfiles) and we can figure out whether to rebase this or open a new PR — sounds good?

Tests/CarthageKitTests/ProjectSpec.swift
```diff
@@ -246,6 +246,19 @@ class ProjectSpec: QuickSpec {
 					fail()
 				}
 			}
+
```
jdhealy: Thanks for taking the time to add tests and fixtures! ☀️

spookyd:
@jdhealy I am willing to pitch in and help tackle this issue. 

As for the traversal of the `git` file validation, I am wondering if that is necessary. For anyone using Carthage they will have this validation upfront for their project. I guess the assumption would be for any given dependency also using Carthage they would have ran through the same upfront validation. I maybe missing the bigger issue so please let me know.

Tests/CarthageKitTests/ProjectSpec.swift
spookyd: 😄

mdiep:
> As for the traversal of the `git` file validation, I am wondering if that is necessary. For anyone using Carthage they will have this validation upfront for their project. I guess the assumption would be for any given dependency also using Carthage they would have ran through the same upfront validation. I maybe missing the bigger issue so please let me know.

I agree. It's already an error if the Cartfile isn't `Cartfile` in a git repository. This PR is focused an consistency between building the local project and building it as a dependency.

It would _also_ be nice to have a better error for the git case, but I think that's a separate issue.

Source/CarthageKit/Project.swift
```diff
@@ -145,6 +144,23 @@ public final class Project { // swiftlint:disable:this type_body_length
...
 
+		// Because the file system Carthage is interacting with may be case sensitive a validation must be
+		// performed to ensure proper file naming. See: https://github.com/Carthage/Carthage/issues/1336
+		func isValidProjectFiles() -> Bool {
+			let workingDirectory = (try? FileManager.default.contentsOfDirectory(at: directoryURL, includingPropertiesForKeys: [], options: .skipsHiddenFiles)) ?? []
+
+			func isValidFile(name: String) -> Bool {
+				let cartfileURL = directoryURL.appendingPathComponent(name)
+				if FileManager.default.fileExists(atPath: cartfileURL.path) {
+					return workingDirectory.filter({ $0.lastPathComponent == name }).count == 1
```
mdiep: This could use `first(where:) != nil`, which would be more efficient.

Source/CarthageKit/Project.swift
```diff
@@ -145,6 +144,23 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+				}
+				return true
+			}
+			// DISCUSS: This could cause some issues. Based on the name `Constants.Project.cartfilePath` it suggests that
+			// the string is a path and not a file name.
```
mdiep: I think it's fine to rename the constants to specify that they're filenames.

jdhealy:
> already an error if the Cartfile isn't Cartfile in a git repository

Not the case if the any of the misnamed Cartfiles is alongside a correctly-cased `./Cartfile`, which would create inconsistency between local project and as dependency.

To have likelihood of this broken project existing, it would probably be created on a case-sensitive file-system — [not the default for APFS conversion in 10.13](https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/APFS_Guide/FAQ/FAQ.html), but we'll likely see some users who have opted for case-sensitivity anyways.

Source/CarthageKit/Project.swift
spookyd: All the cartfilePath constants contain docs specifying that they are relative paths to the project. If this is truly intent I would hate to change it to a file name. From searching the project the only thing I could see being a problem later is if the location of the Cartfile is no longer at project root. I suppose we could deal with that if it ever came up. 

I only want to make sure the intention is clear for what the constant is supposed to represent.

spookyd:
@jdhealy So for anyone using github as their dependency management repo they will already have dealt with resolving this issue because the file system is case sensitive. Is your concern more on the side of people that use their own internal repos as their dependency management repo?

Do you feel as if this change would break the latter group? If not, I would like to think we could push this request through, solving the current problem. And then we could open another PR where we could work on solving the other issue. But if you think this change would put the latter group in a bad situation then I agree with addressing your concerns now.

Source/CarthageKit/Project.swift
jdhealy: Yeah, good looking out, we could comment and use `assert` (which gets compiled out in Release)`.

```swift
[Constants.Project.cartfilePath, Constants.Project.privateCartfilePath]
	.some_function { filename in // relative path used here as file-system resource name
		assert(filename.pathComponents.count == 1)
		…
```

jdhealy: I should ask as well, `Constants.Project.resolvedCartfilePath` should be checked too, right?

Source/CarthageKit/Project.swift
```diff
@@ -145,6 +144,23 @@ public final class Project { // swiftlint:disable:this type_body_length
 		let cartfileURL = directoryURL.appendingPathComponent(Constants.Project.cartfilePath, isDirectory: false)
 		let privateCartfileURL = directoryURL.appendingPathComponent(Constants.Project.privateCartfilePath, isDirectory: false)
 
+		// Because the file system Carthage is interacting with may be case sensitive a validation must be
+		// performed to ensure proper file naming. See: https://github.com/Carthage/Carthage/issues/1336
+		func isValidProjectFiles() -> Bool {
```
jdhealy: maybe more like `func validateFilenamesOfCartfiles`?

jdhealy:
> for anyone using github as their dependency management repo they will already have dealt with resolving this issue because the file system is case sensitive.

Not sure I completely understand here, do you mean a user’s file-system on macOS by default being case-insensitive? Not anything GitHub in particular enforces?

The main issue I see that (admittedly) very few people will likely hit in practice, is that for a Carthage user on a case-sensitive file-system for a repository that only has misnamed Cartfiles, 
`FileManager.default.fileExists(atPath: cartfileURL.path)` returns `false` and the directory contents checking doesn’t happen.

To your point, GitHub listing the root directory contents on the repo’s page might just provide enough context for someone previously inclined to opt in to file-system case-sensitivity — so, all considered, us not implementing the checking with `git ls-tree` is okay by me.

So, next up for us: figure out the user-facing messaging… 👍

Source/CarthageKit/Project.swift
spookyd: Not sure `Constants.Project.resolvedCartfilePath` needs checked since Carthage manages that file. Unless we want to validate that the file has not been tampered with. But I think this scenario would be exposed if `carthage bootstrap` was ran after the file was tampered with.

So, I guess we would need to understand the workflow and at what points this could occur. But again, this scenario means that a user is going in and deliberately messing with the `resolvedCartfilePath` file. And in that case we shouldn't try and solve for people being stupid. Just my opinion...

Source/CarthageKit/Project.swift
spookyd: Yeah naming this function is a little weird but I do like your suggestion better.

spookyd:
> Oh, meaning git handles files case-sensitively (in a broad sense)?
Yes.

As for user-facing messaging, we could simply state something like:
"<Cartfile/ Cartfile.private> not found. However, we detected a file named <cartfile/ cartfile.private>. Did you mean <Cartfile/ Cartfile.private>?"

Or we could add something to suggest changing the name and running the command again.

Source/CarthageKit/Project.swift
jdhealy: > But again, this scenario means that a user is going in and deliberately messing with the resolvedCartfilePath file. And in that case we shouldn't try and solve for people being stupid.

My concern is more bad tooling/scripts messing with filename case — even smart people frequently fail to vet those comprehensively.

jdhealy:
> As for user-facing messaging, we could simply state something like:
> "<Cartfile/ Cartfile.private> not found. However, we detected a file named <cartfile/ cartfile.private>. Did you mean <Cartfile/ Cartfile.private>?"

- Preferable (in my opinion) to avoid “we”/“you” in our messaging. 
- The messaging should also probably include some form of the phrase “case-sensitive”.
- We should include absolute `URL`s to found misnamed Cartfiles — helpful when part of larger CI runs.
- Don’t limit the message to a single misnamed Cartfile. To achieve this, I’d suggest:

If any of the Cartfile paths get `true` from `FileManager.fileExists(atPath:)`, then iterate over `contentsOfDirectory` (once), checking all three Cartfile paths for misnamed ones, creating a single `CarthageError` with a `description` that lists every found misnamed Cartfile.

> we could add something to suggest changing the name

This is likely the hardest part of what we have to message: suggesting to change the name to fix the problem — not just for themselves (on disk) — but, for everybody checking out some future version of the project (meaning publishing changes in `git`, remembering the receiver of the message doesn’t necessarily control the repo).

Source/CarthageKit/Project.swift
spookyd: I am fine with adding in the check for the resolved file. But does it makes sense within the workflow to have the resolved file checked at `loadCombinedCartfiles`?

spookyd:
Yeah I like those suggestions.
Do you think that we need to move the validation to another step in the process? Thinking about avoiding the validation of Cartfiles of our dependencies. And just validating the working project's Cartfiles. To your point, if we validate all Cartfiles a.) what can/ should we do if a dependency's Cartfile fails, and b.) how would we know we are validating the working project's Cartfiles versus the dependency's Cartfile?

spookyd:
Sorry, I got really busy at work... @jdhealy, so where do we stand on this one?

blender:
@spookyd @jdhealy ping

blender:
@spookyd  confident it's desirable :)

PR #2133, opened 2017-08-23
Update resolver signature and update/add tests

BobElDevil:
Precursor to #2122 

I'd recommend viewing the diff with `?w=1`, since an extra level of indentation got added to the tests, but the differences are not actually that large.


mdiep:
This is fantastic! Thanks for opening a separate PR, that made reviewing much easier. ✨

Issue #1853, opened 2017-03-28
How to refer json file placed locally for binary?


shahdhiren:
* `0.20.1`: 
* `Xcode 8.2.1`: 
* Are you using `--no-build`?   NO
* Are you using `--no-use-binaries`?  NO
* Are you using `--use-submodules`?  NO
* Are you using `--cache-builds`?  NO

**Cartfile**
```
binary "myFolder/Library1.json" ~> 1.2
binary "myFolder/Framework1.json" ~> 1.1

```

**Carthage Output**
```
Parse error: non-https URL found for dependency type `binary` in line: Optional("binary \"myFolder/Library1.json\" ~> 1.2")

```

Is it possible to refer json file in local path?

I am referring https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md.

mdiep:
That is not currently possible. Do you have a use case in mind?

shahdhiren:
@mdiep: So right now, we have ready binaries placed on our internal artifact server. Till now we were using Cocoapods to resolve dependency. So we were placing podspec files in our local project directory.

Say if I have to use that library from artifact server which does not contain the json files, then I can create on my own in local machine and I can resolve the dependency. I may not be having permission to upload my json to Artifact server or any other server. (Also this is time consuming process.)

mdiep:
I think it makes sense to relax the requirement to allow local file URLs alongside HTTPS URLs. 👍

shahdhiren:
How long will it take to release with that feature? For now, I have to remove Carthage for my dependency.

mdiep:
That depends… How soon can you open a pull request to enable it? ☺️

Stoff81:
Hi, Im working on a PR for this now, what should the checks for a valid url be? I could make it so that if the first character is "/" then its valid?

Stoff81:
I was thinking of maybe having something like this:

New folder alongside Build/Checkouts called Binary Specs which contains all the json files that you need. So for example if you wanted to include the Fabric SDK, there would be a file "Carthage/Binary Specs/fabric.json". This json would be as spec'ed in the docs here.

This would then be referenced in the Carfile like so:

`binary "spec://fabric.json" ~> 1.6.9`

What do you think? Please let me know if you have a better idea, I am new to Carthage.

shahdhiren:
@Stoff81: From your above comment, I am not able to get where you want to place the json file?

[...]

PR #2142, opened 2017-09-03
[BinaryProject] Add `file` scheme support in `Dependency`  #1853

phimage:
Hi,

I look at PR #2140 for #1853 and see that I can help
Like requested in #1853, json file with this PR could be accessed through file URL

I make unit test, and also see that downloading the json file is already tested, because the one used in test (successful.json) is in `Bundle`

### DIFF
--- a/Source/CarthageKit/Cartfile.swift
+++ b/Source/CarthageKit/Cartfile.swift
@@ -286,10 +286,10 @@ extension Dependency: Scannable {
 		} else if scanner.scanString("binary", into: nil) {
 			parser = { urlString in
 				if let url = URL(string: urlString) {
-					if url.scheme == "https" {
+					if url.scheme == "https" || url.scheme == "file" {
 						return .success(self.binary(url))
 					} else {
-						return .failure(ScannableError(message: "non-https URL found for dependency type `binary`", currentLine: scanner.currentLine))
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Cartfile.swift
```diff
@@ -286,10 +286,10 @@ extension Dependency: Scannable {
 		} else if scanner.scanString("binary", into: nil) {
 			parser = { urlString in
 				if let url = URL(string: urlString) {
-					if url.scheme == "https" {
+					if url.scheme == "https" || url.scheme == "file" {
 						return .success(self.binary(url))
 					} else {
-						return .failure(ScannableError(message: "non-https URL found for dependency type `binary`", currentLine: scanner.currentLine))
+						return .failure(ScannableError(message: "non-https or non-file URL found for dependency type `binary`", currentLine: scanner.currentLine))
```
mdiep: Can you change this to say `non-https, non-file URL`?

`non-https or non-file URL` makes it sound like the the URL needs to be both https and a file URL.

mdiep:
Thanks for the pull request! ✨ 

This is great overall, but I'd like to tweak the error message.

phimage:
It's done, unit test also updated

mdiep:
🎉

PR #2151, opened 2017-09-09
[gardening] Utilize `filterMap` and `flatMap` for map-then-filter (or filter-then-map) patterns

ikesyo:


ikesyo:
Hmm reverted some changes.

Source/CarthageKit/Xcode.swift
```diff
@@ -368,15 +369,13 @@ private func mergeBuildProductsIntoDirectory(
...
+				.filterMap { modulesPath, productsURL in
+					return modulesPath.map { productsURL.appendingPathComponent($0) }
 				}
 
 			let destinationModulesURL = SignalProducer(result: firstProductSettings.relativeModulesPath)
-				.filter { $0 != nil }
-				.map { modulesPath -> URL in
-					return destinationFolderURL.appendingPathComponent(modulesPath!)
+				.filterMap { modulesPath in
+					return modulesPath.map { destinationFolderURL.appendingPathComponent($0) }
```
mdiep: These two make sense to me since a separate filter step requires `!`, but I think the other examples are clearer with a separate filter.

PR #2164, opened 2017-09-20
Migrate to Xcode 9

ikesyo:


Carthage.xcodeproj/project.pbxproj
```diff
@@ -902,7 +902,11 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = D0D1212619E878CC005E4BAA /* Debug.xcconfig */;
 			buildSettings = {
+				CLANG_WARN_BLOCK_CAPTURE_AUTORELEASING = YES;
```
NachoSoto: It drives me insane that Xcode puts these in each target instead of the project level.

ikesyo:
:shipit:

PR #2172, opened 2017-09-22
[VersionFile] Use `Codable`

ikesyo:


Source/CarthageKit/VersionFile.swift
```diff
@@ -1,45 +1,36 @@
...
 }
 
-extension CachedFramework: Argo.Decodable {
-	static func decode(_ json: JSON) -> Decoded<CachedFramework> {
-		return curry(self.init)
-			<^> json <| CachedFramework.nameKey
-			<*> json <| CachedFramework.hashKey
+struct VersionFile: Codable {
+	enum CodingKeys: CodingKey, String {
+		case commitish = "commitish"
```
lightsprint09: is there a reason why you duplicate the case into a string when it is the same? Like:
```swift 
case commitish = "commitish"
```
instead of 
```swift 
case commitish 
```

ikesyo: I think the explicitness is a virtue and durable against some refactorings (renaming).

lightsprint09: Ok. Makes sense

Tests/CarthageKitTests/VersionFileSpec.swift
```diff
@@ -1,7 +1,6 @@
 import Foundation
 import Quick
 import Nimble
-import Argo
```
NachoSoto: Woot! I imagine you're keeping the dependency cause other types still use it?

ikesyo: Yes, Tentacle depends on Argo.

PR #2178, opened 2017-09-23
Migrate to Swift 4

ikesyo:


ikesyo:
Let's wait for #2180.

ikesyo:
It should be ready now.

PR #2185, opened 2017-09-28
Avoid name collision on fetching with appending the hash value of a dependency

### LINKED ISSUE #2143: Version 0.25.0 downgrades frameworks
* `carthage version`: 0.25.0
* `xcodebuild -version`: Xcode 8.3.3, Build version 8E3004b
* Are you using `--cache-builds`? Yes

**Cartfile** (removed entries with no sub dependencies that didn't cause problems)
```
# AKSideMenu is a double side menu library with parallax effect.
github "Dschee/InteractiveSideMenu" "master"

# Network abstraction layer written in Swift.
github "Moya/Moya" ~> 9.0

# Convenient logging during development & release in Swift
github "SwiftyBeaver/SwiftyBeaver" ~> 1.1

# Swifty Sound is a simple library that lets you play sounds with a single line of code.
github "adamcichy/SwiftySound" ~> 0.6

# Drop-in replacement for UISlider with ticks, images & animated labels
github "SwiftArchitect/TGPControls" ~> 4.0
```

**Carthage Output** (command: `carthage update --platform iOS --cache-builds`)
```
*** Fetching InteractiveSideMenu
*** Fetching SwiftyBeaver
*** Fetching SwiftySound
*** Fetching TGPControls
*** Fetching Moya
*** Fetching Result
*** Fetching Alamofire
*** Fetching ReactiveSwift
*** Fetching RxSwift
*** Checking out Moya at "9.0.0-alpha.1"
*** Checking out InteractiveSideMenu at "ac47329a2c8d77e6483fde4d4a5cd13f6f5546d4"
*** Checking out Alamofire at "4.5.0"
*** Checking out TGPControls at "v4.0.0"
*** Checking out SwiftySound at "0.6.0"
*** Checking out ReactiveSwift at "2.0.1"
*** Checking out SwiftyBeaver at "1.3.3"
*** Downloading RxSwift.framework binary at "Oxygen.1"
*** Checking out Result at "3.2.3"
*** xcodebuild output can be found in /var/folders/2l/3tqwt0g91x18hql0jwsz0y280000gn/T/carthage-xcodebuild.2BQ3oQ.log
*** Valid cache found for Alamofire, skipping build
*** Valid cache found for InteractiveSideMenu, skipping build
*** Valid cache found for Result, skipping build
*** Valid cache found for ReactiveSwift, skipping build
*** Valid cache found for RxSwift, skipping build
*** Valid cache found for Moya, skipping build
*** Valid cache found for SwiftyBeaver, skipping build
*** Valid cache found for SwiftySound, skipping build
*** Valid cache found for TGPControls, skipping build
```

My problem is that since I have Carthage 0.25.0 installed a `carthage update --platform ios --cache-builds` downgrades some dependencies to previous versions. I'm not sure if this is an issue with the GitHub API or with Carthage, but I recognized this behavior and my colleagues who upgraded to Carthage 0.25.0 did see the same issue, too. We are now downgrading to 0.24.0 but this should probably be fixed.

Here's an extract of the changed `Cartfile.resolved` (red lines are deleted lines, green lines are the new ones) after running `carthage update`. As you can see several versions have downgraded.

bildschirmfoto 2017-09-05 um 13 01 27

For one of my fellows, as an example, SwiftySound downgraded from 0.6.0 to 0.2.0. So it seems it doesn't behave the same for everybody ...

### LINKED ISSUE #2150: Cache failing to resolve dependencies with name collision
Carthage keeps a cache of fetched dependencies in `~/Library/Caches/org.carthage.CarthageKit/dependencies/…`.

Assuming you have a project with this in its `Cartfile`:

```
github "antitypical/Result"
```

And another project with this:

```
github "somebody/Result"
```

Then due to not using any further subdirectories for `antitypical` or `somebody` carthage ends up fetching the wrong dependency from the local cache for the second project being bootstrapped.

This pretty much defeats the whole point of supporting namespacing in the first place. 😢

ikesyo:
Should fix #2150. Should fix #2143.

This is inspired by https://github.com/apple/swift-package-manager/blob/98cfe1cb24573341488be985fb28d6fe9ecf8920/Sources/SourceControl/Repository.swift#L23-L33.

### DIFF
--- a/Source/CarthageKit/Dependency.swift
+++ b/Source/CarthageKit/Dependency.swift
@@ -7,12 +7,15 @@ import Tentacle
 /// Uniquely identifies a project that can be used as a dependency.
 public enum Dependency {
 	/// A repository hosted on GitHub.com or GitHub Enterprise.
+	/// This is treated as case-insensitive.
 	case gitHub(Server, Repository)
 
 	/// An arbitrary Git repository.
+	/// This is treated as case-sensitive.
 	case git(GitURL)
 
 	/// A binary-only framework
+	/// This is treated as case-sensitive.
 	case binary(URL)
 
 	/// The unique, user-visible name for this project.
--- a/Source/CarthageKit/Dependency.swift
+++ b/Source/CarthageKit/Dependency.swift
@@ -1,5 +1,7 @@
 import Foundation
 import Result
+import ReactiveSwift
+import ReactiveTask
 import Tentacle
 
 /// Uniquely identifies a project that can be used as a dependency.
@@ -28,8 +30,20 @@ public enum Dependency {
 	}
 
 	/// The unique identifier for this project which is useful as a file system path.
-	public var fileSystemIdentifier: String {
-		return name + "-" + String(hashValue)
+	public var fileSystemIdentifier: SignalProducer<String, CarthageError> {
+		let task = Task("/usr/bin/shasum", arguments: ["-a", "256"])
+		return task
+			.launch(standardInput: .init(value: _hashable.data(using: .utf8)!))
+			.ignoreTaskData()
+			.mapError(CarthageError.taskError)
+			.map { data -> String in
+				let output = String(data: data, encoding: .utf8)!
+				let sha = output
+					.components(separatedBy: CharacterSet.whitespaces)[0]
+					.trimmingCharacters(in: .whitespacesAndNewlines)
+
+				return self.name + "-" + sha
+			}
 	}
 
 	/// The path at which this project will be checked out, relative to the
@@ -97,17 +111,21 @@ extension Dependency: Comparable {
 }
 
 extension Dependency: Hashable {
-	public var hashValue: Int {
+	fileprivate var _hashable: String {
 		switch self {
 		case let .gitHub(server, repo):
-			return server.hashValue ^ repo.hashValue
+			return server.url(for: repo).absoluteString.lowercased()
 
 		case let .git(url):
-			return url.hashValue
+			return url.normalizedURLString
 
 		case let .binary(url):
-			return url.hashValue
+			return url.absoluteString
 		}
+	}
+
+	public var hashValue: Int {
+		return _hashable.hashValue
 	}
 }
 
--- a/Source/CarthageKit/Dependency.swift
+++ b/Source/CarthageKit/Dependency.swift
@@ -27,6 +27,11 @@ public enum Dependency {
 		}
 	}
 
+	/// The unique identifier for this project which is useful as a file system path.
+	public var fileSystemIdentifier: String {
+		return name + "-" + String(hashValue)
+	}
+
 	/// The path at which this project will be checked out, relative to the
 	/// working directory of the main project.
 	public var relativePath: String {
@@ -35,7 +40,7 @@ public enum Dependency {
 }
 
 extension Dependency {
-	fileprivate init(gitURL: GitURL) {
+	public init(gitURL: GitURL) {
 		let githubHostIdentifier = "github.com"
 		let urlString = gitURL.urlString
 
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -364,9 +364,8 @@ public final class Project { // swiftlint:disable:this type_body_length
 
 	/// Attempts to resolve a Git reference to a version.
 	private func resolvedGitReference(_ dependency: Dependency, reference: String) -> SignalProducer<PinnedVersion, CarthageError> {
-		let repositoryURL = repositoryFileURL(for: dependency)
 		return cloneOrFetchDependency(dependency, commitish: reference)
-			.flatMap(.concat) { _ in
+			.flatMap(.concat) { repositoryURL in
 				return resolveTagInRepository(repositoryURL, reference)
 					.map { _ in
 						// If the reference is an exact tag, resolves it to the tag.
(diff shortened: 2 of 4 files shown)


jdhealy:
Two areas of thought on this:

1. Apple’s guidance [from Hashable’s documentation](https://developer.apple.com/documentation/swift/hashable): <blockquote><strong>Important</strong>
Hash values are not guaranteed to be equal across different executions of your program. Do not save hash values to use in a future execution.</blockquote>Maybe our use of it (likewise with Swift Package Manager’s) is exceptional in regards to this, @ikesyo? Pulling in a hashing library might give us more dependable behavior, at the expense of having to write code for each property (so, some overhead).

2. Case-sensitivity (filesystems, GitHub request URL’s, GitHub response URL’s): what we are currently relying on, what this PR would rely on, what we can realistically support — something (given some free time) I hope to reexamine cross-project. Eventually and ideally, we would have documentation above `Dependency` regarding case-sensitivity.

ikesyo:
> Maybe our use of it (likewise with Swift Package Manager’s) is exceptional in regards to this, @ikesyo? Pulling in a hashing library might give us more dependable behavior, at the expense of having to write code for each property (so, some overhead).

@jdhealy How does this look like?

cce9143

ikesyo:
> Case-sensitivity (filesystems, GitHub request URLs, GitHub response URLs): what we are currently relying on, what this PR would rely on, what we can realistically support — something (given some free time) I hope to reexamine cross-project. Eventually and ideally, we would have documentation above Dependency regarding case-sensitivity.

@jdhealy I'm sorry but I assume you would do the job, isn't it? Please let me know your knowledge about this atm and I may succeed that.

Source/CarthageKit/Dependency.swift
```diff
@@ -1,16 +1,21 @@
...
 import Result
+import ReactiveSwift
+import ReactiveTask
 import Tentacle
 
 /// Uniquely identifies a project that can be used as a dependency.
 public enum Dependency {
 	/// A repository hosted on GitHub.com or GitHub Enterprise.
+	/// This is treated as case-insensitive.
 	case gitHub(Server, Repository)
```
ikesyo: Both `Server` and `Repository` are case-insensitive for equality:

- https://github.com/mdiep/Tentacle/blob/90c57572d15c56d591e671bbe4fc37f7288df060/Sources/Tentacle/Server.swift#L54-L65
- https://github.com/mdiep/Tentacle/blob/90c57572d15c56d591e671bbe4fc37f7288df060/Sources/Tentacle/Repository.swift#L28-L31

Source/CarthageKit/Dependency.swift
```diff
@@ -1,16 +1,21 @@
...
 
 	/// An arbitrary Git repository.
+	/// This is treated as case-sensitive.
 	case git(GitURL)
```
ikesyo: `GitURL`'s equality is based on `normalizedURLString` and there is no case-insensitive comparison or case conversion in that implementation.

- https://github.com/Carthage/Carthage/blob/a3573d94f97458473b2d8ae0282cd6d34e9908a2/Source/CarthageKit/GitURL.swift#L8-L46
- https://github.com/Carthage/Carthage/blob/a3573d94f97458473b2d8ae0282cd6d34e9908a2/Source/CarthageKit/GitURL.swift#L64-L66

Source/CarthageKit/Dependency.swift
```diff
@@ -1,16 +1,21 @@
...
 
 	/// A binary-only framework
+	/// This is treated as case-sensitive.
 	case binary(URL)
```
ikesyo:
```swift
import Foundation

URL(string: "https://google.com") == URL(string: "https://google.COM") // false
```

Source/CarthageKit/Dependency.swift
ikesyo: https://github.com/Carthage/Carthage/pull/2185/files#diff-d36e2d0823f4b902aeb9274469b117c0R120

Source/CarthageKit/Dependency.swift
ikesyo: https://github.com/Carthage/Carthage/pull/2185/files#diff-d36e2d0823f4b902aeb9274469b117c0R123

Source/CarthageKit/Dependency.swift
ikesyo: https://github.com/Carthage/Carthage/pull/2185/files#diff-d36e2d0823f4b902aeb9274469b117c0R126

mdiep:
This would definitely fix the issue, but I don't think it's the best approach from an end-user perspective. Switching to a fork would require a full clone of the repository.

I think a better solution would be to always use the same repository for a given name, but to use a different git remote for each possible origin. For GitHub.com repositories, the name of the remote could just be the name of the owner. For other repositories, we could do essentially what you've done here.

The origin approach would share the git object database across all forks, leading to faster clones and saved disk space. It would also preserve everyones' caches when they upgrade to get this PR.

What do you think @ikesyo? Does that seem workable? If so, what would you think of pursuing that instead?

Source/CarthageKit/Dependency.swift
```diff
@@ -27,6 +32,23 @@ public enum Dependency {
 		}
 	}
 
+	/// The unique identifier for this project which is useful as a file system path.
+	public var fileSystemIdentifier: SignalProducer<String, CarthageError> {
+		let task = Task("/usr/bin/shasum", arguments: ["-a", "256"])
```
mdiep: I hate shelling out for things like this. But I guess there's probably not an easy Swift option. 😞

ikesyo:
Hmm so let’s revert #2125.

ikesyo:
Reopened and rebased on current `master`.

mdiep:
@ikesyo What do you think of the idea I described in https://github.com/Carthage/Carthage/pull/2286#issuecomment-353429713?

> I think that the best solution here is to move the cache to ~/Library/Caches/org.carthage.CarthageKit/repositories/{ProjectName}/{OwnerOrURL}. That cleans things up quite a bit and makes the cache more navigable and debuggable.

I haven't found Xcode's derived data folder, with hash-appended directories, to be very user-friendly. And although I'm sure the Xcode team would say that you don't need to navigate there, I've found that I often _do_ need too. In other words, there's prior-art for using the hash-appended directory name, but I don't think it's particularly good art.

ikesyo:
https://github.com/Carthage/Carthage/pull/2185#issuecomment-354851146

I'd like to revisit that direction.

PR #2194, opened 2017-10-05
ReadMe: Upgrading frameworks automatically.

SeanRobinson159:
Added a new section in the ReadMe for automatically updating all dependencies on each build.

This is helpful for a one time setup of Carthage. After you set up your Cartfile and the run script, you don't have to think about it again. It just works. Keeps your dependencies updated. Helpful for the initial checkout of a project, since it will automatically get the dependencies. No need to try and build the project in Xcode, finding all sorts of build errors, then have to pull out the command line, figure out where the project file is located, then run `carthage update`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,20 @@ or
 carthage update Box Result
 ```
 
+### Upgrading frameworks automatically
+
+If you want to make sure your dependencies are updated on each build, you can add a Run Script build phase that invokes Carthage like so:
+
+```sh
+{
+    /usr/local/bin/carthage update --cache-builds --platform "$PLATFORM_NAME" --project-directory "$SRCROOT"
+} || {
+    echo "warning: Couldn't update Carthage automatically. Check your internet connection."
+}
+```
+Now you don't need to use the command line to update your dependencies. 
+This is especially helpful when your dependencies are updated frequently, and when multiple people are working on the same project. Now they don't all need to remember to run `carthage update`. Even the first time they checkout the repository.
+
 ### Nested dependencies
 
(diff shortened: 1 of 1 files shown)


mdiep:
Thanks for opening a PR!

I wouldn't recommend running `update` on each build. This could lead to inconsistent behavior when building the same revision at different points in time. That can make it harder to pinpoint regressions, among other things. As a result, I'm 👎 on this addition to the README.

SeanRobinson159:
To me, it seems like that is what version numbers are for. 
What would you suggest for keeping the frameworks updated/synced between team members (And their multiple computers) and CI machines?

dcaunt:
Bootstrap would be more appropriate, but maybe something like [carthage-verify](https://github.com/Carthage/workflows#carthage-verify) meets your needs?

mdiep:
Yeah, it sounds like you actually want to be using `carthage bootstrap`.

SeanRobinson159:
`bootstrap` would work as well, but then you still have to use the command line to update frameworks if you change the cartfile. I am just trying to make it so you don't have to use the command line. 

I add the cartfile to my Xcode project, so that I can change/add frameworks easily. I see that as the preferred way of handling the cartfile. But after adding/changing a framework, it would be great if you could just build your Xcode project, and it would get or update that dependency. No need to open Terminal. 

So I guess some sort of hybrid of `bootstrap` and `update` would work better for this situation.

Does what I am trying to accomplish make sense? It seemed like `update` was the best solution. But I do understand the concern of inconsistent results at different points in time, if the frameworks resolve to different versions.

mdiep:
> bootstrap would work as well, but then you still have to use the command line to update frameworks if you change the cartfile. I am just trying to make it so you don't have to use the command line.

If you don't want to use the command line, I think you'd need to create a separate target/scheme that runs `carthage update`.

SeanRobinson159:
Though I love this idea, the cons outweigh the pros.

PR #2213, opened 2017-10-13
Fix CocoaPods-generated projects support when running `archive` action

ikesyo:
Such projects have unexpected `CONFIGURATION_BUILD_DIR` build setting which is appending a target name to a common `CONFIGURATION_BUILD_DIR` value. That is problematic for our path building.

The issue is found at https://github.com/AliSoftware/Reusable/issues/50#issuecomment-336434345.

Ref: #2158


ikesyo:
Thanks!

PR #2217, opened 2017-10-14
Transitive binary dependencies have invalid symlink

tbrannam:
Binary dependencies are a helpful addition to Carthage.   But for projects that prefer to use `--no-build` or `--no-use-binaries` there are issues with integration and building.

Given a project with a primary dependency to DependencyA  (a normal source dependency from git) which in turn has a binary only dependency.  The primary dependency typically would want to link to the binary-only dependency (DependencyB) that is resolved to $root/Carthage/Build/* 

Currently the binary-only asset is being properly downloaded and placed in $root/Carthage/Build/, but the symbolic reference for $root/Carthage/Checkouts/DependencyA is being created as $root/Carthage/Checkouts/DependencyA/Carthage/Checkouts/DependencyB.  Additionally the symbolic link DependencyB target is not resolvable.

This patch updates the creation of symbolic links for binary dependencies such that 
$root/Carthage/Checkouts/DependencyA/Carthage/Build => $root/Carthage/Build

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -693,6 +693,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 				}
 
 				let symlinkCheckoutPaths = self.symlinkCheckoutPaths(for: dependency, version: version, withRepository: repositoryURL, atRootDirectory: self.directoryURL)
+				let symlinksBinaryPaths = self.symlinkBinaryPaths(for: dependency, version: version, withRepository: repositoryURL, atRootDirectory: self.directoryURL)
 
 				if let submodule = submodule {
 					// In the presence of `submodule` for `dependency` — before symlinking, (not after) — add submodule and its submodules:
@@ -700,10 +701,12 @@ public final class Project { // swiftlint:disable:this type_body_length
 					return addSubmoduleToRepository(self.directoryURL, submodule, GitURL(repositoryURL.path))
 						.startOnQueue(self.gitOperationQueue)
 						.then(symlinkCheckoutPaths)
+						.then(symlinksBinaryPaths)
 				} else {
 					return checkoutRepositoryToDirectory(repositoryURL, workingDirectoryURL, revision: revision)
 						// For checkouts of “ideally bare” repositories of `dependency`, we add its submodules by cloning ourselves, after symlinking.
 						.then(symlinkCheckoutPaths)
+						.then(symlinksBinaryPaths)
 						.then(
 							submodulesInRepository(repositoryURL, revision: revision)
 								.flatMap(.merge) {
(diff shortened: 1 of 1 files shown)


mdiep:
Thanks for opening a pull request! It's always amazing when someone opens a PR after finding an issue. ✨ 

I don't have quite as much time for OSS right now, so it may take me some time to review this PR. Hopefully one of the other maintainers will beat me to it. But if not, please know that the PR is appreciated and _will_ be reviewed!

tbrannam:
The dirty state of `Carthage/Build` didn't particularly occur to me.     And I appreciate the rock and hard place statement.   I certainly can split the filtering of binary dependencies as a separate PR.  

With regards the the meat of this PR, `binary` dependencies clearly fall into an inconvenient space for folks that are using submodules.    Given that there seems to only be two ways of thinking about how dependencies are managed, it does seem that a third path needs to be introduced.  

It is unfortunate that transitive dependencies can introduce workflow complications as their direct dependents evolve over time.

I think I understand your comment about `git check-ignore` - as a linting behavior be an appropriate solution as well, as binary dependencies are found - they could be presented to the developer during update/bootstrap - and the state of .gitignore could be evaluated to determine if the binary dependency is or isn't being hidden by existing rules.     Could you explain a little more about how the issue with bare repos would cause?

Writing to Carthage/Build seems like an expected behavior of `binary` dependencies anyway, and if this was a primary dependency - I think that is what already is happening - and I think works as is.  I'm not sure why a transitive dependency would change what we'd expect developers to have to deal with.

jdhealy:
> Could you explain a little more about how the issue with bare repos would cause?

[Most checkouts of Carthage dependencies are from bare git repos (bare means no working tree, just git objects), mostly from our `Library/Caches` directory.

Without a working tree, `git check-ignore` intentionally exits early, emitting an error message citing its lack of support for bare repositories.

So to use that tool (in its current state) we'd need another checkout somewhere else (which is time consuming, even assuming the drive has space for it).

mdiep:
I'm a little confused about what the problem is and how this PR attempts to solve it.

It sounds like `DependencyA` has a `binary` dependency `DependencyB`, but it's not showing up in `DependencyA`'s `Carthage/Build`. Is that correct? We should already be creating a symlink for that build folder to the root project's build folder.

tbrannam:
The symlink for `DependencyB` is not being created properly - the symlink has the wrong name, and points to an invalid location.   Maybe I'm missing something?   I'll note that in my failcase - `DependencyA` is an open source dependency with a reference to a closed source `DependencyB` - and the project as a whole uses `--no-build` or `--no-use-binaries`

mdiep:
Okay, I think that makes sense. The build symlink only gets created when you're building. Maybe we should create that symlink during `checkout` also? (_Also_, not _instead_, so that you can still build if it's not there.)

That seems like it'd solve your issue. We already create this symlink today, so it shouldn't pose any additional problems wrt ignored files. What do you think @tbrannam @jdhealy?

I'm still a little confused about what the wrong symlink you're seeing is.

tbrannam:
Will review the checkout theory.

I will also setup a sample project demonstrating the problem being addressed - I hope it will help clarify the issue.

Has there been any thoughts about how binary dependencies might be able to describe their dependencies?    In my case a 3rd party `DependencyB` actually includes sub-dependencies as well (in this case those are normal github carthage repos)

mdiep:
> Has there been any thoughts about how binary dependencies might be able to describe their dependencies? In my case a 3rd party DependencyB actually includes sub-dependencies as well (in this case those are normal github carthage repos)

Not really. Putting a `Cartfile` in the `zip` seems unideal. We might be best off putting something inside the JSON? In which case, we'd need to add the option to specify more detailed information about each version, making it a dictionary instead of a string.

blender:
ping

Issue #2229, opened 2017-10-25
How to just download binary dependencies


KoCMoHaBTa:
* `carthage version`: 0.26.2
* `xcodebuild -version`: Xcode 9.0.1
* Are you using `--no-build`? YES
* Are you using `--no-use-binaries`? NO
* Are you using `--use-submodules`? NO
* Are you using `--cache-builds`? NO

Hi, 

For git dependencies, I use `carthage update --no-build` and `carthage checkout` commands in order to respectively update and resolve dependencies.
Then i manually integrate the downloaded projects as references from the `Carthage/Checkouts` directory. For more information, you can check my [simple project](https://github.com/KoCMoHaBTa/CarthageIntegrationSample) 

This works fine, however i need to just download some 3rd party prebuilt frameworks as they are and integrate them manually.
More specifically, there is a zip file that contains multiple frameworks and resources that i just need to download, unzip and manually integrate as usually, giving me the benefit of avoiding copying the frameworks and resources into my repository.
Also i'm looking for a way to manage just resource dependencies in the same way.

I've tried to define the dependency as `binary`, however Carthage appears to do some post processing that fails and never presents the downloaded files. The error is `the DTSDKName key in its plist file is missing`, however i just want Carthage to download the zip file and extract it - then i'll do the integration with its contents myself, no matter what they are.

My question is - is there an existing command that can fulfil my needs, or if not would it be possible to introduce such?

blender:
> there is a zip file that contains multiple frameworks and resources that i just need to download

Why would carthage download a zip file it knows nothing about 

> The error is the DTSDKName key in its plist file is missing

and that is not a framework archive ?

KoCMoHaBTa:
> Why would carthage download a zip file it knows nothing about
For the same reason it can just download any git repository, no matter what kind of code is inside.
What i'm suggesting is to have an additional option that would allow downloading binary without inspecting and copying them in some specific way, as it is now per platform. Similar to `--no-build` - you get the source code and thats it - we miss the behaviour to just download the frameworks.

For more information - the frameworks are static and probably that is the reason for the error.
But Carthage is totally capable of delivering the contents of the zip file, but it does not do it.

The archive is not owned by me, its 3rd party closed source and i have no control over how it is structured.

blender:
I understand. I suggest you use nix for this instead.

KoCMoHaBTa:
[...]

PR #2232, opened 2017-10-26
- added the ability to only download  binary dependencies (#2229)

KoCMoHaBTa:


KoCMoHaBTa:
@mdiep What would be the desired behaviour of Carthage when it encounters a static framework?
Is it only identifying its platform and putting it inside the `Carthage/Build/<platform>` folder?

KoCMoHaBTa:
@mdiep Added support for static frameworks during `checkout`.

Source/carthage/Checkout.swift
```diff
@@ -46,6 +52,8 @@ public struct CheckoutCommand: CommandProtocol {
 				<*> mode <| Option(key: "use-ssh", defaultValue: false, usage: "use SSH for downloading GitHub repositories")
 				<*> mode <| Option(key: "use-submodules", defaultValue: false, usage: "add dependencies as Git submodules")
 				<*> mode <| Option(key: "use-binaries", defaultValue: true, usage: useBinariesUsage)
+				<*> mode <| Option(key: "copy-binaries", defaultValue: false, usage: "copy the contents of the binary depndencies")
+				<*> mode <| Option(key: "copy-frameworks", defaultValue: true, usage: "skip frameworks validations and copying in Build folder, can be used with --copy-binaries in order to only download raw binary dependencies")
```
mdiep: These options should be removed.

Source/CarthageKit/Project.swift
```diff
@@ -1087,8 +1112,41 @@ private func platformForFramework(_ frameworkURL: URL) -> SignalProducer<Platfor
 				let error = Result<(), NSError>.error(message)
 				return .readFailed(frameworkURL, error)
 			}
+			
+			func sdknameFromExecutable() -> Any? {
```
mdiep: The _n_ in _name_ should be capitalized here.

Source/CarthageKit/Constants.swift
```diff
@@ -9,6 +9,10 @@ public struct Constants {
 	/// The name of the folder into which Carthage puts binaries it builds (relative
 	/// to the working directory).
 	public static let binariesFolderPath = "Carthage/Build"
+	
+	/// The name of the folder into which Carthage puts binaries it downloads (relative
+	/// to the working directory).
+	public static let downloadsFolderPath = "Carthage/Downloads"
```
mdiep: This should be removed.

Source/CarthageKit/Project.swift
```diff
@@ -104,6 +104,12 @@ public final class Project { // swiftlint:disable:this type_body_length
 	/// Whether to download binaries for dependencies, or just check out their
 	/// repositories.
 	public var useBinaries = false
+	
+	/// Whenever to copy the downloaded binary dependencies
+	public var copyBinaries: Bool = false
+	
+	/// Whenever to validate and copy the downloaded frameworks to the Build directory
+	public var copyFrameworks: Bool = true
```
mdiep: These options should be removed.

KoCMoHaBTa:
@mdiep I don't understand why you won't accept the feature to expose the contents of the downloaded binary dependency as an option.

What happened with the Carthage statement:

> Carthage builds framework binaries using xcodebuild, but leaves the responsibility of integrating them up to the user.

In this case - the framework binaries are prebuilt, but we don't have the ability to integrate them by ourselves.
**It is not Carthage responsibility to inspect, validate and group prebuilt framework binaries.**
It responsibility should be only to resolve them.

blender:
Please add appropriate tests

mdiep:
> I don't understand why you won't accept the feature to expose the contents of the downloaded binary dependency as an option.

It's not necessary. If a `binary` origin can copy the pre-built static framework to the appropriate `Carthage/Build/` directory, then why do you also need to expose the contents of the zip in `Carthage/Downloads/`?

KoCMoHaBTa:
Mainly, because of:
- the resources that comes with the static library or framework - how would you handle it?
- specific instructions that you might need to follow from the provider of the library
- this will give you the ability to integrate with any kind of dependency for free. Why would you have this limitation?

The `Carthage/Build` folder is intended to put build that Carthage has just builded itself.
Binary dependency means - as is. Carthage should not even try to assume what it is - because it really can be anything.

From the beginning i'm talking about this. 
Since i've wasted too much time, trying to convince you that Carthage can do much more by support anything as dependency with this small addition, i'm considering to move forward.

**So if this is going to be accepted as an idea** - i'm keen to only do changes related to configurations and conventions. Otherwise, just let me know and i'll no longer bother you, as obviously the acceptance of good ideas appears to be very conservative for this repository.

I believe i've made myself, my idea and my reasons clear few times so far. I prefer to do something useful rather than continue spending time arguing.

mdiep:
> the resources that comes with the static library or framework - how would you handle it?

This question doesn't make sense for a static framework—these resources should be bundled _in_ the static framework. Copying that framework gets you the resources.

The solution I've suggested meets the need that you've presented (a vendor-provided static framework) and is consistent with the overall approach for Carthage. If there's another need that it's not meeting, then we should discuss that explicitly.

KoCMoHaBTa:
Static frameworks cannot contain resources, because they are static libraries wrapped to look like frameworks. 
Unfortunately a lot of big companies continue to use this approach - https://github.com/jverkoey/iOS-Framework
And for resources, as these are actually static libraries - they cannot be embedded inside, but are distributed separately, most of the time as bundles, but not always - https://github.com/jverkoey/iOS-Framework#resources

mdiep:
> Static frameworks cannot contain resources, because they are static libraries wrapped to look like frameworks.

Yes they can. ☺️ A static framework is just a special folder with the static library inside. You can put whatever you like inside of it.

But it sounds like that's not how people are actually doing it.

KoCMoHaBTa:
Closing due to inability and unwillingness of the owner to understand the proposed feature and benefits of supporting any kind of binary dependency.

PR #2235, opened 2017-10-30
Correctly handle static framework binaries

blender:
Salvaging good parts of https://github.com/Carthage/Carthage/pull/2232

I made no original contribution and claim no ownership.


Source/CarthageKit/Project.swift
```diff
@@ -1108,6 +1141,11 @@ private func platformForFramework(_ frameworkURL: URL) -> SignalProducer<Platfor
 private func frameworksInDirectory(_ directoryURL: URL) -> SignalProducer<URL, CarthageError> {
 	return filesInDirectory(directoryURL, kUTTypeFramework as String)
 		.filter { url in
+			//skip __MACOSX directory produced from archives
+			if url.pathComponents.contains("__MACOSX") {
```
NachoSoto: I would make this a separate filter:
```swift
.filter { !url.pathComponents.contains("__MACOSX") }
```

blender: Make sense

Source/CarthageKit/Project.swift
```diff
@@ -1087,8 +1087,41 @@ private func platformForFramework(_ frameworkURL: URL) -> SignalProducer<Platfor
 				let error = Result<(), NSError>.error(message)
 				return .readFailed(frameworkURL, error)
 			}
+			
+			func sdkNameFromExecutable() -> Any? {
+				guard let executableURL = bundle?.executableURL else {
+					return nil
+				}
+				
+				let otoolTask = Process()
```
mdiep: We should use ReactiveTask for this.

Source/CarthageKit/Project.swift
```diff
@@ -1087,8 +1087,41 @@ private func platformForFramework(_ frameworkURL: URL) -> SignalProducer<Platfor
...
+				otoolTask.launchPath = "/usr/bin/otool"
+				otoolTask.arguments = ["-lv", executableURL.path]
+				
+				let otoolPipe = Pipe()
+				otoolTask.standardOutput = otoolPipe
+				otoolTask.standardError = otoolPipe
+				otoolTask.launch()
+				
+				let grepTask = Process()
+				grepTask.launchPath = "/usr/bin/grep"
```
mdiep: I think it'd be preferable to do this without `grep` since that's the approach we've taken elsewhere.

Source/CarthageKit/Project.swift
```diff
@@ -1087,8 +1087,41 @@ private func platformForFramework(_ frameworkURL: URL) -> SignalProducer<Platfor
...
+				grepTask.launch()
+				
+				let data = grepPipe.fileHandleForReading.readDataToEndOfFile()
+				let output = String(data: data, encoding: .utf8)?.lowercased()
+				otoolTask.waitUntilExit()
+				grepTask.waitUntilExit()
+				
+				let sdkName = SDK.allSDKs.filter({ output?.contains($0.rawValue) == true }).first?.rawValue
+				return sdkName
+			}
```
mdiep: This code really needs a test.

Source/CarthageKit/Project.swift
```diff
@@ -1107,7 +1140,9 @@ private func platformForFramework(_ frameworkURL: URL) -> SignalProducer<Platfor
 /// Sends the URL to each framework bundle found in the given directory.
 private func frameworksInDirectory(_ directoryURL: URL) -> SignalProducer<URL, CarthageError> {
 	return filesInDirectory(directoryURL, kUTTypeFramework as String)
+		.filter { !$0.pathComponents.contains("__MACOSX") }
```
mdiep: This should have a test too.

This change could be moved out into a separate PR.

Source/CarthageKit/Xcode.swift
```diff
@@ -1200,6 +1200,9 @@ private func binaryURL(_ packageURL: URL) -> Result<URL, CarthageError> {
 		}
 
 	default:
+		if let binaryURL = bundle?.executableURL {
+			return .success(binaryURL)
+		}
```
mdiep: It doesn't seem like this should be necessary because AIUI a static framework should be a `.framework` (which is why we also have a `MachOType` and `FrameworkType`).

Either way this should be tested as well.

blender:
I'll do this.

Source/CarthageKit/Project.swift
blender: I believe this is tested indirectly via any test that touches `checkoutResolvedDependencies`

blender:
@mdiep  I pushed a few changes but I need some guidance.

- For the purpose of testing I have turned a static function from private to public which is not the best idea. I could not find tests that hit this spot otherwise. Any suggestions?
- I have added a framework compiled with Mach-O-Type static library for testing . Any other alternatives I should consider?

mdiep:
> For the purpose of testing I have turned a static function from private to public which is not the best idea. I could not find tests that hit this spot otherwise. Any suggestions?

You should be able to mark it as `internal` and use `@testable import CarthageKit` in the test.

> I have added a framework compiled with Mach-O-Type static library for testing . Any other alternatives I should consider?

That's it as far as I know.

> I'm not sure why the original author wanted to use otool. The DTSDKName key is present in the plist of static frameworks too. At least it is present if all you do is change the Mach-O-Type to "static lib".

If `DTSDKName` is there, then we should just use that and remove the code that calls `otool`. 👍

Tests/CarthageKitTests/ProjectSpec.swift
```diff
@@ -406,6 +406,17 @@ class ProjectSpec: QuickSpec {
 				expect(events) == [.downloadingBinaryFrameworkDefinition(.binary(testDefinitionURL), testDefinitionURL)]
 			}
 		}
+		
+		describe("platformForFramework") {
+			let testStaticFrameworkURL = Bundle(for: type(of: self)).url(forResource: "Alamofire.framework", withExtension: nil)!
+			
+			it("should produce s platform") {
```
mdiep: This looks like a typo?

blender:
Thanks for the review. I am currently on vacations. I will address the comments next week when I am back.

blender:
@mdiep 

Changed function from public to internal for testing

About `DTSDKName` and otool:

Basically the approach with otool is a fallback. If `DTSDKName` is missing form the plist ( #2265 and the original pull request issue) then it's still possible to derive the platform by using otool:

example output:
```
$ /usr/bin/otool -lv Tests/CarthageKitTests/Resources/Alamofire.framework/Alamofire | grep LC_VERSION
      cmd LC_VERSION_MIN_IPHONEOS
      cmd LC_VERSION_MIN_IPHONEOS
      cmd LC_VERSION_MIN_IPHONEOS
      cmd LC_VERSION_MIN_IPHONEOS
      cmd LC_VERSION_MIN_IPHONEOS
      cmd LC_VERSION_MIN_IPHONEOS
      cmd LC_VERSION_MIN_IPHONEOS
      cmd LC_VERSION_MIN_IPHONEOS
```

The LC_VERSION String reveals the platform too. I think it should it's wise to have the fallback because for some people `DTSDKName` is not there.

blender:
@mdiep ping

mdiep:
> My only issue is should it be /user/bin/otool or xcrun otool ? I think xcrun otool is safer

I agree. I think `xcrun otool` would be better. 👍

Source/CarthageKit/Project.swift
```diff
@@ -1095,14 +1095,42 @@ private func platformForFramework(_ frameworkURL: URL) -> SignalProducer<Platfor
...
+				let sdkName: String? = task.launch(standardInput: nil)
+					.ignoreTaskData()
+					.map { String(data: $0, encoding: .utf8) ?? "" }
+					.filter { !$0.isEmpty }
+					.flatMap(.merge) { (output: String) -> SignalProducer<String, NoError> in
+						output.linesProducer
+					}
+					.filter { $0.contains("LC_VERSION") }
+					.take(last: 1)
+					.flatMap(.merge) { lcVersionLine -> SignalProducer<String?, NoError> in
```
mdiep: This looks like it could be a regular `map`?

Source/CarthageKit/Project.swift
```diff
@@ -1095,14 +1095,42 @@ private func platformForFramework(_ frameworkURL: URL) -> SignalProducer<Platfor
...
+						let sdkString = lcVersionLine.split(separator: "_")
+							.last
+							.flatMap(String.init)
+							.flatMap { $0.lowercased() }
+
+						return .init(value: sdkString)
+					}.skipNil()
```
mdiep: `.skipNil()` should be on a new line.

Tests/CarthageKitTests/ProjectSpec.swift
```diff
@@ -406,6 +406,17 @@ class ProjectSpec: QuickSpec {
 				expect(events) == [.downloadingBinaryFrameworkDefinition(.binary(testDefinitionURL), testDefinitionURL)]
 			}
 		}
+		
+		describe("platformForFramework") {
+			let testStaticFrameworkURL = Bundle(for: type(of: self)).url(forResource: "Alamofire.framework", withExtension: nil)!
+			
+			it("should produce a platform") {
```
mdiep: Can you specify whether this is checking the binary or the info plist?

If it's not checking the binary, we should add a test for that.

Source/CarthageKit/Project.swift
```diff
@@ -1095,14 +1095,42 @@ private func platformForFramework(_ frameworkURL: URL) -> SignalProducer<Platfor
...
+					.value
+
+				return sdkName
 			}
 
-			if let sdkName = sdkName as? String {
-				return .success(sdkName)
+			if let sdkNameFromBundle = bundle?.object(forInfoDictionaryKey: "DTSDKName") as? String {
+				return .success(sdkNameFromBundle)
+			} else if let sdkNameFromExecutable = sdkNameFromExecutable() {
```
mdiep: Can you add a comment before the `if`/`else` that explains why we're checking both and in which order?

mdiep:
Just so I'm understanding correctly… there's two improvements here for static frameworks with `binary` origins?

1. Getting the name from the binary if it's not set in the info plist
2. Getting rid of the pesky `__MACOSX` folders

Is that correct? That's all that's needed?

blender:
@mdiep That is correct. Pretty cool right ? :)

Source/CarthageKit/Project.swift
blender: Done

Source/CarthageKit/Project.swift
blender: Done

Source/CarthageKit/Project.swift
blender: Done

Tests/CarthageKitTests/ProjectSpec.swift
blender: It's checking the Binary ;) The .plist is not even present in the bundle.

PR #2252, opened 2017-11-13
Strip string terminator from result of git command

iv-mexx:
This is a strange bug I've encountered: 

Carthage failed, but the failure message was printed incompletely:

screen shot 2017-11-13 at 17 26 54

Turns out, the reason was, that the task description contained the string terminator `\0`. Thats the reason the rest of the error message is not printed:
screen shot 2017-11-13 at 17 27 12

After some digging, I found out that the string terminator is introduced when searching for the repository url of submodules, and if I run the command manually in the terminal, it is printed as well:

screen shot 2017-11-13 at 18 16 06

I wanted to catch this as soon as possible and not just when printing the error, so this change string returned by `launchGitTask`. Not sure if this could also happen for other commands and if this should be applied at other places as well....

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -336,6 +336,7 @@ internal func gitmodulesEntriesInRepository(
 		.flatMap(.concat) { value in parseConfigEntries(value, keyPrefix: "submodule.", keySuffix: ".path") }
 		.flatMap(.concat) { name, path -> SignalProducer<(name: String, path: String, url: GitURL), CarthageError> in
 			return launchGitTask(baseArguments + [ "--get", "submodule.\(name).url" ], repositoryFileURL: repositoryFileURL)
+				.map { $0.stripping(suffix: "\0") }
 				.map { urlString in (name: name, path: path, url: GitURL(urlString)) }
 		}
 }


Source/CarthageKit/Git.swift
```diff
@@ -55,7 +55,7 @@ public func launchGitTask(
 		.ignoreTaskData()
 		.mapError(CarthageError.taskError)
 		.map { data in
-			return String(data: data, encoding: .utf8)!
+			return String(data: data, encoding: .utf8)!.stripping(suffix: "\0")
```
NachoSoto: Could you extract this to a function: `parseGitOutput(_ data: Data) -> String` and add a test that covers this?

mdiep: Maybe we should replace the `"\0"`s with `"\n"`s?

jdhealy:
I'd propose fixing `gitmodulesEntriesInRepository`:

https://github.com/Carthage/Carthage/blob/00431497a74104880a391cd13427fe1949d731dd/Source/CarthageKit/Git.swift#L430-L431

…adding a `.map { $0.stripping(suffix: "\0") }`, so that URLString is passed around without trailing null terminator.

The error message deserves to be fixed too — Carthage grabs that [straight from ReactiveTask](https://github.com/Carthage/ReactiveTask/blob/96e1013e0a6782bf0faba27485fe2d0a1153b531/Sources/Errors.swift#L23-L29), which could be patched to swap in the [“glyph for visual representation of the null character… U+2400 (␀)”](https://en.wikipedia.org/wiki/Null_character) in [the `escaped` computed property which currently only handles whitespace](https://github.com/Carthage/ReactiveTask/blob/96e1013e0a6782bf0faba27485fe2d0a1153b531/Sources/Task.swift#L49-L63).

- - -

Very surprising to me how much was able to work with the trailing null terminator on URLs (not just `git clone`ing), even on APFS case-sensitive volumes:

```
• ${SHELL} --version
zsh 5.3 (x86_64-apple-darwin17.0)

 • print -n 'https://github.com/antitypical/Result.git\x00' | xxd
00000000: 6874 7470 733a 2f2f 6769 7468 7562 2e63  https://github.c
00000010: 6f6d 2f61 6e74 6974 7970 6963 616c 2f52  om/antitypical/R
00000020: 6573 756c 742e 6769 7400                 esult.git.

• git clone "$(print -n 'https://github.com/antitypical/Result.git\x00')"
Cloning into 'Result'...
remote: Counting objects: 1937, done.
remote: Total 1937 (delta 0), reused 0 (delta 0), pack-reused 1937
Receiving objects: 100% (1937/1937), 346.89 KiB | 5.69 MiB/s, done.
Resolving deltas: 100% (1032/1032), done.

• ls | xxd
00000000: 5265 7375 6c74 0a                        Result.

• open -R "file://${PWD}$(print -n '/Result\x00')" # successfully reveals it in a Finder window
```

Source/CarthageKit/Git.swift
jdhealy: @mdiep in [the case of `gitmodulesEntriesInRepository`](https://github.com/Carthage/Carthage/blob/00431497a74104880a391cd13427fe1949d731dd/Source/CarthageKit/Git.swift#L414) (and others) [we’re actually requesting null characters](https://github.com/Carthage/Carthage/blob/00431497a74104880a391cd13427fe1949d731dd/Source/CarthageKit/Git.swift#L414) — not as terminators, but as data separators (because paths are allowed to contain new lines).

Best (IMO) to leave the data unchanged at the `launchGitTask` level (once we Unicode decode it, of course) to not break what some caller expects to receive for parsing.

jdhealy:
@iv-mexx Really exceptional, thanks for compiling and debugging with LLDB, appreciate the effort :clap:

Changing `launchGitTask` how you did would be safe enough for how we currently call it, but could potentially break what some developed-in-the-future caller expects to receive for parsing.

You mind [moving the null terminator stripping to `gitmodulesEntriesInRepository`](https://github.com/Carthage/Carthage/pull/2252#issuecomment-344060192)?

I’d be happy to help you tackle a [ReactiveTask](https://github.com/Carthage/ReactiveTask/pulls?utf8=%E2%9C%93&q=is%3Apr%20null) PR to [swap in that visual replacement glyph](https://github.com/Carthage/Carthage/pull/2252#issuecomment-344060192), if you'd like to head that effort as well 😄

iv-mexx:
Thanks for your comments everyone! 

@jdhealy great, I was not sure what the correct place was to put this fix, what I initially did was just the earliest place I could find. So I happily changed it to your suggestion.

>  thanks for compiling and debugging with LLDB

Weeeeellll.....

screen shot 2017-11-13 at 17 13 46

🤣 


I'll take a look at tomorrow!

PR #2254, opened 2017-11-13
Include incompatible dependencies in `outdated` output

iv-mexx:
`carthage outdated` now lists all dependencies, even those that will not be updated because of the specified version in the Cartfile. 

The output shows the current version, the latest compatible version to which the dependency will be updated when running `carthage update` as well as the latest available version. 

Showing incompatible updates is very helpful to get a complete overview about a projects dependency. This is also the same behaviour as all the other dependency managers that I know of (E.g. Cocoapods, npm, yarn, bundler, ...).

In a follow-up PR (#2257) I will also add colors to the output to make it easier to read (which is also the behaviour of other existing dependency managers).

screen shot 2017-11-13 at 22 17 00

This PR is a continuation of #1369 (resolving #1260) which has not seen progress since more than a year.
I resumed the work from @mattprowse's branch to keep his commits, but in the meantime the codebase has changed considerably so I had to redo the most part. Actually, once I discovered that `Resolver` already does everything necessary, the `latestDependencies` all of a sudden was only 3 lines long 😌 

### Breaking Change

I had to change the signature of `Project.updatedResolvedCartfile` which is a public function in CarthageKit. In Carthage/CarthageKit it is not called from outsied of `Project` but since its public I don't know if it may be used somewhere else.

### Earlier Comments

The two comments by @mdiep mentioned in https://github.com/Carthage/Carthage/pull/1369#issuecomment-229540621 have been discussed further down in this PR:

> It doesn't work with dependencies that have no tagged versions.

Since this issue is not caused by this PR (it also happens before this PR), I will not try to fix it here.

> It prints results for dependencies that are using a branch when there's nothing new on the branch. (Because it's not on the latest version.)

This has been taken care of in 5279bce


iv-mexx:
Hi, I'm having a few questions: 

#### Regarding the points raised by @mdiep in https://github.com/Carthage/Carthage/pull/1369#issuecomment-229540621

> It doesn't work with dependencies that have no tagged versions.

If i'm understanding correctly, the same does not work even before this PR. Carthage throws an error (`No tagged versions found for <Dependency>`) in `Project.versions(for:)`, called by `Project.updatedResolvedCartfile`. 

> It prints results for dependencies that are using a branch when there's nothing new on the branch. (Because it's not on the latest version.)

E.g. `FormatterKit "2ea246..." -> "2ea246..." (Latest: "1.8.0")`. How should this be handled? 
* Keep as is
* Omit from the output
* Print separately (e.g. "There could be newer commits" so the user does not forget about those dependencies... Usually I pin to a branch or commit just when there is no release **yet**)
* I guess we could also parse the git tree from that commit to find the actual newest version / commit (Seems overkill to me)


#### Regarding testing

Any hints how I would mock a Cartfile, Carfile.resolved and Latest Versions with specific Versions for a testcase?

mdiep:
> If i'm understanding correctly, the same does not work even before this PR. Carthage throws an error (No tagged versions found for <Dependency>) in Project.versions(for:), called by Project.updatedResolvedCartfile.

If it's broken before, then I guess fixing it isn't a prerequisite here. But it'd be nice. 😉 

> E.g. FormatterKit "2ea246..." -> "2ea246..." (Latest: "1.8.0"). How should this be handled?

Ideally, it would fetch the branch and show you only if the branch resolves to a different SHA now.

But failing that, I think it could just be omitted.

> Any hints how I would mock a Cartfile, Carfile.resolved and Latest Versions with specific Versions for a testcase?

Maybe look at `ResolverSpec`?

mdiep:
When this is ready to be reviewed, please leave a comment saying so. ☺️

Tests/CarthageKitTests/ProjectSpec.swift
```diff
@@ -424,3 +508,59 @@ extension ProjectEvent {
 		return false
 	}
 }
+
+
+// swiftlint:enable identifier_name
+private struct DB {
```
iv-mexx: This DB struct is copied 1:1 from `ResolverSpec`. Should it rather be extracted into a separate file?

iv-mexx:
@mdiep yup, will do! I still have to fix at least the issue with branches (the second note). 
I will not tackle the first note currently though, maybe in a separate PR later.

iv-mexx:
@mdiep this is now ready for review! 🎉

iv-mexx:
@mdiep 🏓

Tests/CarthageKitTests/ProjectSpec.swift
mdiep: Yes, please extract it and the `github1`, `github2`, etc. 👍

mdiep:
Sorry for the delay!

This is looking good. Just need to extract the `DB` and resolve the conflict. 👍

PR #2257, opened 2017-11-14
Outdated Colors

iv-mexx:
This builds on top of #2254 and includes color output to the `outdated` command to make it easier to read:

**green** for dependencies that can be updated to the latest version
**yellow** for dependencies that can be updated, where the latest update is not compatible
**red** for outdated dependencies where no compatible update is available.

screen shot 2017-11-14 at 13 50 33

### DIFF
--- a/Source/carthage/Formatting.swift
+++ b/Source/carthage/Formatting.swift
@@ -77,11 +77,6 @@ public struct ColorOptions: OptionsProtocol {
 		func quote(_ string: String, quotationMark: String = "\"") -> String {
 			return wrap(isColorful, wrap: Color.Wrap(foreground: .green))(quotationMark + string + quotationMark)
 		}
-
-		/// Wraps a string in a color
-		func colored(_ string: String, color: Color.Named.Color) -> String {
-			return wrap(isColorful, wrap: Color.Wrap(foreground: color))(string)
-		}
 	}
 
 	private init(argument: ColorArgument) {
@@ -96,5 +91,38 @@ public struct ColorOptions: OptionsProtocol {
 				defaultValue: ColorArgument.auto,
 				usage: "whether to apply color and terminal formatting (one of 'auto', 'always', or 'never')"
 			)
+	}
+}
+
+extension OutdatedCommand.UpdateAvailabilityAndApplicability {
+	fileprivate static let dict: [OutdatedCommand.UpdateAvailabilityAndApplicability: Color.Named.Color] = [
+		.updatesAvailableAllApplicable: .green,
+		.updatesAvailableSomeApplicable: .yellow,
+		.updatesAvailableNoneApplicable: .red,
+	]
+}
+
+extension ColorOptions.Formatting {
+	subscript(_ index: OutdatedCommand.UpdateAvailabilityAndApplicability) -> Wrap {
+		guard self.isColorful, let color = OutdatedCommand.UpdateAvailabilityAndApplicability.dict[index] else {
+			return { $0 }
+		}
+		return Color.Wrap(foreground: color).wrap
+	}
+
+	public var legendForOutdatedCommand: String? {
+		guard self.isColorful else { return nil }
+
+		let header = "Legend — <color> • «what happens when you run `carthage update`»:\n"
+		return OutdatedCommand.UpdateAvailabilityAndApplicability.dict.reduce(into: header) {
+			let (situation, color) = $1
+			let tabs = String(
+				repeating: "\t",
+				count: color == .yellow || color == .magenta ? 1 : 2
+			)
+			let colorDescription = Color.Wrap(foreground: color)
+				.wrap("<" + String(describing: color) + ">")
+			$0.append(colorDescription + tabs + "• " + situation.rawValue + "\n")
+		}
 	}
 }
--- a/Source/carthage/Formatting.swift
+++ b/Source/carthage/Formatting.swift
@@ -85,12 +85,20 @@ public struct ColorOptions: OptionsProtocol {
 	}
 
 	public static func evaluate(_ mode: CommandMode) -> Result<ColorOptions, CommandantError<CarthageError>> {
+		return self.evaluate(mode, additionalUsage: nil)
+	}
+
+	public static func evaluate(_ mode: CommandMode, additionalUsage: String? = nil) -> Result<ColorOptions, CommandantError<CarthageError>> {
+		var usage = "whether to apply color and terminal formatting (one of 'auto', 'always', or 'never')"
+		if let additionalUsage = additionalUsage {
+			usage += "\n" + additionalUsage
+		}
 		return curry(self.init)
 			<*> mode <| Option(
 				key: "color",
 				defaultValue: ColorArgument.auto,
-				usage: "whether to apply color and terminal formatting (one of 'auto', 'always', or 'never')"
-			)
+				usage: usage
+		)
 	}
 }
 
@@ -113,21 +121,5 @@ extension ColorOptions.Formatting {
 			return { $0 }
 		}
 		return Color.Wrap(foreground: update.color).wrap
-	}
-
-	public var legendForOutdatedCommand: String? {
-		guard self.isColorful else { return nil }
-
-		let header = "Legend — <color> • «what happens when you run `carthage update`»:\n"
-		return [OutdatedCommand.UpdateType.newest, .newer, .ineligible].reduce(into: header) {
-			let (color, explanation) = ($1.color, $1.explanation)
-			let tabs = String(
-				repeating: "\t",
-				count: color == .yellow || color == .magenta ? 1 : 2
-			)
-			let colorDescription = Color.Wrap(foreground: color)
-				.wrap("<" + String(describing: color) + ">")
-			$0.append(colorDescription + tabs + "• " + explanation + "\n")
-		}
 	}
 }
--- a/Source/carthage/Formatting.swift
+++ b/Source/carthage/Formatting.swift
@@ -94,35 +94,40 @@ public struct ColorOptions: OptionsProtocol {
 	}
 }
 
-extension OutdatedCommand.UpdateAvailabilityAndApplicability {
-	fileprivate static let dict: [OutdatedCommand.UpdateAvailabilityAndApplicability: Color.Named.Color] = [
-		.updatesAvailableAllApplicable: .green,
-		.updatesAvailableSomeApplicable: .yellow,
-		.updatesAvailableNoneApplicable: .red,
-	]
+extension OutdatedCommand.UpdateType {
+	var color: Color.Named.Color {
+		switch self {
+		case .newest:
+			return .green
+		case .newer:
+			return .yellow
+		case .ineligible:
+			return .red
+		}
+	}
 }
 
 extension ColorOptions.Formatting {
-	subscript(_ index: OutdatedCommand.UpdateAvailabilityAndApplicability) -> Wrap {
-		guard self.isColorful, let color = OutdatedCommand.UpdateAvailabilityAndApplicability.dict[index] else {
+	subscript(_ index: OutdatedCommand.UpdateType?) -> Wrap {
+		guard self.isColorful, let update = index else {
 			return { $0 }
 		}
-		return Color.Wrap(foreground: color).wrap
+		return Color.Wrap(foreground: update.color).wrap
 	}
 
 	public var legendForOutdatedCommand: String? {
 		guard self.isColorful else { return nil }
 
 		let header = "Legend — <color> • «what happens when you run `carthage update`»:\n"
-		return OutdatedCommand.UpdateAvailabilityAndApplicability.dict.reduce(into: header) {
-			let (situation, color) = $1
+		return [OutdatedCommand.UpdateType.newest, .newer, .ineligible].reduce(into: header) {
+			let (color, explanation) = ($1.color, $1.explanation)
 			let tabs = String(
 				repeating: "\t",
 				count: color == .yellow || color == .magenta ? 1 : 2
 			)
 			let colorDescription = Color.Wrap(foreground: color)
 				.wrap("<" + String(describing: color) + ">")
-			$0.append(colorDescription + tabs + "• " + situation.rawValue + "\n")
+			$0.append(colorDescription + tabs + "• " + explanation + "\n")
 		}
 	}
 }
--- a/Source/carthage/Formatting.swift
+++ b/Source/carthage/Formatting.swift
@@ -77,6 +77,11 @@ public struct ColorOptions: OptionsProtocol {
 		func quote(_ string: String, quotationMark: String = "\"") -> String {
 			return wrap(isColorful, wrap: Color.Wrap(foreground: .green))(quotationMark + string + quotationMark)
 		}
+
+		/// Wraps a string in a color
+		func colored(_ string: String, color: Color.Named.Color) -> String {
+			return wrap(isColorful, wrap: Color.Wrap(foreground: color))(string)
+		}
 	}
 
 	private init(argument: ColorArgument) {
--- a/Source/carthage/Outdated.swift
+++ b/Source/carthage/Outdated.swift
@@ -7,26 +7,30 @@ import Curry
 
 /// Type that encapsulates the configuration and evaluation of the `outdated` subcommand.
 public struct OutdatedCommand: CommandProtocol {
-	enum UpdateAvailabilityAndApplicability: String {
-		case updatesAvailableAllApplicable =
-		"Will be updated to the newest version."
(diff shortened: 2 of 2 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -413,24 +413,52 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+	/// Attempts to determine the latest version (whether satisfiable or not)
+	/// of the project's Carthage dependencies.
+	///
+	/// This will fetch dependency repositories as necessary, but will not check
+	/// them out into the project's working directory.
+	private func latestDependencies() -> SignalProducer<[Dependency: PinnedVersion], CarthageError> {
+		return loadResolvedCartfile()
+			.flatMap(.merge) { cartfile -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
+				return SignalProducer(cartfile.dependencies.keys)
+					.flatMap(FlattenStrategy.merge) { dependency -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
```
mdiep: This `flatMap` looks like it could be moved out a level, instead of being nested inside the `flatMap`.

Also, this could be `.merge` instead of `FlattenStrategy.merge`.

Source/CarthageKit/Project.swift
```diff
@@ -413,24 +413,52 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+						let versions = self.versions(for: dependency)
+							.filterMap { version -> Version? in
+								guard let semanticVersion = SemanticVersion.from(version).value else { return nil }
+								return (semantic: semanticVersion, pinned: version)
+							}
+							.collect()
+						let latestVersion = versions
+							.filterMap { $0.max(by: { $0.semantic < $1.semantic }) }
+							.map { $0.pinned }
+						return SignalProducer(value: dependency).combineLatest(with: latestVersion)
```
mdiep: Instead of this, I'd suggest having a single `SignalProducer` chain.

```
typealias Version = (semantic: SemanticVersion, pinned: PinnedVersion)
return self
    .versions(for: dependency)
    .filterMap { version -> Version? in
        ...
    }
    .collect()
    .filterMap { ... }
    .map { (dependency, $0.pinned) }
```

Source/CarthageKit/Project.swift
```diff
@@ -413,24 +413,52 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+					}
+			}.reduce(into: [:]) { (working: inout [Dependency: PinnedVersion], next: (Dependency, PinnedVersion)) in
```
mdiep: The `.reduce…` should be on a new line.

Source/carthage/Outdated.swift
```diff
@@ -54,13 +55,40 @@ public struct OutdatedCommand: CommandProtocol {
...
+							carthage.println(formatting.projectName(project.name) + " " + versionSummary)
 						}
 					}
+
+					if options.colorOptions.formatting.isColorful {
+						carthage.println(formatting.path("The color indicates what happens when you run `carthage update`"))
+						carthage.println(formatting.colored("<green>", color: .green) + "\t\t- Will be updated to the newest version")
+						carthage.println(formatting.colored("<yellow>", color: .yellow) + "\t- Will be updated, but not to the newest version"
+							+ "because of the specified version in Cartfile")
+						carthage.println(formatting.colored("<red>", color: .red) + "\t\t- Will not be updated because of the specified version in Cartfile")
```
mdiep: This seems like it should be specified in the help instead of being included in every invocation.

mdiep:
But now I see I should have answered your questions on #2254. 😓

iv-mexx:
@mdiep Thanks for your comments. I opened this PR right away to show that I am working on that. I will keep your comments in mind for #2254 - once that is done, adding the colors is not much more work. 

I could have clarified that this should be reviewed just after #2254 though :sweat_smile:

Medo3oo311:
#2254

Medo3oo311:
294296336915518

Source/carthage/Outdated.swift
iv-mexx: I'm unsure how to do that. The simple solution I see is to add the text to
`public let function = "Check for compatible updates to the project's dependencies"`, then it will be printed at the top. But that seems wrong to me.

I'd like it much better to include the explanation to the help text of the color option, e.g.:

```
[--color (color)]
	whether to apply color and terminal formatting (one of 'auto', 'always', or 'never')
	
	The color indicates what happens when you run `carthage update`
	<green>	- Will be updated to the newest version
	<yellow>	- Will be updated, but not to the newest version
	<red>	- Will not be updated because of the specified version in Cartfile
```

However, I have no Idea how to implement this since the color option help is generated by `ColorOptions.evaluate(mode)` and I don't see a way to add extra text to the help of `ColorOptions`...

mdiep:
Is this ready for review?

iv-mexx:
@mdiep just needed to rebase on master now that #2254 is merged - yes its ready for review now! 😃

Source/carthage/Outdated.swift
```diff
@@ -54,13 +55,40 @@ public struct OutdatedCommand: CommandProtocol {
...
+								+ " (Latest: \(latest))"
+							carthage.println(formatting.projectName(project.name) + " " + versionSummary)
 						}
 					}
+
+					if options.colorOptions.formatting.isColorful {
+						carthage.println(formatting.path("The color indicates what happens when you run `carthage update`"))
+						carthage.println(formatting.colored("<green>", color: .green) + "\t\t- Will be updated to the newest version")
+						carthage.println(formatting.colored("<yellow>", color: .yellow) + "\t- Will be updated, but not to the newest version"
+							+ "because of the specified version in Cartfile")
```
dcaunt: Need a space before `because` to separate it from `version` (see your screenshot of the example output)

iv-mexx: Oh yeah, right 👍

Source/carthage/Outdated.swift
```diff
@@ -54,13 +55,40 @@ public struct OutdatedCommand: CommandProtocol {
...
+							// Update available and applicable
+							versionColor = .green
+						case (current, updated, latest) where current != updated && updated != latest:
+							// Update availabe and applicable, but not to the latest version due to version lock
+							versionColor = .yellow
+						case (current, updated, latest) where current == updated:
+							// Update available, but not applicable due to version lock
+							versionColor = .red
+						default:
+							versionColor = .white
```
jdhealy: Not sure, is this actually down a dead codepath, considering the logic of `outdatedDependencies(_:useNewResolver:resolver:)`?

If actually in the codepath, using `.white` here can’t function as a workaround, considering [Terminal.app](https://www.apple.com/search/terminal.app)’s default theme as of recent years has a white background.

Subsequently, if we actually need to display that text, something worth considering is to separate the formatting from the logic — an enum declared in `Outdated.swift` that has formatting and the legend for colors in `Formatting.swift`:

```swift
enum UpdateAvailabilityAndApplicability: String {
	case updateToLatestAvailableAndApplicable =
		"Will be updated to the newest version."
	case updateAvailableYetLatestUnapplicableDueToVersionSpecifier =
		"Will be updated, but not to the newest version."
	case everyAvailableUpdateUnapplicableDueToVersionSpecifier =
		"Will not be updated because of the specified version in Cartfile."
	case noAvailableUpdates =
		"Will not be updated: no updates available."
}

/* in Formatting.swift */
extension UpdateAvailabilityAndApplicability {
	fileprivate static let dict: [Self: Color.Named.Color] = [
		updateToLatestAvailableAndApplicable: .green,
		updateAvailableYetLatestUnapplicableDueToVersionSpecifier: .yellow,
		everyAvailableUpdateUnapplicableDueToVersionSpecifier: .red
	]
}

extension Formatting {
	public subscript(_ index: UpdateAvailabilityAndApplicability) -> Wrap {
		guard self.isColorful, let color = UpdateAvailabilityAndApplicability.dict[index] else {
			return { $0 }
		}

		return Color.Wrap(foreground: color).wrap
	}

	public var legendForOutdatedCommand: String? {
		guard self.isColorful else { return nil }

		let header = "Legend — <color>\t• «what happens when you run `carthage update`»:\n"
	
		return UpdateAvailabilityAndApplicability.dict.reduce(into: header) {
			let (situation, color) = $1

			let tabs = String(
				repeating: "\t",
				count: color == .yellow || color == .magenta ? 1 : 2
			)

			let colorDescription = Color.Wrap(foreground: color)
				.wrap("<" + String(describing: color) + ">")

			$0.append(colorDescription + tabs + "• " + situation.rawValue + "\n")
		}
	}
}

/* back in Outdated.swift’s `run` for-loop */
let style = formatting[updateAvailabilityAndApplicability]
style(current.description) + " -> " + style(updated.description) + " (Latest: \(latest))"
```

Source/carthage/Outdated.swift
```diff
@@ -54,13 +55,40 @@ public struct OutdatedCommand: CommandProtocol {
...
+						}
+
 						if options.outputXcodeWarnings {
-							carthage.println("warning: \(formatting.projectName(project.name)) is out of date (\(current) -> \(updated)) (Latest: \(latest))")
+							carthage.println("warning: \(formatting.projectName(project.name)) is out of date (\(current) -> \(updated) (Latest: \(latest))_")
```
jdhealy: Why did this change from `) (Latest` to ` (Latest`?

jdhealy:
Thanks for splitting this into two PRs and continuing to work on this! :clap:

Source/carthage/Outdated.swift
iv-mexx: I think that was unintentional, will revert

Source/carthage/Outdated.swift
iv-mexx: Yeah, right now this is not reachable due to the logic of `outdatedDependencies(_:useNewResolver:resolver:)` which will not include dependencies that that are **not** outdated (where no update is available). But that can not be guaranteed and I like your suggestion anyway.

iv-mexx:
@jdhealy Thanks for your feedback!

Source/carthage/Outdated.swift
```diff
@@ -7,6 +7,30 @@ import Curry
...
 public struct OutdatedCommand: CommandProtocol {
+	enum UpdateAvailabilityAndApplicability: String {
+		case updatesAvailableAllApplicable =
+		"Will be updated to the newest version."
+		case updatesAvailableSomeApplicable =
+		"Will be updated, but not to the newest version because of the specified version in Cartfile."
+		case updatesAvailableNoneApplicable =
+		"Will not be updated because of the specified version in Cartfile."
+		case noUpdatesAvailable =
+		"Will not be updated: no updates available."
```
mdiep: I don't think these strings should be specified here. Using a `String`-based enum only makes sense when using the `RawRepresentable` APIs.

Instead, these strings should be moved to a computed property.

Source/carthage/Outdated.swift
```diff
@@ -7,6 +7,30 @@ import Curry
...
+
+		init(currentVersion current: PinnedVersion, applicableVersion applicable: PinnedVersion, latestVersion latest: PinnedVersion) {
+			switch (current, applicable, latest) {
+			case (current, applicable, latest) where current != latest && applicable == latest:
```
mdiep: This `switch`/`case` doesn't make sense: the `case`s are always true because they're matching what's in the `switch`. This should just be an `if`/`else`.

Source/carthage/Outdated.swift
```diff
@@ -7,6 +7,30 @@ import Curry
...
 /// Type that encapsulates the configuration and evaluation of the `outdated` subcommand.
 public struct OutdatedCommand: CommandProtocol {
+	enum UpdateAvailabilityAndApplicability: String {
+		case updatesAvailableAllApplicable =
+		"Will be updated to the newest version."
+		case updatesAvailableSomeApplicable =
+		"Will be updated, but not to the newest version because of the specified version in Cartfile."
+		case updatesAvailableNoneApplicable =
+		"Will not be updated because of the specified version in Cartfile."
+		case noUpdatesAvailable =
```
mdiep: What do you think of renaming this like so?

```swift
public struct OutdatedCommand: CommandProtocol {
    enum UpdateType {
        case newest
        case newer
        case ineligible

        init?(current: PinnedVersion, applicable: PinnedVersion, latest: PinnedVersion) {
            guard current != latest else { return nil }
            if applicable == latest {
                return .newest
            } else if current != applicable {
                return .newer
            } else {
                return .ineligible
            }
        }
    }
}
```

Source/carthage/Formatting.swift
```diff
@@ -93,3 +93,36 @@ public struct ColorOptions: OptionsProtocol {
 			)
 	}
 }
+
+extension OutdatedCommand.UpdateAvailabilityAndApplicability {
+	fileprivate static let dict: [OutdatedCommand.UpdateAvailabilityAndApplicability: Color.Named.Color] = [
+		.updatesAvailableAllApplicable: .green,
+		.updatesAvailableSomeApplicable: .yellow,
+		.updatesAvailableNoneApplicable: .red,
+	]
```
mdiep: With my restructuring of the `enum` below, this could be a computed property instead.

Source/carthage/Outdated.swift
iv-mexx: Looks good. Your other comments will be covered by this change

iv-mexx:
@mdiep Thanks for your feedback! I've addressed your comments.

Source/carthage/Outdated.swift
```diff
@@ -54,13 +82,18 @@ public struct OutdatedCommand: CommandProtocol {
...
+							carthage.println("warning: \(formatting.projectName(project.name)) is out of date (\(current) -> \(applicable)) (Latest: \(latest))")
 						} else {
-							carthage.println(formatting.projectName(project.name) + " \(current) -> \(updated) (Latest: \(latest))")
+							let update = UpdateType(currentVersion: current, applicableVersion: applicable, latestVersion: latest)
+							let style = formatting[update]
+							let versionSummary = "\(style(current.description)) -> \(style(applicable.description)) (Latest: \(latest))"
+							carthage.println(formatting.projectName(project.name) + " " + versionSummary)
 						}
 					}
+					formatting.legendForOutdatedCommand.map(carthage.println)
```
mdiep: I don't think the legend should be prevented on every invocation. IMO this information should be included in `carthage help outdated` instead.

mdiep:
Thanks! That looks good. Just one remaining bit. (Sorry I didn't notice it on the last round of review.)

Source/carthage/Outdated.swift
iv-mexx: I'm unsure how to do that. The simple solution I see is to add the text to
`public let function = "Check for compatible updates to the project's dependencies"`, then it will be printed at the top. But that seems wrong to me.

I'd like it much better to include the explanation to the help text of the color option, e.g.:

```
[--color (color)]
	whether to apply color and terminal formatting (one of 'auto', 'always', or 'never')
	
	The color indicates what happens when you run `carthage update`
	<green>	- Will be updated to the newest version
	<yellow>	- Will be updated, but not to the newest version
	<red>	- Will not be updated because of the specified version in Cartfile
```

However, I have no Idea how to implement this since the color option help is generated by `ColorOptions.evaluate(mode)` and I don't see a way to add extra text to the help of `ColorOptions`...

mdiep: I think it'd be fine to change it to:

```swift
public static func evaluate(_ mode: CommandMode, additionalUsage: String? = nil) -> Result<ColorOptions, CommandantError<CarthageError>>
```

Then you could pass the extra information for `outdated` and have it appended.

iv-mexx: Do you mean actually changing the method in [Commandant.OptionsProtocol](https://github.com/Carthage/Commandant/blob/master/Sources/Commandant/Option.swift#L44) or just adding an override with the extra parameter to `Formatting`?

iv-mexx:
@mdiep Since your comment is now outdated: In 8dbd796 I've implemented the less invasive option by adding `public static func evaluate(_ mode: CommandMode, additionalUsage: String? = nil) -> Result<ColorOptions, CommandantError<CarthageError>>` as an overload to `ColorOptions`. 

The result now looks like this:

screen shot 2017-12-20 at 19 04 51

One side effect is that now the help text is not colored since the help Command itself has no `--color` option... I think thats fine for now?

mdiep:
Looks great. 👍

jdhealy:
three arms and then a bear paw converging and breaking for celebrating teamwork

iv-mexx:
Nice, thanks for the Reviews, @mdiep and @jdhealy!

PR #2307, opened 2018-01-03
Cached lastest remote version to avoid blocked fetching in poor network environment

RingoD:
Since `carthage copy-frameworks` will run in each xcodebuild, blocked fetching remote version will make it slow in poor network environment.
A better way is to cache it and run periodically for such a non-critical thing.

### DIFF
--- a/Source/carthage/BuildVersion.swift
+++ b/Source/carthage/BuildVersion.swift
@@ -5,8 +5,43 @@ import ReactiveTask
 import Result
 import Tentacle
 
+struct ExpirableVersion: Codable {
+	let version: SemanticVersion?
+	let expiredDate: Date
+
+	init(version: SemanticVersion?, shelfLife: TimeInterval) {
+		self.version = version
+		self.expiredDate = Date(timeIntervalSinceNow: shelfLife)
+	}
+
+	init?(fromJSONData data: Data?) {
+		guard let wrappedData = data, let wrapped = try? JSONDecoder().decode(ExpirableVersion.self, from: wrappedData) else {
+			return nil
+		}
+		self = wrapped
+	}
+
+	func encodedJSONData() -> Data? {
+		return try? JSONEncoder().encode(self)
+	}
+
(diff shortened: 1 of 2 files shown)


Source/carthage/BuildVersion.swift
```diff
@@ -5,8 +5,43 @@ import ReactiveTask
...
 public func remoteVersion() -> SemanticVersion? {
+	let userDefaults = UserDefaults.standard
+	let cachedKey = "latestCarthageRemoteVersion"
+
+	let expirableVersionData = userDefaults.data(forKey: cachedKey)
+	let expirableVersion = ExpirableVersion(fromJSONData: expirableVersionData)
+
+	guard expirableVersion == nil || expirableVersion!.isExpired else {
+		return expirableVersion!.version
+	}
```
mdiep: Instead of adding this check to this function, can you create a new function that checks and then calls this one?

Source/carthage/BuildVersion.swift
```diff
@@ -5,8 +5,43 @@ import ReactiveTask
...
+	init?(fromJSONData data: Data?) {
+		guard let wrappedData = data, let wrapped = try? JSONDecoder().decode(ExpirableVersion.self, from: wrappedData) else {
+			return nil
+		}
+		self = wrapped
+	}
+
+	func encodedJSONData() -> Data? {
+		return try? JSONEncoder().encode(self)
+	}
```
mdiep: Instead of building this into this type, can we add a `UserDefaults` extension to decode/encode `Codable` types? That should make it easy to add some tests.

mdiep:
Thank you for the pull request! I'm really sorry it took me so long to review. I've been a little short on time recently.

I think this is a GREAT idea, but there are a few improvements that I'd like to see to the code.

blender:
@RingoD ping

blender:
ping

blender:
ping again

Issue #356, opened 2015-02-23
Treat build platform as a project setting


jspahrsummers:
So it can be [specified in `Cartfile.private`](https://github.com/Carthage/Carthage/issues/262).

kgn:
As mentioned in #877 I'd love a way to specify platforms in the `Cartfile`, something like:

```
platform iOS

github ...
```

marcpalmer:
Initially I was into this but I'm not sure it makes sense. e.g. if you have a project with cross-platform outputs, you really need a single Cartfile that does specify which platforms to build each dependency for.

An example would be an app that has iOS and tvOS versions, and the iOS version has features that are not available in tvOS - and it does this via some dependencies. You would want to build everything for iOS + tvOS, *except* those frameworks that are only desired on iOS. A single Cartfile would not enable this so the "platforms" would need to be specified per dependency.

Otherwise, you'd have no choice but to have separate project folders and separate Cartfile even though these are just variants of the same app?

jspahrsummers:
If implemented, this setting belongs in `Cartfile.private` (as mentioned in the issue body), so it wouldn't apply transitively, but instead be a "local" project configuration. Most useful for applications.

mlilback:
Also useful for those of us who are only building a Mac version. It can be defaulted to all, so there would be no backward compatibility.

Cramsden:
MERGE PLZ!!!!

craigotis:
This would be very great to have. With an iOS app depending on ~10 different libraries, I could save a lot of time not having to build the dependencies for AppleTV, the Apple Watch, and the Mac.

jspahrsummers:
@craigotis You can already provide `--platform` to specify that. This is just about being able to set defaults.

Dschee:
It would still be very nice to see this feature implemented, is there any work done on this already?

mdiep:
No. This is waiting on a more flexible format for Cartfiles. But we're also focused on cleaning up and improving existing code/functionality.

Dschee:
Is there an issue of the discussion about a more flexible Cartfile so we can keep staying updated?

joeblau:
@jspahrsummers You keep mentioning this:

> You can already provide `--platform` to specify that.

But there is no way to default your platform so that you don't have to type it in locally if you're building.  I think that's what people are asking for.

marcvanolmen:
[...]

PR #2312, opened 2018-01-06
[Feature]Add configuration support for Carthage

CodeEagle:
Working with [Commandant PR](https://github.com/Carthage/Commandant/pull/120)

# Features 
- config build options in `Carthage.private`
- can override repo to local project 
- show cost time

# Motivation
- Quit typing config for `Carthage` every time you build your dependencies
- Manual manager dependencies repositories by using `override` feature

# Usage 
Configuration Template(write in `Cartfile.private`)

```
#cache-builds: true/false, default false
#verbose: true/false, default false
#configuration: Release/Debug or other
#new-resolver: true/false, default false
#use-ssh: true/false, default false
#use-submodules: true/false, default false
#platforms:  macOS, iOS, watchOS, tvOS ,remove for all platform
#skip: Alamofire, Alamofire.xcworkspace, Alamofire iOS
#override: Alamofire, ../Alamofire
```

## `skip` 
Can just config `Name` parameter , all build of the `Dependency` will being skilled.
>
> - `Name`:  first parameter, `Dependency`'s property `name`,  when using local path for repo address, last component will treat as `name`, such as `/path/to/your/ProjectName`
`Name` is case sensitive(using as a key).
> - `ProjectName`:  second parameter, using whit third parameter, the name of project 
> - `SchemeName`: third  parameter, the scheme name


mdiep:
Thanks for the pull request, but I don't think this is how we'll want to add support for configuration. I think hijacking comments like this is probably not the best idea.

CodeEagle:
I don't think so.

benjohnde:
For the cirumstance that some other lonely creep is diving into this PR and asking the question if there is some further discussing going on, cf. https://github.com/Carthage/Carthage/issues/356.

Issue #2308, opened 2018-01-03
Cache invalidated when an incompatible binary is available


olejnjak:
Hi, 

I like using Carthage as compared to Cocoapods it speeds up most of our build times especially on our CI server.

One thing that improved times the most are prebuilt binaries which are awesome as long as they are compiled using the same Swift version. Right now we have the issue that Realm provides binary compiled using Swift 4.0.2 and we are using 4.0.3. This means that while running `carthage update --cache-builds` or `carthage bootstrap --cache-builds` it is rebuilt every time which is quite annoying. 

I know this can be solved by adding `--no-use-binaries` to the command but I think it's a shame not to use compatible binaries when they are available.

Is there any chance or desire that `--no-use-binaries` could be implemented on per dependency basis defined in Cartfile? 

Thanks

* `carthage version`: 0.27.0
* `xcodebuild -version`: 9.2 (9C40b)
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? yes

**Cartfile**
```
github "realm/realm-cocoa"
```

**Carthage Output**
```
Jakub-Macbook:ACKReactiveExtensions olejnjak$ carthage update --platform iOS --cache-builds
*** Fetching realm-cocoa
*** Downloading realm-cocoa.framework binary at "v3.0.2"
*** Skipped installing realm-cocoa.framework binary due to the error:
	"Incompatible Swift version - framework was built with 4.0.2 (swiftlang-900.0.69.2 clang-900.0.38) and the local version is 4.0.3 (swiftlang-900.0.74.1 clang-900.0.39.2)."
```

mdiep:
Are you saying that the hosted binary invalidates the cached build? If so, that's a bug.

But there's not a way to specify `--no-use-binaries` on a per-dependency basis.

olejnjak:
> Are you saying that the hosted binary invalidates the cached build? If so, that's a bug.

Yep cache is invalidated, can be reproduced easily

At first it is built which is correct
```
Jakub-Macbook:realm olejnjak$ echo 'github "realm/realm-cocoa"' > Cartfile
Jakub-Macbook:realm olejnjak$ carthage update --platform iOS --cache-builds
*** Fetching realm-cocoa
*** Downloading realm-cocoa.framework binary at "v3.0.2"
*** Skipped installing realm-cocoa.framework binary due to the error:
[...]

PR #2317, opened 2018-01-10
[Proposal] Skip binary download & checkout if valid cache exists

chuganzy:
Currently `carthage bootstrap --cache-builds` overwrites existing frameworks with binaries downloaded without checking version files which causes an issue that it overwrites existing valid binaries with ones downloaded even if they were built with incompatible Swift version..

ex) Using Swift 4.0.2 on the machine and the remote binary is built with Swift 4.0.1. Even if this case, Carthage overwrites the valid binary on the machine with the incompatible one and try to build again so far..

Also I was thinking that `carthage bootstrap` should perform the checkout only if needed since it takes long.

To solve these, this PR changes to check version files for each first and then skip downloading & checking out if `--cache-builds` option is provided and the valid cache is found.

P.S. I’m happy to add tests for this but wanna make sure this gets through before.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -818,6 +818,17 @@ public final class Project { // swiftlint:disable:this type_body_length
 			.zip(with: submodulesSignal)
 			.flatMap(.merge) { dependencies, submodulesByPath -> SignalProducer<(), CarthageError> in
 				return SignalProducer<(Dependency, PinnedVersion), CarthageError>(dependencies)
+					.flatMap(.merge) { dependency, version -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
+						let platforms = buildOptions?.platforms ?? []
+						return versionFileMatches(dependency,
+												  version: version,
+												  platforms: platforms,
+												  rootDirectoryURL: self.directoryURL,
+												  toolchain: buildOptions?.toolchain)
+							.flatMap(.merge) { matched -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
+								matched == true ? .empty : .init(value: (dependency, version))
+						}
+					}
 					.flatMap(.concurrent(limit: 4)) { dependency, version -> SignalProducer<(), CarthageError> in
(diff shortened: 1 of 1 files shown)


olejnjak:
According to description this would be the fix for #2308. I've been looking into it as well and came up with a solution as well, but this looks much better.

EmDee:
I was already wondering why it would always rebuilt my frameworks, although I had it built before. So the binary is invalidated the cache. Would be great, if this PR can be incorporated in the next release, @mdiep :-)

mdiep:
Thanks for the pull request! I appreciate you jumping in and fixing things—here and in your other recent PRs. It's great to have new contributors. ✨ 

I'm not totally comfortable with skipping `checkout` if a valid cached build exists. Skipping checkout could potentially cause issues for people if they're expecting the checkout to have the right version (which seems like a reasonable expectation).

Unfortunately, this code would be quite difficult to write otherwise, since we only perform checkout if we couldn't download the binary.

In light of that, I have a proposal that I'm wondering if you'd like to take on: Let's install binaries during `build` instead of during `checkout`. That simplifies everything, conceptually and in the code. We can perform cache validation in a single spot in the `build` process. `checkout` becomes more straightforward.

This would also fit in well with #2319.

chuganzy:
Hmm.. I am always wondering why Carthage needs to checkout everything again and again when we run `carthage bootstrap` since it takes long and as a heavy user I want at least a option to skip checkout since it wastes our time a lot so far..

mdiep:
> Moving binary download to build process totally makes sense but I also want a way to skip checkout while bootstrap..

That's `carthage build`. 😄 I think what I proposed would let you accomplish that.

chuganzy:
No, I mean I want a way to skip checkout if the valid build cache exists.

chuganzy:
Closing as I opened an issue to disscuss. Thanks!
https://github.com/Carthage/Carthage/issues/2358

chuganzy:
I'm on https://github.com/Carthage/Carthage/pull/2342 to solve #2308

PR #2318, opened 2018-01-10
No longer need to send completed

chuganzy:
From ReactiveSwift v2 it automatically gets interrupted when observer is deallocated.


mdiep:
Thanks for the pull request!

PR #2321, opened 2018-01-15
Update X's dependencies when running carthage update X.

### LINKED ISSUE #2188: `Carthage update X` doesn't build X's dependencies for newly added X
* `carthage version`: 0.25.0
* `xcodebuild -version`: 9.0
* Are you using `--no-use-binaries`?  Yes (Problem doesn't reproduce without this flag if binaries are present)

When adding new dependency X (that depends on Y) to Cartfile. `carthage update X` doesn't build Y.

Sample project:
git clone https://github.com/SanjoDeundiak/carthage-bug-sample
carthage update virgil-sdk-x --no-use-binaries

jwutke:
This fixes #2188. 

I have essentially no experience with ReactiveSwift, so chances are there improvements that can be made to this. 

I haven't taken a look at the tests yet, but I can add some if needed. I should have time later this week for it.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -813,7 +813,21 @@ public final class Project { // swiftlint:disable:this type_body_length
 		return loadResolvedCartfile()
 			.map { resolvedCartfile -> [(Dependency, PinnedVersion)] in
 				return resolvedCartfile.dependencies
-					.filter { dep, _ in dependenciesToCheckout?.contains(dep.name) ?? true }
+					.filter { dep, _ in dependenciesToCheckout?.contains(dep.name) ?? false }
+			}
+			.flatMap(.merge) { dependencies -> SignalProducer<[String], CarthageError> in
+				return SignalProducer<(Dependency, PinnedVersion), CarthageError>(dependencies)
+					.flatMap(.merge) { dependency, version -> SignalProducer<(Dependency, VersionSpecifier), CarthageError> in
+						return self.dependencies(for: dependency, version: version)
+					}
+					.map { $0.0.name }
+					.collect()
+			}
+			.zip(with: loadResolvedCartfile())
+			.map { subDependencies, resolvedCartfile -> [(Dependency, PinnedVersion)] in
+				let dependencies = (dependenciesToCheckout ?? []) + subDependencies
+				return resolvedCartfile.dependencies
+					.filter { dep, _ in dependencies.isEmpty ? true : dependencies.contains(dep.name) }
 			}
 			.zip(with: submodulesSignal)
 			.flatMap(.merge) { dependencies, submodulesByPath -> SignalProducer<(), CarthageError> in
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -362,6 +362,26 @@ public final class Project { // swiftlint:disable:this type_body_length
 		}
 	}
 
+	/// Finds all the transitive dependencies for the dependencies to checkout.
+	func transitiveDependencies(
+		_ dependenciesToCheckout: [String]?,
+		resolvedCartfile: ResolvedCartfile
+	) -> SignalProducer<[String], CarthageError> {
+		return SignalProducer(value: resolvedCartfile)
+			.map { resolvedCartfile -> [(Dependency, PinnedVersion)] in
+				return resolvedCartfile.dependencies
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -813,7 +813,21 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+						return self.dependencies(for: dependency, version: version)
+					}
+					.map { $0.0.name }
+					.collect()
+			}
+			.zip(with: loadResolvedCartfile())
+			.map { subDependencies, resolvedCartfile -> [(Dependency, PinnedVersion)] in
+				let dependencies = (dependenciesToCheckout ?? []) + subDependencies
+				return resolvedCartfile.dependencies
+					.filter { dep, _ in dependencies.isEmpty ? true : dependencies.contains(dep.name) }
```
blender: @jwutke  Thanks for the great work!

Would you be open to split this processing of transitive dependencies in a separate testable function that can be covered by tests in `ProjectSepc.swift` ?

I think it would make this contribution even more substantial and future proof.

blender:
@jwutke ping

Source/CarthageKit/Project.swift
jwutke: Yes, I will look into this. Sorry for the delayed response, apparently I did not have email notifications enabled.

blender:
@jwutke Did you find time to make any progress?

blender:
Thanks, the code looks good to me. I confirmed that this indeed fixes:

- https://github.com/Carthage/Carthage/issues/2349
- https://github.com/Carthage/Carthage/issues/2188

Source/CarthageKit/Project.swift
```diff
@@ -811,9 +831,14 @@ public final class Project { // swiftlint:disable:this type_body_length
 			}
 
 		return loadResolvedCartfile()
-			.map { resolvedCartfile -> [(Dependency, PinnedVersion)] in
+			.flatMap(.latest) { resolvedCartfile in
+				return self.transitiveDependencies(dependenciesToCheckout, resolvedCartfile: resolvedCartfile)
+			}
+			.zip(with: loadResolvedCartfile())
```
mdiep: This loads the resolved cartfile again. Can you reuse `resolvedCartfile` inside the `flatMap` instead?

```swift
.flatMap(.latest) { resolvedCartfile in
    return self
        .transitiveDependencies(dependenciesToCheckout, resolvedCartfile: resolvedCartfile)
        .map { ($0, resolvedCartfile) }
}
```

That'll save reading the cartfile from disk again.

Source/CarthageKit/Project.swift
```diff
@@ -811,9 +831,14 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+			.map { subDependencies, resolvedCartfile -> [(Dependency, PinnedVersion)] in
+				let dependencies = (dependenciesToCheckout ?? []) + subDependencies
 				return resolvedCartfile.dependencies
-					.filter { dep, _ in dependenciesToCheckout?.contains(dep.name) ?? true }
+					.filter { dep, _ in dependencies.isEmpty ? true : dependencies.contains(dep.name) }
```
mdiep: I think we should keep this line the same, where `nil` means everything.

Probably the best way to do that would be to add these inside the `flatMap`.

```swift
.flatMap(.latest) { resolvedCartfile in
    guard let dependenciesToCheckout = dependenciesToCheckout else {
        return SignalProducer(value: (nil, resolvedCartfile))
    }
    return self
        .transitiveDependencies(dependenciesToCheckout, resolvedCartfile: resolvedCartfile)
        .map { (dependenciesToCheckout + $0, resolvedCartfile) }
}
.map { dependenciesToCheckout, resolvedCartfile in
    return resolvedCartfile.dependencies
        .filter { dep, _ in dependenciesToCheckout?.contains(dep.name) ?? true }
}

mdiep:
Awesome. Looks correct, but I think the code could be improved a little. 👍

blender:
@jwutke  ping

Source/CarthageKit/Project.swift
jwutke: This makes sense. It didn't feel right to load the Cartfile again, but I didn't know a better way.

blender:
@mdiep ready for review.

blender:
@jwutke Thanks! Awesome contribution!

PR #2328, opened 2018-01-19
[Bug] Fix a bug that carthage outdated fails

chuganzy:
This PR resolves an issue on master branch that `carthage outdated` fails when git ref is used in Cartfile and no release exists for dependency.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -438,7 +438,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 		}
 	}
 
-	public typealias OutdatedDependency = (Dependency, PinnedVersion, PinnedVersion, PinnedVersion?)
+	public typealias OutdatedDependency = (Dependency, PinnedVersion, PinnedVersion, PinnedVersion)
 	/// Attempts to determine which of the project's Carthage
 	/// dependencies are out of date.
 	///
@@ -474,8 +474,13 @@ public final class Project { // swiftlint:disable:this type_body_length
 			.map { ($0.dependencies, $1.dependencies, $2) }
 			.map { (currentDependencies, updatedDependencies, latestDependencies) -> [OutdatedDependency] in
 				return updatedDependencies.flatMap { (project, version) -> OutdatedDependency? in
-					if let resolved = currentDependencies[project], resolved != version {
-						let latest = latestDependencies[project]
+					if let resolved = currentDependencies[project], let latest = latestDependencies[project], resolved != version || resolved != latest {
+						if SemanticVersion.from(resolved).value == nil, version == resolved {
+							// If resolved version is not a semantic version but a commit
+							// it is a false-positive if `version` and `resolved` are the same
+							return nil
+						}
+
 						return (project, resolved, version, latest)
 					} else {
 						return nil
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -409,10 +409,18 @@ public final class Project { // swiftlint:disable:this type_body_length
 	private func latestDependencies(resolver: ResolverProtocol) -> SignalProducer<[Dependency: PinnedVersion], CarthageError> {
 		return loadResolvedCartfile()
 			.map { $0.dependencies.mapValues { _ in VersionSpecifier.any } }
-			.flatMap(.merge) { resolver.resolve(dependencies: $0, lastResolved: nil, dependenciesToUpdate: nil) }
+			.flatMap(.merge) { dependencies -> SignalProducer<[Dependency: PinnedVersion], NoError> in
+				SignalProducer(dependencies)
+					.flatMap(.merge) { key, value -> SignalProducer<[Dependency: PinnedVersion], NoError> in
+						resolver
+							.resolve(dependencies: [key: value], lastResolved: nil, dependenciesToUpdate: nil)
+							.flatMapError { _ in .empty }
+					}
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -409,10 +409,18 @@ public final class Project { // swiftlint:disable:this type_body_length
...
-			.flatMap(.merge) { resolver.resolve(dependencies: $0, lastResolved: nil, dependenciesToUpdate: nil) }
+			.flatMap(.merge) { dependencies -> SignalProducer<[Dependency: PinnedVersion], NoError> in
+				SignalProducer(dependencies)
+					.flatMap(.merge) { key, value -> SignalProducer<[Dependency: PinnedVersion], NoError> in
+						resolver
+							.resolve(dependencies: [key: value], lastResolved: nil, dependenciesToUpdate: nil)
+							.flatMapError { _ in .empty }
+					}
+					.reduce([Dependency: PinnedVersion]()) { $0.merging($1) { $1 } }
+		}
```
mdiep: Unfortunately, this isn't correct. This method should return the latest set of _compatible_ dependencies.

So if these versions are available:

- A
  - 1.0
  - 2.0
- B
  - 1.0: `A ~> 1.0`

Then it should return `A: 1.0, B: 1.0`, not `A: 2.0, B: 1.0`. The versions of the dependencies can't be considered in isolation.

I don't have an easy solution to this. Probably the easiest thing would be to:

1. Load the combined `Cartfile` in addition to `Cartfile.resolved`
2. If a dependency is specified with a `VersionSpecifier.gitReference`, then use it; otherwise use `.any` (what's currently being used)

That would prevent `outdated` from failing. It's probably not what we'd actually want, which would be to still show versions if possible.

Or… maybe it would be better to:

1. `resolve`
2. catch `CarthageError.requiredVersionNotFound`
    1. pin the erring dependency
    2. Go to (1)

That would be slower in the worst case, but seems like it would return the correct results in all cases. 🤔

mdiep:
Thanks for opening a PR to try to fix this!

Source/CarthageKit/Project.swift
chuganzy: Well, I think this method is to "Attempts to determine the latest version (whether satisfiable or not)" as written here https://github.com/Carthage/Carthage/pull/2328/files#diff-3bd7f6198aa606c5ec6931e77cb3c42bR404. Isn't it correct?

chuganzy: Ah, I misunderstood this method ignores version compatibility😱 I'll fix it.

chuganzy: Addressed https://github.com/Carthage/Carthage/pull/2328/commits/9d96115fe235c414e72876ba3fb9d7cf5357c052.

Source/CarthageKit/Project.swift
```diff
@@ -407,9 +407,35 @@ public final class Project { // swiftlint:disable:this type_body_length
...
-			.flatMap(.merge) { resolver.resolve(dependencies: $0, lastResolved: nil, dependenciesToUpdate: nil) }
+		func resolve(prefersGitReference: Bool) -> SignalProducer<[Dependency: PinnedVersion], CarthageError> {
+			return SignalProducer
+				.combineLatest(loadCombinedCartfile(), loadResolvedCartfile())
+				.map { cartfile, resolvedCartfile in
+					resolvedCartfile
+						.dependencies
+						.reduce(into: [Dependency: VersionSpecifier]()) { result, group in
+							let dependency = group.key
+							switch (prefersGitReference, cartfile.dependencies[dependency]) {
```
mdiep: I think this would be clearer as an `if` instead of a `switch`.

Source/CarthageKit/Project.swift
```diff
@@ -407,9 +407,35 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+							}
+					}
+				}
+				.flatMap(.merge) {
+					resolver.resolve(dependencies: $0, lastResolved: nil, dependenciesToUpdate: nil)
+			}
+		}
+
+		return resolve(prefersGitReference: false).flatMapError {
+			switch $0 {
```
mdiep: Please use a named variable when a closure is over multiple lines.

Source/CarthageKit/Project.swift
```diff
@@ -407,9 +407,35 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+				.map { cartfile, resolvedCartfile in
+					resolvedCartfile
+						.dependencies
+						.reduce(into: [Dependency: VersionSpecifier]()) { result, group in
+							let dependency = group.key
+							switch (prefersGitReference, cartfile.dependencies[dependency]) {
+							case (true, .gitReference(let value)?):
+								result[dependency] = .gitReference(value)
+							default:
+								result[dependency] = .any
```
mdiep: I'd also like to see the assignment to `result[dependency]` in a single place.

```swift
let specifier: VersionSpecifier
if case let .gitReference(value)? = cartfile.dependencies[dependency], prefersGitReference {
    specifier = .gitReference(value)
} else {
    specifier = .any
}
result[dependency] = specifier
```

mdiep:
The general idea here looks good! Thanks for working on this!

Source/CarthageKit/Project.swift
chuganzy: Addressed.

mdiep:
Thanks again!

Issue #902, opened 2015-11-06
Cleanup command


nanoxd:
Thanks for all the hard work on this project :smile:.

With the addition of generating `bcsymbolmap` files, I've seen an influx of things in my Carthage build folders. Before the introduction of the `bcsymbolmap` files, it was easy to go in there myself and delete the files. That is no longer the case with symbol maps using a UUID unless you remove the entire Carthage folder and run `carthage bootstrap`

I suggest an additional command be added to cleanup files for when a user wants to remove the files from their Carthage build/checkout folders that targets the `.framework`, `.bcsymbolmap`, and `.dSYM` artifacts.

ffittschen:
I totally agree! Since the addition of the `bcsymbolmap` files the folder looks like a mess... The `Carthage/Build` folder is on my `.gitignore` but I still like to have things clean in the repo.

The deletion of unneeded files should be automated...

<img width="671" alt="screen shot 2016-04-14 at 10 09 53" src="https://cloud.githubusercontent.com/assets/7734806/14521363/34deffe8-0229-11e6-8b3e-f148f317314b.png">

mdiep:
Maybe Carthage should delete old `.bcsymbolmap`s when building?

ffittschen:
@mdiep This would be perfect 👍🏻
If you use `carthage build` without the specification of a project, all `.bcsymbolmap`s should be deleted, but if you use e.g. `carthage build Alamofire` it should just delete the corresponding `.bcsymbolmap`

BProg:
I would also need this feature for cleanup `bcsymbolmap` files because we keep them in git, so the someone who downloads the project don't need to use `carthage` for the project to run. A `cleanup` command will be very handy

ikesyo:
There is a ongoing PR #2333.

ravitripathi:
Is this still being worked on?

idanzhu:
not find out command `carthage cleanup`, it had been merged ?

ravitripathi:
It has been merged, but carthage cleanup doesn't work as of 0.34.0.

yourshinsuke:
@ravitripathi me too. When??

ravitripathi:
@yourshinsuke no clue. It seems like the feature was never exposed. Weird. We ended up commit those bsymbolmaps too along with the framework files after updating.

PR #2333, opened 2018-01-21
[WIP] Add `carthage cleanup` command

chuganzy:
Add `carthage cleanup` command.
Summary is [here](https://github.com/Carthage/Carthage/pull/2333#issuecomment-369927726).

Addresses https://github.com/Carthage/Carthage/issues/902.

## TODOs
- [x] Error handling
- [x] Logging
- [x] Test


blender:
@chuganzy ping

chuganzy:
@blender Sorry I was busy these days. I'll keep working on this this weekend!

blender:
@chuganzy nice progress you have made.

I have read the linked issue but I'm not sure I follow your implementation.

This adds a new cleanup command that deletes:

- Binaries
- dSYMs
- version files

however it seems to leave bcsymbolmaps? 

Also this seems to halt on the first version file missing. 

Can you go in a little more detail about your implementation?

chuganzy:
This command deletes:
* Binaries
* dSYMs
* version files
which are no longer needed.

I thought all of bcsymbolmaps are not actually used but apparently they are needed to upload the app to iTunes Connect.. I'll figure out how to clear bcsymbolmaps which is no longer used..

blender:
the bcsymbolmaps can be derived from the dsyms or form the framework binaries. There should be code for that.

However I must ask, why would one use this command and not just delete the Carthage/Build folder?

chuganzy:
On CI we cache Carthage directory and run `carthage bootstrap --use-cache` every time. The problem is that carthage keeps old files which are no longer used, so Carthage directory is getting bigger.. Also I want to keep build cache as far as possible.

blender:
So if I understand correctly this would make sure that the artifacts in the build folder are only the ones that should produces according to the current Cartfile.resolved?

chuganzy:
Exactly

blender:
@chuganzy I'm good with that even if it sounds a little more of a workflow problem than a Carthage problem. Should this apply to the checkout folder too?

I think we go the discussion at a good point here and I'm going to summarize for @mdiep:

### Goal
Carthage should be able to clean after itself.

### Motivation
When changing dependencies, a lot of artifacts from non-relevant dependencies are left behind cluttering the `Carthage/Build` folder. This can be solved by removing the `Carthage/Build` folder but this would discard relevant artifacts as well. A smart clean-up is required.

### Current limitations
No limitations

### Proposed solution
Add a cleanup command to make sure that the artifacts in the build folder are only the ones that would be produced according to the current Cartfile.resolved.

mdiep:
I'm good with this if we can correctly figure out which things are no longer needed. 👍

blender:
We are able to figure out the complementary set of things (what to keep) so we should be good.

chuganzy:
Thanks @blender for summarizing!

blender:
@chuganzy I think the cleanup should apply to everything, including the checkouts.

blender:
@chuganzy  ping

chuganzy:
Sorry I could not work on this for a while.. Currently it clears all the `bcsymbolmap` but I'm going to refactor the part. Please hold on!

blender:
ping

blender:
@chuganzy ping

blender:
@chuganzy another ping

chuganzy:
Sorry I had no time to work on this for awhile.. I will continue working on this!

ChristianSteffens:
pong

ravitripathi:
Any updates on this? The build folder is a mess as of now.

sidepelican:
What are the remain works?

chuganzy:
Sorry I have been keeping this open for awhile.. Actually a lot has changed since I opened this PR, and now I have to also put static builds into consideration.. Probably redoing this from scratch..

sidepelican:
I’m working for remaining tasks in this pr. 
fixing 
- bcsymbolmaps
- static framework 
- follow the master branch

Issue #2280, opened 2017-12-05
Carthage strips symbols from dSYM files


naan:
* `carthage version`:  0.26.2
* `xcodebuild -version`: Xcode 9.1 Build version 9B55
* Are you using `--no-build`? No
* Are you using `--no-use-binaries`?  No
* Are you using `--use-submodules`?  No
* Are you using `--cache-builds`?  No

**Cartfile**
N/A

Since Xcode 9, (this is my guess, I can't really tell exactly when this started happen) when I build a framework with Carthage for macOS, dSYM file gets empty even tho Xcode builds it correctly.

Here's what's happening in my log:

- Xcode create dSYM file using `dsymutil` (Default value of `DEBUG_INFORMATION_FORMAT` for release build is `DWARF with dSYM file`
- Xcode then strip symbol (Default value of `STRIP_INSTALLED_PRODUCT` for release build is `Yes`)
- Then Carthage tries to create a dSYM file from the stripped framework, which generates empty dSYM file. (inside of `createDebugInformation` function??)

However, iOS version of framework has a correct dSYM file even the framework itself is stripped. 

I haven't changed any related settings when I updated my project to Xcode 9, but I did change Xcode's recommended changes which might effect something here.

To avoid this, I have to disable `STRIP_INSTALLED_PRODUCT` to create correct dSYM file, then strip the symbol with a separate script.

abrindam:
+1, I am seeing what appears to be the same issue.

When I use Carthage to build frameworks for Mac, I do get a dSYM file but it's very small and doesn't seem to allow for debugging. I can't step into sources, and in Xcode, Debug -> Debug Workflow -> Shared Libraries shows "Debug Symbols" as "no" for any framework built by Carthage.

I reverted to Carthage 0.25 and the problem went away.

I strongly suspect this is caused by the changes from PR #2158 which switched to using "archive" instead of "build" to build frameworks. The "archive" command seems to strip the debug symbols before the dSYM can be created, as described by OP.

Unfortunately, I just convinced my team that breaking up our code into frameworks is the way to go, and now we can't debug anything because of this issue. So there is a minor torches and pitchforks situation. So any help would be greatly appreciated! :-)

abrindam:
#2295 may also be a dup of this?

horse304:
I got the same issue, and after a bit of debug, my suspect is:
- since v26 Carthage, the way build framework is change from "xcodebuild build" -> "xcodebuild archive"
[...]

PR #2334, opened 2018-01-22
Disable the "Strip Linked Product" build setting when using the "arch…

abrindam:
Fix for #2280 and friends.

Disable the "Strip Linked Product" build setting (STRIP_INSTALLED_PRODUCT), when using the "archive" command to build.

This keeps the debug information around so a valid dSYM can be generated.

This change is necessary since Carthage moved to using the "archive" command to build. Most products have `STRIP_INSTALLED_PRODUCT` set to true, but this does not kick in unless you have `DEPLOYMENT_POSTPROCESSING` set to true as well, and that is false for most products. However, using "archive" instead of "build" magically overrides `DEPLOYMENT_POSTPROCESSING` to true, resulting in the product being stripped of debug symbols. As a result dSYM generation produces effectively no information for builds that are built with "archive" (e.g. OSX and iOS device, but not iOS simulator)

This also restores the 0.25 behavior that built frameworks are not stripped. Prior to fix, stripping was happening with the archive command, which may have caused additional issues that were not reported. With this change, we are back to the old behavior and stripping is done at framework copy time.

Testing done:
1. Build a framework using Carthage
2. Use "grep" to check the produced framework for a known source file name.
3. Use "grep" to check the produced dSYM for a known source file name
4. Run an app that uses this framework in Xcode. Open Debug->Debug Workflow->Shared Libraries and ensure the framework has "Debug Symbols" as "Yes"

Prior to fix, all checks above failed, and post fix, all checks above succeed.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -650,6 +650,10 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
 								// setting for GCC as noted in
 								// https://developer.apple.com/library/content/qa/qa1964/_index.html.
 								"CLANG_ENABLE_CODE_COVERAGE=NO",
+
+								// Disable the "Strip Linked Product" build
+								// setting so we can later generate a dSYM
+								"STRIP_INSTALLED_PRODUCT=NO",
 							]
 						}
 


blender:
Hi,

please correct me if I am wrong but I don't see that frameworks will be stripped from debug symbols in the copy-framework phase.

To recap, the release builds of frameworks will now contain desymbolication information by default which might be not preferred by most.

abrindam:
Hi @blender,

It appears you are correct, `carthage copy-frameworks` does not strip debug symbols. I was wrong. Therefore you are correct that in that case the release build will contain debug symbols.

However, I have tested with Carthage 0.25 and this behavior occurs there as well. So I would argue this is not a regression but a restoration to the way things used to work. Changing when debug symbols were stripped was not the intent of the changes that went in between 0.25 and 0.26 (the "archive" change, for short). So I feel it makes sense to put things back to the way they were and then move forward from there.

In other words, nobody complained about the debug symbols when they were included before. However, many are complaining about the invalids dSYMs now. So from a priority perspective I believe the cleanest thing to do is to revert to the 0.25 model.

Additionally, there were comments in the linked issue #2280 that indicated people *wanted* the built frameworks (in `Carthage/Build`) to contain debug information for their own convenience. So this argues for a strip-on-copy rather than strip-on-build approach.

If we want to fix the problem you're raising, I think it would be appropriate for `carthage copy-frameworks` to strip debug symbols based on the Xcode setting `COPY_PHASE_STRIP`. This setting affects the vanilla "Copy Files" phase in Xcode, and as `carthage copy-frameworks` is attempting to be "Copy Files On Steroids", it seems appropriate to honor that as well.

blender:
You read my mind.


Would you be open to try to change copy-frameworks?

This way the consumer of the frameworks can decide whether to strip the symbols or not during copy. I imagine that in debug one would set `COPY_PHASE_STRIP` to no strip and on release to yes.

This would be an elegant way to solve the problem.


If you don't feel like changing this now I'm fine with restoring the old behavior.

abrindam:
Let me give it a try. It's a bit more involved change that this current pull request, but I'm game.

BTW, from what I see `COPY_PHASE_STRIP` is on for release and off for debug by default in Xcode, just as you theorized.

blender:
Thanks, that is very kind of you.

Just as a heads up including tests for the change would be a nice idea. I know other maintainers are going to look for them too when reviewing the PR ;)

mdiep:
Does archiving the app strip the source code from the embedded frameworks?

blender:
@mdiep I just tried building Alamofire with `carthage v0.28.0`, adding a copy-framework phase, archiving an app and it seems that it does strip the symbols. Nice catch.

How to verify: 

```
$otool -tv Carthage/Build/iOS/Alamofire.framework/Alamofire | grep Alamo # Brace for impact
...
$ otool -t "/Users/blender/Library/Developer/Xcode/Archives/2018-02-08/TestsApp 08.02.18, 23.02.xcarchive/Products/Applications/TestsApp.app/Frameworks/Alamofire.framework/Alamofire" | grep Alamo
# no trace
```
So seems that there isn't additional work here @abrindam.

abrindam:
OK, great. Hopefully we can move forward with this pull request then.

I actually coded something up already that seems to work, for symmetry with the standard Xcode copy files we might want to include it anyways. I'll open a separate issue/PR however given this is no longer critical.

blender:
@mdiep  @abrindam 

Hello, sorry to break the news but I just tried with the latest master and archiving the app does not strip the symbols from embedded frameworks.

I got mislead by the fact that before this merge there were no symbols at all for the arm slice so when I checked the embedded library I saw no symbols because it was `lipoed` from the x84 arch which was the only slice with symbols to begin with.

So long story short, the behavior is now restored to `v0.25` but imho the `copy-framework` changes are still needed and relevant.

abrindam:
OK @blender I'll open a separate request with my changes to strip on copy. I'm unfortunately about to embark on a 21 hour flight to India so it'll be a bit before I can get it posted...

blender:
Have a safe flight. No worries.

blender:
@abrindam ping

abrindam:
Hey folks, apologies for totally disappearing. Been a crazy couple weeks. 

I've been soaking my changes by using my local copy of Carthage and I haven't seen any issues. Will open a pull request momentarily for my changes.

PR #2336, opened 2018-01-25
New debugger entitlement on Xcode 9.3

aaroncrespo:
This entitlement is stripped normally via the organizer window. Which might not be used in some Carthage workflows. 
> The entitlement is stripped from apps distributed using the Organizer window.

The PR should maybe be held and link should maybe be replaced with the man page entry until 9.3 is GM?

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -650,6 +650,10 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
 								// setting for GCC as noted in
 								// https://developer.apple.com/library/content/qa/qa1964/_index.html.
 								"CLANG_ENABLE_CODE_COVERAGE=NO",
+								
+								// Disable injecting the debugger entitlement into archive builds
+								// https://download.developer.apple.com/Developer_Tools/Xcode_9.3_beta/Release_Notes_for_Xcode_9.3_beta.pdf
+								"CODE_SIGN_INJECT_BASE_ENTITLEMENTS=NO",
 							]
 						}
 


Source/CarthageKit/Xcode.swift
```diff
@@ -650,6 +650,10 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
 								// setting for GCC as noted in
 								// https://developer.apple.com/library/content/qa/qa1964/_index.html.
 								"CLANG_ENABLE_CODE_COVERAGE=NO",
+								
+								// Disable injecting the debugger entitlement into archive builds
+								// https://download.developer.apple.com/Developer_Tools/Xcode_9.3_beta/Release_Notes_for_Xcode_9.3_beta.pdf
```
mdiep: This link doesn't actually work, so it's not of much use.

mdiep:
Wouldn't this prevent you from debugging your app then? 😕

Source/CarthageKit/Xcode.swift
aaroncrespo: I don't know what apple is doing that link it's what shows in the address bar when I view the patch notes with a logged int account. Since it might not be public until GM here is the section that explains this new build flag. 


>The debugger on macOS now requires the entitlement com.apple.security.get-task- allow to attach to apps built for macOS or for iOS, tvOS or watchOS apps built for Simulator. Xcode automatically injects this entitlement to your builds. The entitlement is stripped from apps distributed using the Organizer window.
>
>To disable Xcode injecting the entitlement, set CODE_SIGN_INJECT_BASE_ENTITLEMENTS to NO in the build settings for the target or app. (34638816)

aaroncrespo:
This section of build settings only affect archive builds IIRC. If you want to retain debugging of archived artifacts AND I interpret the change in Xcode correctly. Then apps will probably not be allowed to submit to the App Store with builds produced from `carthage archive` 

Apple will probably see this existing entitlement and reject due to diagnostic configurations existing in the build.

Source/CarthageKit/Xcode.swift
mdiep: I think they issue you a temporary permission via a redirect when you click on a link from elsewhere.

mdiep:
Carthage uses `archive` to build copies of its frameworks. This would seem to prevent debugging apps that uses frameworks built with Carthage.

But Carthage can't be used to build apps themselves. So building your app for the App Store should be where this entitlement is stripped AFAICT. In other words, I don't Carthage should need to do anything.

aaroncrespo:
we are seeing `CODE_SIGN_INJECT_BASE_ENTITLEMENTS=YES` injected into framework projects built with `xcodebuild` by running `xcodebuild -showBuildSettings | grep INJECT_BASE_ENTITLEMENTS`

Source/CarthageKit/Xcode.swift
aaroncrespo: I hope they have a public link to some documentation ready as a 9.3 release gets closer.

aaroncrespo:
We are still investigating this as I said in the initial PR this might not be ready for merge (Still beta and all) We are still trying to figure out all the repercussions of it.

PR #2342, opened 2018-02-03
Move binary download to build process

chuganzy:
As titled. It got more straightforward✨

Resolves: https://github.com/Carthage/Carthage/issues/2308
Related: https://github.com/Carthage/Carthage/pull/2317#issuecomment-362600637

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -1021,6 +1021,8 @@ public final class Project { // swiftlint:disable:this type_body_length
 					}
 					.collect()
 					.map { installedDependencies -> [(Dependency, PinnedVersion)] in
+						// Filters out dependencies that we've downloaded binaries for
+						// but preserves the build order
 						return dependencies.filter { dependency -> Bool in
 							!installedDependencies.contains { $0 == dependency }
 						}
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -831,27 +831,9 @@ public final class Project { // swiftlint:disable:this type_body_length
 					.flatMap(.concurrent(limit: 4)) { dependency, version -> SignalProducer<(), CarthageError> in
 						switch dependency {
 						case .git, .gitHub:
-
-							let submoduleFound = submodulesByPath[dependency.relativePath] != nil
-							let checkoutOrCloneDependency = self.checkoutOrCloneDependency(dependency, version: version, submodulesByPath: submodulesByPath)
-
-							// Disable binary downloads for the dependency if that
-							// is already checked out as a submodule.
-							if submoduleFound {
-								return checkoutOrCloneDependency
-							}
-
-							return self.installBinaries(for: dependency, pinnedVersion: version, toolchain: buildOptions?.toolchain)
-								.flatMap(.merge) { installed -> SignalProducer<(), CarthageError> in
-									if installed {
-										return .empty
-									} else {
-										return checkoutOrCloneDependency
-									}
-								}
-
-						case let .binary(url):
-							return self.installBinariesForBinaryProject(url: url, pinnedVersion: version, projectName: dependency.name, toolchain: buildOptions?.toolchain)
+							return self.checkoutOrCloneDependency(dependency, version: version, submodulesByPath: submodulesByPath)
+						case .binary:
+							return .empty
 						}
 					}
 			}
@@ -1025,7 +1007,23 @@ public final class Project { // swiftlint:disable:this type_body_length
 				}
 			}
 			.flatMap(.concat) { dependencies -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
-				return .init(dependencies)
+				return SignalProducer(dependencies)
+					.flatMap(.concurrent(limit: 4)) { dependency, version -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
+						switch dependency {
+						case .git, .gitHub:
+							guard options.useBinaries else { return .init(value: (dependency, version)) }
+							return self.installBinaries(for: dependency, pinnedVersion: version, toolchain: options.toolchain)
+								.filterMap { installed -> (Dependency, PinnedVersion)? in
+									guard installed else { return (dependency, version) }
+									return nil
+							}
+						case let .binary(url):
+							return self.installBinariesForBinaryProject(url: url, pinnedVersion: version, projectName: dependency.name, toolchain: options.toolchain)
+								.then(SignalProducer<(Dependency, PinnedVersion), CarthageError>.empty)
+						}
+					}
+					.collect()
+					.flatten()
 			}
 			.flatMap(.concat) { dependency, version -> SignalProducer<BuildSchemeProducer, CarthageError> in
 				let dependencyPath = self.directoryURL.appendingPathComponent(dependency.relativePath, isDirectory: true).path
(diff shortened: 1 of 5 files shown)


Source/carthage/Update.swift
```diff
@@ -73,22 +72,13 @@ public struct UpdateCommand: CommandProtocol {
...
 			return checkoutOptions.loadProject()
-				.on(value: { project in
-					// Never check out binaries if 
-					// 1. we're skipping the build step, or
-					// 2. `--toolchain` option is given
-					// because that means users may need the repository checkout.
-					if !self.buildAfterUpdate || self.buildOptions.toolchain != nil {
-						project.useBinaries = false
-					}
-				})
```
mdiep: I think we need to keep this logic, even though `--use-binaries` has moved to `build`.

Source/CarthageKit/Project.swift
```diff
@@ -1036,7 +1007,24 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+								.filterMap { installed -> (Dependency, PinnedVersion)? in
+									return installed ? nil : (dependency, version)
+							}
+						case let .binary(url):
+							return self.installBinariesForBinaryProject(url: url, pinnedVersion: version, projectName: dependency.name, toolchain: options.toolchain)
+								.then(SignalProducer<(Dependency, PinnedVersion), CarthageError>.empty)
+						}
+					}
+					.collect()
+					.flatten()
```
mdiep: `collect` and `flatten` cancel each other out; these can be removed.

mdiep:
This definitely looks a lot cleaner. 👏 

The code looks good, but I'll need to test this out after the above items are fixed.

Source/carthage/Update.swift
chuganzy: Why do you think so? From now it always checkouts repositories though.

Source/CarthageKit/Project.swift
chuganzy: Well, I thought we should wait all download processes to complete before building as some dependencies might need them to build. Isn't it correct?

mdiep: That's not necessary.

The dependencies are order from leaves to root—i.e. the dependencies always come first. The next line after this is `flatMap(.concat)`. Because we've specified the `concat` strategy, dependencies are built one at a time, sequentially. So it's not important that all the downloads finish before we hit that.

But actually… we need to keep the original order of `dependencies`, which might be affected by the `concurrent(limit: 4)`. So we should do `.then(SignalProducer(dependencies))` here instead.

Source/carthage/Update.swift
mdiep: Oh, I guess you're right!

chuganzy: NP!

Source/CarthageKit/Project.swift
chuganzy: [0] A: Pre-built binary is available
[1] B: Pre-built binary is available
[2] C: Pre-built binary is unavailable, has dependency to A and B

In this case, it starts build C as soon as A _or_ B's download has finished.
So I think we need to wait for all the download to be done.

mdiep: But we also need to retain the original order, which this doesn't do. The order matters when building from scratch because each project's dependencies need to already have been built.

chuganzy: Addressed https://github.com/Carthage/Carthage/pull/2342/commits/fabac4ce9226bb4cfdd5509224692f6962b4a6cc

Source/CarthageKit/Project.swift
```diff
@@ -579,49 +575,42 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+						pinnedVersion: pinnedVersion,
+						fromRepository: repository,
+						client: Client(server: server, isAuthenticated: false)
+					)
+				}
+				.flatMap(.concat) {
+					return self.unarchiveAndCopyBinaryFrameworks(zipFile: $0, projectName: dependency.name, pinnedVersion: pinnedVersion, toolchain: toolchain)
+						.on(value: { _ in
+							// Trash the checkouts folder when we've successfully copied binaries, to avoid building unnecessarily
+							_ = try? FileManager.default.trashItem(at: checkoutDirectoryURL, resultingItemURL: nil)
```
mdiep: We should get rid of this.

The current behavior is to delete the checkout for a dependency after successfully using a downloaded binary. This exists because previously the existence of the checkout directory determined whether or not that dependency was built.

Now that downloading binaries is part of build, we should get right of that line to delete the checkout. It looks like `buildCheckedOutDependenciesWithOptions` can already handle this.

Source/CarthageKit/Project.swift
```diff
@@ -1036,7 +1007,29 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+						case let .binary(url):
+							return self.installBinariesForBinaryProject(url: url, pinnedVersion: version, projectName: dependency.name, toolchain: options.toolchain)
+								.then(.init(value: (dependency, version)))
+						}
+					}
+					.collect()
+					.map { installedDependencies -> [(Dependency, PinnedVersion)] in
+						return dependencies.filter { dependency -> Bool in
+							!installedDependencies.contains { $0 == dependency }
+						}
```
mdiep: This should be able to be `return dependencies.filter { !installedDependencies.contains($0) }`

Source/CarthageKit/Project.swift
```diff
@@ -1036,7 +1007,29 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+								.filterMap { installed -> (Dependency, PinnedVersion)? in
+									return installed ? (dependency, version) : nil
+							}
+						case let .binary(url):
+							return self.installBinariesForBinaryProject(url: url, pinnedVersion: version, projectName: dependency.name, toolchain: options.toolchain)
+								.then(.init(value: (dependency, version)))
+						}
+					}
+					.collect()
+					.map { installedDependencies -> [(Dependency, PinnedVersion)] in
```
mdiep: Can you add a comment here explaining that this filters out dependencies that we've downloaded binaries for but preserves the build order?

mdiep:
Sorry it took me so long to get back to this. I'll be more prompt this time around. 😓

Source/CarthageKit/Project.swift
chuganzy: The type of `installedDependencies` is `[(Dependency, PinnedVersion)]` and `dependency` is `(Dependency, PinnedVersion)` here so cannot use `.contains(_ element: Element)` here because Element is not technically equatable.

chuganzy:
@mdiep NP! Thanks for your review. Addressed / replied to your comments!

Source/carthage/Checkout.swift
```diff
@@ -10,42 +10,31 @@ public struct CheckoutCommand: CommandProtocol {
...
 		private init(useSSH: Bool,
 		             useSubmodules: Bool,
-		             useBinaries: Bool,
 		             colorOptions: ColorOptions,
 		             directoryPath: String,
 		             dependenciesToCheckout: [String]?
 		) {
-			// Disable binary downloads when using submodules.
-			// See https://github.com/Carthage/Carthage/issues/419.
-			let shouldUseBinaries = useSubmodules ? false : useBinaries
```
mdiep: Nice! Glad we could get rid of this. 🤘

mdiep:
Looks good. This is a great improvement. Thanks @chuganzy!

chuganzy:
Yay! Thanks for your review!

mdiep:
Please open a new issue

PR #2343, opened 2018-02-03
Limit parallelism in discovery of build settings

flovilmart:
Currently, on large workspaces, one xcodebuild process is spawned per scheme. This saturates the build host with xcodebuild processes.

This is a proposal to limit the max number of xcodebuild processes to 4 at the buildSettings resolution step.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -114,11 +114,9 @@ public func buildableSchemesInDirectory(
 		.flatMap(.concat) { project -> SignalProducer<(ProjectLocator, [Scheme]), CarthageError> in
 			return project
 				.schemes()
-				.flatMap(.merge) { scheme -> SignalProducer<(Scheme, BuildArguments), CarthageError> in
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -114,9 +114,11 @@ public func buildableSchemesInDirectory(
 		.flatMap(.concat) { project -> SignalProducer<(ProjectLocator, [Scheme]), CarthageError> in
 			return project
 				.schemes()
-				.flatMap(.merge) { scheme -> SignalProducer<Scheme, CarthageError> in
+				.flatMap(.merge) { scheme -> SignalProducer<(Scheme, BuildArguments), CarthageError> in
```
mdiep: This can just be a `map` now.

flovilmart: I'll put it back in then!

flovilmart:
thanks for the quick review @mdiep !

mdiep:
Awesome! Thanks!

PR #2344, opened 2018-02-04
Only check for buildables schemes once workspace filter has occured

flovilmart:
When building a large project, like the FB SDK, Carthage would initially check for the buildability of each scheme of each project, then filter to only build what was exposed on on the main workspace.

This change attempt to reduce the number of xcodebuild calls by first collecting all possible shared schemes, then letting the xcworkspace filtering to the magic in order to reduce of work being done.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -106,56 +106,64 @@ public func buildableSchemesInDirectory(
 	_ directoryURL: URL,
 	withConfiguration configuration: String,
 	forPlatforms platforms: Set<Platform> = []
-) -> SignalProducer<(ProjectLocator, [Scheme]), CarthageError> {
+) -> SignalProducer<(Scheme, ProjectLocator), CarthageError> {
 	precondition(directoryURL.isFileURL)
-
-	return ProjectLocator
-		.locate(in: directoryURL)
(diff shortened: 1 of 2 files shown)


mdiep:
This is a great idea! 👏 

But I think the approach here is a little backwards. I really don't think we should keep both `buildableSchemesInDirectory` and `schemesInDirectory`—we should have one obvious way to get the list of schemes. (The other use of this function is in `carthage archive`, which needs to pick up all the same schemes.)

So I think instead of moving the buildability check _out_ of the function, we should move the workspace code _in_. That will keep everything using the same function, and I think it will have the nice side-effect of making `buildInDirectory` clearer.

flovilmart:
I was actually wondering for the archive step. 

So in summary IIUC, the buildableSchemesInDirectory should do the workspace reduction instead of the caller.

mdiep:
> So in summary IIUC, the buildableSchemesInDirectory should do the workspace reduction instead of the caller.

Yup! 👍

flovilmart:
Awesome!

flovilmart:
Made the changes but it feels this function is quite crowded now. Let me know if it's what you had in mind. I'm not sure what I can / should strip out the flow.

mdiep:
This is great! I agree that the function is a bit crowded now, but it's an improvement over what we had. Any step in the right direction is ✨!

flovilmart:
Awesome!

PR #2353, opened 2018-02-14
Speed up build command

blender:
Reduce the amount of calls to `xcodebuild` by `min(1, number_of_sdks_in_scheme * n_schemes)` by returning `BuildSettings` and `SDK` at the same time.

### DIFF
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -117,10 +117,10 @@ public struct BuildSettings {
 	///
 	/// If an SDK is unrecognized or could not be determined, an error will be
 	/// sent on the returned signal.
-	public static func SDKsForScheme(_ scheme: Scheme, inProject project: ProjectLocator) -> SignalProducer<SDK, CarthageError> {
+	public static func SDKsForScheme(_ scheme: Scheme, inProject project: ProjectLocator) -> SignalProducer<(BuildSettings,SDK), CarthageError> {
 		return load(with: BuildArguments(project: project, scheme: scheme))
 			.take(first: 1)
-			.flatMap(.merge) { $0.buildSDKs }
+			.flatMap(.merge) { settings in settings.buildSDKs.map { (settings, $0) } }
 	}
 
 	/// Returns the value for the given build setting, or an error if it could
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -453,20 +453,13 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 	)
 
 	return BuildSettings.SDKsForScheme(scheme, inProject: project)
-		.flatMap(.concat) { sdk -> SignalProducer<SDK, CarthageError> in
-			var argsForLoading = buildArgs
-			argsForLoading.sdk = sdk
-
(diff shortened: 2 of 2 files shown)


ikesyo:
Is this actually a safe operation? Looks like the build setting is loaded without setting a sdk as opposed to the before. This may affect some frameworks which uses universal target (a target which supports macOS, iOS, ... by the single target).

blender:
@ikesyo to my understanding it was loaded also before without an SDK just to be loaded again. `BuildSettings.SDKsForScheme` calls `loads` without SDK

ikesyo:
> I just shortcut the second load, the purpose of which is just to get the BuildSettings to filter the SDKs

I'd say the second part with specifying a sdk is important. The build settings for a universal target will be different if different sdk is specified and that will affects the value of `settings.bitcodeEnabled.value`.

blender:
I see, @ikesyo If you have an example of such I can test

ikesyo:
I don't have an example atm unfortunately.

blender:
@ikesyo I tried with Quick and Alamofire everything seems to work fine.

 ENABLE_BITCODE seems to per a per-target setting, not per SDK. Please correct me if I am wrong.

I can add additional SDKs to project (infact iOS targets have both iphoneos and iphonesimulator) but in the target settings ENABLE_BITCODE is a global flag for the target that does not care about the SDK.

ikesyo:
The config can be easily changed by xcconfig file, e.g. https://github.com/ishkawa/APIKit/blob/a9b682850eae516e164e56fed0ce7f0256a57247/Configurations/APIKit.xcconfig#L15-L17

`xcodebuild clean -sdk iphoneos -showBuildSettings -skipUnavailableActions` for a universal target will use `macosx` sdk as a default and if an xccofing file defines `ENABLE_BITCODE = YES` for `macosx` and `ENABLE_BITCODE = NO` for other sdks, this will fail.

ikesyo:
Another viewpoint: I realized that bitcode is not supported on macOS (yet), so `settings.bitcodeEnabled.value == true` will be always `false` and `![.tvOS, .watchOS].contains(sdk)` will be always evaluated. So a universal target's `appletvos` and `watchos` slices may not be build.

blender:
I see, thanks for the insights. What is really needed here is then another way of getting all sdks or getting all bitcode flags at once.

blender:
There is really no way to get around this by using `xcodebuild`. And @ikesyo had a really nice catch here. 

The solution here would be not to rely on `xcodebuild` at all for these sort of discoveries. 

For example carthage uses "xcodebuild -list" to discover schemes when it could just read the shared schemes from the project's directory.

Also, carthage could read the project file itself without relying on xcodebuild.

This however is a greater change and needs some thinking. Would it be a direction the project is interested to take?

mdiep:
I'd prefer to rely on `xcodebuild` than to read undocumented, implementation-detail file formats that can change.

blender:
I think that is a wise direction but as a drawback carthage is dependent on the quirks of xcodebuild.

I'll give it some more though and see if I missed anything here and can make it work some other way. Otherwise I will close the pr.

Thanks everyone for take the time to review.

blender:
Closing the PR since there is really no way around this without calling xcodebuild again. Bummer :(

PR #2355, opened 2018-02-21
Improve Issue Template

blender:
Add some more fields to issue template:

- Carthage install method (something might be wrong with the packaging)
- `which carthage`, the location of the binary might make people realize they are not using the binary they intended to use
- Add --new-resolver filed (some people might not know this is an option)
- Ask to describe what happened and what they expect should happen


mdiep:
Awesome. 👍

PR #2360, opened 2018-02-27
[BinaryProject] Relative file Support

dmiluski:
Related Issue: https://github.com/Carthage/Carthage/issues/1853

### Goal

Support relative file paths for binary (possibly also git?) sources in Cartfile.

### Example:
```
binary "my/domain/com/framework.json"
```

Motivation

Currently only absolute file paths are supported via the `file:// scheme`.

### Example:
```
binary "file:///my/domain/com/framework.json" // this however is absolute.
```

### Current limitations

The file:// scheme does not support relative file paths. See discussion at https://github.com/Carthage/Carthage/pull/2360#issuecomment-368889406

### Proposed solution

Drop the scheme. If the location of the source has no scheme treat it as a file path.

Example:
```
binary "my/domain/com/framework.json" // Relative to cwd
binary "/my/domain/com/framework.json" // absolute
```
------------------------------------------------------------

Use Case:
I am scripting the auto creation of JSON Carthage specs for several repos which create/store local json files rather than hosting on a repo somewhere an manually updating. Once I have these files, the issue with existing file: scheme support is those files are now at a specific path of the user who created it.

Alternative:
Use script to periodically run and keep personal spec repos up to date. Github api
eg. https://api.github.com/repos/mParticle/mparticle-apple-sdk/releases

Expectation:
Support relative path from currently executing carthage function

Provide a path as: (no scheme)
/my/domain/com/framework.json

Which will resolve to {currentDir}/my/domain/com/framwork.json

Notes:
~I struggled to find a readily identifiable scheme for relative paths. In Bash I use paths directly without a scheme. As a proof of concept I included relativePath: for now and will hopefully get feedback on a better approach? Would no scheme be an option?~ I ended up leaning back towards a scheme less url.

Changes:
+ Add binary parsing for nil scheme url which constructs relative path
+ Added test case

Notes:
Rather than include FileManager, potential to inject value to boost testability?


mdiep:
I'm a little confused by what you're trying to do. Does this not do what you want?

```
# no slash should be needed here
binary "my/domain/com/framework.json"
```

dmiluski:
Hi @mdiep,

Here's my use case. Ever since the binary Carthage support came out, I've been happy pulling in binary dependencies via Carthage spec json files, rather than committing binaries directly to the repo. It makes me feel all warm and cozy keeping the repo specifically our own code.

https://github.com/dmiluski/CarthageSpecs

The problem with this approach has been that it's been very manual. I manually add releases to these files.

When I noticed local file support for Carthage spec support for binary Carthage dependencies, I reasoned it could be useful to package these json files locally to my repo, then pull in the binaries given those definitions?

Given Repos & GithubAPI, Generate CarthageSpec Files //Locally to project directory
eg. https://api.github.com/repos/mParticle/mparticle-apple-sdk/releases

Given a script which hits GitHub release api, I could automate the mapping of the files to Carthage Spec style

The problem I ran into was that Carthage currently only support https vs file schemes. If I used the file scheme the path would be specific to my machine, and after reviewing the issue linked, it appeared others were also looking for relative support. Given the relative path + the executing directory, I could now provide support to access Carthage Spec files local to the project directory (Say in a Dir named "CarthageSpecs")

When I read through, I noticed that current implementation specified support for only https or file
https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/Dependency.swift#L126

Given this, was this approach poorly implemented? Or is there another obvious solution for converting GitHub release objects to Carthage Spec json file which I was just unaware of?

Thanks for your time and feedback

awgeorge:
@mdiep - Exactly what I was trying - but fails with error: `Parse error: non-https, non-file URL found for dependency type `binary` in line: binary "framework.json"`.

A change was made to only allow local .json files using the `file://` schema - but this only allows absolute paths - rendering it effectively useless if you're in a distributed team.

This fails with:
```
*** Downloading binary-only framework  at "file://framework.json"
Failed to read file or folder at : Error Domain=Result.AnyError Code=1 "file is directory"
```

I tried various attempts at tricking carthage to read the .json file at a relative directory but all failed:

```
binary "file:framework.json"
binary "file:.framework.json"
binary "file:./framework.json"
binary "file:/framework.json"
binary "file://./framework.json"
```

dmiluski:
@awgeorge , the original first commit in this PR (I left for context) allowed for a relative path support and passing all tests. Still need to fix these tests. But I wasn't  happy with that scheme. It just didn't feel right.

Remember the 'file:///' requires three '/' to let it know you're starting from root directory? That tripped me up a bit until understanding that.

dmiluski:
@mdiep , totally misunderstood your feedback.

slash is now removed in the code with latest push:
https://github.com/Carthage/Carthage/pull/2360#issuecomment-368889406

Source/CarthageKit/Dependency.swift
```diff
@@ -125,8 +125,11 @@ extension Dependency: Scannable {
 				if let url = URL(string: urlString) {
 					if url.scheme == "https" || url.scheme == "file" {
 						return .success(self.binary(url))
+					} else if url.scheme == nil,
+						let url = URL(string: FileManager.default.currentDirectoryPath)?.appendingPathComponent(url.relativePath) {
+						return .success(self.binary(url))
 					} else {
-						return .failure(ScannableError(message: "non-https, non-file URL found for dependency type `binary`", currentLine: scanner.currentLine))
+						return .failure(ScannableError(message: "non-https, non-file, non-relative-file URL found for dependency type `binary`", currentLine: scanner.currentLine))
```
dmiluski: Would there be a better term for this message?

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -214,22 +214,13 @@ class DependencySpec: QuickSpec {
...
 					let dependency = Dependency.from(scanner).value
 
-					let path = URL(string: FileManager.default.currentDirectoryPath)!.appendingPathComponent("/my/domain/com/framework.json").absoluteString
-					let url = "file://\(path)"
+					let url = URL(string: FileManager.default.currentDirectoryPath)!.appendingPathComponent("my/domain/com/framework.json")
 
-					expect(dependency) == .binary(URL(string: url)!)
-				}
-
-				it("should fail with non-https URL") {
```
dmiluski: Removed this as this now conflicts with relative file path

awgeorge:
@dmiluski - Would you not want to keep `file://` but allow it to read relative paths. IE. Fix the error further down the path.

blender:
I think the schema is requirement here. Removing the schema is not a good solution.

dmiluski:
@awgeorge, I chose not to keep file b/c I had assumed that verbosely specified an absolute path. I felt a lack of schema assumed local path?

@blender , Although I can understand the requirement of a schema, would you have a proposal for relative vs absolute path differentiation? Potentially, the difference could be file:/// vs file:// ? Any other thoughts?

blender:
A schema is roughly defined as `<schema-id>://`. THhe third `/` you see in `file:///somefile` is an absolute path from the root `/`. Relative paths from `cwd` would be `file://somefile`(note 2 `/`). 


The problem does not sit here but definitely down the chain like @awgeorge suggested. It seems paths are always resolved as absolute.

awgeorge:
@blender - If we go down the `3` slashes route `///` it will be a breaking change - However - I do think it's the best approach. `///` = absolute `//` = relative

blender:
Maybe I'm missing something but if 2 slashes are resolving as absolute currently, this is definitely a bug and people should not rely on it.

tbrannam:
Relative paths aren't part of the file:// standard from what I recall.  But I've seen example use cases where the slashes are omitted entirely if relative paths are supported.   This avoids `//` vs `///` argument, if 0 slashes, then it's relative?

awgeorge:
@blender - 2 slashes fail with this: 
```
*** Downloading binary-only framework  at "file://framework.json"
Failed to read file or folder at : Error Domain=Result.AnyError Code=1 "file is directory"
```
So it doesn't work - and you're right - it can't be relied on.

But - if people have used `file://my/absolute/path/framwork.json` then, it will suddenly stop working, as they're missing an extra `/`.

blender:
https://tools.ietf.org/html/draft-ietf-appsawg-file-scheme-03#appendix-F

I think this documents leaves much to interpretation however to in appendix F it mentions `file:/some/path` as absolute.


The other forms `file:///` `file://` are also mentioned as absolute.

I think there is a lot of room for wiggling here. I think @tbrannam suggestion is viable.

awgeorge:
Could we check if `URL.hasPrefix('.')`?

blender:
After reading the [RFC](https://tools.ietf.org/html/rfc8089) a couple of times it seems to me that relative URLs are out of question in the specification.

Also about the path component I read:

> In a POSIX file system, the root of the file system is represented as
   a directory with a zero-length name, usually written as "/"; the
   presence of this root in a file URI can be taken as given by the
   initial slash in the "path-absolute" rule.

So both `file:///path/to/file` and `file://path/to/file` should be absolute.
By the same rule `file://.relative/path`should expand to `file:///.relative/path` so this won't work either.

So, either we don't follow the schema specification (carthage is not a browser after all and I doubt that we want to support non-local _auth-path_ for the file schema like `file://my.domain.com/path/to/file`) or we find another way without schema (but I find this weird). Also as a side note in the [Cartfile](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#example-cartfile) documentation the only supported file scheme has 3 `///`.

blender:
From Foundation:
```
let url = URL(fileURLWithPath: "./help", isDirectory: false, relativeTo: nil) // or "help"
let path = url.absoluteString
print(path)
```
produces a resolved absolute path against `cwd`

dmiluski:
@blender I really like the 'fileWithPath' approach, it makes the path construction so much less hacky. 

As for scheme vs no scheme, after reading this thread I'm re-itterating that the scheme seems to enforce too much specificity to the local file system vs no scheme relative path.

If this is okay with the greater group? I can also update the binary examples to provide a scheme less example?

blender:
I think we go the discussion at a good point here and I'm going to summarize for @mdiep:

### Goal
Support relative file paths for binary (possibly also git?) sources in Cartfile.

Example:
```
binary "my/domain/com/framework.json"
```

### Motivation
Currently only absolute file paths are supported via the `file://` scheme.

Example:
```
binary "file:///my/domain/com/framework.json" // this however is absolute.
```

### Current limitations
The `file://` scheme does not support relative file paths. See discussion at https://github.com/Carthage/Carthage/pull/2360#issuecomment-369031008

### Proposed solution
Drop the scheme. If the location source has no scheme treat it as a file path.

Example:
```
binary "my/domain/com/framework.json" // Relative to cwd
binary "/my/domain/com/framework.json" // absolute
```

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -213,12 +213,14 @@ class DependencySpec: QuickSpec {
 					expect(dependency) == .binary(URL(string: "file:///my/domain/com/framework.json")!)
 				}
 
-				it("should fail with non-https URL") {
-					let scanner = Scanner(string: "binary \"nope\"")
+				it("should read a URL with relativeFile scheme") {
+					let scanner = Scanner(string: "binary \"my/domain/com/framework.json\"")
 
-					let error = Dependency.from(scanner).error
+					let dependency = Dependency.from(scanner).value
```
dmiluski: @blender what are you thoughts on injecting current dir to provide dependency injection of CWD for these relative paths? This way I wouldn't have to insert duplicated URL construction logic between test vs Dependency being responsible?

Source/CarthageKit/Dependency.swift
```diff
@@ -125,8 +129,10 @@ extension Dependency: Scannable {
 				if let url = URL(string: urlString) {
 					if url.scheme == "https" || url.scheme == "file" {
 						return .success(self.binary(url))
+					} else if url.scheme == nil {
+						return .success(self.binary(URL(fileURLWithPath: url.relativePath, isDirectory: false, relativeTo: base)))
```
dmiluski: Hi @blender , just realized this API required MacOS 10.11, where as this projects target is 10.10 (Current prod available 10.13), are you aware of other easy alternatives for backwards compatibility? I'm not aware of this projects compatibility goals?

dmiluski: Potentially I could leverage NSURL interface, then case back to as URL
let url = NSURL(fileURLWithPath: "help", isDirectory: false, relativeTo: nil) // Optional init vs required vs latest URL API currently in code

Source/CarthageKit/Dependency.swift
```diff
@@ -151,6 +158,26 @@ extension Dependency: Scannable {
 			return .failure(ScannableError(message: "empty string after dependency type", currentLine: scanner.currentLine))
 		}
 	}
+
+	// MacOS 10.10 Compatibility Helper
+	private static func absoluteFileURL(for relativePath: String, base: URL?) -> URL? {
+		if #available(macOS 10.11, *) {
+			return URL(fileURLWithPath: relativePath, isDirectory: false, relativeTo: base).standardizedFileURL
+		} else {
+			guard let url = (base ?? URL(string: FileManager.default.currentDirectoryPath)),
```
blender: isn't there are currentDirectoryURL ?

blender:
Hi, 

I don't want to damp anyone's enthusiasm for this but the solution still needs approval by @mdiep  

However, I see that currently only the relative paths are supported. The solution should support asbolute paths without `file://` scheme too.

mdiep:
Thanks for the summary, @blender. That sounds good to me. 👍

Source/CarthageKit/Dependency.swift
```diff
@@ -151,6 +158,26 @@ extension Dependency: Scannable {
 			return .failure(ScannableError(message: "empty string after dependency type", currentLine: scanner.currentLine))
 		}
 	}
+
+	// MacOS 10.10 Compatibility Helper
```
mdiep: Maybe we should just drop 10.10 at this point? 🤔

blender: Yeah :D I think 10.11 is acceptable

blender:
@dmiluski Full steam ahead. Please see my comments.

Source/CarthageKit/Dependency.swift
dmiluski: Woohoo! That makes this way easier. I played around this for way to long learning about the intricacies of URL comparisons.

dmiluski:
Hi @blender , I wanted to double check I satisfied your comment about absolute url support:
https://github.com/Carthage/Carthage/pull/2360#issuecomment-369892078

Given that by removing 10.10 support, I believe the URL(fileWithPath:isDirectory:relativeTo) implementation this is now covered by it's implementation and tests?

dmiluski:
Hm.. any thoughts on these Test failures?
I updated minimum deploy target to by 10.11, but some of the tests ares still failing with availability 10.11 warning. The worker information for Travis appears to be MacOS 10.12, so I'm confused what I may have missed.

Worker information
hostname: ee9de18c-adbe-48e7-b68a-8de43b509754@416.wjb-1
version: v3.5.0 https://github.com/travis-ci/worker/tree/77dbc57c72d00592aeb754773b712da843c7e00d
instance: c583778b-35e2-441a-a358-9fc7e973513e travis-ci-macos10.12-xcode9.2-1512491837 (via amqp)
startup: 2m3.924193391s

Documentation/Artifacts.md
```diff
@@ -31,10 +31,11 @@ git "https://enterprise.local/desktop/git-error-translations2.git"
 
 ##### Binary only frameworks
 
-Dependencies that are only available as compiled binary `.framework`s are specified with the `binary` keyword and an https address that returns a binary project specification:
+Dependencies that are only available as compiled binary `.framework`s are specified with the `binary` keyword and as as an https address, or a relative path that returns a binary project specification:
```
hyperspacemark: Grammar: "and as as an https address" should be "and an https address, or a relative path, that"

Source/CarthageKit/Dependency.swift
```diff
@@ -125,6 +129,9 @@ extension Dependency: Scannable {
 				if let url = URL(string: urlString) {
 					if url.scheme == "https" || url.scheme == "file" {
 						return .success(self.binary(url))
+					} else if url.scheme == nil {
+						let abosoluteURL = URL(fileURLWithPath: url.relativePath, isDirectory: false, relativeTo: base).standardizedFileURL
```
hyperspacemark: Typo: "abosoluteURL"

Documentation/Artifacts.md
```diff
@@ -84,6 +85,9 @@ git "file:///directory/to/project" "branch"
 
 # A binary only framework
 binary "https://my.domain.com/release/MyFramework.json" ~> 2.3
+
+# A binary only framework via local relative path to binary project specification
+binary "relativePath/MyFramework.json" ~> 2.3
```
blender: also add example for `/absolute/path`

Documentation/Artifacts.md
```diff
@@ -31,10 +31,11 @@ git "https://enterprise.local/desktop/git-error-translations2.git"
 
 ##### Binary only frameworks
 
-Dependencies that are only available as compiled binary `.framework`s are specified with the `binary` keyword and an https address that returns a binary project specification:
+Dependencies that are only available as compiled binary `.framework`s are specified with the `binary` keyword and as an https address, or a relative path, that returns a binary project specification:
```
blender: I would change this to:


Dependencies that are only available as compiled binary `.framework`s are specified with the `binary` keyword and as an `https://` URL, a `file://` URL, a relative path or an absolute path with no scheme, that returns a binary project specification

Documentation/Artifacts.md
```diff
@@ -31,10 +31,11 @@ git "https://enterprise.local/desktop/git-error-translations2.git"
...
 
 ```
-binary "https://my.domain.com/release/MyFramework.json"
+binary "https://my.domain.com/release/MyFramework.json"   // Remote Hosted
+binary "relativePath/MyFramework.json"					   				// Locally hosted at relative path
```
blender: also add example for `/absolute/path`

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -213,12 +213,14 @@ class DependencySpec: QuickSpec {
 					expect(dependency) == .binary(URL(string: "file:///my/domain/com/framework.json")!)
 				}
 
-				it("should fail with non-https URL") {
-					let scanner = Scanner(string: "binary \"nope\"")
+				it("should read a URL with relativeFile scheme") {
```
blender: should succeed with a relative file path

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -213,12 +213,14 @@ class DependencySpec: QuickSpec {
...
+					let relativePath = "my/relative/path/framework.json"
+					let scanner = Scanner(string: "binary \"\(relativePath)\"")
 
-					let error = Dependency.from(scanner).error
+					let workingDirectory = URL(string: "file:///current/working/directory/")!
+					let dependency = Dependency.from(scanner, base: workingDirectory).value
 
-					expect(error) == ScannableError(message: "non-https, non-file URL found for dependency type `binary`", currentLine: "binary \"nope\"")
+					expect(dependency) == .binary(URL(string: "\(workingDirectory)\(relativePath)")!)
 				}
```
blender: Please also add test for absolute file path

blender:
@dmiluski @mdiep  I have checked the error on Travis and it seems that when building for swiftpm the right OSX sdk is not picked up despite the fact that 10.11 is set for both Carthage and CarthageKit targets

ikesyo:
FYI xcodeproj files are not related to SwiftPM.

blender:
Thanks @ikesyo, do you have insight of how to solve the issue here?

ikesyo:
Looks like passing `-Xswiftc "-target" -Xswiftc "x86_64-apple-macosx10.11" to `swift build` and `swift test` would do the trick.

dmiluski:
@blender , I added the documentation and additional test. Now looking into @ikesyo 's recommendation for args in swift build/test. I added what I interpreted, but don't necessarily understand it, so walking through PM docs to see if I can find that.

Makefile
```diff
@@ -92,15 +92,15 @@ package: installables
 	(cd "$(CARTHAGE_TEMPORARY_FOLDER)$(FRAMEWORKS_FOLDER)" && zip -q -r --symlinks - "$(OUTPUT_FRAMEWORK)") > "$(OUTPUT_FRAMEWORK_ZIP)"
 
 swiftpm:
-	swift build -c release -Xswiftc -static-stdlib
+	swift build -c release -Xswiftc -static-stdlib -Xswiftc "-target" -Xswiftc "x86_64-apple-macosx10.11"
```
dmiluski: Hi @ikesyo , was this the location/arg format you recommended?

dmiluski:
@blender , All tests now passing. Thanks @ikesyo for the fix pointer.

dmiluski:
Hi @mdiep, is this PR approach/goal okay? Was there a timeline on merges/releases?

mdiep:
Looks good to me! Thank you for your work on this. It's always really appreciated when people open PRs instead of just complaining in an issue. 😄

dmiluski:
@mdiep , it's been about 45 days since last release? Was there a doc on how often/milestones you prefer to include in each release?

mdiep:
We don't have a defined release schedule. We've been trying to get a release out the door for the past week or so, but there are some regressions that haven't quite been cleared up yet. (See #2372)

dmiluski:
Coming back to this, I discovered I missed a requirement for support:

Cartfile
```
# Binary Only Third Party
binary "CarthageSpecs/AppboyFramework.json"
binary "CarthageSpecs/mParticle-Appboy.json"
```

Cartfile.resolved
```
binary "file:///Users/dmiluski/code/venmo/CarthageSpecs/AppboyFramework.json" "3.3.2"
binary "file:///Users/dmiluski/code/venmo/CarthageSpecs/mParticle-Appboy.json" "7.2.0"
```

Although Carthage is able to resolve the locally provided json file, the resolved file is still tied to an absolute path vs a relative path. 🤦‍♂️ Looking into it now.

dmiluski:
Hi @blender , for best practices, should I open an issue to track? I'm thinking through how to address now, but didn't yet have a structure defined to support yet.

Current Status:
`Carthage update` with current relative implementation works... but leaves an absolute file path for binary json specs, which will fail on others computers

`Carthage install` fails if not run on computer which ran update b/c path is not going to be the same.

Current Option I'm thinking through:
for Dependency.binary? Potentially separate the associated value to an enum differentiating relative vs absolute model items?

PR #2361, opened 2018-03-01
Strip debug symbols when copying frameworks using "carthage copy-frameworks"

abrindam:
Strip debug symbols when copying frameworks using "carthage copy-frameworks" if COPY_PHASE_STRIP is enabled.
    
Now "carthage copy-frameworks" behaves the same as vanilla Xcode copy task as far as debug symbols go.

See PR #2334 for full discussion on why this is a good thing.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -873,10 +873,11 @@ public func buildInDirectory(
 	}
 }
 
-/// Strips a framework from unexpected architectures, optionally codesigning the
-/// result.
-public func stripFramework(_ frameworkURL: URL, keepingArchitectures: [String], codesigningIdentity: String? = nil) -> SignalProducer<(), CarthageError> {
+/// Strips a framework from unexpected architectures and potentially debug symbols,
+/// optionally codesigning the result.
+public func stripFramework(_ frameworkURL: URL, keepingArchitectures: [String], strippingDebugSymbols: Bool, codesigningIdentity: String? = nil) -> SignalProducer<(), CarthageError> {
 	let stripArchitectures = stripBinary(frameworkURL, keepingArchitectures: keepingArchitectures)
+	let stripSymbols = strippingDebugSymbols ? stripDebugSymbols(frameworkURL) : .empty
 
 	// Xcode doesn't copy `Headers`, `PrivateHeaders` and `Modules` directory at
 	// all.
@@ -887,6 +888,7 @@ public func stripFramework(_ frameworkURL: URL, keepingArchitectures: [String], 
 	let sign = codesigningIdentity.map { codesign(frameworkURL, $0) } ?? .empty
 
 	return stripArchitectures
+		.concat(stripSymbols)
 		.concat(stripHeaders)
 		.concat(stripPrivateHeaders)
 		.concat(stripModules)
@@ -1080,6 +1082,17 @@ public func architecturesInPackage(_ packageURL: URL) -> SignalProducer<String, 
 		}
 }
 
+/// Strips debug symbols from the given framework
+public func stripDebugSymbols(_ frameworkURL: URL) -> SignalProducer<(), CarthageError> {
+	return SignalProducer<URL, CarthageError> { () -> Result<URL, CarthageError> in binaryURL(frameworkURL) }
+		.flatMap(.merge) { binaryURL -> SignalProducer<TaskEvent<Data>, CarthageError> in
+			let stripTask = Task("/usr/bin/xcrun", arguments: [ "strip", "-S","-o", binaryURL.path, binaryURL.path])
+			return stripTask.launch()
+				.mapError(CarthageError.taskError)
+		}
+		.then(SignalProducer<(), CarthageError>.empty)
+}
+
 /// Strips `Headers` directory from the given framework.
 public func stripHeadersDirectory(_ frameworkURL: URL) -> SignalProducer<(), CarthageError> {
 	return stripDirectory(named: "Headers", of: frameworkURL)
@@ -1185,7 +1198,7 @@ private func UUIDsFromDwarfdump(_ url: URL) -> SignalProducer<Set<UUID>, Carthag
 }
 
 /// Returns the URL of a binary inside a given package.
-private func binaryURL(_ packageURL: URL) -> Result<URL, CarthageError> {
+public func binaryURL(_ packageURL: URL) -> Result<URL, CarthageError> {
 	let bundle = Bundle(path: packageURL.path)
 	let packageType = (bundle?.object(forInfoDictionaryKey: "CFBundlePackageType") as? String).flatMap(PackageType.init)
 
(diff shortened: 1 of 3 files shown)


abrindam:
Well I totally forgot the tests. Dealing with that now...

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,6 +1082,17 @@ public func architecturesInPackage(_ packageURL: URL) -> SignalProducer<String,
 		}
 }
 
+/// Strips debug symbols from the given framework
+public func stripDebugSymbols(_ frameworkURL: URL) -> SignalProducer<(), CarthageError> {
+	return SignalProducer<URL, CarthageError> { () -> Result<URL, CarthageError> in binaryURL(frameworkURL) }
+		.flatMap(.merge) { binaryURL -> SignalProducer<TaskEvent<Data>, CarthageError> in
+			let stripTask = Task("/usr/bin/xcrun", arguments: [ "strip", "-S","-o", binaryURL.path, binaryURL.path])
```
blender: is not -x needed?

```
 -S     Remove the debugging symbol table entries (those created by the -g option to cc(1) and other
              compilers).
 -x     Remove all local symbols (saving only global symbols).
```

mdiep:
The original intention in #2334 was to keep symbols in the `.dSYM`, but not the the binary within the framework, correct?

If so, can we remove the symbols from the binary during build? Since `copy-frameworks` is run frequently, it's best to keep it as quick as possible.

blender:
Good point. 

My bad I indicated copy-frameworks as the correct place. Sorry @abrindam :(

abrindam:
No problem folks. I raised this very question during my long post here: https://github.com/Carthage/Carthage/issues/2280#issuecomment-356825509

I received a single reply indicating a preference for keeping the debug symbols in the built binaries because it maximized developer flexibility (they could then choose whether or not to strip, rather than forcing stripping).

Also keep in mind this is the behavior of the default Xcode copy operation you can add in Build Phases. So if you weren't using Carthage, this would be the behavior anyways. So I'm still a fan of this because I think copy-frameworks should always be a strict superset of the vanilla copy operation (aka copy-frameworks should do more, but never less). Since Xcode is willing to do this on every build, it seems reasonable it can't be that expensive.

Any thoughts on this line of thinking?

Source/CarthageKit/Xcode.swift
abrindam: So to figure out what flags to pass, I reverse-engineered Xcode by replacing the `strip` binary on my system with one that logs all the arguments passed to it. Xcode passes this flag and only this flag when using a vanilla copy. So I figured we should be doing what Xcode does for symmetry.

I am by far not an expert on the intricacies of stripping symbols, so I'm happy to listen to alternatives however.

blender:
Last word is on @mdiep :) 

I'm of your same opinion @abrindam . With debug symbols in the binaries I think debugging would also be easier (or am I wrong?)

abrindam:
I think an ideal setup would be to have COPY_PHASE_STRIP off for debug and on for release. Then there is absolutely no performance penalty for the usual developer iteration.

Not sure what most people do however...

blender:
The default for a project is set to `NO` for COPY_PHASE_STRIP

Source/CarthageKit/Xcode.swift
abrindam: Further info... there is a setting STRIP_STYLE which seems to affect this. Default for me was "Debugging Symbols", but "Non-Global Symbols" seems to be what brings in "-x".

The documentation for this property seems to indicate it applies only to the linked product and not to strip on copy, but I'm suspicious that this is not the case. Will investigate further.

blender: Maybe then it should respect STRIP_STYLE ? I'm not an expert either so I guess that respecting the style is safe bet.

mdiep:
> I think an ideal setup would be to have COPY_PHASE_STRIP off for debug and on for release. Then there is absolutely no performance penalty for the usual developer iteration.

I like that. 👍

blender:
@abrindam I think we're good here then. The only change that I would like to see is to respsect the STRIP_STYLE

Source/CarthageKit/Xcode.swift
```diff
@@ -1080,6 +1082,35 @@ public func architecturesInPackage(_ packageURL: URL) -> SignalProducer<String,
...
+				.mapError(CarthageError.taskError)
+		}
+		.then(SignalProducer<(), CarthageError>.empty)
+}
+
+/// Returns whether a given framework contains debug symbols
+/// There are many suggestions on how to do this but no one single
+/// accepted way. This seems to work best:
+/// https://lists.apple.com/archives/unix-porting/2006/Feb/msg00021.html
+public func packageContainsDebugSymbols(_ packageURL: URL) -> SignalProducer<Bool, CarthageError> {
```
mdiep: It looks like this is only used in the tests. Let's move it there.

abrindam: @mdiep Agree and my first attempt had it there - the one problem was I need to user `binaryURL()` which is private to Xcode.swift. 

I could make `binaryURL()` public, or I could just copy `binaryURL()` into the test. Thoughts?

blender: I don't see an issue with making it `public` or  `internal`

abrindam:
All good on the unit test changes, I will move the function and make the helper function public.

On the `STRIP_STYLE` business, bad news. I've confirmed that vanilla copy is NOT affected by `STRIP_STYLE` - by once again replacing the `strip` binary with my own which observes what arguments are passed to it.

I might have reached this conclusion faster if I noticed the "nice name" in Xcode for `COPY_PHASE_STRIP` is "Strip Debug Symbols During Copy" (i.e. it mentions debug symbols explicitly). D'oh!

This actually makes sense though. In my case, the default for my project for "Strip Style" was "All Symbols". This would be disastrous for a dynamic framework as global symbols are needed to use the dylib (I think/I'm pretty sure).

To further confound things, I can't find a single source that explains why you should NOT strip non-global symbols along with debug symbols. So I think optimality-wise, it would be ideal to do `strip -S -x` as suggested by @blender and strip both debug and non-global symbols, but this would cause the behavior of `carthage copy-frameworks` to diverge from the vanilla copy (which does `strip -S` only).

Apple strikes again! Thoughts?

blender:
Let's stick with the minimal subset then which is `-S` thus respecting `COPY_PHASE_STRIP` "Strip Debug Symbols During Copy" . I'm wondering then what `STRIP_STYLE` applies too.

I leave this here for further references: https://www.objc.io/issues/6-build-tools/mach-o-executables/

abrindam:
OK, I've made the changes to the unit tests, and now that's we've had that entertaining diversion on strip style, assuming the CI gods don't curse me, I think we should be good to go here.

Source/CarthageKit/Xcode.swift
blender: Decision was made to leave -S only.

See:
- https://github.com/Carthage/Carthage/pull/2361#issuecomment-370429346
- https://github.com/Carthage/Carthage/pull/2361#issuecomment-370436508

mdiep:
Thanks @abrindam!

PR #2368, opened 2018-03-09
Don't create Carthage/Build symlink if project doesn't use Carthage

mdiep:
This prevents us from polluting those dependencies with a `Carthage/` directory. (This can be annoying if you're using submodules and they haven't ignored that directory.)

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -965,7 +965,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 		_ options: BuildOptions,
 		dependenciesToBuild: [String]? = nil,
 		sdkFilter: @escaping SDKFilterCallback = { sdks, _, _, _ in .success(sdks) }
-	) -> SignalProducer<BuildSchemeProducer, CarthageError> {
+	) -> BuildSchemeProducer {
 		return loadResolvedCartfile()
 			.flatMap(.concat) { resolvedCartfile -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
 				return self.buildOrderForResolvedCartfile(resolvedCartfile, dependenciesToInclude: dependenciesToBuild)
@@ -1013,7 +1013,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 							return self.installBinaries(for: dependency, pinnedVersion: version, toolchain: options.toolchain)
 								.filterMap { installed -> (Dependency, PinnedVersion)? in
 									return installed ? (dependency, version) : nil
-							}
+								}
 						case let .binary(url):
 							return self.installBinariesForBinaryProject(url: url, pinnedVersion: version, projectName: dependency.name, toolchain: options.toolchain)
 								.then(.init(value: (dependency, version)))
@@ -1029,7 +1029,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 					}
 					.flatten()
 			}
-			.flatMap(.concat) { dependency, version -> SignalProducer<BuildSchemeProducer, CarthageError> in
+			.flatMap(.concat) { dependency, version -> BuildSchemeProducer in
 				let dependencyPath = self.directoryURL.appendingPathComponent(dependency.relativePath, isDirectory: true).path
 				if !FileManager.default.fileExists(atPath: dependencyPath) {
 					return .empty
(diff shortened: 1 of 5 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -965,7 +965,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 		_ options: BuildOptions,
 		dependenciesToBuild: [String]? = nil,
 		sdkFilter: @escaping SDKFilterCallback = { sdks, _, _, _ in .success(sdks) }
-	) -> SignalProducer<BuildSchemeProducer, CarthageError> {
+	) -> BuildSchemeProducer {
```
mdiep: Simplified these because they didn't need to be a producer of a producer.

Source/CarthageKit/Project.swift
```diff
@@ -1042,26 +1042,73 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+				return self
+					.dependencySet(for: dependency, version: version)
+					.flatMap(.concat) { dependencies -> SignalProducer<(), CarthageError> in
+						// Don't create symlink the build folder if the dependencies doesn't have
+						// any Carthage dependencies.
+						if dependencies.isEmpty {
+							return .empty
+						}
+						return symlinkBuildPath(for: dependency, rootDirectoryURL: self.directoryURL)
+					}
```
mdiep: Here's the actual fix. We use similar logic for not creating `Carthage/Checkouts/`.

This call to `symlinkBuildPath` used to be inside the `build` call right after this.

Source/CarthageKit/Project.swift
```diff
@@ -1042,26 +1042,73 @@ public final class Project { // swiftlint:disable:this type_body_length
...
 						}
 					}
 			}
 	}
 }
 
+/// Creates symlink between the dependency build folder and the root build folder
+///
+/// Returns a signal indicating success
+private func symlinkBuildPath(for dependency: Dependency, rootDirectoryURL: URL) -> SignalProducer<(), CarthageError> {
```
mdiep: This just moved from another file.

Source/CarthageKit/Xcode.swift
```diff
@@ -493,7 +494,7 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 			return !sdks.isEmpty
 		}
 		.flatMap(.concat) { platform, sdks -> SignalProducer<TaskEvent<URL>, CarthageError> in
-			let folderURL = workingDirectoryURL.appendingPathComponent(platform.relativePath, isDirectory: true).resolvingSymlinksInPath()
+			let folderURL = rootDirectoryURL.appendingPathComponent(platform.relativePath, isDirectory: true).resolvingSymlinksInPath()
```
mdiep: This also changed. Before, projects would copy their built binaries into their own `Carthage/Build/`, relying on the fact that it was symlinked to the root build dir. But if a project has no Carthage dependencies, this directory no longer exists. So this copies directly to the root dir.

Source/carthage/Build.swift
```diff
@@ -131,7 +130,7 @@ public struct BuildCommand: CommandProtocol {
 		if options.skipCurrent {
 			return buildProducer
 		} else {
-			let currentProducers = buildInDirectory(directoryURL, withOptions: options.buildOptions)
+			let currentProducers = buildInDirectory(directoryURL, withOptions: options.buildOptions, rootDirectoryURL: directoryURL)
```
mdiep: When building the root project, the root directory URL *is* the directory URL.

sharplet:
Besides the test failures, this looks good to me 👍

Tests/CarthageKitTests/XcodeSpec.swift
```diff
@@ -394,28 +379,6 @@ class XcodeSpec: QuickSpec {
...
-				.ignoreTaskData()
-				.on(value: { project, scheme in // swiftlint:disable:this end_closure
-					NSLog("Building scheme \"\(scheme)\" in \(project)")
-				})
-				.wait()
-
-			expect(result.error).to(beNil())
-
-			expect(dependencyBuildURL).to(beRelativeSymlinkToDirectory(buildURL))
-		}
```
mdiep: I deleted this test because symlinks are no longer handled by the `build` method that's being tested here.

PR #2372, opened 2018-03-12
Check if schemes are buildable with projects, not workspaces

### LINKED ISSUE #2369: Carthage builds from master are attempting to build realm-cocoa's example projects as dependencies.
Maybe this is a known issue and intentional, but I thought I'd mention it just in case it isn't. I tried to look through the recent PR history to determine that on my own.

* carthage install method: [ ] .pkg, [✅ ] homebrew, [ ] source
* `carthage version`:
```
brew info carthage
carthage: stable 0.28.0 (bottled), HEAD
Decentralized dependency manager for Cocoa
https://github.com/Carthage/Carthage
/usr/local/Cellar/carthage/HEAD-375cb56 (69 files, 26.7MB) *
  Built from source on 2018-03-09 at 11:10:50
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/carthage.rb
==> Requirements
Build: xcode ✔
==> Options
--HEAD
	Install HEAD version
==> Caveats
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

zsh completions have been installed to:
  /usr/local/share/zsh/site-functions

```
* `xcodebuild -version`: 9C40b
* Are you using `--no-use-binaries`? Yes

**Cartfile**
```
github "SwiftyJSON/SwiftyJSON" ~> 4.0

github "ReactiveX/RxSwift" ~> 4.0

github "realm/realm-cocoa" ~> 3.0
github "RxSwiftCommunity/RxRealm" ~> 0.7

github "bustoutsolutions/siesta" ~> 1.0

github "sparkle-project/Sparkle" ~> 1.17

github "insidegui/ChromeCastCore" "swift4"

# ChromeCastCore's dependencies are currently met manually
github "alexeyxo/protobuf-swift" == 3.0.23
# SwiftJSON is compiled directly into ChromeCastCore
```

**Carthage Invocation**

```
carthage bootstrap --no-use-binaries --platform macOS
```

**Carthage Output**
```
*** Checking out ChromeCastCore at "899055fd3b44411a637eb7ce0a656041adf3f2c3"
*** Checking out siesta at "1.3.1"
*** Checking out realm-cocoa at "v3.1.1"
*** Checking out Sparkle at "1.18.1"
*** Checking out SwiftyJSON at "4.0.0"
*** Checking out RxSwift at "4.1.2"
*** Checking out protobuf-swift at "3.0.23"
*** Checking out RxRealm at "0.7.5"
*** xcodebuild output can be found in /var/folders/9c/21k0gfj9379dn7c5hqff7x2n5h8681/T/carthage-xcodebuild.a5yN7W.log
*** Building scheme "ChromeCastCore" in ChromeCastCore.xcodeproj
*** Building scheme "ProtocolBuffers(OSX)" in ProtocolBuffers.xcodeproj
*** Building scheme "Realm" in Realm.xcworkspace
*** Building scheme "RealmSwift" in Realm.xcworkspace
*** Building scheme "JSONImport" in RealmExamples.xcworkspace
*** Building scheme "DownloadCache" in RealmExamples.xcworkspace
Build Failed
	Task failed with exit code 65:
	/usr/bin/xcrun xcodebuild -workspace /Users/allen.humphreys/projects/WWDC/Carthage/Checkouts/realm-cocoa/examples/tvos/objc/RealmExamples.xcworkspace -scheme DownloadCache -configuration Release -derivedDataPath /Users/allen.humphreys/Library/Caches/org.carthage.CarthageKit/DerivedData/9.2_9C40b/realm-cocoa/v3.1.1 ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES archive -archivePath /var/folders/9c/21k0gfj9379dn7c5hqff7x2n5h8681/T/realm-cocoa SKIP_INSTALL=YES GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=NO CLANG_ENABLE_CODE_COVERAGE=NO STRIP_INSTALLED_PRODUCT=NO (launched in /Users/allen.humphreys/projects/WWDC/Carthage/Checkouts/realm-cocoa)

This usually indicates that project itself failed to compile. Please check the xcodebuild log for more details: /var/folders/9c/21k0gfj9379dn7c5hqff7x2n5h8681/T/carthage-xcodebuild.a5yN7W.log
```

**Actual outcome**
Carthage attempts to build the example projects for `realm-cocoa`, located in the top-level `examples` directory.

**Expected outcome**
Carthage should not do this for many reasons, but most importantly, those projects can't be built for macOS resulting in an error.

mdiep:
Fixes #2369.

This broke in #2344. The check for whether a scheme should be built needs to occur against the project it's from, not against the workspace that includes it.

cc @jdhealy @allenhumphreys

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -132,35 +132,38 @@ public func buildableSchemesInDirectory(
 		.flatMap(.merge) { (projects: [(ProjectLocator, [Scheme])]) -> SignalProducer<(Scheme, ProjectLocator), CarthageError> in
 			return schemesInProjects(projects).flatten()
 		}
-		.flatMap(.merge) { scheme, project -> SignalProducer<(Scheme, ProjectLocator, ProjectLocator), CarthageError> in
-			return locator
-				// This scheduler hop is required to avoid disallowed recursive signals.
-				// See https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2042.
-				.start(on: QueueScheduler(qos: .default, name: "org.carthage.CarthageKit.Xcode.buildInDirectory"))
-				// Pick up the first workspace which can build the scheme.
-				.filter { project, schemes in
-					switch project {
-					case .workspace where schemes.contains(scheme):
-						return true
-
-					default:
-						return false
-					}
-				}
-				// If there is no appropriate workspace, use the project in
-				// which the scheme is defined instead.
-				.concat(value: (project, []))
-				.take(first: 1)
-				.map { host, _ in (scheme, project, host) }
-		}
-		.flatMap(.concurrent(limit: 4)) { (scheme: Scheme, project: ProjectLocator, host: ProjectLocator) -> SignalProducer<(Scheme, ProjectLocator), CarthageError> in
+		.flatMap(.concurrent(limit: 4)) { scheme, project -> SignalProducer<(Scheme, ProjectLocator), CarthageError> in
 			/// Check whether we should the scheme by checking against the project. If we're building
 			/// from a workspace, then it might include additional targets that would trigger our
 			/// check.
 			let buildArguments = BuildArguments(project: project, scheme: scheme, configuration: configuration)
 			return shouldBuildScheme(buildArguments, platforms)
 				.filter { $0 }
-				.map { _ in (scheme, host) }
+				.map { _ in (scheme, project) }
+		}
+		.flatMap(.concurrent(limit: 4)) { scheme, project -> SignalProducer<(Scheme, ProjectLocator), CarthageError> in
+			return locator
+				// This scheduler hop is required to avoid disallowed recursive signals.
+				// See https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2042.
+				.start(on: QueueScheduler(qos: .default, name: "org.carthage.CarthageKit.Xcode.buildInDirectory"))
+				// Pick up the first workspace which can build the scheme.
+				.flatMap(.concat) { project, schemes -> SignalProducer<ProjectLocator, CarthageError> in
+					switch project {
+					case .workspace where schemes.contains(scheme):
+						let buildArguments = BuildArguments(project: project, scheme: scheme, configuration: configuration)
+						return shouldBuildScheme(buildArguments, platforms)
+							.filter { $0 }
+							.map { _ in project }
+
+					default:
+						return .empty
+					}
+				}
+				// If there is no appropriate workspace, use the project in
+				// which the scheme is defined instead.
+				.concat(value: project)
+				.take(first: 1)
+				.map { project in (scheme, project) }
 		}
 		.collect()
 		.flatMap(.merge) { (schemes: [(Scheme, ProjectLocator)]) -> SignalProducer<(Scheme, ProjectLocator), CarthageError> in
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -138,7 +138,7 @@ public func buildableSchemesInDirectory(
 						return .init(error: .noSharedFrameworkSchemes(.git(GitURL(directoryURL.path)), platforms))
 					}
 				}
-		}.flatMap(.merge) { scheme, project -> SignalProducer<(Scheme, ProjectLocator), CarthageError> in
+		}.flatMap(.merge) { scheme, project -> SignalProducer<(Scheme, ProjectLocator, ProjectLocator), CarthageError> in
 			return locator
 				// This scheduler hop is required to avoid disallowed recursive signals.
 				// See https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2042.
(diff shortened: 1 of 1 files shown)


allenhumphreys:
For what it's worth. This change does fix my original problem in #2369.

mdiep:
Updated to fix #2373 too.

jdhealy:
Okay, I think I found the disconnect between #2344 and the code that preceded it —

Here’s a bug: if there's an shared scheme `Example` building a macOS app `Example.app`, the code:
https://github.com/Carthage/Carthage/blob/26fb61a1ca61e5a2eb11fb1520bee9b4cd1633a6/Source/CarthageKit/Xcode.swift#L142-L160
…will eschew any actually buildable-by-carthage `Example` schemes in favor of passing only the scheme for the macOS app to https://github.com/Carthage/Carthage/blob/26fb61a1ca61e5a2eb11fb1520bee9b4cd1633a6/Source/CarthageKit/Xcode.swift#L162-L166

- - - 

To a larger point, the following probably holds false-by-convention for most repos, but would cause enough failures to be non-optimal for us: scheme names don’t hold universal·meaning/non-duplication across the boundaries of different xcodeproj/xcworkspace. 

We avoided this issue in the past, largely, by potentially-duplicated scheme names being treated as identifiers only for the subset of shared schemes that `carthage` deemed buildable.

mdiep:
I don't see how the logic here in #2372 is different from the behavior before #2344.

In both cases:

1. We only built shared schemes that built a framework in the container that holds them
2. We build schemes from workspaces if they're contained in one

#2344 broke (1) by checking build ability against the workspace instead of the container, but this PR fixes it. So I think this fix restores the previous behavior while retaining the potential performance benefits of #2344.

Am I missing something @jdhealy?

jdhealy:
Unlike when `locator` was decidedly producing deemed-buildable-by-carthage buildable schemes:

https://github.com/Carthage/Carthage/blob/19e584623e548de68f424ea5c5dc4c081d750112/Source/CarthageKit/Xcode.swift#L779

…the [`.take(first: 1)`](https://github.com/Carthage/Carthage/commit/26fb61a1ca61e5a2eb11fb1520bee9b4cd1633a6#diff-f66644e8cfbcad3c3826395a690ddfbcR159) in #2372 or #2344 can grab a single deemed-unbuildable-by-carthage scheme with no recourse (where previously recourse wasn’t needed because the deemed-buildable-by-carthage determination was done beforehand by `locator`).

jdhealy:
> given scheme that's shared from a given project

> The .take(first: 1) existed before too

Yeah, difference being: what’s given to the `.take(first: 1)` originated from a `locator` that previously (differently than #2372 or #2344) was [filtering out schemes deemed-nonbuildable-by-carthage beforehand](https://github.com/Carthage/Carthage/blob/19e584623e548de68f424ea5c5dc4c081d750112/Source/CarthageKit/Xcode.swift#L117-L123).

mdiep:
Okay, let's try this again. 😅 

@jdhealy provided a very helpful reproduction of the issue he described. 👏 
[WorkspaceWithDependency·ModifiedToTest«b395568».zip](https://github.com/Carthage/Carthage/files/1815500/WorkspaceWithDependency.ModifiedToTest.b395568.zip)

mdiep:
_bump_

Source/CarthageKit/Xcode.swift
```diff
@@ -130,40 +130,48 @@ public func buildableSchemesInDirectory(
...
 				// Pick up the first workspace which can build the scheme.
-				.filter { project, schemes in
+				.flatMap(.concat) { project, schemes -> SignalProducer<ProjectLocator, CarthageError> in
 					switch project {
 					case .workspace where schemes.contains(scheme):
-						return true
+						let buildArguments = BuildArguments(project: project, scheme: scheme, configuration: configuration)
+						return shouldBuildScheme(buildArguments, platforms)
+							.filter { $0 }
+							.map { _ in project }
```
jdhealy: Know you just relocated this from elsewhere, but this could be `.filterMap { $0 ? project : nil }`, if that sits right with you.

Source/CarthageKit/Xcode.swift
```diff
@@ -130,40 +130,48 @@ public func buildableSchemesInDirectory(
...
-			return shouldBuildScheme(buildArguments, platforms)
-				.filter { $0 }
-				.map { _ in (scheme, project) }
+		.collect()
+		.flatMap(.merge) { (schemes: [(Scheme, ProjectLocator)]) -> SignalProducer<(Scheme, ProjectLocator), CarthageError> in
+			if !schemes.isEmpty {
+				return .init(schemes)
+			} else {
+				return .init(error: .noSharedFrameworkSchemes(.git(GitURL(directoryURL.path)), platforms))
+			}
```
jdhealy: Less readably intuitive, but more performant (potentially ↜ untested), might be:
```swift
.lift {
	func noSharedFrameworkSchemesError(_ directoryURL: URL) -> CarthageError {
		return .noSharedFrameworkSchemes(.git(GitURL(directoryURL.path)), platforms)
	}

	return Signal.merge( // merge in error iff no values are sent
		$0,
		$0.materialize()
			.reduce(into: false) { if case .value = $1 { $0 = true } }
			.flatMap(.race) {
				$0 ? SignalProducer.empty : .init(error: noSharedFrameworkSchemesError(directoryURL))
			}
	)
}
```

mdiep:
I'd say let's just merge so we can release. ☺️

jdhealy:
thumbs-up

PR #2374, opened 2018-03-13
Resolve all swiftlint violations.

jakeadams:
Some clean up of the compilation warnings. Most of this is either adding `swiftlint:disable` hints or resolving the issue.

In collaboration with @mdiep, I disabled 2 rules in `.swiftlint.yml` definition.
I also removed conformance to `CustomStringConvertible` for `Platform` as it only returned the `rawValue` and the call site was accessing `description` directly, which defeats the purpose of the protocol. I updated the call site to use `rawValue` directly. I added a test around this to make sure nothing changed.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -434,7 +434,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 								specifier = .any
 							}
 							result[dependency] = specifier
-					}
+						}
 				}
 				.flatMap(.merge) { resolver.resolve(dependencies: $0, lastResolved: nil, dependenciesToUpdate: nil) }
 		}
@@ -595,8 +595,6 @@ public final class Project { // swiftlint:disable:this type_body_length
 	///
 	/// Sends a boolean indicating whether binaries were installed.
 	private func installBinaries(for dependency: Dependency, pinnedVersion: PinnedVersion, toolchain: String?) -> SignalProducer<Bool, CarthageError> {
-		let checkoutDirectoryURL = self.directoryURL.appendingPathComponent(dependency.relativePath, isDirectory: true)
-
 		switch dependency {
 		case let .gitHub(server, repository):
 			let client = Client(server: server)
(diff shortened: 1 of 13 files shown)


Tests/CarthageKitTests/XcodeSpec.swift
```diff
@@ -316,6 +316,7 @@ class XcodeSpec: QuickSpec {
 			}
 
 			expect(isExpectedError) == true
+			expect(result.error?.description) == "Dependency \"Swell-0.5.0\" has no shared framework schemes for any of the platforms: Mac"
```
jdhealy: Was this change intended for this PR?

Source/CarthageKit/Project.swift
```diff
@@ -919,10 +917,10 @@ public final class Project { // swiftlint:disable:this type_body_length
 		let fileManager = FileManager.default
 
 		return self.dependencySet(for: dependency, version: version)
-			.zip(with: // file system objects which might conflict with symlinks
-				list(treeish: version.commitish, atPath: carthageProjectCheckoutsPath, inRepository: repositoryURL)
-					.map { (path: String) in (path as NSString).lastPathComponent }
-					.collect()
+			// file system objects which might conflict with symlinks
+			.zip(with: list(treeish: version.commitish, atPath: carthageProjectCheckoutsPath, inRepository: repositoryURL)
```
jdhealy: The linting message I see (*“Colon Violation: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. (colon)”*) is mistaken in that this is a method call with an argument label.

We should just use `// swiftlint:disable:previous colon` on the next line to workaround this errant message from the linter.

jdhealy:
Great changes overall 👏

Tests/CarthageKitTests/XcodeSpec.swift
blender: Yes. This is indented to test the `description` -> `rawValue` change to Platform

jdhealy: Ah, okay. Well, divorced from the potential of these `XcodeSpec` changes being removed, I'm fine enough with the changes to [the `zip(with: //` thing](https://github.com/Carthage/Carthage/pull/2374#discussion_r174268071).

jdhealy:
bravo

jakeadams:
Thank you, all!

Issue #2375, opened 2018-03-14
Segmentation fault when running `carthage bootstrap` 


vytis:
* carthage install method: [ ] .pkg, [x] homebrew, [] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.28
* `xcodebuild -version`: Xcode 9.0.1
* Are you using `--no-build`? No
* Are you using `--no-use-binaries`? No
* Are you using `--use-submodules`? No
* Are you using `--cache-builds`? No
* Are you using `--new-resolver`? Yes

**Cartfile.resolved**
```
github "wireapp/AFNetworking" "2.6.3-iOS8"
github "wireapp/Cartography" "3.0.1"
github "wireapp/Classy" "v0.2.6-swift3.0.2"
github "wireapp/CocoaLumberjack" "3.3.0"
github "wireapp/Down" "v1.0.2-test"
github "wireapp/FLAnimatedImage" "1.0.12-wire"
github "wireapp/FormatterKit" "1.8.1-swift3.0.2"
github "wireapp/HockeySDK-iOS" "5.0.0"
github "wireapp/PINCache" "2.3-swift3.1"
github "wireapp/PureLayout" "v3.0.0"
github "wireapp/TTTAttributedLabel" "2.0.0-swift3.0.2"
github "wireapp/ZipArchive" "v2.0.8"
github "wireapp/avs-ios-binaries" "4.1.131"
github "wireapp/ios-snapshot-test-case" "2.1.4-swift3.1"
github "wireapp/libPhoneNumber-iOS" "0.9.2-swift3.1"
github "wireapp/mixpanel-swift" "v2.2.2"
github "wireapp/ocmock" "v3.4"
github "wireapp/onepassword-app-extension" "framework/1.8.2-wire-swift3.0.2"
github "wireapp/ono" "1.4.0"
github "wireapp/protobuf-objc" "1.9.14"
github "wireapp/wire-ios-canvas" "5.0.2"
github "wireapp/wire-ios-cryptobox" "11.0.1"
github "wireapp/wire-ios-data-model" "93.2.2"
github "wireapp/wire-ios-images" "14.0.0"
github "wireapp/wire-ios-link-preview" "7.0.1"
github "wireapp/wire-ios-message-strategy" "79.0.0"
github "wireapp/wire-ios-protos" "13.0.0"
github "wireapp/wire-ios-request-strategy" "77.0.0"
github "wireapp/wire-ios-share-engine" "66.0.0"
[...]

PR #2376, opened 2018-03-15
Fix for segmentation fault when bootstrapping

vytis:
This is my attempt in fixing the issue discussed in #2375. Thanks @blender for consultations! 

Basically the code was converted from capturing and sharing `platformCaches` dictionary to reducing the results of calculating all hashes to a single dictionary. It is then used to initialize `VersionFile` as before.

This is my first time working with a ReactiveSwift codebase, so I am not sure if I followed all the good practices. Things I think could be improved:

* `createVersionFile` function doesn't do that much but has many parameters...
* Nested tuple in `SignalProducer<(String, (String, String)), CarthageError>`. My first try was passing down `url` instead of `(frameworkName, platformName)`, but then I need to calculate `frameworkName` two times.

Let me know what you think!

### DIFF
--- a/Source/CarthageKit/VersionFile.swift
+++ b/Source/CarthageKit/VersionFile.swift
@@ -248,16 +248,17 @@ public func createVersionFileForCommitish(
 
 	if !buildProducts.isEmpty {
 		return SignalProducer<URL, CarthageError>(buildProducts)
-			.flatMap(.merge) { url -> SignalProducer<(String, URL), CarthageError> in
-				let frameworkName = url.deletingPathExtension().lastPathComponent
-				let frameworkURL = url.appendingPathComponent(frameworkName, isDirectory: false)
-				return SignalProducer.zip(hashForFileAtURL(frameworkURL), SignalProducer<URL, CarthageError>(value: url))
-			}
-			.reduce(into: platformCaches) { (platformCaches: inout PlatformCaches, values: (String, URL)) in
-				let hash = values.0
-				let url = values.1
+			.flatMap(.merge) { url -> SignalProducer<(String, (String, String)), CarthageError> in
 				let frameworkName = url.deletingPathExtension().lastPathComponent
 				let platformName = url.deletingLastPathComponent().lastPathComponent
+				let frameworkURL = url.appendingPathComponent(frameworkName, isDirectory: false)
+				let details = SignalProducer<(String, String), CarthageError>(value: (frameworkName, platformName))
+				return SignalProducer.zip(hashForFileAtURL(frameworkURL), details)
+			}
+			.reduce(into: platformCaches) { (platformCaches: inout PlatformCaches, values: (String, (String, String))) in
+				let hash = values.0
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/VersionFile.swift
```diff
@@ -216,47 +237,42 @@ public func createVersionFileForCommitish(
...
+			.reduce(into: platformCaches) { (platformCaches: inout PlatformCaches, values: (String, (String, String))) in
+				let hash = values.0
+				let platformName = values.1.0
+				let frameworkName = values.1.1
+
+				let cachedFramework = CachedFramework(name: frameworkName, hash: hash)
+				if var frameworks = platformCaches[platformName] {
+					frameworks.append(cachedFramework)
+					platformCaches[platformName] = frameworks
+				}
```
mdiep: This seems like it needs an `else` case to create the array for the first time?

Source/CarthageKit/VersionFile.swift
```diff
@@ -216,47 +237,42 @@ public func createVersionFileForCommitish(
...
+			}
+			.flatMap(.merge) { platformCaches -> SignalProducer<(), CarthageError> in
+				createVersionFile(commitish, dependencyName: dependencyName, rootDirectoryURL: rootDirectoryURL, platformCaches: platformCaches)
 			}
-			.then(writeVersionFile)
 	} else {
 		// Write out an empty version file for dependencies with no built frameworks, so cache builds can differentiate between
 		// no cache and a dependency that has no frameworks
-		return writeVersionFile
+		return createVersionFile(commitish, dependencyName: dependencyName, rootDirectoryURL: rootDirectoryURL, platformCaches: platformCaches)
```
mdiep: Can we pass `[:]` instead of `platformCaches` here to make it obvious that it's an empty list?

Then we can move the declaration of `platformCaches` to be closer to where it's used.

mdiep:
Thanks for the PR! It's always great to have a new contributor.

This looks pretty good! This is exactly the sort of structure I was hoping for. ✨

Source/CarthageKit/VersionFile.swift
vytis: Good point!

Source/CarthageKit/VersionFile.swift
vytis: As I understand in the code above (https://github.com/Carthage/Carthage/pull/2376/files#diff-6206d1bf8ed3195b3b646bc3d2b6553dR244) it creates empty arrays for each platform we support. Then for platforms that are not supported nothing gets added to the dict. Not sure if that is the behaviour that we need, but this was the way original code was structured.

mdiep: Oops! You're totally right. I was looking for code like that and just couldn't find it.

blender:
A small segmentation fault fix for @vytis, a giant bug fix for Carthage. 🎉

PR #2378, opened 2018-03-17
Fix swiftlint comma warnings

yoching:
Hi!. Fixed swiftlint comma warnings in VersionFile.swift


mdiep:
Thanks for the PR! Fixes like this are really appreciated. ✨

PR #2383, opened 2018-03-22
[BinaryProject] Relative Path Cartfile.resolved Support

dmiluski:
## Problem:
PR https://github.com/Carthage/Carthage/pull/2360 which allows anyone to update and install required binaries via relative path. BUT.. once installed, the resolved file is tied to a user specific filepath

## Current Functionality/Limitation (0.29.0)
`carthage update` with current relative implementation works... but leaves an absolute file path for binary json specs, which will fail on others computers

`carthage bootstrap` fails if not run on computer which ran update b/c path is not going to be the same.

Cartfile
```
binary "CarthageSpecs/AppboyFramework.json"
```

Cartfile.resolved
```
binary "file:///Users/dmiluski/code/folder/CarthageSpecs/AppboyFramework.json" "3.3.2"
```

This means the resolved file path would only work on the original computer used to run.

## Proposed Solution/Expectation:
binary urls should appear the same for both Cartfile & Cartfile.resolved

Cartfile
```
binary "CarthageSpecs/AppboyFramework.json"
```

Cartfile.resolved
```
binary "CarthageSpecs/AppboyFramework.json"  "3.3.2"
```

This allows relative paths to still be used on multple devices/users.

## Approach:
- Package Binary dependency model with a url + representation string
- Update resolve write out to use binary representable description

### Proof of Concept: Automated Github Releases Fetch & Construct Spec file
https://github.com/dmiluski/CarthageSpecUpdater/blob/master/update_carthage_specs

### DIFF
--- a/Source/CarthageKit/Dependency.swift
+++ b/Source/CarthageKit/Dependency.swift
@@ -2,15 +2,29 @@ import Foundation
 import Result
 import Tentacle
 
+/// Protocol which supports URL + potential custom description
+public protocol URLConvertible: CustomStringConvertible {
+	var url: URL { get }
+}
+
 /// Uniquely identifies a project that can be used as a dependency.
 public enum Dependency {
+	/// Uniquely identifies a Binary Spec's resolved URL and its description
+	public struct BinaryURL: URLConvertible {
+		/// A Resolved URL
+		public let url: URL
 
-	public struct Binary {
-		// Actual Absolute URL
-		let absoluteURL: URL
+		/// An Optional custom description
+		public let overloadedDescription: String?
 
-		// Representative Value
-		let representation: String
+		public var description: String {
+			return overloadedDescription ?? url.description
+		}
+
+		init(url: URL, overloadedDescription: String? = nil) {
+			self.url = url
+			self.overloadedDescription = overloadedDescription
+		}
 	}
 
 	/// A repository hosted on GitHub.com or GitHub Enterprise.
@@ -20,7 +34,7 @@ public enum Dependency {
 	case git(GitURL)
 
 	/// A binary-only framework
-	case binary(Binary)
+	case binary(BinaryURL)
 
 	/// The unique, user-visible name for this project.
 	public var name: String {
@@ -31,8 +45,8 @@ public enum Dependency {
 		case let .git(url):
 			return url.name ?? url.urlString
 
-		case let .binary(binary):
-			return binary.absoluteURL.lastPathComponent.stripping(suffix: ".json")
+		case let .binary(url):
+			return url.url.lastPathComponent.stripping(suffix: ".json")
 		}
 	}
 
@@ -110,21 +124,8 @@ extension Dependency: Hashable {
 			return url.hashValue
 
 		case let .binary(binary):
-			return binary.hashValue
+			return binary.url.hashValue
 		}
-	}
-}
-
-extension Dependency.Binary: Equatable {
-	public static func == (_ lhs: Dependency.Binary, _ rhs: Dependency.Binary) -> Bool {
-		return lhs.absoluteURL == rhs.absoluteURL &&
-				lhs.representation == rhs.representation
-	}
-}
-
-extension Dependency.Binary: Hashable {
-	public var hashValue: Int {
-		return absoluteURL.hashValue ^ representation.hashValue
 	}
 }
 
@@ -150,10 +151,10 @@ extension Dependency: Scannable {
 			parser = { urlString in
 				if let url = URL(string: urlString) {
 					if url.scheme == "https" || url.scheme == "file" {
-						return .success(self.binary(Binary(absoluteURL: url, representation: url.absoluteString)))
+						return .success(self.binary(BinaryURL(url: url)))
 					} else if url.scheme == nil {
 						let absoluteURL = URL(fileURLWithPath: url.relativePath, isDirectory: false, relativeTo: base).standardizedFileURL
-						return .success(self.binary(Binary(absoluteURL: absoluteURL, representation: url.absoluteString)))
+						return .success(self.binary(BinaryURL(url: absoluteURL, overloadedDescription: url.absoluteString)))
 					} else {
 						return .failure(ScannableError(message: "non-https, non-file URL found for dependency type `binary`", currentLine: scanner.currentLine))
 					}
@@ -200,7 +201,7 @@ extension Dependency: CustomStringConvertible {
 			return "git \"\(url)\""
 
 		case let .binary(binary):
-			return "binary \"\(binary.representation)\""
+			return "binary \"\(binary)\""
 		}
 	}
 }
@@ -222,5 +223,17 @@ extension Dependency {
 		case .binary:
 			return nil
 		}
+	}
+}
+
+extension Dependency.BinaryURL: Equatable {
+	public static func == (_ lhs: Dependency.BinaryURL, _ rhs: Dependency.BinaryURL) -> Bool {
+		return lhs.description == rhs.description
+	}
+}
+
+extension Dependency.BinaryURL: Hashable {
+	public var hashValue: Int {
+		return description.hashValue
 	}
 }
--- a/Source/CarthageKit/Dependency.swift
+++ b/Source/CarthageKit/Dependency.swift
@@ -115,14 +115,10 @@ extension Dependency: Hashable {
 	}
 }
 
-extension Dependency.Binary: Comparable {
+extension Dependency.Binary: Equatable {
 	public static func == (_ lhs: Dependency.Binary, _ rhs: Dependency.Binary) -> Bool {
 		return lhs.absoluteURL == rhs.absoluteURL &&
 				lhs.representation == rhs.representation
-	}
-
-	public static func < (_ lhs: Dependency.Binary, _ rhs: Dependency.Binary) -> Bool {
-		return lhs.representation.caseInsensitiveCompare(rhs.representation) == .orderedAscending
 	}
 }
 
--- a/Source/CarthageKit/Dependency.swift
+++ b/Source/CarthageKit/Dependency.swift
@@ -7,16 +7,16 @@ public struct BinaryURL: CustomStringConvertible {
 	/// A Resolved URL
 	public let url: URL
 
-	/// An Optional custom description
-	public let overloadedDescription: String?
+	/// A custom description
+	public let resolvedDescription: String
 
 	public var description: String {
-		return overloadedDescription ?? url.description
+		return resolvedDescription
 	}
 
-	init(url: URL, overloadedDescription: String? = nil) {
+	init(url: URL, resolvedDescription: String) {
 		self.url = url
-		self.overloadedDescription = overloadedDescription
+		self.resolvedDescription = resolvedDescription
 	}
 }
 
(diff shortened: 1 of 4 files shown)


Source/CarthageKit/Dependency.swift
```diff
@@ -100,12 +109,29 @@ extension Dependency: Hashable {
...
 	}
 }
 
+extension Dependency.Binary: Comparable {
+	public static func == (_ lhs: Dependency.Binary, _ rhs: Dependency.Binary) -> Bool {
+		return lhs.absoluteURL == rhs.absoluteURL &&
+				lhs.representation == rhs.representation
+	}
+
+	public static func < (_ lhs: Dependency.Binary, _ rhs: Dependency.Binary) -> Bool {
```
blender: Why not just Equatable?

Source/CarthageKit/Dependency.swift
```diff
@@ -100,12 +109,29 @@ extension Dependency: Hashable {
...
+		return lhs.representation.caseInsensitiveCompare(rhs.representation) == .orderedAscending
+	}
+}
+
+extension Dependency.Binary: Hashable {
+	public var hashValue: Int {
+		return absoluteURL.hashValue ^ representation.hashValue
```
dmiluski: Not sure if this needs to hash both absoluteURL and representation if the representation is considered a unique id?

Source/CarthageKit/Dependency.swift
```diff
@@ -4,14 +4,23 @@ import Tentacle
 
 /// Uniquely identifies a project that can be used as a dependency.
 public enum Dependency {
+
+	public struct Binary {
+		// Actual Absolute URL
+		let absoluteURL: URL
```
dmiluski: Potential for better names options? This is the value resolved to load the resource

Source/CarthageKit/Dependency.swift
```diff
@@ -4,14 +4,23 @@ import Tentacle
...
+
+		// Representative Value
```
dmiluski: This is the value used in the cartfile

Source/CarthageKit/Dependency.swift
dmiluski: You're right, I misread the Dependency comparable as a requirement for this implementation. Will mark just equatable.

Source/CarthageKit/Dependency.swift
blender: What do you think of:

```swift
protocol URLConvertible: CustomStringConvertible {
    var url: URL {get}
}

struct BinaryURL: URLConvertible {
    
    let url: URL
    let overloadedDescription: String?
    var description: String {
        
        return overloadedDescription ?? url.description
    }
    
}


BinaryURL(url: URL.init(string: "https://google.com")!, overloadedDescription: "mom!")
```

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -1,3 +1,4 @@
+@testable import CarthageKit
```
blender: one of the two imports is redundant

Source/CarthageKit/Project.swift
```diff
@@ -226,23 +226,25 @@ public final class Project { // swiftlint:disable:this type_body_length
 			.startOnQueue(cloneOrFetchQueue)
 	}
 
-	func downloadBinaryFrameworkDefinition(url: URL) -> SignalProducer<BinaryProject, CarthageError> {
+	// Dane - what is this value used for?
```
blender: are you trying to ask something or did you just forget the comments in 😄  ?

Source/CarthageKit/Project.swift
```diff
@@ -226,23 +226,25 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+	func downloadBinaryFrameworkDefinition(binary: Dependency.Binary) -> SignalProducer<BinaryProject, CarthageError> {
 		return SignalProducer<Project.CachedBinaryProjects, CarthageError>(value: self.cachedBinaryProjects)
 			.flatMap(.merge) { binaryProjectsByURL -> SignalProducer<BinaryProject, CarthageError> in
-				if let binaryProject = binaryProjectsByURL[url] {
+				if let binaryProject = binaryProjectsByURL[binary.absoluteURL] {
 					return SignalProducer(value: binaryProject)
 				} else {
-					self._projectEventsObserver.send(value: .downloadingBinaryFrameworkDefinition(.binary(url), url))
+					// What arguments create this?
```
blender: same here 😄  ?

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -1,3 +1,4 @@
...
 import CarthageKit
```
blender: one of the two imports is redundant

Tests/CarthageKitTests/DependencySpec.swift
blender: one of the two imports is redundant, see next line

Source/CarthageKit/Dependency.swift
dmiluski: Interesting. I like the Convertible naming friendliness with Swift convertible protocol.

Will experiment with this approach.

Source/CarthageKit/Project.swift
dmiluski: totally forgot to clean up some notes I was following around code

Source/CarthageKit/Dependency.swift
blender: Does not look like to me that the developer using the code nor the code structure requires it to be distinct. That's my opinion however. A few more eyes on this would be nice.

Source/CarthageKit/Dependency.swift
```diff
@@ -2,16 +2,39 @@ import Foundation
...
+
 	/// A repository hosted on GitHub.com or GitHub Enterprise.
 	case gitHub(Server, Repository)
 
 	/// An arbitrary Git repository.
 	case git(GitURL)
 
 	/// A binary-only framework
-	case binary(URL)
+	case binary(BinaryURL)
```
dmiluski: I was attempting to leverage generic protocol URLConvertible, but ran into:
https://www.bignerdranch.com/blog/why-associated-type-requirements-become-generic-constraints/

In this case, I feel the separate protocol does not provide enough benefit to exist on it's own. So I feel I can toss the protocol in favor with the publicly available BinaryURL

Tests/CarthageKitTests/DependencySpec.swift
dmiluski: Remove initial import

dmiluski:
@mdiep , if you have time for review of the proposal?

Source/CarthageKit/Dependency.swift
```diff
@@ -23,7 +41,7 @@ public enum Dependency {
 			return url.name ?? url.urlString
 
 		case let .binary(url):
-			return url.lastPathComponent.stripping(suffix: ".json")
+			return url.url.lastPathComponent.stripping(suffix: ".json")
```
mdiep: Can me move this logic into a method on `BinaryURL`?

Source/CarthageKit/Dependency.swift
```diff
@@ -100,8 +118,8 @@ extension Dependency: Hashable {
 		case let .git(url):
 			return url.hashValue
 
-		case let .binary(url):
-			return url.hashValue
+		case let .binary(binary):
+			return binary.url.hashValue
```
mdiep: This can be `return binary.hashValue`.

Source/CarthageKit/Dependency.swift
```diff
@@ -4,14 +4,32 @@ import Tentacle
 
 /// Uniquely identifies a project that can be used as a dependency.
 public enum Dependency {
+	/// Uniquely identifies a Binary Spec's resolved URL and its description
+	public struct BinaryURL: CustomStringConvertible {
```
mdiep: Let's make this a top-level type.

Source/CarthageKit/Dependency.swift
```diff
@@ -4,14 +4,32 @@ import Tentacle
...
+		/// A Resolved URL
+		public let url: URL
+
+		/// An Optional custom description
+		public let overloadedDescription: String?
```
mdiep: I'm not wild about this name. Maybe something with `resolved` in it?

Source/CarthageKit/Dependency.swift
```diff
@@ -4,14 +4,32 @@ import Tentacle
...
+
+		public var description: String {
+			return overloadedDescription ?? url.description
+		}
+
+		init(url: URL, overloadedDescription: String? = nil) {
+			self.url = url
+			self.overloadedDescription = overloadedDescription
+		}
```
mdiep: Do we need to get the absolute URL when reading from `Cartfile.resolved`? Will it cause problems if this binary URL isn't equal to the one that's read from the `Cartfile`?

Source/CarthageKit/Dependency.swift
dmiluski: Will address in upcoming commit

Source/CarthageKit/Dependency.swift
dmiluski: Will address in upcoming commit

Source/CarthageKit/Dependency.swift
dmiluski: Will address in upcoming commit, wasn't sure if it was useful outside Dependency.

Source/CarthageKit/Dependency.swift
dmiluski: In this case, I wasn't sure what to call it: (Likely only beneficial for relative paths 'short hand')

Alternatives:
customDescription
identifier
...

Source/CarthageKit/Dependency.swift
dmiluski: That is one thing I wasn't sure of.

My expectation was:
If Cartfile uses a relative path, so should a Cartfile.resolved, otherwise, an absolute file page would break between machines. Otherwise, it would appear folks would leverage a more universal https instead?

In my use case, I was assuming Spec files would be local to the repo (scripted), and updated when we update our dependencies.
https://github.com/dmiluski/CarthageSpecUpdater/blob/master/update_carthage_specs

Why:
I'm tired of maintaining my own Spec repo for binary only dependencies, which don't already offer a Binary Spec json file. Thus trying to automate. Potential for alternate approaches? Also tired of manually downloading, and adding to our repo (bloating).

So, I had lead with the approach of the `overriddenDescription` as being the unique identifier to a transitive absoluteURL that was used for actual file loading?

Source/CarthageKit/Dependency.swift
dmiluski: I was unsure about this here?

In this case, this struct is for Binary URLs, but the only use case I need this override is for relative Binary Spec paths.

relative is a tricky thing to consider b/c most urls have a relative path, But this isn't necessary for all urls?

mdiep: How about `resolvedDescription`? Maybe we could make it non-optional and include the correct value for absolute URLs as well?

Source/CarthageKit/Dependency.swift
mdiep: After thinking about it a bit more, I don't think this is an issue.

blender:
@dmiluski Where do we stand on this?

dmiluski:
@blender, currently on vacation with only a phone. Will update with previous recommendations once I get back on Monday

dmiluski:
@mdiep , updated PR with:
+ Move BinaryURL to toplevel, renamed required usage
+ Implement name in BinaryURL
- overloadedDescription -> resolvedDescription (required)

dmiluski:
@mdiep ping?

dmiluski:
Hi @mdiep @blender , seeing if there are any other items here I still need to address?

mdiep:
This is on my todo list and I plan to get to it in the next couple days.

mdiep:
Looks good! Thanks @dmiluski!

gabrielmaldi:
Great to see this work done and merged, thanks! When can we expect a release?

blender:
@gabrielmaldi if you want to try this out you can always install via brew with --HEAD. This will build current master.

marcvanolmen:
Thanks you guys this is an important enhancement on the original feature!

dmiluski:
@blender , I just installed via brew 0.29.0, but found my Cartfile.resolved still wasn't showing relative path. Instead including file:///***

Was there an absoluteURL writeout I may have missed? Grepping code now, but only see 
```
		case let .downloadingBinaryFrameworkDefinition(dependency, url):
			carthage.println(formatting.bullets + "Downloading binary-only framework " + formatting.projectName(dependency.name)
				+ " at " + formatting.quote(url.absoluteString))
```
Which I'm not too worried about showing absoluteString.

dmiluski:
nm, I'm an idiot, I read the 0.29.0 change as including this PR

PR #2390, opened 2018-03-25
Provide information on linker warning workaround for static framework script

mfclarke:
I encountered some linker warnings after following the instructions here, and wanted to provide information on a workaround.

I think this would be ultimately solved by somehow providing a custom modulemap for the larger dynamic framework that defines correct modules for each of the static frameworks that make it up, but that's beyond me at the moment.


mdiep:
Thanks for the PR! Documentation updates are really appreciated. ✨

PR #2392, opened 2018-03-26
 Add support for pre-release and build metadata, as per SemVer specs

marcoconti83:
According to [SemVer specs](http://semver.org), pre-release and build metadata are valid components of a version specification. Adding support for then in the version parser.

This PR is derived from the work on https://github.com/Carthage/Carthage/pull/1553, and it's the first of the tree discussed there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ cache:
 branches:
   only:
     - master
+    - fix/semver-support
 matrix:
   include:
     - os: osx
--- a/Source/CarthageKit/Version.swift
+++ b/Source/CarthageKit/Version.swift
@@ -8,6 +8,7 @@ import ReactiveSwift
 public protocol VersionType: Hashable {}
 
 /// A semantic version.
+/// - Note: See <http://semver.org/>
 public struct SemanticVersion: VersionType {
 	/// The major version.
 	///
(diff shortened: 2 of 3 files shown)


blender:
Good job! Looks good to me 👍

blender:
@marcoconti83 I changed this to point to fix/semver-support which is 1:1 with master

marcoconti83:
Thanks @blender . I will start working on the next step when both this and #2394 are merged

blender:
ok, since we're now on a branch I think that @jdhealy can merge once he has reviewed and we can take in from there

mdiep:
These changes look good to me, but I'd like to see a merge from the branch after #2394 is merged so we can see the affect, if any, on those tests.

marcoconti83:
@mdiep I rebased this branch on `fix/semver-support` after merging  #2394, and updated the test that were already affected by this.
I will update the other tests in a following PR.

marcoconti83:
@mdiep I rebased this branch on `fix/semver-support` after merging  #2394, and updated the test that were already affected by this. See https://github.com/Carthage/Carthage/pull/2392/commits/4e89e8042b92565e55e3892c32c20500ca691b87. 
I will update the other tests in a following PR.

jdhealy:
https://github.com/Carthage/Carthage/blob/b44c294a8c5ddfded864e84d9802b394db4befdc/.travis.yml#L8-L10

@marcoconti83 Adding `- fix/semver-support` prefixed with four spaces (under the above line) will (I believe) allow Travis CI to execute the tests and report back on this PR and and future PRs to [it](https://github.com/Carthage/Carthage/tree/fix/semver-support).

Adding a commit for that in this PR would be fine. Later on, the last PR to target [that branch](https://github.com/Carthage/Carthage/tree/fix/semver-support) can include a commit to restore just `master` under `branches: only:` — should be sufficient :+1:

jdhealy:
My bad for not recognizing that earlier.

marcoconti83:
@jdhealy I modified the travis configuration accordingly.

mdiep:
👏 

On PR to go, right?

marcoconti83:
Here it is: https://github.com/Carthage/Carthage/pull/2406

PR #2396, opened 2018-03-28
JSON specification file name recommendation for binary projects

TwiterZX:
As there is a lot of binary framework developers that call their specification file Carthage.json, updating frameworks becomes difficult and confusing
Updating those frameworks needs the command `carthage update Carthage` that first sounds weird, but also does update all binary dependencies with this same specification name.


mdiep:
Absolutely! Thanks for opening a documentation change for this. Hopefully it'll help.

TwiterZX:
👍

PR #2406, opened 2018-04-04
Implement comparison and intersection of pre-release and build versions

marcoconti83:
This commit introduces the logic for comparison of pre-release versions and versions with build metadata.

For the purpose of comparison:
- pre-release versions follow the specifications at http://semver.org/
- build metadata is ignored when performing comparisons

For the purpose of intersection of requirements (i.e. find a version that satisfies two requirements):
- pre-release versions follow the specifications at http://semver.org/
- build metadata is ignored when using a `.atLeast`, `.any` or `.compatibleWith` requirement, but taken into account when using an `.exactly` requirement.

### DIFF
--- a/Source/CarthageKit/Version.swift
+++ b/Source/CarthageKit/Version.swift
@@ -42,6 +42,11 @@ public struct SemanticVersion: VersionType {
 	public var components: [Int] {
 		return [ major, minor, patch ]
 	}
+	
+	/// Whether this is a prerelease version
+	public var isPreRelease : Bool {
+		return self.preRelease != nil
+	}
 
 	public init(major: Int, minor: Int, patch: Int, preRelease: String? = nil, buildMetadata: String? = nil) {
 		self.major = major
@@ -69,6 +74,14 @@ public struct SemanticVersion: VersionType {
 				return .failure(ScannableError(message: "syntax of version \"\(version)\" is unsupported", currentLine: scanner.currentLine))
 			}
 		}
+	}
+	
+	/// The same SemanticVersion as self, except that the build metadata is discarded
+	public var discardingBuildMetadata: SemanticVersion {
+		return SemanticVersion(major: self.major,
+		                       minor: self.minor,
+		                       patch: self.patch,
+		                       preRelease: self.preRelease)
 	}
 }
 
@@ -139,15 +152,18 @@ extension Scanner {
 	}
 }
 
-
-
 extension SemanticVersion: Comparable {
 	public static func < (_ lhs: SemanticVersion, _ rhs: SemanticVersion) -> Bool {
+		if lhs.components == rhs.components {
+			return lhs.isPreReleaseLesser(preRelease: rhs.preRelease)
+		}
 		return lhs.components.lexicographicallyPrecedes(rhs.components)
 	}
 
 	public static func == (_ lhs: SemanticVersion, _ rhs: SemanticVersion) -> Bool {
 		return lhs.components == rhs.components
+			&& lhs.preRelease == rhs.preRelease
+			&& lhs.buildMetadata == rhs.buildMetadata
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Version.swift
```diff
@@ -70,6 +75,14 @@ public struct SemanticVersion: VersionType {
 			}
 		}
 	}
+	
+	/// The same SemanticVersion as self, except that the build metadata is discarded
+	public var discardingBuildMetadata: SemanticVersion {
```
mdiep: Do we even care about the build metadata?

Let's just remove it from `SemanticVersion` entirely. (But allow it in the parsing code.)

Source/CarthageKit/Version.swift
```diff
@@ -221,34 +332,54 @@ public enum VersionSpecifier: VersionType {
 			if let semanticVersion = SemanticVersion.from(version).value {
 				return predicate(semanticVersion)
 			} else {
-				// Consider non-semantic versions (e.g., branches) to meet every
-				// version range requirement.
-				return true
+				// Consider non-semantic versions (e.g., branches) not to meet
+				// any requirement as we can't guarantee any ordering nor
+				// compatibility
+				return false
```
mdiep: We don't want to change this because it's considered a feature: You can specify a branch at the root of your dependency tree and it will override other versions in the tree.

Source/CarthageKit/Version.swift
marcoconti83: What about `.exactly`? 

I think it makes no sense to consider build metadata for `.atLeast` (also, no clear ordering is defined with build metadata anyway) and `.compatibleWith` (any other version with same major-minor-patch, without the build metadata, is compatible). 

However I would find if weird if when specifying `exactly 1.0.2+build2`, it would resolve to `1.0.2+build8997`. Is there any case in which you see this as desired?

Source/CarthageKit/Version.swift
marcoconti83: Got it. I will update all tests.

marcoconti83: If I understand correctly, this is actually the original issue that I was trying to solve when I started working on these PR one year ago.

On our repo, someone pushed a tag with pre-release or build metadata. Because it did not parse as a valid version, it will be considered non-sematic. And because of this line you commented on, it would always pick that non semantic one.

If we revert this line, we would go back to that behavior - now pre-releases will be parsed properly, but anyone pushing any other random (non-semantic) tag to the repo would case that tag to always be picked, no matter what is specified in the Cartfile.

E.g. if I specify `>= 5.0.0` in the Cartfile, and the tags on the repo are:

- `0.1.0`
- `Demo`
- `1.0.0`
- ...
- `5.0.1`

It would resolve to `Demo` instead of `5.0.1`. 

It might be the responsibility of the repo mantainers to not allow any tag with non-semantic versioning, but this is something that is easy to overlook. Once one of those has been pushed, it will disrupt everyone working with that repository.

If I understand correctly, the desired behavior is that if I specifiy a commit SHA, a branch name, or a non-sematic tag in the Cartfile in the root project, it will pick that specific commit for this repo and for all dependencies, ignoring whatever the dependency specifies. It makes sense, however this code achieves that with the nasty side effect that I described above. 

Did I get this right?

If this is the case, I would like to implement the desired functionality in a different manner not to have the side effect.

Source/CarthageKit/Version.swift
mdiep: Hmm… that's tricky. I'm not sure this makes sense, so my preference would be to disallow it for now. Maybe we should just not even parse it? I don't know if anyone even really uses this.

Source/CarthageKit/Version.swift
mdiep: > If I understand correctly, this is actually the original issue that I was trying to solve when I started working on these PR one year ago.

Yes and no. This _would_ solve it for tags pre-release builds because those would no longer parse as non-semantic versions. But you're right that a `Demo` tag could still run into the bug you encountered.

The non-semantic version bug really needs to be fixed in the resolver. Even though a non-semantic version is compatible with any version, the resolver shouldn't pick a non-semantic version unless one is specifically requested.

Source/CarthageKit/Version.swift
mdiep: I'm definitely open to ideas on this, because I'm not really sure what the right answer is.

Source/CarthageKit/Version.swift
marcoconti83: My suggestion is then: 
- revert this line to keep the previous behavior
- merge this and the other pre-release PRs to master to mitigate the issue (i.e. at least pre-releases tags will not cause the bug)
- investigate how to fix the resolver so that it does not pick a non-semantic version unless one is specified in a completely separate branch, without having that fix tied to the release of pre-release version parsing

Source/CarthageKit/Version.swift
marcoconti83: I considered and discarded the following alternatives:

1. discarding build metadata completely. This feels broken, as when I say that I want *exactly* build `1.2.0+b32`, Carthage will instead pick `1.2.0+b8934`.
1. consider build metadata when resolving for `.compatibleWith`. This makes no sense: by definition, all version with the same major-minor-patch are compatible, so a different build metadata should not make a version not compatible.
1. consider build metadata when resolving for `.atLeast`. This goes against SemVer specifications as there is no ordering defined between build versions. From the [specs](https://semver.org/#spec-item-10): 
> Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.

One alternative I can think of would be to modify the first rule in the following way:

- if required to take `exactly 1.2.0+b32` (i.e. the requirement has build metadata), then only the version with that exact build metadata should be picked.
- if required to take `exactly 1.2.0` (i.e. no build metadata), then it will pick *any* `1.2.0` version even if they have a build metadata, e.g. it would pick `1.2.0+b42`. If there are multiple ones, it would pick the first one matching.

Source/CarthageKit/Version.swift
mdiep: That sounds perfect. 👍

blender:
@mdiep I think then this one is ready to go. 3 down, one more on the resolved.

marcoconti83:
@blender I would work the resolver as a completely separate PR that I suggest should not block the merge of these 3 PR to master, as I have yet to understand how much work is required and when I can find the time to do it.

We still have an open discussion on this PR regarding what to do with build metadata though.

Source/CarthageKit/Version.swift
mdiep: We can roll with it as-is and make adjustments down the road if needed.

mdiep:
Thanks for this! Want to open a PR against `master` from this branch?

blender:
@marcoconti83 can't do that I believe. I'll do it for him.

PR #2407, opened 2018-04-05
Prefer SSH over HTTPS by default

Nirma:
I think that SSH should be the default with the option to use HTTPS via `--use-https` available.

This saves time when a user has Private dependancies written in their Cartfile and goes ahead with cloning the repository as long as SSH access is properly setup. 

If SSH is the new default having `--use-ssh` does not make sense anymore but having an option for `--use-https` does. 

What do you think? Am I far off the mark on this one?

### DIFF
--- a/Source/Scripts/carthage-bash-completion
+++ b/Source/Scripts/carthage-bash-completion
@@ -28,7 +28,7 @@ _carthage() {
                 ;;
             bootstrap|update)
                 COMPREPLY=($(compgen -W '--no-checkout --no-build --verbose --configuration \
-                    --platform --toolchain --derived-data --use-ssh --use-submodules \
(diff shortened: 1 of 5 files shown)


mdiep:
GitHub actually recommends using HTTPS over SSH, which is why Carthage prefers HTTPS.

Nirma:
@mdiep This now makes sense! Sorry for prematurely sending this!

blender:
@Nirma Can we close this pull request or do you think you want to make further suggestions? 

if you want to make further suggestions I suggest you create an issue to discuss the changes first. You can follow this template:  https://github.com/Carthage/Carthage/pull/2360#issuecomment-369279533

PR #2394, opened 2018-03-26
Expand `VersionSpec` tests to include pre-release and non-semantic versions

marcoconti83:
At the moment, Carthage does not comply with the [semantic versioning](https://https://semver.org/) specifications for pre-release versions, build number versions and non-semantic versions.

This PR aims to document this non-standard behavior. In a following PR, I will make Carthage comply with the semantic versioning specifications and I can invert the logic of these tests.

I explicitly marked the tests that are highlighting the non-standard behavior with the following comment:
> `/// The following behavior does not follow the SemVer <https://semver.org/> specifications`

PR #2412, opened 2018-04-09
Semantic Version Support Enhancements

blender:
This chunk of work by @marcoconti83 builds upon and supersedes #1553 

It contains:
- #2392 
- #2394 
- #2406 
- #2439

Please feel free to fill in the details where missing @marcoconti83  @mdiep


.travis.yml
```diff
@@ -8,6 +8,7 @@ cache:
 branches:
   only:
     - master
+    - fix/semver-support
```
marcoconti83: I think this has to be removed. See https://github.com/Carthage/Carthage/pull/2392#issuecomment-377610046

marcoconti83: @blender can you remove this line? I can't edit it

Dschee:
I have one question to the changes made here:
Will the full SemVer support allow something like this?

```
github "Moya/Moya" ~> 11.0.2+rxmoya
```

Or would it need to be specificed with the exact version definition like so?

```
github "Moya/Moya" "11.0.2+rxmoya"
```

It would be great to be able to finally have at least a workaround solution for #1227 by allowing something like the the first rather than the latter (which might work for projects but not for frameworks). See the discussion in https://github.com/Moya/Moya/pull/1642 for more details.

Dschee:
And here's another approach that just got into my mind which will definitely work and would tackle all problems I have in mind but seems like an old idea:

What if we create new repos that serve just as workarounds just like my branches solution did? In those repos we could use plain old semantic versioning just as in Moya, and the usage side would be this:

```
github "Moya/PlainMoya" ~> 11.0
```

or 

```
github "Moya/RxMoya" ~> 11.0
```

or 

```
github "Moya/ReactiveMoya" ~> 11.0
```

Please note that I'm **not** talking about separating the repo into multiple projects and updating cod e in all of them. That's not what's in the `carthage/...` branches I suggested either, I'm merely talking about repos that only the CI commits to when a new release is made to ensure we have "clean" repos for Carthage users.

I'm just not sure yet about the naming. Using `PlainMoya`, `RxMoya` and `ReactiveMoya` as suggested above might lead to some confusion for non-Carthage users. So maybe something like `Moya/CarthageMoya`, `Moya/CarthageRxMoya` and `Moya/CarthageReactiveMoya`?

In case you agree with this approach, I'm not sure if I have the rights to create such projects and integrate the CI with them. I'd be happy about hints on what I need for implementing this restriction-wise.

What do you think?

marcoconti83:
> I have one question to the changes made here:
> Will the full SemVer support allow something like this?
> 
>     github "Moya/Moya" ~> 11.0.2+rxmoya
> 
> Or would it need to be specificed with the exact version definition like so?
> 
>     github "Moya/Moya" "11.0.2+rxmoya"

Are you asking if by specifying `11.0.2+rxmoya` it will pick any version compatible with `11.0.2` (so `11.0.2` or greater) that has the "+rxmoya" build metadata? 

I believe this is not how build metadata is intended to be used according to [semantic versioning](https://semver.org/#spec-item-10), so we would deviate from the standard behavior, which I would advise against.

I had a brief look at the (very long 😄 ) discussion you linked, and I see how your problem is relevant but I would suggest solving it without using semantic versioning.

Dschee:
Yes, that's exactly what I meant. And two things I want to note:

1. I already suggested an alternative solution that's not based on SemVer metadata. Actually I accidentally posted that solution here as a comment. I just deleted it and moved it [to the thread it belongs to](https://github.com/Moya/Moya/pull/1642#issuecomment-382500110).

2. Semantic Versioning doesn't make any statements about what you use the metadata for. It merely states what the structure would look like and that it shouldn't be considered part of the version specification when it comes to updating to the latest version. That's it. This means we _could_ specify that the metadata serves as a filter when used as part of a `~> 11.0.2+rxmoya` version specification. Since I found another solution though, I'm no longer suggesting that we need it. Nor am I saying I'm against it. But I don't think SemVer forbids it.

mdiep:
@jdhealy Do you still want to review this?

jdhealy:
Yeah, sorry, hitting an issue — that seems local to my machine — compiling this with Xcode 9.3, so I haven't been able to dive into this as deeply as I'd like yet.

I've got more reading to do off <https://semver.org/#spec-item-10>, but 
> Identifiers [for build metadata] MUST NOT be empty

…stood out to me; has this been mentioned before in this PR chain?

Whether something like `PinnedVersion("v0.7.7+")` (or more generally an empty string for `SemanticVersion.buildMetadata`) would _actually_ cause issue or unexpectedness, I've yet to fully consider. Thoughts?

marcoconti83:
@jdhealy I can add a unit test for "v0.7.7+". Since SemVer specifies that it MUST NOT be empty, I would parse it as invalid.

blender:
@jdhealy @marcoconti83 @mdiep ping 

... runs away

mdiep:
A unit test would be great

marcoconti83:
@jdhealy tested (and fixed!) here: https://github.com/Carthage/Carthage/pull/2439

blender:
@marcoconti83 @jdhealy @mdiep ... and we're back. 

Seems like this is good to go now.

marcoconti83:
Thanks for taking care of this PR, @blender !

mdiep:
Looks good to me. But let's wait for @jdhealy to review too since he's expressed interest.

blender:
> Changing what tags satisfy .any cannot be under the ‘slight tweak’ model, in my opinion; it needs more initial scrutiny. Such a major change to package graph resolution has to be a measure-twice→cut-once release considering that old carthage executables can remain in use on miscellaneous CI servers for a lot longer than expected, and the difference from what’s on one’s local machine can often be easily missed.

True, however only `update` will try to resolve the graph again given that a `Cartfile.resolved`exists for other cases and it's checked in. So old versions of Carthage will exhibit the new behavior.

Given that it is always possible to roll back carthage to a previous version on one's personal machine I don't see it as a big problem. I would imagine however for this change to be clearly stated in the release notes along with a 0.XX._ -> 0.YY.0 version bump.

Last I would argue that if a dependency is floating with `.any` you're kind of asking for trouble, as we see over and over in the issues that get opened.

marcoconti83:
@jdhealy, for as much as I would love to have this PR finally merged in, I see why you are cautious about changing the behavior for `.any`. I'll let the maintainers of Carthage discuss on a roll-out strategy.

Regarding `v7.1.1+build01-alpha`, I would stick to the SemVer specs, which specify that this is a valid build metadata. It's not confusing if the user understands how SemVer works, which I think it's less cognitive load than understanding SemVer and remembering the exceptions that Carthage implemented.

jdhealy:
@marcoconti83 thanks for your understanding. This PR is definitely a priority for me this week, to work on squaring this away, and readying it for release. I'll try to post my thoughts frequently. With the nature of open-source, if you reach limits on having time to respond or make changes, feel free to delegate any task to me, and I can post potential patches or do some testing or whatever. This is getting closer, can't thank you enough.

@blender:

> True, however only update will try to resolve the graph again given that a `Cartfile.resolved` exists for other cases and it's checked in. So old versions of Carthage will exhibit the new behavior.

That mitigates for the most part, but the corollary is the slight concern — in the situation of `bootstrap` without `Cartfile.resolved` — that we misparse, resolve something unintended which gets output in `Cartfile.resolved` and that creates confusion with use of a future tweaked-not-to-misparse Carthage release.

###### A misparse on comments:

Didn't think to test with [comments (`# …`)](https://github.com/Carthage/Carthage/blob/0.29.0/Source/CarthageKit/Cartfile.swift#L29-L36) until today, but for lines with a `+` character, the `# …` gets scanned into `buildMetadata`:

￼
comment scanned into «buildmetadata» · carthage «ffa502c8b7a7d8693361664dc2246b4f35fb16b1»

jdhealy:
and, you’re right on: `v7.1.1+build01-alpha` is 👌 with semver spec — I was misremembering it instead of something else… maybe, it comes to mind later…

marcoconti83:
@jdhealy I see the error with `#`, however I'm not sure how to fix it - it seems to me like this is an issue with the code that reads the line from the file rather than the version parser - I think it should be stripped even before passing it to the version parser. 

I can try to fix it but I didn't touch that part of the code, could you give me some pointers on where to look?

jdhealy:
@marcoconti83 I believe the totality of it is here: https://github.com/Carthage/Carthage/blob/0.29.0/Source/CarthageKit/Cartfile.swift#L29-L36

jdhealy:
The scanner object (as a reference) gets passed here: https://github.com/Carthage/Carthage/blob/0.29.0/Source/CarthageKit/Cartfile.swift#L43

blender:
@jdhealy Yes, `boostrap` will have different outcomes. 

However, __if you don't pin dependencies AND don't have a Carthage.resolved file__,  bootstrap can anywhere produce two different resolved files. The user is at the mercy of the developers of the dependencies anyways.

Please correct me if this is not the direction you where point at.

marcoconti83:
@jdhealy here's my fix for the comment being included in the SemVer parser: https://github.com/Carthage/Carthage/pull/2452

marcoconti83:
Please also see https://github.com/Carthage/Carthage/pull/2453

marcoconti83:
@jdhealy @mdiep so what's the next step for this PR? Waiting for your input

blender:
@marcoconti83 This node here has to resolved https://github.com/Carthage/Carthage/issues/2443

From a technical perspective I think we're good to go.

marcoconti83:
I addressed the fullwidth characters issue here: https://github.com/Carthage/Carthage/pull/2458

marcoconti83:
And the comment parsing issue mentioned by @jdhealy in https://github.com/Carthage/Carthage/pull/2452 was addressed in https://github.com/Carthage/Carthage/pull/2459

marcoconti83:
@jdhealy is this what you meant? https://github.com/Carthage/Carthage/pull/2464

jdhealy:
@marcoconti83 https://github.com/Carthage/Carthage/pull/2459#issuecomment-391613329 👍

marcoconti83:
Addressed the feedback on the `ignoreCartfileComments` parameter here: https://github.com/Carthage/Carthage/pull/2469

marcoconti83:
Anything else that needs to be fixed on this PR?

mdiep:
Looks good to me. @jdhealy?

mdiep:
_bump_

mdiep:
Alright, let's go for it. 👍 

Thanks for all your work on this @marcoconti83!

jdhealy:
Shipped with [Carthage v0.30.0](https://github.com/Carthage/Carthage/releases/tag/0.30.0) — great work on this @marcoconti83! 🎉

marcoconti83:
Thanks @jdhealy  and @mdiep for reviewing and to @blender for pushing this along. It took one and a half year but we managed :)

PR #2415, opened 2018-04-10
Complete support building static frameworks

Aranoledur:
Decided to open this PR, because #1935 seems to stuck. Allows carthage to build static frameworks and fixes when more than one xcode target has the same Product name for same Deployment target and configuration. Really would like to see this feature in next release, going to support it till it's done. Can't see any downsides for now.


blender:
@Aranoledur Thank you very much for your effort, I believe I speak for everyone what I say it's very much appreciated 🥇 

In order to speed up the review and aid understanding, can you give a short recap of the approach that you too to solve the challenges posed in #1935

MikeSilvis:
@Aranoledur @blender would love to see this pushed through

Aranoledur:
I don't know why travis fails, maybe because of Xcode 9.2

Tests/CarthageKitTests/XcodeSpec.swift
```diff
@@ -380,11 +380,94 @@ class XcodeSpec: QuickSpec {
 			let result = ProjectLocator.locate(in: directoryURL.appendingPathComponent("ReactiveCocoaLayout")).first()
 			expect(result).to(beNil())
 		}
+
+		it("should build static library and place result to subdirectory") {
```
artemnovichkov: `make swiftpm_test` fails in this test with error:
`Fatal error: Unexpectedly found nil while unwrapping an Optional value`

artemnovichkov:
What's about resources in static frameworks built with Carthage?

Aranoledur:
Had to bump osx_image, but now all checks are passing 🎉

Tests/CarthageKitTests/XcodeSpec.swift
Aranoledur: Fixed by changing copy-fixtures script

Source/CarthageKit/Xcode.swift
```diff
@@ -222,6 +222,16 @@ internal enum FrameworkType {
...
 		}
 	}
+
+	/// Folder name for subdirectory
+	var folderName: String {
+		switch self {
+		case .static:
+			return "Static"
+		case .dynamic:
+			return "Dynamic"
```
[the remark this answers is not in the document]
blender: ah right I missed https://github.com/Carthage/Carthage/pull/2415/files/661fb613af5778410709f5fab25351ea34d7014e#diff-2c8b811867554fe0cd24adead2bd16daR295

Aranoledur:
Fixed Xcode 9.2 support

mdiep:
There's a conflict that needs to be resolved.

Aranoledur:
Done

ikesyo:
Looks like Curry submodule is rolled back. Please revert tha change.

Aranoledur:
Done again 😆

Source/CarthageKit/SwiftShims.swift
```diff
@@ -4,4 +4,12 @@ extension Sequence {
 		return try flatMap(transform)
 	}
 }
+
+func == (lhs: FrameworkType??, rhs: FrameworkType) -> Bool {
```
mdiep: This `??` is a code smell and I think we should avoid it. Instead of adding this `==` let's get rid of the `??` in the caller code.

mdiep:
Reviewing this as time allows. I need to look deeper at it, but I'll leave whatever feedback I have as I go.

Source/CarthageKit/SwiftShims.swift
Aranoledur: The problem is, that `BuildOptions.frameworkType` returns `Result<FrameworkType?, ...>`, so if you get the value of this result, it became `FameworkType??`. What should I do with this?

Source/CarthageKit/Xcode.swift
```diff
@@ -664,6 +692,10 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
 					return !dependencyCheckoutDir.hasSubdirectory(projectURL)
 				}
 				.collect()
+				.flatMap(.concat) { settings -> SignalProducer<[BuildSettings], CarthageError> in
+					let producers = settings.map { resolveSameTargetName(for: $0) }
+					return SignalProducer(value: producers).flatten().flatten(.concat).collect()
+				}
```
mdiep: I think moving this before the `.collect` would let you get rid of most of these operators.

```swift
.flatMap(.concat) { settings in resolveSameTargetName(for: settings) }
.collect()
```

mdiep:
Sorry it took me a week to get back to this. I think it looks pretty good. It just needs to be cleaned up in a few places.

Aranoledur:
Update

mdiep:
@jdhealy Do you have any interest in taking a look at this?

mdiep:
Let's go for it. 👍

mendesbarreto:
I love you guys.......

chuganzy:
I guess this breaks `--cache-build`..
https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/VersionFile.swift#L73

Probably we should also store framework URLs into version files..?

Aranoledur:
@chuganzy  Did you test it? I think it will work normal, until you will have built dynamic framework and failed static for the same dependency. I will test and report an issue

chuganzy:
Sorry I have not tested it yet but from what I see in the code, static libs should always be invalid cache.
To fix it, we might need to make some changes to VersionFile I believe.
https://github.com/Carthage/Carthage/blob/master/Documentation/VersionFile.md

isaac-weisberg:
Works perfectly!

PR #2429, opened 2018-04-22
[WIP] Build dependencies as paths

Aranoledur:
What do you think about this feature? Calling `carthage build /path/to/framework` will build this framework without `git clone`.

### Goal
It's a good known practice to split your Big project to simple local frameworks (NetworkClient, DBClient,...). But this frameworks become bigger and can be time-consuming when building after `Clean` or for publishing.

### Why not use git dependency type in Cartfile?
Git dependency type requires git repo, version and will copy sources to `Carthage/Checkouts`. It's great for outer frameworks, but not ok for local ones, because you already know the version, you don't need to copy sources, because they are already in repo as a `Carthage/Checkouts` folder, you don't want to create submodule, because, again, sources are already in the repo.

Also, this feature can be improved to use `--cache-build`, version can be taken from Project. This is not a complete feature, just an idea.

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -63,7 +63,12 @@ public struct BuildCommand: CommandProtocol {
 			.flatMap(.merge) { stdoutHandle, temporaryURL -> SignalProducer<(), CarthageError> in
 				let directoryURL = URL(fileURLWithPath: options.directoryPath, isDirectory: true)
 
-				let buildProgress = self.buildProjectInDirectoryURL(directoryURL, options: options)
+				let buildProgress: BuildSchemeProducer
+				if let buildAsPaths = self.buildDependenciesAsPaths(directoryURL, options: options) {
+					buildProgress = buildAsPaths
+				} else {
+					buildProgress = self.buildProjectInDirectoryURL(directoryURL, options: options)
+				}
 
 				let stderrHandle = options.isVerbose ? FileHandle.standardError : stdoutHandle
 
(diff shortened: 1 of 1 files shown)


blender:
Are you aware you can point Carthage to the local file system and specify a branch name?

Aranoledur:
Yes, I've described it at "Why not use git dependency type in Cartfile" section. Imagine you have this repo structure

```
RepoRoot
- Project
- Project.xcodeproj
- Cartfile
- LocalFrameworkA
- LocalFrameworkB
```

Why I have to make LocalFrameworkA and LocalFrameworkB git repo? I only want to build them

blender:
Why would you want carthage to handle these code locations? I am saying code locations because they don't appear to be dependencies to me.

- If they are not git repositories how do you get them in via carthage?
- What is the version of these dependecies?
- what trace would this leave in the Cartfile.resolved?

This to me seems more related to your workflow  rather than a a carthage concern

Aranoledur:
Because I want to save some time using `--cache-builds` for developers and CI. These frameworks change very rarely, so there is no need to recompile them over and over again.

- By path
- Can be taken from .xcodeproj
- path "DependencyName" "version from xcodeproj"

blender:
Can you explain how this is supposed to work with `--cache-builds` ? 

As per your answers:
- If they are not git repositories how do you get them in via carthage? By path

Can you explain how carthage is supposed to grab the dependency looking at the Cartifle

Aranoledur:
Carthage can create `.version` file after building and then compare version from xcodeproj to this one. And if it's the same version, skip building.

I think it can be done by keyword `path` in Cartfile. E.g. 
```
github "onevcat/Kingfisher" ~> 4.0
path "some/relative/path/to/Dependency"
```

Aranoledur:
It's not just about time shaving, don't want to create git repo where it's not needed and duplicate sources. Well, ok maybe it's little superfluous for this specific case. Will try to look at Rome, maybe it will help to deal with this.
About `Info.plist`. Thanks for pointing this out never thought about leaving it unchanged

blender:
> Thanks for pointing this out never thought about leaving it unchanged

Yeah well, don't do that! 😃

Aranoledur:
😃 
Will close this PR then, will try Rome or `--no-skip-current`

Issue #2430, opened 2018-04-23
Bootstrapping binary frameworks with same .json filename overwrites single .version file


juraj-blahunka:
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.29.0
* `xcodebuild -version`: Xcode 9.3 Build version 9E145
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? yes
* Are you using `--new-resolver`? no

**Cartfile**
```
binary "https://raw.githubusercontent.com/kayak/Specs/fabric-1.7.6/Carthage/Fabric.json" "1.7.6"
binary "https://raw.githubusercontent.com/kayak/Specs/fabric-1.7.6/Carthage/tvOS/Fabric.json" "1.7.6"
```

**Carthage Output**
Command: `carthage bootstrap --cache-builds`
```
*** No cache found for Fabric, building with all downstream dependencies
*** No cache found for Fabric, building with all downstream dependencies
*** Downloading binary-only framework Fabric at "https://raw.githubusercontent.com/kayak/Specs/fabric-1.7.6/Carthage/Fabric.json"
*** xcodebuild output can be found in /var/folders/48/sy74dwmd3_n9ym1dc0fh1fhhj37rwj/T/carthage-xcodebuild.fNTiBG.log
*** Downloading binary-only framework Fabric at "https://raw.githubusercontent.com/kayak/Specs/fabric-1.7.6/Carthage/tvOS/Fabric.json"
```

**Actual outcome**
Carthage overwrites the existing .version file and only shows a framework entry for one platform.

```
{
  "Mac" : [

  ],
  "watchOS" : [

  ],
  "tvOS" : [
    {
      "name" : "Fabric",
      "hash" : "e2ce2f8be8375a7089590fd9050a2144a06e85c0ecdfa0b6bb6364acffc94eb1"
    }
  ],
  "commitish" : "1.7.6",
  "iOS" : [

  ]
}
```

**Expected outcome**
Carthage realizes, that multiple `binary` entries in `Cartfile.resolved` represent the same framework and instead of overwriting the same .version file, it appends platform entries:
```
{
  "Mac" : [

  ],
  "watchOS" : [

  ],
  "tvOS" : [
    {
      "name" : "Fabric",
[...]

Issue #2036, opened 2017-06-29
Use Fabric, Crashlytics and Answers with Carthage


Zyphrax:
Below are binary definitions for Fabric, Crashlytics and Answers.
Each JSON points to the latest zip file directly provided by Fabric.

Just add the following to your Cartfile.
```ruby
binary "https://raw.githubusercontent.com/Building42/Specs/master/Carthage/Fabric.json"
binary "https://raw.githubusercontent.com/Building42/Specs/master/Carthage/Crashlytics.json"
binary "https://raw.githubusercontent.com/Building42/Specs/master/Carthage/Answers.json"
```

I've mailed the Fabric team asking them if they're willing to maintain the JSONs.

Zyphrax:
Most likely you won't need the Answers framework because it seems to be integrated in the Crashlytics framework. But it is available if you want to use it without Crashlytics.

Don't forget to add the items to your Plist. For more info see:
https://fabric.io/kits/ios/crashlytics/install
https://fabric.io/kits/ios/answers/install

**Do not** add the frameworks to the Carthage copy script!
Otherwise archiving your app will fail with a Mach-O error about entitlements.

Can I maybe create a wiki for this? A lot of people are still using CocoaPods because Fabric doesn't officially support Carthage. Otherwise they can find this post for more information.

Prince2k3:
@Zyphrax I think this should be added some where so that is it known there is some support. Right now this is only found through Google search at best

eyeoashephard:
FYI @Zyphrax - I had to install the Answers framework in order to successfully reference it within my code, but only had to import Crashlytics as a module.

Other than that, works well! Thanks.

dmiluski:
@Zyphrax Is it possible to just point to a GitHub releases to drive this rather than a separately hosted son file?

Zyphrax:
@dmiluski Carthage needs a JSON file to describe the versions of the framework. As far as I know it isn't possible to directly use a zip file or GitHub releases page reference.

ncatta:
Hello !  

Is it possible to provide JSON files for MacOS release ?

Thanks !

mdiep:
Yes. Please refer to the Carthage documentation about the `binary` dependency specifier.

Zyphrax:
@ncatta The MacOS specs are available at:
https://github.com/Building42/Specs/tree/master/Carthage/macOS

RafaelPlantard:
[...]

PR #2432, opened 2018-04-24
Lenient framework platform check. 

blender:
If platform can't be determined, report error and don't fail entire chain.

See https://github.com/Carthage/Carthage/issues/2430#issuecomment-383837595 

Note: I would like to rebase in one commit before anyone merges

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -586,16 +586,16 @@ public final class Project { // swiftlint:disable:this type_body_length
 					.flatMap(.concat) { frameworkURLsResults -> SignalProducer<(), CarthageError> in
 
 						let (frameworkURLs, errors) = frameworkURLsResults.reduce(([URL](), [CarthageError]()), { acc, frameworkURLResult in
-							var mutalbeAcc = acc
+							var mutableAcc = acc
 
 							switch frameworkURLResult {
 							case .success(let url):
-								mutalbeAcc.0.append(url)
+								mutableAcc.0.append(url)
 							case .failure(let error):
-								mutalbeAcc.1.append(error)
+								mutableAcc.1.append(error)
 							}
 
-							return mutalbeAcc
+							return mutableAcc
 						})
 
 						for error in errors {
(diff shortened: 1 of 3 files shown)


mdiep:
I think failing the whole build is actually the right thing to do here. 😕 It ensures that (1) framework authors fix up issues and (2) framework consumers don't miss it if the framework isn't valid.

blender:
Good points @mdiep . 

However I think that then [this check](https://github.com/Carthage/Carthage/blob/b18a85dfefe4e1a640453a6721807e8b9d751b4a/Source/CarthageKit/Project.swift#L1281) is a little too loose. That function is only used when deciding what `.frameworks` to copy.

I could rework it to check if there is binary, and if the binary is a dynamic (and soon static) framework. Filter out false positives otherwise.

The fact that Carthage doesn't enforce a specific shape (i.e. directory structure) for the downloaded archive is very powerful because any archive goes. The motivation for the current change is that directories that are named `.framework` currently are picked up here and it's a bit of shame that the [archive Fabric provides (see comment(https://github.com/Carthage/Carthage/issues/2430#issuecomment-383837595))] fails because it contains a bogus directory called`Fabric.framework` containing a shell script.

I though that since Google, despite it's size,  [doesn't  want to play ball](https://github.com/firebase/firebase-ios-sdk/issues/9#issuecomment-381746071) I'd skip invalid archives instead of filtering them out silently as I suggested above. Or maybe, I should filter them out and just log that they are being ignored. 

I think that, whatever implementation is chosen in the end if any at all, this is a change for good for the following reasons:

- Naming a bundle `.framework` doesn't make it a framework (mistake in Carthage's logic here) and bogus bundles should be ignored.
- Increase the end user's security by making it possible to use and archive provided by Goolge instead of an archive [hosted by 3rd parties](https://github.com/Carthage/Carthage/issues/2036#issue-239346388)

mdiep:
I think I like this alternative approach. It seems simpler.

But we should make sure that we reuse/share any logic there if possible. There's an existing `PackageType` that might be of use, e.g.

blender:
@mdiep PackageType is in Carthage, while this logic is in CarthageKit. I can move it ofc.

blender:
Closing. Superseded by #2436

PR #2434, opened 2018-04-27
Try to improve `skippedInstallingBinaries` error message

ikesyo:
This would clarify the intention of the error (actually an warning) and help users to understand what will happen after the error.

How does this look like?


mdiep:
Nice. 🤘

PR #2435, opened 2018-04-28
Remove BITCODE_GENERATION_MODE handling which should not be needed an…

ikesyo:
…ymore (because of `xcodebuild archive`)

Refs: 

- https://github.com/Carthage/Carthage/issues/535
- https://github.com/Carthage/Carthage/pull/839
- https://github.com/Carthage/Carthage/pull/2158


blender:
So, if I understand correctly this is not needed anymore because for device t"archive" action is used instead of build and for simulator it does not impact anything?

ikesyo:
@blender Exactly.

mdiep:
I love to delete code. 💖

PR #2436, opened 2018-04-30
Filter out improper directories passing as frameworks

blender:
Same final purpose of #2432 but different implementation.

Note that in this version the user is not informed about the filtering. 
It would be possible to pass in `self._projectEventsObserver` for logging.


mdiep:
This approach looks good to me. 👍 

Do you want to try writing some tests for that function? If not, you can go ahead and merge.

blender:
@mdiep
- Made `frameworksInDirectory` and similar functions internal
- Added test 
- Added fixture for tests
- Had to modify copy-fixtures scripts (somehow there is `__MACOSX` directory, not part of my fixture, that needs to be detelet before re-running the tests)

PR #2439, opened 2018-05-02
Discard invalid build metadata and pre-release semantic version

marcoconti83:
Any version with `-` but without pre-release, or with `+` but without build metadata, is discarded as invalid


blender:
@jdhealy  I'm just going to to ahead and merge this one into the working branch. Everything seems fine to me.

Issue #2444, opened 2018-05-08
The 'copy-frameworks' command fails when archiving an macOS application on an APFS volume


cmittendorf:
We use Carthage 0.29.0 installed via homebrew.

I recently updated our macOS project and added a `carthage copy-frameworks` build step for packaging our application. But Jenkins failed building the project with the following error message:

    Exit status: 65

In the details it showed:

    A shell task (/usr/bin/xcrun codesign --force --sign []…]/Result.framework) failed with exit code 1:
    […]/Contents/Frameworks/Result.framework: 
    bundle format is ambiguous (could be app or framework)

The optional code signing step was failing. The underlying problem was, that the frameworks have not been copied into the target directory and therefore `codesign` found nothing it could sign.

blender:
Closing since #2445 got merged

PR #2445, opened 2018-05-08
To destination is a symlink when archiving on macOS

cmittendorf:
This PR resolves issue #2444.

Our archive build steps for a macOS application were failing when doing the `carthage copy-frameworks` step due to a problem with code signing. The error message shown in the log was:

    […]/Contents/Frameworks/Result.framework: bundle format is ambiguous (could be app or framework)

However, the real problem was that the framework has not been copied at all and therefore `codesign` had nothing to sign at the destination path.

The `copyItem` method in `FrameworkExtension.swift` contains a workaround for copying files on an APFS disk and it falls back to the `copyFile` function from the standard c lib instead of using `NSFileManager`.

When executing an archive step on macOS, the `to` target resolves to a symlink:
    …/DerivedData/…/BuildProductsPath/Release/foobar.app@ ->
    …/DerivedData/…/InstallationBuildProductsLocation/Applications/foobar.app

The `to` destination is a symlink to the app bundle within an `…/Application` folder and thus the frameworks were not copied as the `copyfile` function was called with the `COPYFILE_NOFOLLOW` option. 

The solution is to use `COPYFILE_NOFOLLOW_SRC` instead. Now the frameworks are copied to the  destination directory.

### DIFF
--- a/Source/CarthageKit/FrameworkExtensions.swift
+++ b/Source/CarthageKit/FrameworkExtensions.swift
@@ -324,7 +324,9 @@ extension FileManager: ReactiveExtensionsProvider {
 		try from.path.withCString { fromCStr in
 			try to.path.withCString { toCStr in
 				let state = copyfile_state_alloc()
-				let status = copyfile(fromCStr, toCStr, state, UInt32(COPYFILE_ALL | COPYFILE_RECURSIVE | COPYFILE_NOFOLLOW))
+				// Can't use COPYFILE_NOFOLLOW. Restriction relaxed to COPYFILE_NOFOLLOW_SRC
+				// http://openradar.appspot.com/32984063
+				let status = copyfile(fromCStr, toCStr, state, UInt32(COPYFILE_ALL | COPYFILE_RECURSIVE | COPYFILE_NOFOLLOW_SRC))
 				copyfile_state_free(state)
 				if status < 0 {
 					throw NSError(domain: NSPOSIXErrorDomain, code: Int(errno), userInfo: nil)


blender:
@cmittendorf Thank you very much for your contribution! You did a  very nice job both in researching the issue and in fixing it yourself. 🎖 

This sounds like it only happens on APFS so, do you think any tests should be added to the suit?

cmittendorf:
I don‘t know the exact problem behind rdar://32984063 and why the workaround has been added. But copying to a symlinked folder seems to be necessary on the macOS platform and a test that could reproduce rdar://32984063 and validate the workaround might also be helpful once Apple declares the radar being resolved.

blender:
Reference: http://openradar.appspot.com/32984063

@cmittendorf so if I understand correctly, the way to test this would be to read the binary and make sure it's not just all zero bytes.

I don't think it's necessary for us to test this but please add a comment like:

```
// Can't use COPYFILE_NOFOLLOW. Restriction laxed to  COPYFILE_NOFOLLOW_SRC
// http://openradar.appspot.com/32984063
```

cmittendorf:
@blender Thanks for the info. I updated my commit. However, I tried to reproduce the mentioned bug and I wasn't able to reproduce it (using 10.13.4, not 10.12.5 as in the bug description).

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -324,7 +324,9 @@ extension FileManager: ReactiveExtensionsProvider {
 		try from.path.withCString { fromCStr in
 			try to.path.withCString { toCStr in
 				let state = copyfile_state_alloc()
-				let status = copyfile(fromCStr, toCStr, state, UInt32(COPYFILE_ALL | COPYFILE_RECURSIVE | COPYFILE_NOFOLLOW))
+                // Can't use COPYFILE_NOFOLLOW. Restriction relaxed to COPYFILE_NOFOLLOW_SRC
+                // http://openradar.appspot.com/32984063
```
mdiep: Can you fix the indentation here? 🙈

cmittendorf: Done, changed spaces to tabs.

PR #2448, opened 2018-05-09
Allow binary assets with .zip extension.

bwhiteley:
Sometimes framework.zip files (created with `carthage archive`) are uploaded with the incorrect MIME type. For example, if the archives are uploaded with the `github-release` utility, they end up with a MIME type of `application/octect-stream` instead of `application/zip`. 

This patch supports archives with a `.zip` extension but which may not have the expected MIME type.

### DIFF
--- a/Source/CarthageKit/Constants.swift
+++ b/Source/CarthageKit/Constants.swift
@@ -96,12 +96,8 @@ public struct Constants {
 		/// tried as a binary framework.
 		public static let binaryAssetPattern = ".framework"
 
-		/// The extension of a zip archive. Sometimes zip archives added to a release
-		/// have an incorrect MIME type, so check for a mime type or an extension.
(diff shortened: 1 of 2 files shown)


blender:
Also "gz", "tgz", "bz2" are [valid extension](https://github.com/Carthage/Carthage/blob/f029456d4dddbd311e574f1f74ae34f08ef07ac5/Source/CarthageKit/Archive.swift#L23) for `binary` releases.

I think what should really be challenged here is why `application/zip` is required in the response from github.

blender:
I would advocate to remove the MIME type check. I think it's an artificial restriction as well as requiring the archive to contain the word "framework". 

I'll take another look and get back to you.

blender:
@bwhiteley I would definitely think that the way forward is to add "application/octet-stream" to the allowed content-types or remove the check completely

PR #2450, opened 2018-05-12
Generate a version file for projects with no shared schemes

jasonboyle:
We have a number of internal projects, all of which depend on a project with no shared schemes.  We recently updated this project, and as a result, the --cache-builds flag always identified its cache as invalid because Carthage currently does not write out an updated version file for projects with no schemes.  I believe this PR addresses this issue.


mdiep:
Awesome! Thanks @jasonboyle.

Would you mind resolving the conflict? It should be easy. Another fixture was added in another PR, so the list just needs to be updated.

mdiep:
Thanks!

PR #2453, opened 2018-05-16
Stricter validation of pre-release and build metadata according to SemVer

marcoconti83:
This commit implements validation of pre-release and build metadata according to SemVer specifications (https://semver.org/):

> [pre-release and build metadata] identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]

Additionally, for pre-release

> [...] MAY be denoted by appending a hyphen and a series of dot separated identifiers.  [...]  Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes.

### DIFF
--- a/Source/CarthageKit/Version.swift
+++ b/Source/CarthageKit/Version.swift
@@ -55,10 +55,7 @@ public struct SemanticVersion: VersionType {
 		self.preRelease = preRelease
 		self.buildMetadata = buildMetadata
 	}
-
-	/// The set of all characters present in valid semantic versions.
-	fileprivate static let versionCharacterSet = CharacterSet(charactersIn: "0123456789.")
-
+	
 	/// Attempts to parse a semantic version from a PinnedVersion.
 	public static func from(_ pinnedVersion: PinnedVersion) -> Result<SemanticVersion, ScannableError> {
 		let scanner = Scanner(string: pinnedVersion.commitish)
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Version.swift
```diff
@@ -83,6 +80,23 @@ public struct SemanticVersion: VersionType {
 		                       patch: self.patch,
 		                       preRelease: self.preRelease)
 	}
+	
+	/// Set of valid characters for SemVer major.minor.patch section
+	fileprivate static let versionCharacterSet = CharacterSet(charactersIn: ".")
+		.union(CharacterSet.decimalDigits)
```
jdhealy: Our fault for not documenting this to begin with, but `CharacterSet.decimalDigits` was avoided because of the weirdness that `CharacterSet.decimalDigits.contains("\u{FF13}") == true` for full-width scalars like [`FULLWIDTH DIGIT THREE`](https://www.unicode.org/charts/nameslist/n_FF00.html#FF13)

marcoconti83: No problem, I can change it back to numbers and add a comment to avoid this in future.

marcoconti83: Addressed here: https://github.com/Carthage/Carthage/pull/2458

Issue #2000, opened 2017-06-18
Add option to save inputs to resolver


mdiep:
Resolving dependencies—choosing compatible versions of all the dependencies—is one of the most central and hardest things that Carthage does. These problems are also the hardest to debug when users report issues.

Our tests for the resolver now use a very simple database of tags/versions:

```swift
private struct DB {
	// Each available dependency has a dictionary of versions to its dependencies.
	// So if A has version 1.0 that has `B ~> 2.0` in its 1.0 Cartfile, then you'd see this:
	// [ A: [ 1.0: [B: .compatibleWith(2.0)] ] ]
	var versions: [Dependency: [PinnedVersion: [Dependency: VersionSpecifier]]]

        // Dependencies can also have random git references (non-version tags or just SHAs)
	// that point to SHAs.
	var references: [Dependency: [String: PinnedVersion]] = [:]
}
```

We should add a flag that creates and saves a database like this for the current project's Cartfile. We could change the dependency names to be simple things like `A`, `B`, etc., which would anonymize the data a bit.

Then, when users have resolver issues, we could ask them to run this, which would let us easily create and reduce a test.

Dschee:
This is issue # 2000 🎉

PR #2455, opened 2018-05-17
Completely new resolver implementation which is faster and eliminates a lot of the bugs both in the original and new resolver

werner77:
I created a new resolver implementation based on an optimized backtrack algorithm which doesn't suffer from the performance problems and bugs in the original and new resolver implementations.

I ensured the resolver successfully meets all the tests from the ResolverSpec test case.

In this pull request a flag has been added called '--fast-resolver' to enable this new resolver. However, if the review has been done I would like to eliminate both older buggy resolvers and substitute this one instead.

At the company I currently work for we make use of Carthage a lot and we suffer from the buggy implementation of both older resolvers, that's why I came up with this completely rewritten implementation inspired by other resolvers in different languages and we have been using this new version here already successfully.


blender:
Hi, thanks for you contribution.

The more documentation you provide on the changes that you have made the quicker this will go.

Can you summarize and link any appropriate piece of information?

werner77:
Ok, here is a summary of my changes with some details about the algorithm: 

- created a new class 'BackTrackingResolver' which implements the ResolverProtocol based on an optimized backtracking algorithm. See the method 'backtrack' inside this class.
- this algorithm basically recursively tries to eliminate paths from the dependency tree with a fail fast approach until a complete set has been found without conflicts. The algorithm is inspired by looking at resolvers for other languages and wikipedia.
- the BackTrackingResolver class uses a DependencySet which uses a ConcreteVersionSet, which is based on SortableSet of ConcreteVersions (which wraps around PinnedVersion/SemanticVersion)
- the sorted set is a TreeSet implementation, with O(log(N)) for lookup, delete and add operations. 
- the ConcreteVersionSet defines a method retainVersions(compatibleWith versionSpecifier: VersionSpecifier) which also has O(log(N)) because of the underlying sorted set
- because of the decreased time complexity for elimination of version sets (O(N) -> O(log(N)) operations on large dependency sets go much faster
- also the path for resolving is much more optimized which can be observed if a carthage update is run for a specified set of dependencies (e.g. carthage update on 1 dependency without sub-dependency will only clone that dependency whereas the original implementation would clone all dependencies)
- all operations done by the versionsForDependency, resolvedGitReference, dependenciesForDependency are cached by the class DependencyRetriever
- test cases have been slightly extended (ResolverSpec) to check whether an error is returned for some operations (was not checked before in those tests)
- a test case has been added for cyclic dependencies which has been marked pending. The original resolver goes into an infinite loop for these kind of dependencies. Both --new-resolver and the BackTrackingResolver give a result. Ideally the resolver would throw an error for these kinds of dependencies (may be fixed in the future still, but the BacktrackingResolver is at least not worse than the older ones in this respect)
- a flag --fast-resolver has been added to trigger the BackTrackingResolver. I would Ideally want to remove this flag and --new-resolver and just substitute the BackTrackingResolver for the older ones.

Let me know if you need more info.

mdiep:
I don't see that you've added any tests that your new resolver passes but the old resolvers fail. An essential part of upgrading to a better resolver is to add tests. Are you able to add any that show the benefit of your new resolver?

werner77:
I will try to add some tests

Source/CarthageKit/BackTrackingResolver.swift
```diff
@@ -0,0 +1,512 @@
...
+					}
+				} else {
+					// All done
+					result = (.rejected, dependencySet)
+				}
+			}
+		}
+
+		// By definition result is not nil at this point, so it's ok to force unwrap
+		return result!
```
blender: Question: Would asserting in debug and throwing an error in release also do here?

Tests/CarthageKitTests/ConcreteVersionTest.swift
```diff
@@ -0,0 +1,245 @@
+//
```
blender: Remove the lines before the import please

bwhiteley:
We use Carthage heavily and it had become nearly unusable with the existing resolvers. An `update` will run forever and never exit. We had often resorted to manual edits of `Cartfile.resolved` to work around this. The project uses many internal frameworks so we could not share public test cases. 

I've been testing `--fast-resolver` today and it has been working fine with projects where the other resolvers fail.

Source/CarthageKit/BackTrackingResolver.swift
```diff
@@ -139,9 +139,12 @@ public final class BackTrackingResolver: ResolverProtocol {
 				}
 			}
 		}
-
-		// By definition result is not nil at this point, so it's ok to force unwrap
-		return result!
+		
+		// By definition result is not nil at this point (while loop only breaks when result is not nil)
+		guard let finalResult = result else {
+			preconditionFailure("Expected result to not be nil")
```
blender: What I mean here is `preconditionFailure` will crash carthage and (I think) won't give any good reason to the user.

I would rather prefer to throw a CarhageError

werner77: This is like an assertion, as you can see from the program flow this should never happen

werner77:
I'm working on a full-fledged test case right now, one from a real-life scenario where the other resolvers fail.

werner77:
See LocalRepositoryTest, if you change 'pending' to 'it' you can run that test to record a local mock repository from a live Carthage project (just specify the correct projectURL).

To add a test for this mocked repository see the implementation of ResolverSpec:

it("should correctly resolve complex conflicting dependencies")

For some reason, like I said in the commit comment, the mocked repo does not give a problem for the older resolvers, while the live repo does. I'm not sure what the exact issue is there. 

I created a screen recording here with the live behaviour of the three resolvers for my project:

https://youtu.be/5q7OEFPzozQ

werner77:
Timed the --new-resolver crunch on the same project. It is "slightly" slower:

 time /Users/werneraltewischer/Library/Developer/Xcode/DerivedData/Carthage-hkpgwaktgfvnhadtarxuprasxkrm/Build/Products/Debug/carthage update --no-build --new-resolver
*** Fetching INGWebtrekkAnalytics
*** Fetching INGCordova
*** Fetching INGBiometrics
*** Fetching ocmock
*** Fetching INGFoundation
*** Fetching SwiftLint
*** Fetching ScanToolsNL
*** Fetching INGHybridPackage
*** Fetching HockeySDK-iOS
*** Fetching INGTransactions
*** Fetching BuildScripts
*** Fetching CommunicationNL
*** Fetching INGMoneyTransfers
*** Fetching INGFormFields
*** Fetching NinaNL
*** Fetching INGCore
*** Fetching INGCommunication
*** Fetching INGProducts
*** Fetching OHHTTPStubs
*** Fetching webtrekk-ios-sdk
*** Fetching INGStyleKitV2
Could not pick a version for git "ssh://git@gitlab.ing.net:2222/MobileBanking-iOS/INGCore.git", due to mutually incompatible requirements:
	~> 3.0.0
	~> 4.0.0 (git "ssh://git@gitlab.ing.net:2222/NL-Bankieren-iOS/ScanToolsNL.git")

real	499m29.654s
user	498m48.440s
sys	0m15.927s

werner77:
Please let me know what I can do more to get this merged.

nickmshelley:
Two problems with this change that I've found. To reproduce, create a test project with a Cartfile that only includes `github "facebook/facebook-ios-sdk"`

1. The normal resolver gets the latest release version, version `4.33.0`. The fast resolver gets the tagged version of the same commit (but not the release version) `sdk-version-4.33.0`. Not sure it's a huge problem, but it seems wrong and is different than the previous behavior.

2. The current Carthage release builds four schemes and succeeds. This one tries to build the `FBSDKIntegrationTests` scheme and fails. Not sure why this PR would change how things build, but it seems to.

werner77:
@nickmshelley I will look into that issue and also create a unit test for it.

werner77:
I made a test case which indeed shows the failure for the facebook/facebook-ios-sdk version. The reason lies however in the fact that the SemanticVersion parser ignores leading characters before the semantic part. Since the versions are stored in a set, they are identical regardless of their actual tag name, so 4.33.0 would give the same version as sdk-version-4.33.0. 

So I am not sure this is actually a bug. Maybe some of the others can shed some light on this. 

There was a bug in the test case for the complex conflicting dependencies test case where the resolver type was not correcly passed. Now the test case shows that it fails for the original Resolver (the test is not enabled for that resolver to pass the build). It still passes for the NewResolver, but it does it about 25 times slower. I'm still not sure why the NewResolver does pass with the simulated repository while with a live repository it takes 8.5 hours to complete.

blender:
@marcoconti83 I summon you. This is using the current implementation of Semantic Version.

werner77:
I tried #2412 but that does not fix the problem (if it really is a problem). That new parser still ignores leading characters as I can judge from the code.

let scanner = Scanner(string: pinnedVersion.commitish)

```
		// Skip leading characters, like "v" or "version-" or anything like
		// that.
		scanner.scanUpToCharacters(from: versionCharacterSet, into: nil)
```

nickmshelley:
I just checked and the `sdk-version-4.33.0` problem seems to be gone with the latest from this PR. Does anyone want to verify that I'm not just doing something stupid?

If it's the case that that is fixed, is there anything else holding up this PR?

nickmshelley:
We have an actual Cartfile that fails on all three resolvers (including this one). Unfortunately, it seems like our internal dependencies are the cause, so I can't just paste the Cartfile and have it be useful. I'll try to provide a minimal repro case (or preferably, a fix), but it will probably take some time.

blender:
@nickmshelley did you see this comment? https://github.com/Carthage/Carthage/pull/2455#issuecomment-390769944

werner77:
You can create a test case from your repo by running the diagnostic resolver with your Cartfile. Please run LocalRepositoryTest with your Cartfile to do that. This will record all dependencies and it’s versions in JSON format so a test case may be created without requiring live access to the private git repos

hollanderbart:
What's the status with this merge request? It seems to have a high potential..

nickmshelley:
It seems like the problem with the Cartfile that never finishes is that backtracking isn't aggressive enough. For example, let's say you have five direct dependencies, A through E. A and E depend on different versions of F. When A is processed, it adds version 2 of F and considers F resolved. When E is processed, it requires version 1 of F, so that tree is rejected. What should happen is that we backtrack all the way back to A (or right after A where F was considered) and pick a new version of F to try. But what actually happens is that it will backtrack to D, pick the next version of D, then try all versions of E to see if a compatible version of F is available. Once all versions of D are exhausted, it will backtrack to C, pick the next version, then try _all versions_ of D and E again. And so on all the way back up the chain.

In our case we have a lot of dependencies, and an early one points to a Result library with no version specifier, so it picks 4.0. About 15 dependencies later, we have a library that specifies a 3.x version of Result. That starts the whole explosion of recursion I described above.

However, when I hard-coded a version when trying to reproduce the problem with some test repos, it finished very quickly. So order plays a big role.

Which brings me to my next concern: the order in which the fast resolver processes dependencies is undefined. To get the next dependency, it just calls `first` on a set of unresolved dependencies. This order happens to be consistent across runs because hash values are consistent, but that will change in Xcode 10/Swift 4.2: https://github.com/apple/swift-evolution/blob/master/proposals/0206-hashable-enhancements.md#proposed-solution.

I think we should also create a defined order to process dependencies. I'm not sure if the actual order matters as long as it's consistent, so I was thinking either use the order in which they appear in the Cartfile, or just something simple like alphabetical order.

werner77:
@nickmshelley Thanks for debugging. Could you please record the dependencies as described above with the LocalRepositorySpec. It will result in a directory with JSON files describing all the dependencies which can be taken as input for a test case so I can debug your scenario.

Regarding the order: the speed of the algorithm is definitely dependent on the order in which the dependencies are resolved, but regardless of the order, always the same result should be given. Right now the algorithm gives branches of the dependency tree where previous conflicts were encountered more precedence as they are more likely to fail for subsequent evaluations as well. The whole idea of the backtrack algorithm is to discard as many branches as possible in an early stage so no unnecessary work has to be done.  It's the same algorithm which is used to solve sudoku puzzles for example.

nickmshelley:
@werner77 The output is unaffected by the order only in the case where there is a single solution (as with "well-formed" sudoku puzzles). But with dependency graphs (and many poorly-crafted sudoku puzzles), there are usually many solutions, so the order does affect the output.

The Swift package manager seems to use the manifest file order: https://github.com/apple/swift-package-manager/blob/master/Sources/PackageGraph/DependencyResolver.swift#L755

```
This solver currently implements a basic depth-first greedy backtracking algorithm, and honoring the order of dependencies as specified in the manifest file.
...
There is currently no external control offered by the solver over _which_ solution satisfying the properties above is selected, if more than one are possible. In practice, the algorithm is designed such that it will effectively prefer (i.e., optimize for the newest version of) dependencies which are earliest in the depth-first, pre-order, traversal.
```

nickmshelley:
Also FYI, I have a local repository test that is hanging with the current implementation and finishes fine when dependencies are ordered alphabetically. We're going to see if we can anonymize it and still have it fail.

werner77:
Anomization is easily done by just replacing the names of the repositories and file/directory names with different ones in the recordings. I did the same for the test cases I created. 

Verstuurd vanaf mijn iPhone

nickmshelley:
It turns out that anonymizing breaks the local tests because it changes the order in which dependencies are processed. So the resolver will process versions of dependencies that weren't processed when the local repository was created, and then it will look for and fail to find a `master.json` or `default.json` on some branch of the tree that wasn't hit when the order was different.

I think this will be fixed by having a concrete ordering, which is probably a good thing in general anyway. I'll make it order by where they are specified in the Cartfile, which should make the local test work after anonymizing.

nickmshelley:
In fact, since our problem Cartfile is a problem for all three resolvers, I don't think I should have even started that discussion here. My opinion (based on our experience) is that this resolver has performed better than the others in most or all cases, and it would be great to merge as soon as possible. I'll log separate issues for the problems I've found.

mdiep:
This PR is too big (and with too many files) to review on GitHub.

A good first step would be to introduce any new tests in a separate PR. They can be disabled if the current resolvers don't pass them. Failing resolver tests are one of the most important contributions.

Second, it seems like you might have a script that runs against a repo to create test cases. That'd be an AMAZING separate contribution. I opened https://github.com/Carthage/Carthage/issues/2000 for this a while ago and it'd make a great option on the `carthage` tool to let users upload failing sets of dependencies. (I also think we'd want to combine these so we output 1 JSON file per environment.)

Those would dramatically reduce the number of changes here, making it easier to review.

blender:
@werner77 hello! How do you feel about addressing @mdiep 's concerns? This would make for a great addition to the next release

werner77:
I’m on vacation for yet one week after that I will try to pick it up.

werner77:
I created a separate branch for the resolver diagnostics: https://github.com/Carthage/Carthage/pull/2519

blender:
At a minimum: 
1. https://github.com/Carthage/Carthage/pull/2519
1. Code review once this is cleaned up from #2519

werner77:
How exactly is the update command with concrete dependencies supposed to work:

1. Update ONLY the dependencies specified, keep the transitive dependencies of those dependencies at their current version
2. Update the dependencies specified, including their transitive dependencies

Even while only a subset of dependencies is specified should still the entire Cartfile be validated?

Say the Cartfile contains two dependencies:

A ~> 1.0, resolved: A = 1.1
B ~> 1.0, resoved: B = 1.0

Now both A and B are updated in the Cartfile:

A ~> 2.0 
B ~> 2.0

and carthage update is run with A: carthage update A

A would update, but the cartfile.resolved does not satisfy B ~> 2.0 which would result in invalid dependencies according to the Cartfile.
It gets even more interesting if B is also a dependency of A and A(2.0) specifies dependency B(1.1).

How should the resolver cope with these issues?

werner77:
Please checkout our fork if you want a working resolver: https://github.com/nsoperations/Carthage. This Pull Request have been closed due to too long review time.

blender:
@werner77 Carthage is run by volunteers, everyone puts the time they have into it. 

Your effort is very much appreciated and I still look forward to see it merged 👍

werner77:
I understand, it’s just that the current rate of progress is too slow for the needs of me and the company I work for. I suggest to take a more active approach for pull requests: if it’s 90% ok: merge it and make the required modifications to fit your code style yourself. Otherwise it just goes back and forth.

david6p2:
I think this PR won't add any breaking changes and will add another option when the current resolver won't work. It should have been merged, or mark specific issues to know what needs to be resolved for this to be merged. I'm more than happy to help reviewing it, if this help to unblock us make the project move faster.

bwhiteley:
I would like to see this PR merged. In the meanwhile I've moved to @werner77 's fork because neither of the resolvers in Carthage today work with my projects.

PR #2459, opened 2018-05-19
Use `#` token for comment detection instead of space when parsing semver

marcoconti83:
Previously, the code was relying on there being a space between a version specified and a comment. This commit removes that assumption.

### DIFF
--- a/Source/CarthageKit/Cartfile.swift
+++ b/Source/CarthageKit/Cartfile.swift
@@ -7,6 +7,10 @@ public let carthageProjectCheckoutsPath = "Carthage/Checkouts"
 /// Represents a Cartfile, which is a specification of a project's dependencies
 /// and any other settings Carthage needs to build it.
 public struct Cartfile {
+	
+	/// Any text following this character is considered a comment
+	static let commentIndicator = "#"
+	
 	/// The dependencies listed in the Cartfile.
 	public var dependencies: [Dependency: VersionSpecifier]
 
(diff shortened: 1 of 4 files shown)


jdhealy:
Nice, especially like the expanded comments and the `Cartfile.commentIndicator` static variable.

Looks like `ignoreCartfileComments` is never non-defaulted with `false` in CarthageKit itself, only CarthageKitTests…

The [current tests are passing](https://travis-ci.org/Carthage/Carthage/builds/382190197), despite (I suspect) errantly ignoring post-`#` characters in determining SemVer conformance for tags — failure on our part, I guess I was preconceived that test already existed. I'm tied up for the next handful of hours, but I'll develop and commit a test for this scenario at some point after that.

marcoconti83:
I will add that test and see if it passes or not.

jdhealy:
> Looks like `ignoreCartfileComments` is never non-defaulted with `false` in CarthageKit itself, only CarthageKitTests…

Based on that, I suspected something like:

https://github.com/Carthage/Carthage/blob/92efa51c77afce3097737b937e9fdaf64eab213c/Tests/CarthageKitTests/VersionSpec.swift#L73

…would fail… (knew that already existed).

Figured out, during the process of writing a similar integration test, that `.isAtEnd` on `scanner`:

https://github.com/Carthage/Carthage/blob/92efa51c77afce3097737b937e9fdaf64eab213c/Source/CarthageKit/Version.swift#L59-L74

…correctly handles post-`#` characters in determining SemVer conformance for tags.

So, that suspicion was mistaken, that case behaves as expected.

The `else` case of the `ignoreCartfileComments` conditional never being hit by CarthageKit code is probably confusing enough to be worth addressing, though.

Props for continuing to work on this, can't thank you enough…

marcoconti83:
So should I remove the `ignoreCartfileComments` and assume it's always true?

marcoconti83:
Addressed the feedback on the ignoreCartfileComments parameter here: #2469

PR #2464, opened 2018-05-24
Add test for non-semver reference with number sign

marcoconti83:
See comment in https://github.com/Carthage/Carthage/pull/2412#pullrequestreview-122235889


marcoconti83:
@jdhealy should we still merge this? Even if it doesn't fix anything, it adds some missing tests.

mdiep:
This is great! Thanks for adding all these tests!

You've done a ton of great work on #2412. Thank you for being so willing to add all these tests and helping make sure all our bases are covered. ✨

PR #2465, opened 2018-05-26
Fix errant UTF-16 offset use by ranging with converted `NSRange`.

jdhealy:
Refactor `Scanner.scanStringWithPrefix` with `prefix` as `Character`.

Refactor `Scanner.remainingString` → `Scanner.remainingSubstring` which returns an optional.

See code comments and <https://twitter.com/UINT_MIN/status/968200980599746560>.

### DIFF
--- a/Source/CarthageKit/Version.swift
+++ b/Source/CarthageKit/Version.swift
@@ -241,35 +241,33 @@ extension Scanner {
 	/// string is encountered.
 	/// - returns: the scanned string without the prefix. If the string does not start with the prefix,
 	/// or the scanner is at the end, it returns `nil` without advancing the scanner.
-	fileprivate func scanStringWithPrefix(_ prefix: String, until: String) -> String? {
-		guard self.remainingString.hasPrefix(prefix) else {
+	fileprivate func scanStringWithPrefix(_ prefix: Character, until: String) -> String? {
+		guard !self.isAtEnd, self.remainingSubstring?.first == prefix else { return nil }
+
+		var buffer: NSString? = nil
+		self.scanUpTo(until, into: &buffer)
+		guard let stringWithPrefix = buffer as String?, stringWithPrefix.first == prefix else {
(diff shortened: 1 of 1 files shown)


jdhealy:
Targets #2412	/cc @marcoconti83

Source/CarthageKit/Version.swift
```diff
@@ -241,35 +241,33 @@ extension Scanner {
...
+	fileprivate func scanStringWithPrefix(_ prefix: Character, until: String) -> String? {
+		guard !self.isAtEnd, self.remainingSubstring?.first == prefix else { return nil }
+
+		var buffer: NSString? = nil
+		self.scanUpTo(until, into: &buffer)
+		guard let stringWithPrefix = buffer as String?, stringWithPrefix.first == prefix else {
 			return nil
 		}
+
+		return String(stringWithPrefix.dropFirst())
```
marcoconti83: This looks way cleaner than before! 👍

Source/CarthageKit/Version.swift
```diff
@@ -241,35 +241,33 @@ extension Scanner {
...
-	fileprivate var remainingString: String {
-		guard !self.isAtEnd else {
-			return ""
+
+	/// The string (as `Substring?`) that is left to scan.
+	///
+	/// Accessing this variable will not advance the scanner location.
+	///
+	/// - returns: `nil` in the unlikely event `self.scanLocation` splits an extended grapheme cluster.
+	fileprivate var remainingSubstring: Substring? {
```
marcoconti83: This change of API will clash with: https://github.com/Carthage/Carthage/pull/2469

To make it compatible again it would be just a one line change here: https://github.com/Carthage/Carthage/pull/2469/files#diff-75029f8fcb86c4555bca7116130d72b6R48 so it's not a big deal, we just need to decide which one to merge first.

PR #2469, opened 2018-05-31
Refactor detection of Cartfile comments

marcoconti83:
## Problem
Having the detection of Cartfile comments inside the version parser makes the parser unnecessarily tied to the Cartfile syntax. Additionally, since there is no case in which we would parse a version specifier and not ignore the comments, either we have to live with a parameter that is always set to `true` in the entire codebase, or remove the parameter and make strong assumptions on how the version parsing is used.

## Solution
This PR moves comment filtering to the Cartfile parsing code. Comments are stripped from each line before passing lines to the version parser.

The detection of comments is done with the following strategy:
- any `#` character that is not inside "quotes" is considered the beginning of the comment
- for the purpose of determining whether a character is inside quotes, the code relies on the fact that there should not be nested quotes (e.g. `quotes "like \"this\""`) inside a cartfile specification.

Even if not necessary at the moment, as the comment indicator is a single character, the implementation is designed to handle multi-character comment indicators. This has influenced the choice of using `String.components(separatedBy:)` over other solutions.

All the above is documented in code comments, and the PR includes tests to cover the comment stripping.

### DIFF
--- a/Source/CarthageKit/Cartfile.swift
+++ b/Source/CarthageKit/Cartfile.swift
@@ -31,24 +31,30 @@ public struct Cartfile {
 		var result: Result<(), CarthageError> = .success(())
 
 		string.enumerateLines { line, stop in
-			let scanner = Scanner(string: line)
+			let scannerWithComments = Scanner(string: line)
 
-			if scanner.scanString(Cartfile.commentIndicator, into: nil) {
+			if scannerWithComments.scanString(Cartfile.commentIndicator, into: nil) {
 				// Skip the rest of the line.
 				return
 			}
 
-			if scanner.isAtEnd {
+			if scannerWithComments.isAtEnd {
 				// The line was all whitespace.
 				return
 			}
+			
+			let scannerWithoutComments = Scanner(
+				string: scannerWithComments
+					.remainingString.strippingTrailingCartfileComment
+					.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)
+			)
 
-			switch Dependency.from(scanner).fanout(VersionSpecifier.from(scanner)) {
+			switch Dependency.from(scannerWithoutComments).fanout(VersionSpecifier.from(scannerWithoutComments)) {
 			case let .success((dependency, version)):
 				if case .binary = dependency, case .gitReference = version {
 					result = .failure(
 						CarthageError.parseError(
-							description: "binary dependencies cannot have a git reference for the version specifier in line: \(scanner.currentLine)"
+							description: "binary dependencies cannot have a git reference for the version specifier in line: \(scannerWithComments.currentLine)"
 						)
 					)
 					stop = true
@@ -67,12 +73,7 @@ public struct Cartfile {
 				return
 			}
 
-			if scanner.scanString(Cartfile.commentIndicator, into: nil) {
-				// Skip the rest of the line.
-				return
-			}
-
-			if !scanner.isAtEnd {
+			if !scannerWithoutComments.isAtEnd {
 				result = .failure(CarthageError.parseError(description: "unexpected trailing characters in line: \(line)"))
 				stop = true
 			}
@@ -163,5 +164,45 @@ extension ResolvedCartfile: CustomStringConvertible {
 			.sorted { $0.key.description < $1.key.description }
 			.map { "\($0.key) \($0.value)\n" }
 			.joined(separator: "")
+	}
+}
+
+
+extension String {
+	
+	/// Returns self without any potential trailing Cartfile comment. A Cartfile
+	/// comment starts with the first `commentIndicator` that is not embedded in any quote
+	var strippingTrailingCartfileComment: String {
+		
+		// since Cartfile syntax doesn't support nested quotes, such as `"version-\"alpha\""`
+		// I can just consider any odd-number occurence of a quote as a quote-start, and any
+		// even-numbered occurrence of a quote as quote-end.
+		// The comment indicator (e.g. `#`) is the start of a comment if it's not nested in quotes.
+		// The following code works also for comment indicators that are are more than one character long (e.g. "//")
+		
+		let quote = "\""
+		
+		// Splitting the string by quote will make odd-numbered chunks outside of quotes, and
+		// even-numbered chunks inside of quotes.
+		// `omittingEmptySubsequences` is needed to maintain this property even in case of empty quotes.
+		let quoteDelimitedChunks = self.split(separator: quote.first!,
+											  maxSplits: Int.max,
+											  omittingEmptySubsequences: false)
+		
+		
+		for (offset, chunk) in quoteDelimitedChunks.enumerated() {
+			let isInQuote = offset % 2 == 1 // even chunks are not in quotes, see comment above
+			if isInQuote {
+				continue // we don't consider comment indicators inside quotes
+			}
+			if let range = chunk.range(of: Cartfile.commentIndicator) {
+				// there is a comment, return everything before its position
+				let previousChunks = quoteDelimitedChunks[..<offset]
+				let chunkBeforeComment = chunk[..<range.lowerBound]
+				return (previousChunks + [chunkBeforeComment])
+					.joined(separator: quote) // readd the quotes that were removed in the initial split
+			}
+		}
+		return self
 	}
 }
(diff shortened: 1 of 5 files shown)


marcoconti83:
@blender: previously, the execution flow was:
- in `Cartfile`, parse the Cartfile by:
   - read Cartfile line by line, parse the first part of a line (host, repository and organization)
   - call to `SemanticVersion(from:)` to parse the version from the rest of the line
- in `SemanticVersion(from:)`, parse the **rest of the line** (not just the version), by:
   - strip Cartfile comments from the line
   - parse what's before the comment

With this PR, the execution flow is:
- in `Cartfile`, parse the Cartfile by:
   - read Cartfile line by line, stripping comments
   - parse the first part of a line (host, repository and organization)
   - call to `SemanticVersion(from:)` to parse the version
- in `SemanticVersion(from:)`, parse the version

I believe this makes `SemanticVersion` decoupled from the Cartfile syntax and separates concerns.

blender:
Looks good

marcoconti83:
PR updated because of internal API changes of https://github.com/Carthage/Carthage/pull/2465

Source/CarthageKit/Cartfile.swift
```diff
@@ -31,24 +31,30 @@ public struct Cartfile {
...
 
-			if scanner.scanString(Cartfile.commentIndicator, into: nil) {
+			if scannerWithComments.scanString(Cartfile.commentIndicator, into: nil) {
 				// Skip the rest of the line.
 				return
 			}
 
-			if scanner.isAtEnd {
+			guard !scannerWithComments.isAtEnd,
+				let remainingString = scannerWithComments.remainingSubstring else {
```
jdhealy: I think we have to stop parsing and surface an error if `Scanner.remainingSubstring` returns `nil`:

```swift
guard let remainingString = scannerWithComments.remainingSubstring.map(String.init) else {
	result = .failure(CarthageError.internalError(
		description: "Can NSScanner split an extended grapheme cluster? If it does, this will be the error…"
	))
 	stop = true
 	return
}
```

→ also removing the `String(…)` before `. strippingTrailingCartfileComment`

marcoconti83:
Fixed!

Source/CarthageKit/Cartfile.swift
```diff
@@ -165,3 +166,43 @@ extension ResolvedCartfile: CustomStringConvertible {
...
+		
+		
+		for (offset, chunk) in quoteDelimitedChunks.enumerated() {
+			let isInQuote = offset % 2 == 1 // even chunks are not in quotes, see comment above
+			if isInQuote {
+				continue // we don't consider comment indicators inside quotes
+			}
+			if let range = chunk.range(of: Cartfile.commentIndicator) {
+				// there is a comment, return everything before its position
+				let previousChunks = quoteDelimitedChunks[..<offset]
```
jdhealy: Even though this code is well-considered as it stands — in that we could assert `startIndex` for `quoteDelimitedChunks` always being `0` based on the singular way we initialize it — just to prevent any potential of breaking on someone else’s refactor, let’s:

```swift
				let advancedOffset = (..<offset).relative(to: quoteDelimitedChunks)
				let previousChunks = quoteDelimitedChunks[advancedOffset]
```

jdhealy:
Sorry, I'm slow in typing a few more (very minor) things 😄

Tests/CarthageKitTests/CartfileCommentsSpec.swift
```diff
@@ -0,0 +1,51 @@
...
+					expect($0.strippingTrailingCartfileComment) == $0
+				}
+			}
+			
+			it("should not alter strings with comment marker in quotes") {
+				[
+					"foo bar \"#baz\"",
+					"\"#quotes\" is the new \"quotes\"",
+					"\"#\""
+					]
```
jdhealy: Nit: indentation.

Source/CarthageKit/Version.swift
```diff
@@ -106,44 +106,12 @@ public struct SemanticVersion: VersionType {
 }
 
 extension SemanticVersion: Scannable {
-	/// Attempts to parse a semantic version from a human-readable string of the
-	/// form "a.b.c" from a string scanner. It assumes the string might contain a
-	/// comment coming from a Cartfile and will ignore the comment if present
```
jdhealy: Nit: grammar: `the comment, if present.`

Source/CarthageKit/Cartfile.swift
```diff
@@ -165,3 +173,43 @@ extension ResolvedCartfile: CustomStringConvertible {
...
+
+
+extension String {
+	
+	/// Returns self without any potential trailing Cartfile comment. A Cartfile
+	/// comment starts with the first `commentIndicator` that is not embedded in any quote
+	var strippingTrailingCartfileComment: String {
+		
+		// since Cartfile syntax doesn't support nested quotes, such as `"version-\"alpha\""`
+		// I can just consider any odd-number occurence of a quote as a quote-start, and any
```
jdhealy: Nit: grammar: end the `such as` phrase in the `since` line with a comma.

Nit: `I can just consider any` → `simply consider` to remove first-person.

Source/CarthageKit/Cartfile.swift
```diff
@@ -165,3 +173,43 @@ extension ResolvedCartfile: CustomStringConvertible {
...
+		// even-numbered occurrence of a quote as quote-end.
+		// The comment indicator (e.g. `#`) is the start of a comment if it's not nested in quotes.
+		// The following code works also for comment indicators that are are more than one character long (e.g. "//")
```
jdhealy: Really like the note on future abstraction here :clap:

But, _apologies for putting our buggy tooling on to you_, some off the maintainers run [SwiftLint](https://github.com/realm/SwiftLint/) which (with our configuration) misparses an actual comment out of the second `//` here and emits a linting warning:

```
Whitespace Comment Start Violation: A comment should always start with a whitespace. (whitespace_comment_start)
```

We could do `"//"` → `"/" + "/"` or maybe, just easier to remove the line all together, considering it’s not describing current behavior…

Sorry to bring you really into the weeds on that one 🙃

marcoconti83:
@jdhealy thanks for taking the time to review, let me know when you are done so I can fix all of them in one go

Source/CarthageKit/Version.swift
```diff
@@ -106,44 +106,12 @@ public struct SemanticVersion: VersionType {
...
-				)
-			)
-		} else {
-			numericVersionScanner = scanner
-		}
+	public static func from(_ scanner: Scanner) -> Result<SemanticVersion, ScannableError> {
 		
 		var versionBuffer: NSString?
-		guard numericVersionScanner.scanCharacters(from: versionCharacterSet, into: &versionBuffer),
+        guard scanner.scanCharacters(from: versionCharacterSet, into: &versionBuffer),
```
jdhealy: Nit: indentation.

jdhealy:
Last comment added. Massive thanks for the entirety of this :clap:

Source/CarthageKit/Version.swift
marcoconti83: This comment line is now gone anyway.

marcoconti83:
I think I got all of them

Source/CarthageKit/Version.swift
jdhealy: 🤦‍♂️  — misread my note about the comment [here](https://github.com/Carthage/Carthage/pull/2469/files#diff-75029f8fcb86c4555bca7116130d72b6R181)

so minor… no need to do anything about that…

Source/CarthageKit/Cartfile.swift
```diff
@@ -165,3 +173,45 @@ extension ResolvedCartfile: CustomStringConvertible {
...
+		// long (e.g. double slashes)
+		
+		let quote = "\""
+		
+		// Splitting the string by quote will make odd-numbered chunks outside of quotes, and
+		// even-numbered chunks inside of quotes.
+		// `omittingEmptySubsequences` is needed to maintain this property even in case of empty quotes.
+		let quoteDelimitedChunks = self.split(separator: quote.first!,
+											  maxSplits: Int.max,
+											  omittingEmptySubsequences: false)
```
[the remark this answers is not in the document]
jdhealy: Yeah, every web browser that I've view GitHub diffs with displays a tab as eight space-lengths. That, along with tabs being a configurable number of space-lengths in Xcode, has has been cause for me to steer clear of using tabs to align with characters (and spaces, for the most part too.)

Looks good! 👍

jdhealy:
Above and beyond, thanks!

Issue #2442, opened 2018-05-06
Static Linking question


dimazen:
Thats not an issue with the Carthage but rather a help request in order to better understand compiler behaviour.

I was repeating steps described in your [StaticFrameworks](https://github.com/Carthage/Carthage/blob/master/Documentation/StaticFrameworks.md) guide. My case is a shared framework, that links multiple static dependencies in itself and then gets linked by other targets. This allows us to share common dependencies without duplicating them. 
One of the steps is: 

> To ensure that this new merged framework is a true merge of all of its dependent static frameworks, you should include the -all_load flag in its OTHER_LDFLAGS build setting.

When I'm following this step, project fails to compile. Most of the warnings are:
```
ld: warning: Auto-Linking library not found for -lswiftCore
ld: warning: Auto-Linking library not found for -lswiftDispatch
ld: warning: Auto-Linking library not found for -lswiftObjectiveC
ld: warning: Auto-Linking library not found for -lswiftCoreFoundation
ld: warning: Auto-Linking library not found for -lswiftDarwin
ld: warning: Auto-Linking library not found for -lswiftFoundation

Undefined symbols for architecture x86_64:
  "Swift.String.init<A>(stringInterpolationSegment: A) -> Swift.String", referenced from:
...
and so on
```

In other words looks like that it fails to link Swift's standard libraries. Adding single internal class resolves the issue. Looks like that compiler adds auto-link of the standard libraries in this case. 

Other case is to exclude `-all_load` flag. In this case even with empty `Shared` target is succeed to compile the project. However, my shared dependency (Alamofire) gets linked with other target for some reason. 

You can check out project [here](https://github.com/dimazen/static-linking). To test how compiler works run
```
./carthage-build-static.sh build --platform iOS
```

Then open `SharedTestClass.swift` and include / exclude it from the `Shared` target to see how it affects on the compilation process.

Also, it might be useful to add hint about empty class to the StaticFrameworks guide. 

Thanks in advance!

blender:
Maybe you want to suggest improvements to #2497

marcusdesantis:
Thank's @dimazen, this works for me,

PR #2471, opened 2018-06-03
Add additional StaticFrameworks linker workarounds

dimazen:
This pull request adds extra workaround that addresses issue https://github.com/Carthage/Carthage/issues/2442.

Project where you can check described behaviour can be found [here](https://github.com/dimazen/static-linking).


blender:
@dimazen thanks for the PR and keeping the docs up to date! 🥇

PR #2473, opened 2018-06-05
Ignore errors when getting EXPANDED_CODE_SIGN_IDENTITY

### LINKED ISSUE #2472: Xcode 10 Build Fail: missing EXPANDED_CODE_SIGN_IDENTITY
* carthage install method: homebrew
* `which carthage`: `/usr/local/bin/carthage`
* `carthage version`: `0.29.0`
* `xcodebuild -version`: Xcode 10.0, Build version 10L176w

**Actual outcome**
When building in Xcode 10, the Copy Frameworks script phase fails with the error:

> Environment variable not set: `EXPANDED_CODE_SIGN_IDENTITY`

**Expected outcome**
Carthage should be able to find the `EXPANDED_CODE_SIGN_IDENTITY` environment variable but it isn't being set by `xcodebuild`.

clayellis:
Resolves #2472 

Xcode 10 is apparently omitting environment variables whose value is an empty string where previous versions would include the variable. This PR accounts for that by ignoring the missing environment variable error (in this case `EXPANDED_CODE_SIGN_IDENTITY`).

#### Other options considered (adjusting `getEnvironmentVariable`):

1. Specifying that the environment variable is optional, in which case an empty string would be returned.
```
func getEnvironmentVariable(_ variable: String, isOptional: Bool = false) -> Result<String, CarthageError> {
    let environment = ProcessInfo.processInfo.environment

    if let value = environment[variable] {
        return .success(value)
    } else if isOptional {
        return .success("")
    } else {
        return .failure(CarthageError.missingEnvironmentVariable(variable: variable))
    }
}
```
This didn't seem like the right solution since an empty string didn't seem like it had enough precedence for being the fallback value. So next I considered the following:

2. Providing a fallback value (that defaults to `nil` so that no existing code would be affected by the change.)
```
func getEnvironmentVariable(_ variable: String, fallbackValue: String? = nil) -> Result<String, CarthageError> {
    let environment = ProcessInfo.processInfo.environment

    if let value = environment[variable] {
        return .success(value)
    } else if let fallback = fallbackValue {
        return .success(fallback)
    } else {
        return .failure(CarthageError.missingEnvironmentVariable(variable: variable))
    }
}
```
This could have been the best solution, but in this specific case, we would have to provide an empty string as the fallback value, which would get ignored by the `.isEmpty` check in the `map` which seemed a little too tightly coupled to that logic.

### DIFF
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -46,13 +46,12 @@ public struct CopyFrameworksCommand: CommandProtocol {
 }
 
 private func copyFramework(_ source: URL, target: URL, validArchitectures: [String]) -> SignalProducer<(), CarthageError> {
-	return SignalProducer.combineLatest(copyProduct(source, target), codeSigningIdentity())
-		.flatMap(.merge) { url, codesigningIdentity -> SignalProducer<(), CarthageError> in
+	return copyProduct(source, target).flatMap { url -> SignalProducer<(), CarthageError> in
 			let strip = stripFramework(
 				url,
 				keepingArchitectures: validArchitectures,
 				strippingDebugSymbols: shouldStripDebugSymbols(),
-				codesigningIdentity: codesigningIdentity
+				codesigningIdentity: codeSigningIdentity()
 			)
 			if buildActionIsArchiveOrInstall() {
 				return strip
@@ -100,14 +99,8 @@ private func copyBCSymbolMapsForFramework(_ frameworkURL: URL, fromDirectory dir
 		}
 }
 
-private func codeSigningIdentity() -> SignalProducer<String?, CarthageError> {
-	return SignalProducer { () -> Result<String?, CarthageError> in
-		if codeSigningAllowed() {
-			return .success(getEnvironmentVariableIfPresent("EXPANDED_CODE_SIGN_IDENTITY"))
-		} else {
-			return .success(nil)
-		}
-	}
+private func codeSigningIdentity() -> String? {
+	return getEnvironmentVariableIfPresent("EXPANDED_CODE_SIGN_IDENTITY")
 }
 
 private func codeSigningAllowed() -> Bool {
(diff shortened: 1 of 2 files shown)


clayellis:
A different approach, which I think I might actually like best, is introducing a `getEnvironmentVariableIfPresent(_:)` method which would result in:

*Environment.swift*
```
func getEnvironmentVariableIfPresent(_ variable: String, isOptional: Bool = false) -> String? {
    let environment = ProcessInfo.processInfo.environment

    return environment[variable]
}
```

*CopyFrameworks.swift*
```
private func codeSigningIdentity() -> SignalProducer<String?, CarthageError> {
    return SignalProducer { () -> Result<String?, CarthageError> in
        if codeSigningAllowed() {
            return .success(getEnvironmentVariableIfPresent("EXPANDED_CODE_SIGN_IDENTITY"))
        } else {
            return .success(nil)
        }
    }
}
```

I'm going to revise the PR to use this solution over flat mapping the error away.

clayellis:
I won't make any more changes now, I think we've reached the best solution.

blender:
@clayellis  I believe you altered the logic by ignoring `codeSigningAllowed`

clayellis:
@blender You're right. I'll fix that.

Source/carthage/CopyFrameworks.swift
```diff
@@ -100,13 +99,12 @@ private func copyBCSymbolMapsForFramework(_ frameworkURL: URL, fromDirectory dir
 		}
 }
 
-private func codeSigningIdentity() -> SignalProducer<String?, CarthageError> {
-	return SignalProducer { () -> Result<String?, CarthageError> in
-		if codeSigningAllowed() {
-			return getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY").map { $0.isEmpty ? nil : $0 }
```
ikesyo: I think a simpler solution is changing this line to:

```swift
return getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY")
    .flatMapError { _ in .success("") } 
    .map { $0.isEmpty ? nil : $0 }
```

mdiep:
We've taken #2476 instead. Thanks for opening a PR to fix this!

PR #2474, opened 2018-06-06
Truncate log file before writing new one

ChristopherRogers:
Carthage was just writing on top of the log file so you could see old logs if the new output was smaller than the previous output. This forces the file to be truncated first. (Alternatively, you could also use `handle.truncateFile(atOffset: 0)` method to achieve the same thing.)


mdiep:
Nice catch/fix. 👏

PR #2475, opened 2018-06-06
Fix build error with Xcode 10

ChristopherRogers:
This fixes the error, "Raw type 'String' must appear first in the enum inheritance clause".


blender:
Thanks! 🎉

PR #2476, opened 2018-06-06
Consider `XCODE_PRODUCT_BUILD_VERSION` alongside `EXPANDED_CODE_SIGN_IDENTITY` to address Xcode 10 beta.

jdhealy:
Fixes <https://github.com/Carthage/Carthage/issues/2472>.

Treat `EXPANDED_CODE_SIGN_IDENTITY` — when `EXPANDED_CODE_SIGN_IDENTITY` is unset and `XCODE_PRODUCT_BUILD_VERSION` is set — with the same behavior as having empty string.

Xcode beta release `10L176w` (and potentially more) seem to omit environment variables set to an empty string in Run Script Phases (and potentially other places). So with the heuristic of `XCODE_PRODUCT_BUILD_VERSION` being set signifying a Run Script Phases (or other within Xcode scenario), ignore the error by returning `nil`, as is done when `codeSigningAllowed()` is false.

Credit and thanks to @clayellis for reporting the behavior change and [iterating on code changes](https://github.com/Carthage/Carthage/pull/2473).

### DIFF
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -101,13 +101,29 @@ private func copyBCSymbolMapsForFramework(_ frameworkURL: URL, fromDirectory dir
 }
 
 private func codeSigningIdentity() -> SignalProducer<String?, CarthageError> {
-	return SignalProducer { () -> Result<String?, CarthageError> in
-		if codeSigningAllowed() {
-			return getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY").map { $0.isEmpty ? nil : $0 }
-		} else {
-			return .success(nil)
+	return SignalProducer(codeSigningAllowed)
+		.attemptMap { codeSigningAllowed in
+			guard codeSigningAllowed == true else { return .success(nil) }
+
+			return getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY")
+				.map { $0.isEmpty ? nil : $0 }
+				.flatMapError {
+					// See https://github.com/Carthage/Carthage/issues/2472#issuecomment-395134166 regarding Xcode 10 betas
+					// … or potentially non-beta Xcode releases of major version 10 or later.
+
+					switch getEnvironmentVariable("XCODE_PRODUCT_BUILD_VERSION") {
+					case .success(_):
+						// See the above issue.
+						return .success(nil)
+					case .failure(_):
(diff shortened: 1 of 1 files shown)


jdhealy:
@clayellis I gave you push access to the [jdhealy/Carthage](https://github.com/jdhealy/Carthage) fork — if there’s any comments or code changes you’d like to push to this PR.

blender:
`XCODE_PRODUCT_BUILD_VERSION` is only set on Xcode >= 10 ?

jdhealy:
`XCODE_PRODUCT_BUILD_VERSION` isn’t anything new, it gets included in much older Xcodes.

The thought is: _all_ Xcode ‘Run Script Phases’ get the same behavior, but outside of Xcode (via the heuristic of `XCODE_PRODUCT_BUILD_VERSION`) — where there’s actually utility in guiding the user through what’s necessary to add to achieve (for what most is the goal) of ensuring that code signing happens — we surface the error of unset `EXPANDED_CODE_SIGN_IDENTITY`

Source/carthage/CopyFrameworks.swift
```diff
@@ -101,13 +101,29 @@ private func copyBCSymbolMapsForFramework(_ frameworkURL: URL, fromDirectory dir
...
+			guard codeSigningAllowed == true else { return .success(nil) }
+
+			return getEnvironmentVariable("EXPANDED_CODE_SIGN_IDENTITY")
+				.map { $0.isEmpty ? nil : $0 }
+				.flatMapError {
+					// See https://github.com/Carthage/Carthage/issues/2472#issuecomment-395134166 regarding Xcode 10 betas
+					// … or potentially non-beta Xcode releases of major version 10 or later.
+
+					switch getEnvironmentVariable("XCODE_PRODUCT_BUILD_VERSION") {
+					case .success(_):
```
kballard: BTW, I don't know if this is some particular Carthage style you're following here, but in general you can omit the `(_)` and just say `case .success:` if you don't want any of the associated values.

jdhealy:
> Xcode beta release 10L176w (and potentially more) seem to omit environment variables set to an empty string in Run Script Phases (and potentially other places).

This was confirmed intentional by Apple: ⌘F for 40843445 in [Xcode 10 Beta Release Notes (via Apple Developer Sign-In)](https://developer.apple.com/go/?id=xcode-10-beta-rn).

Issue #2449, opened 2018-05-09
By Moving Binary Download to Build process, Binaries are no longer pulled if using submodules


dmiluski:
# PR Reference:
Follow up question on this change: Move binary download to build process #2342

Given this change, and use case:
carthage update --no-build --use-ssh --use-submodules $@

A mix of submodule internal/external Carthage Dependencies, plus a few additional Carthage Binary only dependencies based on those developer's project distribution preferences. This allows us to pull in dependencies and only build required architectures while developing.

Previously (0.28.0), this worked, with the issue of cached binary as referenced in PR summary.

From what I'm reading here, if we skill build via --no-build, this would prevent binary download? Any recommendations for how to approach this use case? Perhaps I'm approaching project dependency management in a way that is not optimal? Or if there is another solution of managing submodules vs binaries I had missed?

# Details
* carthage install method: [ ] .pkg, [ ] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`:  0.29.0
* `xcodebuild -version`:  Xcode 9.3 Build version 9E145
* Are you using `--no-build`? YES
* Are you using `--no-use-binaries`?  NO
* Are you using `--use-submodules`?  YES
* Are you using `--cache-builds`? NO
* Are you using `--new-resolver`? NO

carthage update --no-build --use-ssh --use-submodules

**Cartfile**
```
# Interanal
github "https://github.braintreeps.com/venmo-ios/VenmoKit" ~> 12.0

# Binary Only Third Party
binary "https://raw.githubusercontent.com/dmiluski/CarthageSpecs/master/AppboyFramework.json"
binary "https://raw.githubusercontent.com/dmiluski/CarthageSpecs/master/mParticle-Appboy.json"
```

**Carthage Output**
```
>> Successful Carthage run, but Binary Only dependencies aren't pulled
```

**Actual outcome**
Carthage did or did not ...
>> Successful Carthage run, but Binary Only dependencies aren't pulled

**Expected outcome**
Carthage should pull dependencies specified as submodules as well as pull binaries even if --no-build is passed in.

ikesyo:
Actually this is not related to `--use-submodules` but related to `--no-build`.

KoCMoHaBTa:
@ikesyo What is the progress of this issue? Are there any directions for solving it?

hydra1983:
Any command to download binaries only?

samritchie:
[...]

PR #2478, opened 2018-06-07
Proposed Behavior Change: Binary Only Downloads

dmiluski:
## 🔨 Issue :
[Binaries are no longer pulled if using submodules](https://github.com/Carthage/Carthage/issues/2449)

Which occurred after this change: 
Move binary download to build process [#2342](https://github.com/Carthage/Carthage/pull/2342)

## 📦 Use Case:
```
# Third Party
github "SwiftKickMobile/SwiftMessages" ~> 4.0

# Binary Only Third Party
binary "https://raw.githubusercontent.com/dmiluski/CarthageSpecs/master/AppboyFramework.json"
```
Given this change, and use case:
`carthage update --no-build --use-ssh --use-submodules`

A users directly is not propagated with the binary-only dependency.

A mix of submodule internal/external Carthage Dependencies, plus a few additional Carthage Binary only dependencies based on those developer's project distribution preferences. This allows us to pull in dependencies and only build required architectures while developing.

## ❓ Why:
Given Binary only downloads has moved to building phase, build patterns which opt to not build do not get the benefit of being able to download BinaryOnly.

## 💍 Proposal:
Provide additional option to verbosely specify downloading binaries

## 🗒  Options:
`--download-binaries-and-skip-build`     // I felt this one could conflict with `--no-build`
`--download-binaries-if-skip-build`         // Conveys we still want to download binaries even if build skipped
`--download-binaries`                                // Simpler, but does this conflict with `--use-binaries`
`--download-binary-only-dependencies` // Too Long?
   others?

## ❓ Questions:
Is this a step completely separate from sync, build? ✅ 
Will this affect build? 🚫 
Is this option naming confusing given the parameter: --use-binaries ❓
When should this now happen: Checkout : Build : Other ❓ 

## Additional Comments:
Current PR Diff does not accurately represent proposal yet

## Initial Attempt: 🤧 
[First Commit](https://github.com/Carthage/Carthage/pull/2478/commits)
Goal:
Not require --build in order to provide a buildProducer. Instead, inspect --build within buildProducer to make the determination.

Problem
Per @ikesyo 
```That’s because the build option key is already defined and consumed by UpdateCommand.Options so BuildOptions would not get the option and the default value will be used.```

Option:
- Pass correctly vs currying (didn't understand this part of parsing)
- move forward with build approach, or use additional flag

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -1034,7 +1034,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 				return SignalProducer(dependencies)
 					.flatMap(.concurrent(limit: 4)) { dependency, version -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
 						switch dependency {
-						case .git, .gitHub:
+						case .git where options.build, .gitHub where options.build:
 							guard options.useBinaries else {
 								return .empty
 							}
@@ -1042,6 +1042,9 @@ public final class Project { // swiftlint:disable:this type_body_length
 								.filterMap { installed -> (Dependency, PinnedVersion)? in
 									return installed ? (dependency, version) : nil
 								}
+						case .git, .gitHub:
+							return .empty // If "--no-build" no need to build
+
 						case let .binary(binary):
 							return self.installBinariesForBinaryProject(binary: binary, pinnedVersion: version, projectName: dependency.name, toolchain: options.toolchain)
 								.then(.init(value: (dependency, version)))
@@ -1088,7 +1091,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 							// not to error out with `.noSharedFrameworkSchemes`
 							// to continue building other dependencies.
 							self._projectEventsObserver.send(value: .skippedBuilding(dependency, error.description))
-							
+
 							if options.cacheBuilds {
 								// Create a version file for a dependency with no shared schemes
 								// so that its cache is not always considered invalid.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Source/CarthageKit/Project.swift [hunk 1]
```diff
@@ -1034,14 +1034,17 @@ public final class Project { // swiftlint:disable:this type_body_length
 				return SignalProducer(dependencies)
 					.flatMap(.concurrent(limit: 4)) { dependency, version -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
 						switch dependency {
-						case .git, .gitHub:
+						case .git where options.build, .gitHub where !options.build:
```
dmiluski: This may be a misunderstanding, but with installing binaries in the build process I assumed rather than check for no-build ahead of time on:
https://github.com/Carthage/Carthage/pull/2478/files#diff-46708ead17dbc4a66dd1b8eb0e73b9a9L39

I assumed this check could instead be done here? Otherwise, perhaps installBinaries may not be appropriate in the build step?

🤦‍♂️
Now that I'm writing this, there may have been another location to do this the whole time.

ikesyo:
I will take a look 👀

dmiluski:
As it stands, this is still incorrect since it’s still building when I pass no build so i’ve Definitely missed something.

Working my way through each of the command arguments now to get a better idea of how each arg works where to get an understanding of checkout vs build vs download

blender:
Referencing rationale for behavior change in 0.29.0 https://github.com/Carthage/Carthage/pull/2317#issuecomment-362600637

ikesyo:
> this is still incorrect since it’s still building when I pass no build so i’ve Definitely missed something.

That’s because the `build` option key is already defined and consumed by `UpdateCommand.Options` command so `BuildOptions` would not get the option and the default value will be used.

blender:
> When I run make install, then install the package I was still seeing 0.29.0 even though I was updating the plist and added a local tag 

```
`brew unlink carthage`
rm `which carthage`
make install
```

PR #2486, opened 2018-06-14
Make use of --new-resolver flag in bootstrap command

blender:


Source/carthage/Bootstrap.swift
```diff
@@ -17,7 +17,9 @@ public struct BootstrapCommand: CommandProtocol {
 				if !FileManager.default.fileExists(atPath: project.resolvedCartfileURL.path) {
 					let formatting = options.checkoutOptions.colorOptions.formatting
 					carthage.println(formatting.bullets + "No Cartfile.resolved found, updating dependencies")
-					return project.updateDependencies(shouldCheckout: options.checkoutAfterUpdate, buildOptions: options.buildOptions)
+					return project.updateDependencies(shouldCheckout: options.checkoutAfterUpdate,
+													  useNewResolver: options.useNewResolver,
```
blender: No idea why it looks like this on Github, looks fine in Xcode.

ikesyo: I think this should be formatted like this:

```swift
return project.updateDependencies(
	shouldCheckout: options.checkoutAfterUpdate,
	useNewResolver: options.useNewResolver,
	buildOptions: options.buildOptions
)
```

blender:
@ikesyo Done

PR #2496, opened 2018-06-21
Sanitizing dependency names

Mov1s:
#### Description
This removes percent escaped characters when trying to determine a dependency name from a git URL.

#### Motivation
Since the dependency name for a git URL was derived from the last path component it could potentially contain URL percent escaped characters, since the dependency name is used throughout the build processes those special characters show up in a few places. Specifically there was a problem with `xcodebuild`'s `-derivedDataPath` when the path contained a percent sign.

I considered just sanitizing the derived data path but decided to clean up the dependency name the best I could so that those characters don't work their way into the build system later on. Plus it makes the console output look better 😄

### DIFF
--- a/Source/CarthageKit/GitURL.swift
+++ b/Source/CarthageKit/GitURL.swift
@@ -52,7 +52,8 @@ public struct GitURL {
 		return components
 			.last
 			.map(String.init)
-			.map(strippingGitSuffix)
+			.map(strippingGitSuffix)?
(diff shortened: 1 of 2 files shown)


blender:
@Mov1s Great! Thanks for the contribution! Looks good to me!

mdiep:
Spaces aren't ideal, but they're probably better than `%20`. 🙃

PR #2501, opened 2018-06-26
[CI] Use Xcode 9.4 image.

jdhealy:
At least, temporarily.

Maybe, no need to even merge — just checking what this the run for this PR looks like might be enough.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,18 +12,18 @@ matrix:
   include:
     - os: osx
       language: objective-c
-      osx_image: xcode9.2
+      osx_image: xcode9.4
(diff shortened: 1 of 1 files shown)


mdiep:
It wouldn't hurt to run both 9.2 and 9.4 on CI.

blender:
Do we want to duplicate everything to run on Xcode 9.2 and 9.4 ?

jdhealy:
Duplicate approach with `osx_image`s for both Xcode 9.2 and 9.4 detailed in issue https://github.com/Carthage/Carthage/issues/2508.

Issue #2500, opened 2018-06-26
Frameworks without Info.plist are rejected


blender:
* carthage install method: [ ] .pkg, [] homebrew, [x] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`:  master (pre 0.30.0)
* `xcodebuild -version`:  Xcode 9.2
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? no
* Are you using `--new-resolver`? no

**Cartfile**
```
N/A
```

**Carthage Output**
```
Failed to write to /Users/Tommaso/Code/<redacted>/iOS/<redacted>/Carthage/Build/.Fabric.version: Error Domain=NSCocoaErrorDomain Code=4 "The folder “.Fabric.version” doesn’t exist." UserInfo={NSFilePath=/Users/Tommaso/Code/<redacted>/iOS/<redacted>/Carthage/Build/.Fabric.version, NSUserStringVariant=Folder, NSUnderlyingError=0x604000055d50 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
```

**Actual outcome**
Carthage did not write a `.version` file nor copy the contents of the archive to `Carthage/Build`

**Expected outcome**
Carthage should .write a `.version` file and copy the contents of the archive to `Carthage/Build`

blender:
See discussion in https://github.com/Carthage/Carthage/issues/2491

blender:
Fixed in 0.30.0

PR #2502, opened 2018-06-26
Use `objdump` to determine whether a binary blob from a compressed archive should be handled by Carthage

blender:
Some frameworks don't have `Info.plists` and are being currently rejected as binary dependencies.

This PR addresses #2500 and also makes sure that .version files are written.

### DIFF
--- a/Source/CarthageKit/MachHeader.swift
+++ b/Source/CarthageKit/MachHeader.swift
@@ -1,0 +1,146 @@
+import Foundation
+import MachO.loader
+import ReactiveTask
+import ReactiveSwift
+import Result
+
+/// Represents a Mach header
+///
+/// Provides a unified structure for
+/// `MachO.loader.mach_header` and `MachO.loader.mach_header_64`
+
+struct MachHeader {
+
+	enum Endianness {
+		case little
+		case big
+	}
+
+	let magic: UInt32
+	let cpuType: cpu_type_t
+	let cpuSubtype: cpu_type_t
+	let fileType: UInt32
+	let ncmds: UInt32
+	let sizeofcmds: UInt32
+	let flags: UInt32
+	let reserved: UInt32?
+
+	var is64BitHeader: Bool {
+
+		return magic == MH_MAGIC_64 || magic == MH_CIGAM_64
+	}
+
+	var is32BitHeader: Bool {
+
+		return !is64BitHeader
+	}
+
+	var endianess: Endianness {
+
+		return magic == MH_CIGAM_64 || magic == MH_CIGAM ? .big : .little
+	}
+}
+
+extension MachHeader {
+
+	static let carthageSupportedFileTypes: Set<UInt32> = {
+		return Set([
+			MH_OBJECT, // Carthage accepts static libraries
+			MH_BUNDLE, // Bundles https://github.com/ResearchKit/ResearchKit/blob/1.3.0/ResearchKit/Info.plist#L15-L16
+			MH_DYLIB, // or dynamic shared libraries
+			].map { UInt32($0) }
+		)
+	}()
+}
+
+extension MachHeader {
+
+	/// Reads the Mach headers from a Mach-O file.
+	/// - Parameter url: The url of the Mach-O file
+	/// - Remark: Uses `objdump` to read the header and parse the output.
+	///           The output is composed of one or more sets of lines like the following:
+	///
+	///           Mach header
+	///                 magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
+	///            0xfeedfacf 16777223          3  0x00           1     8       1720 0x00002000
+	///
+	/// - See Also:  [LLVM MachODump.cpp](https://llvm.org/viewvc/llvm-project/llvm/trunk/tools/llvm-objdump/MachODump.cpp?view=markup&pathrev=225383###see%C2%B7line%C2%B72745)
+
+	static func headers(forMachOFileAtUrl url: URL) -> SignalProducer<MachHeader, CarthageError> {
+
+		// This is the command `otool -h` actually invokes
+		let task = Task("/usr/bin/xcrun", arguments: [
+			"objdump",
+			"-macho",
+			"-private-header",
+			"-non-verbose",
+			url.resolvingSymlinksInPath().path
+			]
+		)
+
+		return task.launch(standardInput: nil)
+			.ignoreTaskData()
+			.map { String(data: $0, encoding: .utf8) ?? "" }
+			.filter { !$0.isEmpty }
+			.flatMap(.merge) { (output: String) -> SignalProducer<(String, String), NoError> in
+				output.linesProducer.combinePrevious()
+			}.filterMap { (previousLine, currentLine) -> MachHeader? in
+
+				let previousLineComponents = previousLine
+					.components(separatedBy: CharacterSet.whitespaces)
+					.filter { !$0.isEmpty }
+				let currentLineComponents = currentLine
+					.components(separatedBy: CharacterSet.whitespaces)
+					.filter { !$0.isEmpty }
+
+				let strippedComponents = currentLineComponents
+					.map { $0.stripping(prefix: "0x")}
+
+				let magicIdentifiers = [
+					MH_MAGIC_64,
+					MH_CIGAM_64,
+					MH_MAGIC,
+					MH_CIGAM,
+					].lazy
+
+				guard previousLineComponents == [
+					"magic",
+					"cputype",
+					"cpusubtype",
+					"caps",
+					"filetype",
+					"ncmds",
+					"sizeofcmds",
+					"flags"
+					]
+					, !strippedComponents.isEmpty
+					, let magic = UInt32(strippedComponents.first!, radix:16)
+					, magicIdentifiers.first(where: { $0 ==  magic }) != nil else {
+						return  nil
+				}
+
+				guard
+					let cpuType = cpu_type_t(strippedComponents[1], radix: 10),
+					let cpuSubtype =  cpu_subtype_t(strippedComponents[2], radix: 10),
+					let fileType = UInt32(strippedComponents[4], radix: 10),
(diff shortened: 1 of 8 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -1289,14 +1290,78 @@ internal func frameworksInDirectory(_ directoryURL: URL) -> SignalProducer<URL,
...
+					}
+					.filter {
+						$0 == MH_OBJECT // Carthage accepts static libraries
+						|| $0 == MH_BUNDLE // Bundles https://github.com/ResearchKit/ResearchKit/blob/1.3.0/ResearchKit/Info.plist#L15-L16
+						|| $0 == MH_DYLIB // or dynamic shared libraries
+					}
+					.single()?
+					.value
+
+					return fileType != nil
```
mdiep: I think this code ought to be split out into its own function.

Source/CarthageKit/VersionFile.swift
```diff
@@ -49,7 +49,13 @@ struct VersionFile: Codable {
...
 	}
 
-	init(commitish: String, macOS: [CachedFramework]?, iOS: [CachedFramework]?, watchOS: [CachedFramework]?, tvOS: [CachedFramework]?) {
+	init(
+		commitish: String,
+		macOS: [CachedFramework]?,
+		iOS: [CachedFramework]?,
+		watchOS: [CachedFramework]?,
+		tvOS: [CachedFramework]?
+	) {
```
mdiep: Are there any functional changes in this file? I'm having trouble spotting anything with the reformatting.

Source/CarthageKit/Project.swift
mdiep: I'd want to see tests for this as well.

Source/CarthageKit/VersionFile.swift
blender: No there are no functional changes.

Source/CarthageKit/Project.swift
blender: ok

Source/CarthageKit/Project.swift
```diff
@@ -1289,14 +1290,78 @@ internal func frameworksInDirectory(_ directoryURL: URL) -> SignalProducer<URL,
...
+					.flatMap(.merge) { (output: String) -> SignalProducer<String, NoError> in
+						output.linesProducer
+					}.filter {
+						// Keep the lines with Mach-O magic number indefiers
+						$0.contains(String(format: "%x", MH_CIGAM))
+							|| $0.contains(String(format: "%x", MH_MAGIC))
+							|| $0.contains(String(format: "%x", MH_CIGAM_64))
+							|| $0.contains(String(format: "%x", MH_MAGIC_64))
+					}
+					.take(last: 1)
```
jdhealy: So, on the unstated assertion here (‘every `filetype` on a magic number containing line will be the same, so just grab any of them — as implemented: the last one’) — I said earlier that I’d post an implementation where a `reduce` makes this more explicit:

```swift
let acceptedFiletypesAsSpecifiedByMachHeader: Set<Int32> = [
	MH_OBJECT, // Carthage accepts static libraries
	MH_BUNDLE, // Bundles https://github.com/ResearchKit/ResearchKit/blob/1.3.0/ResearchKit/Info.plist#L15-L16
	MH_DYLIB, // or dynamic shared libraries
]

let øøø: Bool = task.launch(standardInput: nil)
	.ignoreTaskData()
	.map { String(data: $0, encoding: .utf8) ?? "" }
	.filter { !$0.isEmpty }
	.flatMap(.merge) { (output: String) -> SignalProducer<(String, String), NoError> in
		output.linesProducer.combinePrevious()
	}
	.filterMap {
		guard
			$0.0.hasPrefix("      magic cputype cpusubtype  caps    filetype ncmds"),
			[MH_CIGAM, MH_MAGIC, MH_CIGAM_64, MH_MAGIC_64]
				.lazy.map({ " " + String(format: "%x", $0) })
				.first(where: $0.1.hasPrefix) != nil
		else { return nil }

		let components = $0.1
			.components(separatedBy: CharacterSet.whitespaces)
			.filter { !$0.isEmpty }

		precondition(components.count == 8, "∆∆∆")

		return Int32(components[4])
	}
	.filter(acceptedFiletypesAsSpecifiedByMachHeader.contains)
	.reduce(into: Set<Int32>()) { $0.insert($1); return }
	.map { $0.count == 1 }
	.single()?
	.value ?? false
```

What do you think?

blender: I like the reduce but I think that this line with is maybe too "whitespace" sensitive?

`$0.0.hasPrefix("      magic cputype cpusubtype  caps    filetype ncmds"),`

jdhealy: Whitespace is fixed according to <https://llvm.org/viewvc/llvm-project/llvm/trunk/tools/llvm-objdump/MachODump.cpp?view=markup&pathrev=225383###see·line·2745> — IMO, we’d want a silent non-consideration if that output structure were to change (adding a field).

jdhealy: I guess we could strip the whitespace from both the output line and our conditional and compare, that would still catch an added field.

blender: I'm going to push a few changes according to @mdiep desire to have a separate function.

I'll try to put the reduce too

jdhealy:
Great work tracking this down (I should have said upfront) — massive thanks 👏

blender:
I have updated the PR:
- Introduced `MachHeader` strcture
- refactored out function to read mach headers in static method on `MachHeader`
- Added tests
- Remove checking for package type. It's superfluous give that the headers always work.

blender:
Any opinion on the refactored version?

mdiep:
The code looks fine to me, but I'm not totally confident whether this is a good idea. What do you think, @jdhealy?

Source/CarthageKit/Project.swift
```diff
@@ -1,6 +1,7 @@
 // swiftlint:disable file_length
 
 import Foundation
+import MachO.loader
```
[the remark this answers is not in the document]
blender: No need. Good catch.

blender:
The alternative that I have discussed with @jdhealy is to read the filetype by hand. I think it would be more error prone than using existing tools that know how to read Mach-O files correctly.

Source/CarthageKit/Project.swift
```diff
@@ -1282,23 +1283,33 @@ internal func frameworksInDirectory(_ directoryURL: URL) -> SignalProducer<URL,
...
 			default:
-				return false
+				// In case no Info.plist exists check the Mach-O fileType
+				guard let executableURL = bundle?.executableURL else {
+					return false
+				}
+
+				return MachHeader.headers(forMachOFileAtUrl: executableURL)
+					.filter { MachHeader.carthageSupportedFileTypes.contains($0.fileType) }
+					.reduce(into: Set<UInt32>()) { $0.insert($1.fileType); return }
```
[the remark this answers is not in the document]
jdhealy: That `.insert` doesn’t return `Void` which necessitates either a `_ = ` prefix or a `return` suffix to match the type signature of the `into:` `reduce`.

Source/CarthageKit/MachHeader.swift
```diff
@@ -0,0 +1,147 @@
...
+
+				guard
+					let cpuType = cpu_type_t(strippedComponents[1], radix: 10),
+					let cpuSubtype =  cpu_subtype_t(strippedComponents[2], radix: 10),
+					let fileType = UInt32(strippedComponents[4], radix: 10),
+					let ncmds = UInt32(strippedComponents[5], radix: 10),
+					let sizeofcmds = UInt32(strippedComponents[6], radix: 10),
+					let flags = UInt32(strippedComponents[7], radix: 16)
+					else {
+
```
jdhealy: Nit: I’d prefer this without this empty line.

Source/CarthageKit/Project.swift
```diff
@@ -1282,23 +1283,33 @@ internal func frameworksInDirectory(_ directoryURL: URL) -> SignalProducer<URL,
...
-
 			// Skip nested frameworks
 			let frameworksInURL = url.pathComponents.filter { pathComponent in
 				return (pathComponent as NSString).pathExtension == "framework"
 			}
 			return frameworksInURL.count == 1
 		}.filter { url in
-
-			let packageType: PackageType? = Bundle(url: url)?.packageType
+			// Try reading the Info.plist of the Framework
```
jdhealy: I’m confident in the structure of the CLI output from `objdump` rarely changing (and if it does, we should be able to catch via open-sourcedness before it’s widely distributed), but a simple Plist read is probably faster than a process execution. If you agree, could we leave this the default and add a comment with the reasoning?

```
// For reasons of speed and the fact that CLI-output structures can change,
// first try the safer method of reading the ‘Info.plist’ from the Framework’s bundle.
```

blender: Sure

Source/CarthageKit/MachHeader.swift
```diff
@@ -0,0 +1,147 @@
...
+extension MachHeader {
+
+	/// Reads the Mach headers from a Mach-O file.
+	/// - Parameter url: The url of the Mach-O file
+	/// - Remark: Uses `objdump` to read the header and parse the output.
+	/// 	      The output is composed of one or more sets of lines like the following:
+	///
+	///			Mach header
+	///		          magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
```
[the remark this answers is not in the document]
blender: Something my new installation of Xcode is wrong and it's driving me crazy with spaces and tabs :D . Fixed.

Source/CarthageKit/MachHeader.swift
blender: Done

blender:
I would like to rebase before merging to remove superfluous commits

blender:
Rebased. All Done.

PR #2518, opened 2018-07-18
 Fix SemVer parsing with too many dots

marcoconti83:
The code was incorrectly parsing invalid [SemVer](https://semver.org) version that contained too many dots (e.g. `1.8.0.1` or `1.8..1`) as valid.

This PR adds unit test to highlight those cases, and fixes the issue.


mdiep:
Awesome!

PR #2519, opened 2018-07-19
Resolver diagnostics

werner77:
Adds the diagnose command to store dependency information and Cartfile for problematic dependency trees so offline diagnosis can be done e.g. with unit tests.


blender:
@werner77 Great contribution! Thanks!

I would like to start reviewing this asap. Where should I start from to follow the flow the most accurately ?

Carthage.xcodeproj/xcshareddata/xcschemes/carthage.xcscheme
```diff
@@ -65,7 +65,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      launchStyle = "0"
```
blender: is this intentional?

werner77: No this is not correct, will revert that change

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,155 @@
...
+				try localRepository?.storePinnedVersions(pinnedVersions, for: storedDependency, gitReference: gitReference)
+
+				versionSet = pinnedVersions
+			}
+
+			let filteredVersionSet = versionSet.filter { (pinnedVersion) -> Bool in
+				versionSpecifier.isSatisfied(by: pinnedVersion)
+			}
+			
+			print("Versions for dependency '\(dependency)' compatible with versionSpecifier \(versionSpecifier): \(versionSet)")
```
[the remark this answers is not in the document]
werner77: yes good point

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,155 @@
...
+			return filteredVersionSet
+		} catch let error {
+			print("Caught error while retrieving versions for \(dependency): \(error)")
```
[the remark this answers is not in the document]
werner77: yes good point

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,155 @@
...
+			let storedTransitiveDependencies = transitiveDependencies.map { (transitiveDependency, versionSpecifier) -> (Dependency, VersionSpecifier) in
+				let storedTransitiveDependency = self.dependencyMappings?[transitiveDependency] ?? transitiveDependency
+				return (storedTransitiveDependency, versionSpecifier)
+			}
+			try localRepository?.storeTransitiveDependencies(storedTransitiveDependencies, for: storedDependency, version: version)
+
+			print("Dependencies for dependency '\(dependency)' with version \(version): \(transitiveDependencies)")
+			return transitiveDependencies
+		} catch let error {
+			print("Caught error while retrieving dependencies for \(dependency) at version \(version): \(error)")
```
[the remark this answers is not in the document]
werner77: yes good point

werner77:
To elaborate on the Pull Request and the features it adds:

- Adds a "diagnose" command (name is up for discussion) for storing the Cartfile and dependencies and associated versions into json files into a specifiable output directory. 
- This action takes three parameters:
    - --output-path: The directory path where to output the diagnosis information. If not specified a new temporary directory is created.
    - --project-directory: The directory containing the Carthage project, defaults to the current directory.
    - --mappings-file: An optional file containing mappings for dependencies to anonymize during the storage. The format should be: <dependency string>=<mapped dependency string> (one mapping per line).
- The information stored by the diagnose command can be used subsequently as input for test cases as shown by some of the pending test cases in ResolverSpec (e.g. pending("should correctly resolve items with conflicting names, giving precedence to pinned versions"))
- The methods have been added to the Project class to support this functionality:

```
public func storeDependencies(to repository: LocalRepository, ignoreErrors: Bool = false, dependencyMappings: [Dependency: Dependency]?) -> SignalProducer<Cartfile, CarthageError>

public func resolveUpdatedDependencies(
		from repository: LocalRepository,
		resolverType: ResolverProtocol.Type,
		dependenciesToUpdate: [String]? = nil) -> SignalProducer<ResolvedCartfile, CarthageError>
```

- Supporting classes are LocalRepository (which is responsible for storing or loading the dependencies to/from the local filesystem) and DiagnosticResolver (implementation of ResolverProtocol which is responsible for using the local repository to store its output).
- Some test cases have been added by the Pull Request, which are set to pending. The BackTrackingResolver introduced by the Pull Request https://github.com/Carthage/Carthage/pull/2455 solves these issues (and other performance issues).

blender:
Ok, so despite the scary __906__ File changed, the changes are additive and limited to a few classes. Most of the file changes are json files of recorded dependencies.

I did a first pass for "side effect changes to existing logic" and din't find any.

I left a few comments to address. Also it would greatly help if you could rebase this in 2 commits:

- commit 1: Purely the new command and it's pieces
- commits 2: Tests & jsons.

I know this is yet another request but it would make reviewing this much much easier.

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,170 @@
...
+		) -> SignalProducer<[Dependency: PinnedVersion], CarthageError> {
+		let result: Result<[Dependency: PinnedVersion], CarthageError>
+
+		do {
+			try traverse(dependencies: Array(dependencies))
+			result = .success([Dependency: PinnedVersion]())
+		} catch let error {
+			let carthageError: CarthageError = (error as? CarthageError) ?? CarthageError.internalError(description: error.localizedDescription)
+			result = .failure(carthageError)
+		}
```
chuganzy: [nits] I think it's better to split error handling by types:
```
do {
    try traverse(dependencies: Array(dependencies))
    result = .success([Dependency: PinnedVersion]())
} catch let error as CarthageError {
    result = .failure(error)
} catch {
    result = .failure(CarthageError.internalError(description: error.localizedDescription))
}

File: Source/CarthageKit/DiagnosticResolver.swift [hunk 6]
```diff
@@ -0,0 +1,170 @@
...
+	func diagnosticResolver(_ resolver: DiagnosticResolver, didFindTransitiveDependencies transitiveDependencies: [(Dependency, VersionSpecifier)], for dependency: Dependency, withVersion version: PinnedVersion)
+	func diagnosticResolver(_ resolver: DiagnosticResolver, didEncounterError error: Error, whileRetrievingTransitiveDependenciesFor dependency: Dependency, withVersion version: PinnedVersion)
+	func diagnosticResolver(_ resolver: DiagnosticResolver, didEncounterError error: Error, whileRetrievingVersionsFor dependency: Dependency, compatibleWith versionSpecifier: VersionSpecifier)
+}
+
+/**
+Resolver which logs all dependencies it encounters, and optionally stores them in the specified local repository.
+*/
+public final class DiagnosticResolver: ResolverProtocol {
+	public weak var delegate: DiagnosticResolverDelegate?
```
chuganzy: I don't think Carthage uses `delegate` pattern. Please consider to use `Signal` instead.

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,170 @@
...
+					handledDependencies.insert(pinnedDependency)
+
+					let transitiveDependencies = try findDependencies(for: dependency, version: version)
+					try traverse(dependencies: transitiveDependencies)
+				}
+			}
+		}
+	}
+
+	private func findAllVersions(for dependency: Dependency, compatibleWith versionSpecifier: VersionSpecifier) throws -> [PinnedVersion] {
```
chuganzy: I feel this function should return `SignalProducer`

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,170 @@
...
+				default:
+					pinnedVersionsProducer = versionsForDependency(dependency)
+				}
+
+				let concreteVersionsProducer = pinnedVersionsProducer.filterMap { pinnedVersion -> PinnedVersion? in
+					pinnedVersions.append(pinnedVersion)
+					return nil
+				}
+
+				_ = try concreteVersionsProducer.collect().first()!.dematerialize()
```
chuganzy: Hmm? I have not tested yet but this should crash as `filterMap` always returns `nil` so far..

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,170 @@
...
+
+				switch versionSpecifier {
+				case .gitReference(let hash):
+					pinnedVersionsProducer = resolvedGitReference(dependency, hash)
+					gitReference = hash
+				default:
+					pinnedVersionsProducer = versionsForDependency(dependency)
+				}
+
+				let concreteVersionsProducer = pinnedVersionsProducer.filterMap { pinnedVersion -> PinnedVersion? in
```
chuganzy: Abuse of `filterMap`.

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,170 @@
...
+		public let pinnedVersion: PinnedVersion
+		private let hash: Int
+
+		init(dependency: Dependency, pinnedVersion: PinnedVersion) {
+			self.dependency = dependency
+			self.pinnedVersion = pinnedVersion
+			self.hash = 37 &* dependency.hashValue &+ pinnedVersion.hashValue
+		}
+
+		public var hashValue: Int {
```
chuganzy: How about making use of synthesizing hashable conformance?

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,170 @@
...
+			return hash
+		}
+
+		public static func == (lhs: PinnedDependency, rhs: PinnedDependency) -> Bool {
```
chuganzy: Same here.

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,170 @@
...
+				return [PinnedVersion]()
+			} else {
+				throw error
+			}
+		}
+	}
+
+	private func findDependencies(for dependency: Dependency, version: PinnedVersion) throws -> [(Dependency, VersionSpecifier)] {
+		do {
+			let transitiveDependencies: [(Dependency, VersionSpecifier)] = try dependenciesForDependency(dependency, version).collect().first()!.dematerialize()
```
chuganzy: This would crash when the dependency has no dependency.

Source/CarthageKit/DiagnosticResolver.swift
ikesyo: That only works for Xcode 9.3 or later (Swift 4.1) but we still support Xcode 9.2 (Swift 4.0).

Once we migrate to ReactiveSwift 4.0 which requires Swift 4.1, we can rely on that feature.

ikesyo: I have a WIP branch locally to update the dependencies so stay tuned :wink:

chuganzy: Aha, got it.

chuganzy:
Sorry I got busy these days.. I have not done yet with reviewing but will keep doing when I am free.

werner77:
[**@blender**](https://github.com/blender) commented on [22 Jul 2018, 09:36 CEST](https://github.com/Carthage/Carthage/pull/2519#issuecomment-406847406 "2018-07-22T07:36:52Z - Replied by Github Reply Comments"):
> Ok, so despite the scary **906** File changed, the changes are additive and limited to a few classes. Most of the file changes are json files of recorded dependencies.
> 
> I did a first pass for "side effect changes to existing logic" and din't find any.
> 
> I left a few comments to address. Also it would greatly help if you could rebase this in 2 commits:
> 
> *   commit 1: Purely the new command and it's pieces
> *   commit 2: Tests &amp; jsons.
> 
> I know this is yet another request but it would make reviewing this much much easier.

I'm not really sure rebasing is a good idea. Quoting from this site: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

> Don’t include any commit you’ve already pushed to a central server – doing so will confuse other developers by providing an alternate version of the same change.

Source/CarthageKit/DiagnosticResolver.swift
werner77: Applied this change

Source/CarthageKit/MachHeader.swift
```diff
@@ -111,17 +111,14 @@ extension MachHeader {
 					"filetype",
 					"ncmds",
 					"sizeofcmds",
-					"flags"
-					]
-					, !strippedComponents.isEmpty
-					, let magic = UInt32(strippedComponents.first!, radix:16)
-					, magicIdentifiers.first(where: { $0 ==  magic }) != nil else {
+					"flags",
+					], !strippedComponents.isEmpty, let magic = UInt32(strippedComponents.first!, radix:16), magicIdentifiers.first(where: { $0 == magic }) != nil else {
```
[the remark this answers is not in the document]
werner77: Reverted this change

Carthage.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
werner77: Reverted this file

Source/CarthageKit/DiagnosticResolver.swift
werner77: I am not quite sure it is such a great idea to let this reactive style of programming trickle through to every component (rather than in the assembly part of the app, glueing the different components together). Not only does every developer need to understand the reactive patterns instead of just ordinary Swift, it also makes the API more cumbersome in my opinion. 

The above may be due to me having too little experience with this idiom. However, as can be witnessed by the resolver performance issues, it is hard to get a grasp on algorithm time complexities while using reactive patterns everywhere.

That being said, I'm open to concrete suggestions how to improve the code in a reactive manner if that really is the wish. Maybe you could propose how to rewrite the code above?

blender: I am not so religious about the reactive style but that has been the standard so far. This is a question for @mdiep .

Source/CarthageKit/DiagnosticResolver.swift
blender: I agree. You can find several examples of how to use Signal for this purpose in the existing code base

Source/CarthageKit/DiagnosticResolver.swift
blender: Yes,  this should be just .collect for example

mdiep:
> I'm not really sure rebasing is a good idea.

That's good general advice, but it's okay to rebase here in a PR. No one is depending on this history yet. 🙂

Source/CarthageKit/DiagnosticResolver.swift
werner77: I don't understand how to apply this without changing the resolver protocol. Please either make the change yourself or provide a more elaborate code example.

Tests/CarthageKitTests/ResolverSpec.swift
```diff
@@ -355,23 +494,85 @@ class ResolverBehavior: Behavior<ResolverProtocol.Type> {
...
+                    .v2_0_0: [
+                        github2: .compatibleWith(.v2_0_0),
+                    ],
+                ],
+                github2: [
+                    .v1_0_0: [ github3: .compatibleWith(.v1_0_0) ],
+                ],
+                github3: [
+                    .v1_0_0: [ github1: .compatibleWith(.v1_0_0)],
+                ],
```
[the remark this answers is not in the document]
ikesyo: Now https://github.com/Carthage/Carthage/pull/2586 has been merged, so `trailing_comma` should be okay.

Tests/CarthageKitTests/ResolverSpec.swift
```diff
@@ -355,23 +494,85 @@ class ResolverBehavior: Behavior<ResolverProtocol.Type> {
...
+                case .dependencyCycle(_):
+                    print("Dependency cycle error: \(error)")
+                default:
+                    fail("Expected error to be of type .dependencyCycle")
+                }
+            }
+        }
+
+        pending("should correctly resolve items with conflicting names, giving precedence to pinned versions") {
+            let testCartfileURL = Bundle(for: ResolverBehavior.self).url(forResource: "Resolver/ConflictingNames/Cartfile", withExtension: "")!
```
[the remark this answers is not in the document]
ikesyo: This is okay: https://github.com/Carthage/Carthage/blob/58b703f45e8dca867085c8c255bc4fcf3d092129/.swiftlint.yml#L32-L33

Source/CarthageKit/DiagnosticResolver.swift
werner77: A problem with relying on auto-synthesis of hashable is performace. The hashValue should only be calculated once in the initializer and not every time hashValue is called. This actually is a big improvement in performance I did in my other pull request with the fast resolver.

mdiep:
Can you start by rebasing or squashing this to remove the unwanted commits? Then we can review.

werner77:
I squashed all commits into a single commit

Tests/CarthageKitTests/ResolverSpec.swift
```diff
@@ -56,11 +56,17 @@ class ResolverBehavior: Behavior<ResolverProtocol.Type> {
...
+				
+				switch resolved {
+				case .success(let value):
+					expect(value) == [
+						github2: .v1_0_0,
+						github1: .v0_1_0,
+					]
+				case .failure(let error):
+					fail("Expected no error to occur: \(error)")
+				}
```
mdiep: Why did these change?

Tests/CarthageKitTests/ResolverSpec.swift
```diff
@@ -193,15 +224,105 @@ class ResolverBehavior: Behavior<ResolverProtocol.Type> {
...
+						.v2_0_0: [:],
+					],
 				]
+
+				let resolved = db.resolve(resolverClass, [ github1: .any, github2: .compatibleWith(.v1_0_0), github3: .compatibleWith(.v1_0_0) ])
+				expect(resolved.value).to(beNil())
+				expect(resolved.error).notTo(beNil())
+			}
+
+			pending("should correctly resolve the latest version") {
```
mdiep: Can you pull out any added `pending` tests into a separate PR? 🙏

Tests/CarthageKitTests/Resources/Resolver/ConflictingDependencies/Repository/dependencies/BMBiometrics/0.1.0.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  {
+    "versionSpecifier" : "~> 0.13.0",
+    "dependency" : {
+      "git" : {
+        "urlString" : "https:\/\/git.behindmedia.com\/Common\/BMFoundation.git"
+      }
+    }
+  }
+]
```
mdiep: Where are all these fixtures used? Do we really need _all_ of them?

blender:
@werner77  @mdiep let's push this forward.

There are some changes to make and a crucial question that should be addressed. 

To the point: not every part of the PR follows FRP patterns. What do we do? I can help in that if needed.

werner77:
@blender @mdiep 

I'm going to put some renewed effort in to try to get this merged. 
Plan of action:

- Merge upstream/master and ensures the build works again
- Resolve the issues related to the reactive style programming patterns
- Any more?

In reply to @mdiep 

- The fixtures are automatically recorded, they comprise the full dependency set of all dependencies, for a complete test case yes they should all be there
- The pending tests are exactly the tests which are problematic for the other resolvers and are fixed by my new resolver implementation (https://github.com/Carthage/Carthage/pull/2455). We exclusively use the new resolver at my current work already for a year because of the issues with the old ones. I'd like to get them merged right away so we can go ahead with that pull request afterwards unless you have really strong objections to this.

blender:
The plan looks good to me. 

As far as I understand the diagnose command would work for all resolvers we currently have. Correct @werner77 ?

werner77:
@blender Yes it works for all resolvers. You can run any resolver against the local repository in a test case like:

```
let resolverType = ... // Any of Resolver.self, NewResolver.self, BackTrackingResolver.self
let project = Project(directoryURL: projectDirectoryURL)
let repository = LocalRepository(directoryURL: repositoryURL)
let signalProducer = project.resolveUpdatedDependencies(from: repository,
                                                                    resolverType: resolverType.self,
                                                                    dependenciesToUpdate: nil)
```

werner77:
Apparently there is an error in the tests with a force unwrap of a nil value, but I am unable to pinpoint where that occurs. Somebody has any idea?

```
Test Suite 'RemoteVersionSpec' started at 2019-02-08 09:57:45.662
Test Case '-[CarthageKitTests.RemoteVersionSpec remoteVersion, should time out]' started.
Fatal error: Unexpectedly found nil while unwrapping an Optional value
Exited with signal code 4
```

werner77:
It seems to build successfully now, I replaced a whole lot of force casts, force unwraps and force tries in the test suite.

Is there any reason though why swift test --filter "some-regex" does not seem to work? It does list the tests but for some reason no actual tests are run, e.g.:

```
swift test -l --filter "ResolverSpec"
```

Correctly lists some tests but when trying to run them with:

```
swift test --filter "ResolverSpec"
```

no actual test is run.

There is no other option than to run the whole test suite now unfortunately (since unit tests with resources do not work from Xcode)

Source/CarthageKit/Cartfile.swift
```diff
@@ -121,6 +120,15 @@ public struct Cartfile {
 	}
 }
 
+extension Cartfile: CustomStringConvertible {
+	public var description: String {
+		return dependencies
+			.sorted { $0.key.description < $1.key.description }
+			.map { "\($0.key) \($0.value)\n" }
+			.joined(separator: "")
```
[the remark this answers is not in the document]
werner77: Fixed this

werner77:
For some reason the Test Suite 'ProjectSpec' took longer than the allowed time.

I feel this test is not really a unit test since it takes so long. To just check the build order the builds don't necessarily have to be triggered, right?

Anyway, any chance this PR can get through any time soon? I think I cleaned up everything that needed to be done.

Source/CarthageKit/LocalRepository.swift
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	public func loadTransitiveDependencies(for dependency: Dependency, version: PinnedVersion) throws -> [(Dependency, VersionSpecifier)] {
+		let fileURL = try self.transitiveDependenciesURL(for: dependency, version: version)
+		let data = try Data(contentsOf: fileURL)
+		let versionSpecs = try JSONDecoder().decode([DependencyVersionSpecification].self, from: data)
+		return versionSpecs.map { ($0.dependency, $0.versionSpecifier) }
+	}
+
+	private func pinnedVersionsURL(for dependency: Dependency, gitReference: String? = nil, createDirs: Bool = false) throws -> URL {
```
blender: I find this method confusing. 

The method is really just constructing a path for a JSON file but I don't find the name suggests that. 
Maybe naming it `canonicalPinnedVersionFileURL` would be a better choice (also possibly in an extension of Dependency).

Also it has 2 responsibilities. I would prefer it not to create any path, which is the only reason why this method throws.

Source/CarthageKit/LocalRepository.swift
```diff
@@ -0,0 +1,238 @@
...
+		let fileName = (gitReference.map { "\($0).json" } ?? "default.json")
+		let fileURL = directoryURL.appendingPathComponent("versions").appendingPathComponent(dependency.name).appendingPathComponent(fileName)
+
+		if createDirs {
+			try FileManager.default.createDirectory(at: fileURL.deletingLastPathComponent(), withIntermediateDirectories: true, attributes: nil)
+		}
+		return fileURL
+	}
+
+	private func transitiveDependenciesURL(for dependency: Dependency, version: PinnedVersion, createDirs: Bool = false) throws -> URL {
```
blender: same here

Source/CarthageKit/LocalRepository.swift
```diff
@@ -0,0 +1,238 @@
+// swiftlint:disable vertical_parameter_alignment_on_call
+import Foundation
+import Tentacle
+import Result
+import ReactiveSwift
+
+/**
+Local repository for storing/loading dependencies and their versions. This is for testing without requiring live connection to real repositories.
+*/
+public final class LocalRepository {
```
blender: The name is a bit confusing to me as it overlaps very much with git repository.

If I understand correctly this is just is just a wrapper around a URL for where to get the serialized dependency information from.

Maybe a name like `ResolverDiagnsticStore` suits this better.

Source/CarthageKit/Project.swift
```diff
@@ -1521,3 +1521,67 @@ public func cloneOrFetch(
...
+
+// Diagnostic methods to be able to diagnose problems with the resolver with dependencies which cannot be tested 'live', e.g. for private repositories
+extension Project {
+	// Function which outputs all possible dependencies and versions of those dependencies to the repository specified
+    public func storeDependencies(to repository: LocalRepository, ignoreErrors: Bool = false, dependencyMappings: [Dependency: Dependency]? = nil, eventObserver: ((DiagnosticResolverEvent) -> Void)? = nil) -> SignalProducer<(Cartfile, ResolvedCartfile?), CarthageError> {
+		let resolver = DiagnosticResolver(
+			versionsForDependency: versions(for:),
+			dependenciesForDependency: dependencies(for:version:),
+			resolvedGitReference: resolvedGitReference,
+            localRepository: repository
```
blender: line indentation seems off here

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,197 @@
...
+
+        let (signal, observer) = Signal<DiagnosticResolverEvent, NoError>.pipe()
+        diagnosticResolverEvents = signal
+        diagnosticResolverEventPublisher = observer
+    }
+
+    public func resolve(
+        dependencies: [Dependency: VersionSpecifier],
+        lastResolved: [Dependency: PinnedVersion]? = nil,
+        dependenciesToUpdate: [String]? = nil
```
blender: this argument is not used. What do you think or asserting if not nill?

Source/CarthageKit/DiagnosticResolver.swift
```diff
@@ -0,0 +1,197 @@
...
+        ) -> SignalProducer<[Dependency: PinnedVersion], CarthageError> {
+        let result: Result<[Dependency: PinnedVersion], CarthageError>
+        do {
+            try traverse(dependencies: Array(dependencies))
```
blender: The signature of `traverse` seems a missed opportunity. It could return the successful processed dependencies while instead it's just waiting to fail or otherwise return an empty map.

Overall I'm not a fan of `throw`ing. I would rather use Result.

werner77: I decided to let this class not implement the resolver protocol, since it is not really a resolver. Renamed it to DependencyCrawler instead. It now returns an empty result or a CarthageError. See the updated class.

Source/CarthageKit/DiagnosticResolver.swift
werner77: Removed lastResolved, dependenciesToUpdate. The class is now not an implementation of Resolver anymore.

Source/CarthageKit/Project.swift
werner77: Corrected all indentations

Source/CarthageKit/LocalRepository.swift
werner77: Renamed it to LocalDependencyStore

blender:
LGTM.

@Carthage/carthage  this is a separate command that does not impact current functionality. I consider it safe to merge.

werner77:
Again RemoteVersionSpec failed, which happens randomly about 30% of the time. Is it ok to disable the "should time out" test until it has properly been fixed.

blender:
It would be even better to fix it :D

werner77:
Believe me, I tried. However, this error does not popup when running tests inside Xcode, it may have something todo with the worse performance of the Travis VM. 
Also 'swift test' does not print a stack trace for fatal errors which is fairly annoying (I haven't found a solution for this searching on the web).

A good start would be to get rid of all the occurences of force unwraps in the code (which are 100+),  e.g. by replacing them with a utility unwrap function which at least can print the stack trace in case something goes wrong. I can do this, but this may unnecessarily expand the scope of this Pull Request.

werner77:
Converted the Spec to an ordinary XCTestCase, which seems to have fixed the issue, but it could also be a random success, not sure.

werner77:
Can this Pull Request now be merged, or is there something else to be done still?

blender:
@mdiep if no further objections I would like to merge this PR

mdiep:
Some of my feedback wasn't addressed. I'll leave them as fresh comments.

Tests/CarthageKitTests/Resources/Resolver/ConflictingDependencies/Repository/dependencies/BMBiometrics/0.1.0.json
```diff
@@ -0,0 +1,18 @@
...
+    }
+  },
+  {
+    "versionSpecifier" : "~> 0.13.0",
+    "dependency" : {
+      "git" : {
+        "urlString" : "https:\/\/git.behindmedia.com\/Common\/BMFoundation.git"
+      }
+    }
+  }
```
mdiep: Where are all these fixtures used? Do we really need _all_ of them?

Source/CarthageKit/Cartfile.swift
```diff
@@ -121,6 +120,16 @@ public struct Cartfile {
 	}
 }
 
+extension Cartfile: CustomStringConvertible {
+	public var description: String {
+		return dependencies
+			.sorted { $0.key.description < $1.key.description }
+			.map { "\($0.key) \($0.value)" }
+			.joined(separator: "\n")
+            .appending("\n")
```
mdiep: Indentation is off here

Source/CarthageKit/Cartfile.swift
```diff
@@ -169,16 +178,16 @@ extension ResolvedCartfile: CustomStringConvertible {
 	public var description: String {
 		return dependencies
 			.sorted { $0.key.description < $1.key.description }
-			.map { "\($0.key) \($0.value)\n" }
-			.joined(separator: "")
+			.map { "\($0.key) \($0.value)" }
+			.joined(separator: "\n")
+            .appending("\n")
```
mdiep: Indentation is off here

Source/CarthageKit/Cartfile.swift
```diff
@@ -169,16 +178,16 @@ extension ResolvedCartfile: CustomStringConvertible {
...
 }
 
 extension String {
-
-	/// Returns self without any potential trailing Cartfile comment. A Cartfile
-	/// comment starts with the first `commentIndicator` that is not embedded in any quote
-	var strippingTrailingCartfileComment: String {
+	  /// Returns self without any potential trailing Cartfile comment. A Cartfile
+	  /// comment starts with the first `commentIndicator` that is not embedded in any quote
+	  var strippingTrailingCartfileComment: String {
```
mdiep: Indentation is off here

Source/CarthageKit/DependencyCrawler.swift
```diff
@@ -0,0 +1,226 @@
+import Foundation
+import Result
+import ReactiveSwift
+
+/**
+Signals for DependencyCrawler events
+*/
```
mdiep: This comment doesn't make sense. How are these _signals_?

Also, we should use the `///` comments that we use everywhere else.

Source/CarthageKit/DependencyCrawler.swift
```diff
@@ -0,0 +1,226 @@
...
+public enum DependencyCrawlerEvent {
+	case foundVersions(versions: [PinnedVersion], dependency: Dependency, versionSpecifier: VersionSpecifier)
+	case foundTransitiveDependencies(transitiveDependencies: [(Dependency, VersionSpecifier)], dependency: Dependency, version: PinnedVersion)
+	case failedRetrievingTransitiveDependencies(error: Error, dependency: Dependency, version: PinnedVersion)
+	case failedRetrievingVersions(error: Error, dependency: Dependency, versionSpeficier: VersionSpecifier)
```
mdiep: We shouldn't use `Error` directly—we should use a concrete error type.

mdiep: _specifier_ is misspelled here

Source/CarthageKit/DependencyCrawler.swift
```diff
@@ -0,0 +1,226 @@
...
+
+	If ignoreErrors is true, any error during retrieval of the dependencies will not be fatal but will result in an empty array instead.
+	*/
+	public init(
+			versionsForDependency: @escaping (Dependency) -> SignalProducer<PinnedVersion, CarthageError>,
+			dependenciesForDependency: @escaping (Dependency, PinnedVersion) -> SignalProducer<(Dependency, VersionSpecifier), CarthageError>,
+			resolvedGitReference: @escaping (Dependency, String) -> SignalProducer<PinnedVersion, CarthageError>,
+			store: LocalDependencyStore,
+			mappings: [Dependency: Dependency]? = nil,
+			ignoreErrors: Bool = false) {
```
mdiep: `) {` should be on its own line

Source/CarthageKit/DependencyCrawler.swift
```diff
@@ -0,0 +1,226 @@
...
+			result = .failure(error)
+		} catch {
+			result = .failure(CarthageError.internalError(description: error.localizedDescription))
+		}
+		return result
+	}
+
+	private func traverse(dependencies: [(Dependency, VersionSpecifier)],
+						  handledDependencies: inout Set<PinnedDependency>,
+						  cachedVersionSets: inout [Dependency: [PinnedVersion]]) throws {
```
mdiep: We shouldn't use `throws`. Carthage uses typed errors. This should use `Result` with a concrete error type.

Source/CarthageKit/DependencyCrawler.swift
```diff
@@ -0,0 +1,226 @@
...
+private struct PinnedDependency: Hashable {
+	public let dependency: Dependency
+	public let pinnedVersion: PinnedVersion
+	private let hash: Int
+
+	init(dependency: Dependency, pinnedVersion: PinnedVersion) {
+		self.dependency = dependency
+		self.pinnedVersion = pinnedVersion
+		// Pre-compute hash for efficiency
+		self.hash = 37 &* dependency.hashValue &+ pinnedVersion.hashValue
```
mdiep: I don't think we should be doing this. Did you actually instrument and find this to be a source of slowness?

Source/CarthageKit/LocalDependencyStore.swift
```diff
@@ -0,0 +1,237 @@
+// swiftlint:disable vertical_parameter_alignment_on_call
```
mdiep: We shouldn't be disabling swiftlint rules. They're there for a reason: fix the violations.

Tests/CarthageKitTests/ArchiveSpec.swift
```diff
@@ -7,9 +7,11 @@ import ReactiveSwift
...
 	override func spec() {
 		describe("unzipping") {
-			let archiveURL = Bundle(for: type(of: self)).url(forResource: "CartfilePrivateOnly", withExtension: "zip")!
-
-			it("should unzip archive to a temporary directory") {
+            it("should unzip archive to a temporary directory") {
+                guard let archiveURL = Bundle(for: type(of: self)).url(forResource: "CartfilePrivateOnly", withExtension: "zip") else {
+                    fail("Could not find CartfilePrivateOnly.zip in resources")
+                    return
+                }
```
ikesyo: Wrong indentation (spaces are used).

Tests/CarthageKitTests/CartfileSpec.swift
```diff
@@ -13,13 +13,22 @@ import struct Foundation.URL
 class CartfileSpec: QuickSpec {
 	override func spec() {
 		it("should parse a Cartfile") {
-			let testCartfileURL = Bundle(for: type(of: self)).url(forResource: "TestCartfile", withExtension: "")!
-			let testCartfile = try! String(contentsOf: testCartfileURL, encoding: .utf8)
+            guard let testCartfileURL = Bundle(for: type(of: self)).url(forResource: "TestCartfile", withExtension: "") else {
+                fail("Could not find TestCartFile in resources")
+                return
+            }
```
ikesyo: Wrong indentation (spaces are used).

Tests/CarthageKitTests/MachHeaderSpec.swift
```diff
@@ -11,7 +11,10 @@ class MachHeaderSpec: QuickSpec {
 
 		describe("headers") {
 			it("should list all mach headers for a given Mach-O file") {
-				let directoryURL = Bundle(for: type(of: self)).url(forResource: "Alamofire.framework", withExtension: nil)!
+                guard let directoryURL = Bundle(for: type(of: self)).url(forResource: "Alamofire.framework", withExtension: nil) else {
+                    fail("Could not load Alamofire.framework from resources")
+                    return
+                }
```
ikesyo: Wrong indentation (spaces are used).

Tests/CarthageKitTests/ProjectSpec.swift
```diff
@@ -10,28 +10,47 @@ import XCDBLD
 
 // swiftlint:disable:this force_try
 
+private enum ProjectSpecError: Error {
+    case assertion(message: String)
```
ikesyo: Wrong indentation (spaces are used).

Tests/CarthageKitTests/RemoteVersionSpec.swift
```diff
@@ -12,20 +13,29 @@ import struct Foundation.URL
...
-				expect(version).notTo(beNil())
-				expect(version).toEventually(beNil(), timeout: 0.6)
-			}
-
 			it("should return version") {
-				let release = Release(id: 0, tag: "0.1.0", url: URL(string: "about:blank")!, assets: [])
+                guard let aboutURL = URL(string: "about:blank") else {
+                    fail("Expected aboutURL to not be nil")
+                    return
+                }
```
ikesyo: Wrong indentation (spaces are used).

Tests/CarthageKitTests/ResolverSpec.swift
```diff
@@ -23,7 +23,7 @@ private func equal<A: Equatable, B: Equatable>(_ expectedValue: [(A, B)]?) -> Pr
 			}
 			return PredicateResult(status: .fail, message: message)
 		}
-		return PredicateResult(bool: expectedValue! == actualValue!, message: message)
+        return PredicateResult(bool: expectedValue! == actualValue!, message: message)
```
ikesyo: Wrong indentation (spaces are used).

Source/CarthageKit/DependencyCrawler.swift
werner77: As part of the resolver improvement MR I indeed found that hash calculation is one of the bottlenecks for the performance issues.

werner77:
Thanks for the review, I'm trying to incorporate it. With regards to the throws/Result discussion: I'm trying to convert that, but it results in rather obscure code when dealing with loops/recursive functions. Throws automatically goes up the call chain and deals with fail fast and handling the error in one block (the upper catch block), while using Result requires multiple ifs (or guards) littered throughout the functions which results in rather unreadable code IMHO.

Is it good enough to only use Result for the public functions so the public contract specifies a CarthageError?

Source/CarthageKit/Cartfile.swift
werner77: Fixed

Source/CarthageKit/Cartfile.swift
werner77: Fixed

Source/CarthageKit/DependencyCrawler.swift
werner77: Updated doc, and now use /// everywhere

Source/CarthageKit/DependencyCrawler.swift
werner77: Now uses CarthageError

werner77: Fixed typo

Source/CarthageKit/DependencyCrawler.swift
werner77: Fixed

Source/CarthageKit/DependencyCrawler.swift
werner77: Removed it to rely on auto-conformance of Hashable. For this particular case performance is not that critical.

Source/CarthageKit/LocalDependencyStore.swift
werner77: Removed the swiftlint rule disable. However, are you aware that there are many instances of this throughout the code base?

Tests/CarthageKitTests/ArchiveSpec.swift
werner77: Fixed

Tests/CarthageKitTests/CartfileSpec.swift
werner77: Fixed

Tests/CarthageKitTests/MachHeaderSpec.swift
werner77: Fixed

Tests/CarthageKitTests/RemoteVersionSpec.swift
werner77: Fixed

Source/CarthageKit/DependencyCrawler.swift
werner77: Is it ok to keep the throws for private methods since it results in a more complicated flow when using Result (multiple exit points that I need to guard against, which require the same handler). These methods are not seen or used from the outside. I did change all public signatures (also in LocalDependencyStore) to use result instead of throws.

Tests/CarthageKitTests/Resources/Resolver/ConflictingDependencies/Repository/dependencies/BMBiometrics/0.1.0.json
werner77: All these fixtures are used in the pending test cases in ResolverSpec. They are direct results of using the diagnose command for concrete problematic dependency trees.

mdiep: Can we please leave any pending test cases and related fixtures for a separate PR?

Source/CarthageKit/DependencyCrawler.swift
mdiep: No, we'd like to use `Result` everywhere for the additional type safety.

werner77:
Just to let you know: we have forked the entire project within a new open source organization (https://github.com/nsoperations) to allow us to move forward more rapidly than is currently the case. High on our list of improvements to fix are:

1. Fixing the resolver
2. Fixing concurrency issues
3. Fixing caching issues
4. Implement support for different caching back-ends (e.g. Artifactory)

This and the resolver PR have been open for a year already, which just doesn't cut it for us, we need to move forward with this.

PR #2523, opened 2018-07-20
Proposal for (and implementation of) a "carthage validate" command

jasonboyle:
This PR implements a "carthage ~verify~ validate" command which inspects the Cartfile of each dependency at the version specified in the current project's Cartfile.resolved in order to determine whether the resolved versions of all dependencies are compatible.

Using "carthage update" to update Cartfile.resolved should always result in a valid Cartfile.resolved.  However, there are some situations where it may not be desirable or possible to run "carthage update":

* "carthage update" will occasionally unnecessarily bump the versions of other dependencies even when run with a specific dependency as an argument, e.g. "carthage update Framework1".  This may not be appropriate in a release branch that has been tested with a specific versions of those dependencies.
* "carthage update" sometimes hangs when provided with a specific dependency as an argument.  As a result, the only way to reliably update Cartfile.resolved is to run "carthage update" with no arguments which may bump the versions of additional dependencies.  I believe this is a deficiency of the current resolver implementation.
* There is currently no easy way to verify that a modification of Cartfile.resolved was the result of running "carthage update" rather than a developer editing the file by hand, either by accident or due to a lack of Carthage knowledge.  When this happens it can be difficult to determine which dependencies are incompatible in order to restore Cartfile.resolved to a valid state.

The "carthage ~verify~ validate" command proposed in this PR could be used to validate Cartfile.resolved in a continuous integration pipeline, or after editing the file by hand due to one of the reasons listed above.  This command could also be used to verify the correctness of a resolver implementation.


Source/CarthageKit/Project.swift
```diff
@@ -1103,6 +1133,64 @@ public final class Project { // swiftlint:disable:this type_body_length
...
 			}
 	}
+
+	/// Returns an array of CompatibilityInfo representing the pinned version
+	/// of each dependency in Cartfile.resolved which is incompatible with
+	/// one or more of the versions specified in a transitive dependency's
+	/// Cartfile.  The incompatible versions are stored in the
+	/// dependencyVersions property of the CompatibilityInfo object.
+	public func verify(resolvedCartfile: ResolvedCartfile? = nil) -> SignalProducer<(), CarthageError> {
+		let resolvedProducer = (resolvedCartfile != nil) ? SignalProducer(value: resolvedCartfile!) : loadResolvedCartfile()
```
blender: I'm not so keen on this loading the Cartfile.resolved by itself. I would not expect this from the signature.

blender:
Should #2519 and this PR live under the same CLI command  like:

carthage diagnose resolved
carthage diagnose dump-deps

mdiep:
> Should #2519 and this PR live under the same CLI command like:

I think these are separate things.

* #2519 helps diagnose problems with Carthage
* #2523 (this PR) would catch user errors, not Carthage bugs.

Source/CarthageKit/Errors.swift
```diff
@@ -87,6 +87,9 @@ public enum CarthageError: Error {
 
 	/// An internal error occurred
 	case internalError(description: String)
+
+	/// Cartfile.resolved contains incompatible versions
+	case verifyFailed([CompatibilityInfo])
```
mdiep: I think this could use a better name. Maybe `invalidResolvedCartfile`?

Source/CarthageKit/CompatibilityInfo.swift
```diff
@@ -0,0 +1,24 @@
...
+/// Identifies a dependency, its pinned version, and versions of this dependency with which it may or may not be compatible
+public struct CompatibilityInfo: Equatable {
+	/// The dependecy
+	public let dependency: Dependency
+
+	/// The pinned version of the dependency
+	public let pinnedVersion: PinnedVersion
+	
+	/// Versions of this dependency with which it may or may not be compatible
+	public let dependencyVersions: [(Dependency, VersionSpecifier)]
```
mdiep: I think this property could have a better name. It seems like this is actually the requirements for this dependency. And it probably makes more sense as a dictionary.

```swift
/// The projects that list this dependency as a requirement and the version specifiers they use.
public let requirements: [Dependency: VersionSpecifier]
```

Source/CarthageKit/CompatibilityInfo.swift
```diff
@@ -0,0 +1,24 @@
...
+	/// The versions which are not compatible with the pinned version of this dependency
+	public var incompatibleVersions: [(Dependency, VersionSpecifier)] {
+		return dependencyVersions.filter { _, version in !version.isSatisfied(by: pinnedVersion) }
+	}
+
+	public static func == (lhs: CompatibilityInfo, rhs: CompatibilityInfo) -> Bool {
+		return lhs.dependency == rhs.dependency &&
+			lhs.pinnedVersion == rhs.pinnedVersion &&
+			lhs.dependencyVersions.elementsEqual(rhs.dependencyVersions, by: { $0.0 == $1.0 && $0.1 == $1.1 })
+	}
```
mdiep: By using a dictionary above, we should be able to use a synthesized `==` here.

Source/CarthageKit/Errors.swift
```diff
@@ -334,6 +337,17 @@ extension CarthageError: CustomStringConvertible {
...
+		case let .verifyFailed(incompatibilities):
+			var message = "The following incompatibilities were found in Cartfile.resolved:\n"
+			var lines: [String] = []
+			incompatibilities.forEach { incompatibility in
+				for (dependency, version) in incompatibility.dependencyVersions {
+					lines.append("* \(incompatibility.dependency.name) is incompatible with \(dependency.name) \(version)")
+				}
+			}
+			message += lines.joined(separator: "\n")
+			return message
```
mdiep: I'd like to see a test for this formatting.

Source/carthage/Verify.swift
```diff
@@ -0,0 +1,21 @@
+import CarthageKit
+import Commandant
+import Foundation
+import ReactiveSwift
+import Result
+
+/// Type that encapsulates the configuration and evaluation of the `verify` subcommand.
+public struct VerifyCommand: CommandProtocol {
+	public let verb = "verify"
+	public let function = "Verify that the dependencies in Cartfile.resolved are compatible"
```
mdiep: This could be clearer.

> Verify that the versions in Cartfile.resolved are compatible with the Cartfile requirements

Source/CarthageKit/Project.swift
mdiep: I agree. I think we should always pass in the resolved cartfile instead of loading it here.

Source/CarthageKit/Project.swift
```diff
@@ -371,6 +371,36 @@ public final class Project { // swiftlint:disable:this type_body_length
 					}
 					.map { $0.0.name }
 					.collect()
+		}
+	}
+
+	/// Finds all the transitive dependencies for the current project, grouped by their parent dependency.
```
mdiep: Reading this and looking at the return value, I'm not sure which `Dependency` is which. I think it deserves some clarification.

mdiep:
Also, I think this is a great improvement! Thanks for opening a PR!

ikesyo:
It seems that `carthage verify` is confusing with https://github.com/Carthage/workflows/blob/master/README.md#carthage-verify.

Source/CarthageKit/CompatibilityInfo.swift
ikesyo: The operator still should be defined by hand for Xcode 9.2.

jasonboyle:
@mdiep Thank you!

I made the changes you and @blender requested.  Please take another look when you have the chance.  Thanks!

chuganzy:
OK, I originally thought this command has the same purpose with https://github.com/Carthage/workflows/blob/master/README.md#carthage-verify but not.. Hmm, to me, the command name is confusing..

jasonboyle:
If `verify` is the best name for this command, it might make sense to rename the `carthage-verify` workflow.  Other naming possibilities I can think of include `check-compatibility` and `validate`.

A related question is whether this should be its own command or a flag on the update/bootstrap commands such as `--check-compatibility`.  In the future it might make sense to always print an incompatibility warning when these commands are run even if the flag is off.

blender:
Renaming the workflow is fine with me, although I am not sure we have metrics on how many people will be affected

mdiep:
Maybe `validate` would be a better name?

> The "carthage verify" command proposed in this PR could be used to validate Cartfile.resolved in a continuous integration pipeline

_Validate_ seems to be more about _validity_ and _verify_ seems to be more about _accuracy_. Since I think we're after _validity_ and not _accuracy_ (how can we know what versions you intended?), I think we want _validate_.

I think this makes sense as a separate command. I'd suggest starting with that. If we want, we can add a bootstrap flag also.

jasonboyle:
Makes sense, thanks!  I changed the name of the command to "validate".

Source/CarthageKit/Project.swift
```diff
@@ -371,7 +371,43 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+
+	/// Finds the required dependencies and their corresponding version specifiers for each dependency in Cartfile.resolved.
+	func requirementsByDependency(
+		resolvedCartfile: ResolvedCartfile,
+		tryCheckoutDirectory: Bool
+	) -> SignalProducer<[Dependency: [Dependency: VersionSpecifier]], CarthageError> {
+		return SignalProducer(value: resolvedCartfile)
+			.map { resolvedCartfile -> [(Dependency, PinnedVersion)] in
+				return resolvedCartfile.dependencies
+					.map { k, v in (k, v) }
```
mdiep: I think this could be written `return Array(resolvedCartfile.dependencies)`

Source/CarthageKit/Errors.swift
```diff
@@ -334,6 +337,19 @@ extension CarthageError: CustomStringConvertible {
...
+			var lines: [String] = []
+			let sortedIncompatibilities = incompatibilities.sorted(by: { $0.dependency.name < $1.dependency.name })
+			sortedIncompatibilities.forEach { incompatibility in
+				let sortedRequirements = incompatibility.incompatibleRequirements.sorted(by: { $0.0.name < $1.0.name })
+				for (dependency, version) in sortedRequirements {
+					lines.append("* \(incompatibility.dependency.name) \(incompatibility.pinnedVersion) is incompatible with \(dependency.name) \(version)")
+				}
+			}
+			message += lines.joined(separator: "\n")
+			return message
```
mdiep: I think this could be written more clearly in a functional style:

```swift
let lines = incompatibilities
  .sorted { $0.dependency.name < $1.dependency.name }
  .flatMap { incompatibility in
    let sortedRequirements = incompatibility
      .incompatibleRequirements
      .sorted { $0.0.name < $1.0.name } }
    return sortedRequirements.map { dependency, version in
      return "* \(incompatibility.dependency.name) \(incompatibility.pinnedVersion) is incompatible with \(dependency.name) \(version)"
    }
  }
  .joined(separator: "\n")
```

Source/CarthageKit/Project.swift
```diff
@@ -371,7 +371,43 @@ public final class Project { // swiftlint:disable:this type_body_length
...
 			}
+			.flatMap(.merge) { dependencies -> SignalProducer<[Dependency: [Dependency: VersionSpecifier]], CarthageError> in
+				return SignalProducer(dependencies)
+					.flatMap(.merge) { dependencyAndPinnedVersion -> SignalProducer<(Dependency, (Dependency, VersionSpecifier)), CarthageError> in
```
mdiep: We shouldn't use the `merge` strategy here. If you have 50 dependencies, this would try to process all 50 at once. Since this is reading from disk, that could result in thrashing. Elsewhere we use `.concurrent(4)`; I think that'd be appropriate here.

Source/CarthageKit/Project.swift
```diff
@@ -1109,6 +1145,60 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+						dict[requiredDependency] = requirements
+					}
+				}
+				return SignalProducer(value: dict)
+			}
+
+		return resolvedCartfileProducer.map { (resolved: ResolvedCartfile) -> [Dependency: PinnedVersion] in
+				return resolved.dependencies
+			}
+			.combineLatest(with: dependencyRequirements)
```
mdiep: `.zip(with:)` would better communicate the intent here.

Source/CarthageKit/Project.swift
```diff
@@ -1109,6 +1145,60 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+			.map { (dependencyInfo: ([Dependency: PinnedVersion], [Dependency: [Dependency: VersionSpecifier]])) -> [CompatibilityInfo] in
+				let (resolved, requirements) = dependencyInfo
+				var result: [CompatibilityInfo] = []
+				resolved.forEach { dependency, pinnedVersion in
+					// Skip non-semantic resolved versions and transitive dependencies without semantic versions
+					if case .success = SemanticVersion.from(pinnedVersion), let requirements = requirements[dependency] {
+						result.append(CompatibilityInfo(dependency: dependency, pinnedVersion: pinnedVersion, requirements: requirements))
+					}
+				}
+				return result
```
mdiep: It'd like to see this logic and the dictionary inversion above move into its own method that can be tested. Since it's going to create a collection of `CompatibiltyInfo`s, it probably makes sense as a `static func` on that type.

Source/CarthageKit/Project.swift
```diff
@@ -1109,6 +1145,60 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+			}
+			.flatMap(.merge) { (compatibilityInfos: [CompatibilityInfo]) -> SignalProducer<(), CarthageError> in
+				let incompatibleInfos: [CompatibilityInfo] = compatibilityInfos.filter { compatibilityInfo in
+					return !compatibilityInfo.incompatibleRequirements.isEmpty
+				}
+				return incompatibleInfos.isEmpty ? SignalProducer(value: ()) : SignalProducer(error: .invalidResolvedCartfile(incompatibleInfos))
+			}
```
mdiep: I think this would be better expressed if the `filter` was part of the signal chain.

You can do that by turning the `map` above this into a `flatMap`.

```swift
.flatMap(.concat) { dependencyInfo -> [CompatibilityInfo] in … }
.filter { !$0.incompatibleRequirements.isEmpty }
.collect()
.flatMap { incompatible in
  return incompatible.isEmpty ? SignalProducer(value: ()) : SignalProducer(error: …)
}
```

jasonboyle:
@mdiep Done!

mdiep:
Sorry, I've been very distracted by some things in my personal life. I'll try to review this PR this week.

Source/CarthageKit/Project.swift
```diff
@@ -371,7 +371,42 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+	}
+
+	/// Finds the required dependencies and their corresponding version specifiers for each dependency in Cartfile.resolved.
+	func requirementsByDependency(
+		resolvedCartfile: ResolvedCartfile,
+		tryCheckoutDirectory: Bool
+	) -> SignalProducer<CompatibilityInfo.Requirements, CarthageError> {
+		return SignalProducer(value: resolvedCartfile)
+			.map { resolvedCartfile -> [(Dependency, PinnedVersion)] in
+				return Array(resolvedCartfile.dependencies)
```
mdiep: Instead of `map` and `flatMap`, this can create a `SignalProducer` directly:

```swift
return SignalProducer(resolvedCartfile.dependencies)
  .flatMap(.concurrent(limit: 4)){ dependencyAndPinnedVersion ->
```

This is important because we have the `concurrent` in the wrong spot right now. (That's probably my fault. Sorry!)

Source/CarthageKit/Project.swift
```diff
@@ -1109,6 +1144,24 @@ public final class Project { // swiftlint:disable:this type_body_length
...
 	}
+
+	/// Determines whether the requirements specified in this project's Cartfile.resolved
+	/// are compatible with the versions specified in the Cartfile for each of those projects.
+	///
+	/// Either emits a value to indicate success or an error.
+	public func validate(resolvedCartfile: ResolvedCartfile) -> SignalProducer<(), CarthageError> {
+		return SignalProducer(value: resolvedCartfile)
+			.flatMap(.concat) { (resolved: ResolvedCartfile) -> SignalProducer<([Dependency: PinnedVersion], CompatibilityInfo.Requirements), CarthageError> in
+				return SignalProducer(value: resolved.dependencies).zip(with: self.requirementsByDependency(resolvedCartfile: resolved, tryCheckoutDirectory: true))
```
mdiep: This would be clearer if it were written with the `static func` version of `zip`.

```swift
return SignalProducer.zip(
  SignalProducer(value: resolved.dependencies),
  self.requirementsByDependency(…)
)
```

mdiep:
Sorry again that I've been so slow on this. Once that one issue is fixed, this will be good to merge. 👍

Tests/CarthageKitTests/ValidateSpec.swift
```diff
@@ -0,0 +1,169 @@
...
+		// These tuples represent the desired version of a dependency, paired with its parent dependency;
+		// moya_3_1_0 indicates that Moya expects a version compatible with 3.1.0 of *another* dependency
+		let moya_3_1_0 = (moyaDependency, VersionSpecifier.compatibleWith(SemanticVersion(major: 3, minor: 1, patch: 0)))
+		let moya_4_1_0 = (moyaDependency, VersionSpecifier.compatibleWith(SemanticVersion(major: 4, minor: 1, patch: 0)))
+		let reactiveSwift_3_2_1 = (reactiveSwiftDependency, VersionSpecifier.compatibleWith(SemanticVersion(major: 3, minor: 2, patch: 1)))
+
+		describe("requirementsByDependency") {
+			it("should group dependencies by parent dependency") {
+				let resolvedCartfile = ResolvedCartfile.from(string: validCartfile)
+				let project = Project(directoryURL: URL(string: "file://fake")!)
```
jdhealy: There’s a directory — `/var/empty` — which (by implementation) cannot contain real files; would look like `URL(string: "file:///var/empty/fake")!`. Let’s use that.

jasonboyle:
@mdiep @jdhealy Thanks for the reviews! I've made the requested changes.

mdiep:
Thank you for all your work and patience on this!

jasonboyle:
No problem!  Thanks again for reviewing!  🎉

PR #2438, opened 2018-05-02
Add --archive option to the build command.

sunshinejr:
Fixes #1411. 

To be honest I just finished the works by @3lvis so credits to him for starting that one 😉

PR #2524, opened 2018-07-20
Improve instructions for prebuilding frameworks

3lvis:
[0.30.1](https://github.com/Carthage/Carthage/releases/tag/0.30.1) introduces a simpler way of archiving. The command was introduced in https://github.com/Carthage/Carthage/pull/2438 and it works great!

```
carthage build --archive
```


blender:
I would rather have:

 ```sh
 carthage build --no-skip-current
 carthage archive YourFrameworkName
 ```

or simply

``sh
carthage build --archive
```

blender:
Great! Thanks for grooming the docs!

PR #2542, opened 2018-07-30
Use `Bundle.executableURL` instead of `object(forInfoDictionaryKey: "CFBundleExecutable")`

### LINKED ISSUE #2491: Carthage support for a prebuilt .framework (built with static library + aggregator)
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `carthage version`: 0.29.0
* `xcodebuild -version`: Xcode 9.4.1, Build version 9F2000
* Are you using `--cache-builds`? no

**Cartfile**
```
binary "./<framework-name>.json" == 4.1.3
```

**Carthage Output**
```
$ carthage update
*** Downloading binary-only framework <framework-name> at "file:///Users/Daniel/.../<framework-name>.json"
*** Downloading binary-only framework <framework-name> at "file:///Users/Daniel/.../<framework-name>.json"
*** xcodebuild output can be found in /var/folders/3w/g8q6qv_93k15g87227vqwhbc0000gn/T/carthage-xcodebuild.KT8zF1.log
Failed to read file or folder at /Users/Daniel/.../Carthage/Build/iOS/<framework-name>.framework
```

**Background**
We are trying to add Carthage support to a closed source / private iOS SDK. We build the SDK / .framework from a static library target with an aggregator target, which means that there is no dynamic framework target in the project -> 
```
carthage build --no-skip-current
carthage archive YourFrameworkName
```
fails (since there is no shared framework). We also do not use a .plist file anywhere in the app (in reference to https://github.com/Carthage/Carthage/issues/1077).
Because of this, and because we already have the `<framework-name>.framework` file, I tried creating a binary only release (https://github.com/Carthage/Carthage/pull/1760/).

Since there is no shared framework target, I need to skip `carthage build --no-skip-current`, as this leads to this error message:
```
*** Skipped building <framework-name> due to the error:
Dependency "<framework-name>" has no shared framework schemes
```
So I mocked the Carthage structure by putting the pre-built .framework to `<project-root>/Carthage/Build/iOS` and calling `carthage archive <framework-name>`. This gives me the following output:
```
*** Found Carthage/Build/iOS/<framework-name>.framework
*** Found Carthage/Build/iOS/<framework-name>.framework.dSYM
Failed to read file or folder at /Users/Daniel/.../Carthage/Build/iOS/<framework-name>.framework
```
Archiving it manually into a .zip file that only contains `Carthage/Build/iOS/<framework-name>.framework` and trying to get it via `carthage update` leads to the error message in the "Carthage Output" point above.

**Actual outcome**
On `carthage update`, Carthage downloads the (closed source / private) binary only release / .framework file. It copies it to `Carthage/Buidls/iOS` but then fails with the above error message.

**Expected outcome**
Is there a way to offer Carthage support with the described project setup (i.e. skipping `carthage build --no-skip-current` and `carthage archive YourFrameworkName`)? 

(we are btw. looking into adding a new framework target to the project so we can go the reqular `build` and `archive` way, but here we're running into the problem of `carthage build --no-skip-current` not including the (public) header files to the .framework - this might very well be a problem on our side, but: is this a known issue?)

ikesyo:
This would fix #2491 (should work for frameworks who does not contain the `Info.plist` file).

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -1188,21 +1188,16 @@ private func UUIDsFromDwarfdump(_ url: URL) -> SignalProducer<Set<UUID>, Carthag
 public func binaryURL(_ packageURL: URL) -> Result<URL, CarthageError> {
 	let bundle = Bundle(path: packageURL.path)
 
-	switch bundle?.packageType {
-	case .framework?, .bundle?:
-		if let binaryName = bundle?.object(forInfoDictionaryKey: "CFBundleExecutable") as? String {
-			return .success(packageURL.appendingPathComponent(binaryName))
-		}
+	if let executableURL = bundle?.executableURL {
+		return .success(executableURL)
+	}
 
-	case .dSYM?:
+	if bundle?.packageType == .dSYM {
 		let binaryName = packageURL.deletingPathExtension().deletingPathExtension().lastPathComponent
 		if !binaryName.isEmpty {
 			let binaryURL = packageURL.appendingPathComponent("Contents/Resources/DWARF/\(binaryName)")
 			return .success(binaryURL)
 		}
-
-	default:
-		break
 	}
 
 	return .failure(.readFailed(packageURL, nil))


chezzdev:
Thanks for fixing this. Is there a plan to make a release to include this fix?

Issue #2544, opened 2018-07-31
Timeout for version check does not work


Sebastian-Hojas:
* carthage install method: homebrew
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.30.1

When invoking any carthage command with a bad network connection (also simulated), the version check blocks operations without respecting the timeout of the version check request.

**Carthage Output**
```bash
# Invoking carthage with a bad network connection
> time carthage help
carthage help  0.06s user 0.05s system 0% cpu 19.325 total
```

Carthage took 19 seconds to complete `carthage help`.

**Expected outcome**

- Carthage correctly times out version check request after 0.5s
- I was also hoping for an invocation flag that would bypass the version check for CI builds. For a build process that would invocate carthage multiple times, even a timeout of 0.5s per call slows our build process down unnecessarily.

**Investigation**

The function `remoteVersion` in `BuildVersion.swift` gets the latest remote version with a timeout:
```swift
.timeout(after: 0.5, raising: CarthageError.gitHubAPITimeout, on: QueueScheduler.main)
```
This timeout doesn't seem to work.

blender:
Do you want to give this a shot?

blender:
And thanks to OP this is closed. 👏

PR #2546, opened 2018-07-31
Use different queue for version check timeouts

Sebastian-Hojas:
Using `first()` on the same queue causes the timeout
event to never be executed.

See https://github.com/Carthage/Carthage/issues/2544


blender:
Wonderful! Thanks for the contribution! 

Do you think we can add a quick test for this?

Sebastian-Hojas:
In order to make `remoteVersion()` testable, I assume we would need to mock the http request. Can you pinpoint me to some tests where you have already done this? Or did you have some other test strategy in mind?

blender:
There are ideas of course, if you have better ones feel free to suggest.  I am also fine without tests for something till small but it would be a nice to have.

Sebastian-Hojas:
Yeah, I can definitely see the value in this and would support said change. I can also think a bit about it, but I would suggest to not piggy-back the introduction of a mocking web server for tests on this tiny PR in particular. :)

blender:
Maybe introduce another "firable" event that will never get fired in tests?

blender:
A mock `Repository(owner: "Carthage", name: "Carthage").releases, perPage: 2)` ?

Sebastian-Hojas:
You can't mock `Client` unfortunately as it's `final`, but we could mock the producer itself. Something like `public func remoteVersion(_ remoteVersionProducer: SignalProducer<_, Client.Error>) -> SemanticVersion? `

As far as I understand that, we would also need to move the public function from `BuildVersion.swift` to `CarthageKit`. Is my assumption correct?
I am wondering whether that semantically makes sense.

Sebastian-Hojas:
Renamed to `RemoteVersion`, moved mockable `remoteVersion` to CarthageKit and added simple tests. Waiting for feedback now.

Source/CarthageKit/RemoteVersion.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+import ReactiveSwift
+import ReactiveTask
+import Result
+import Tentacle
+
+/// Synchronously returns the newest Carthage release version,
```
blender: Is the doc mentioning "Carthage" still relevant?

Source/carthage/RemoteVersion.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+import CarthageKit
+import ReactiveSwift
+import ReactiveTask
+import Result
+import Tentacle
+
+/// The latest online version as a SemanticVersion object.
+public func remoteVersion() -> SemanticVersion? {
+	let remoteVersionProducer = Client(.dotCom, urlSession: URLSession.proxiedSession)
```
blender: awesome

Tests/CarthageKitTests/RemoteVersionSpec.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+import Nimble
+import Quick
+import ReactiveSwift
+import Tentacle
+
+@testable import CarthageKit
+
+class RemoteVersionSpec: QuickSpec {
```
blender: great! Now it shall never break again!

blender:
Just one question on the docs, then this is good to merge

Source/CarthageKit/RemoteVersion.swift
Sebastian-Hojas: Yeah, very good point. I updated the docs.

Sebastian-Hojas:
Everything should be resolved now. Thanks for the feedback and the guidance - I am happy we got that issue resolved so easily!

blender:
Awesome. Well done! Thanks for the contribution!

PR #2547, opened 2018-07-31
Add option to skip remote version check

Sebastian-Hojas:
I strongly believe that we should allow carthage to operate without synchronously making a blocking request for the version check. A timeout of 0.5 seconds sounds harmless, but can be really annoying when carthage is used during the build process multiple times. I can see the strong case to keep the check, but would at least want to offer users an option (e.g `--skip-remote-version-check`) to skip it for scenarios where it really doesn't make any sense.

See issue https://github.com/Carthage/Carthage/issues/2544 and related PR https://github.com/Carthage/Carthage/pull/2546

One problem I am facing at the moment is to introduce the idea of `GlobalOptions` which didn't exist before. I want to discuss the general approach before suggesting a more rigid solution.

### DIFF
--- a/Source/carthage/GlobalArguments.swift
+++ b/Source/carthage/GlobalArguments.swift
@@ -1,0 +1,31 @@
+import Foundation
+import Curry
+import Commandant
+import Result
+
+struct GlobalOptions: OptionsProtocol {
+
+	let skipRemoteVersionCheck: Bool
+
+	typealias ClientError = NoError
+
+	static func evaluate(_ m: CommandMode) -> Result<GlobalOptions, CommandantError<NoError>> {
+		return curry(self.init)
+			<*> m <| Option(key: "skip-remote-version-check", defaultValue: false, usage: "avoid checking for version remotely")
+	}
+
+	static func consume(globalArguments: inout [String]) -> GlobalOptions {
+		let options = GlobalOptions.evaluate(.arguments(ArgumentParser(globalArguments)))
+		globalArguments.pop { $0 == "--skip-remote-version-check" }
+		return options.value!
+	}
+}
+
+extension Array {
+
+	@discardableResult
+	mutating func pop(where predicate: (Element) -> Bool) -> Element? {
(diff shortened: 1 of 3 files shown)


blender:
I am ok with the proposal, (does not mean go head, let's wait for @mdiep ). However 0.5s is the max you're going to wait. Hopefully it won't take that long to get a response.

Source/carthage/GlobalArguments.swift
```diff
@@ -0,0 +1,31 @@
...
+	let skipRemoteVersionCheck: Bool
+
+	typealias ClientError = NoError
+
+	static func evaluate(_ m: CommandMode) -> Result<GlobalOptions, CommandantError<NoError>> {
+		return curry(self.init)
+			<*> m <| Option(key: "skip-remote-version-check", defaultValue: false, usage: "avoid checking for version remotely")
+	}
+
+	static func consume(globalArguments: inout [String]) -> GlobalOptions {
```
Sebastian-Hojas: suggest to rename to `GlobalOptions.options(consuming globalArguments:)`

Source/carthage/GlobalArguments.swift
```diff
@@ -0,0 +1,31 @@
...
+		let options = GlobalOptions.evaluate(.arguments(ArgumentParser(globalArguments)))
+		globalArguments.pop { $0 == "--skip-remote-version-check" }
+		return options.value!
+	}
+}
+
+extension Array {
+
+	@discardableResult
+	mutating func pop(where predicate: (Element) -> Bool) -> Element? {
```
Sebastian-Hojas: suggest to rename to `remove(where:)`

mdiep:
I don't see a problem with adding this feature if the implementation is reasonable.

Does this work as-is if you add do `carthage bootstrap --skip-remote-version-check --use-submodules`? (Handling a mix of global and non-global options)

blender:
@Sebastian-Hojas ping

PR #2551, opened 2018-08-09
Symlink the Build folder for binary downloads

jasonboyle:
Carthage creates a symlink to the main Build folder from the Carthage directory of each dependency.  However, these symlinks are created in the process of building each dependency, and as a result they are not created for binary downloads (which are not built by Carthage).  This PR fixes this inconsistency.

I am not sure if this change will be useful for normal Carthage users; however, it would be useful for my team as our project has two workspaces, one of which includes the contents of Carthage/Checkout as sub-projects.  Xcode needs all of the Build folders to be symlinked in order to build the project.

I would appreciate advice as to how this change could be tested.  As far as I can tell there are currently no tests that verify the Build symlinks are created, nor tests that check out a binary framework (although I might not be looking in the right places).

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -1049,6 +1049,10 @@ public final class Project { // swiftlint:disable:this type_body_length
 								.then(.init(value: (dependency, version)))
 						}
 					}
+					.flatMap(.merge) { dependency, version -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
+						return self.symlinkBuildPathIfNeeded(for: dependency, version: version)
+							.then(.init(value: (dependency, version)))
+					}
 					.collect()
 					.map { installedDependencies -> [(Dependency, PinnedVersion)] in
 						// Filters out dependencies that we've downloaded binaries for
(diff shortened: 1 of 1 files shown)


mdiep:
Can you share more about why you need this? If you have a pre-built binary for dependency _X_, then why do you need it to have the build symlinks? It's already built.

Are other dependencies that depend on _X_ looking at its build directory? Because it has transitive dependencies that it doesn't explicitly list?

jasonboyle:
The project I work on has two Xcode workspaces.  The first workspace is set up as an ordinary Carthage project and uses the frameworks built by Carthage.  We use this project for CI and release.

We have a second workspace which references our (in-house) dependencies as sub-projects located in the Carthage/Checkouts directory.  This workspace builds these dependencies as part of the Xcode build process instead of using the pre-built Carthage frameworks.  This allows us to replace one or more of these sub-projects with a symlink to the corresponding project's repository so that we can make changes in multiple repositories from within the same Xcode workspace.

The sub-projects in the Checkouts directory (which are not replaced with a symlink to a repo) need their Carthage Build directory to be symlinked to the main project's Build directory so they can find frameworks built by Carthage for dependencies which are not sub-projects of our Xcode workspace, e.g. not-in-house dependencies.

We currently use a script to set up these missing Build directory symlinks and can continue to do so but this wasn't necessary before we started using binary downloads.

As I see it this is mainly a consistency issue.  Carthage does symlink the contents of the Checkouts directory for binary downloads (as it is part of the checkout process), but does not symlink the Build directory for binary downloads (as it is part of the build process).

mdiep:
> The sub-projects in the Checkouts directory (which are not replaced with a symlink to a repo) need their Carthage Build directory to be symlinked to the main project's Build directory so they can find frameworks built by Carthage for dependencies which are not sub-projects of our Xcode workspace, e.g. not-in-house dependencies.

Do your sub-projects in the Checkouts directory have `Cartfile`s that list the binary dependencies that they depend on?

jasonboyle:
Yes, all of the sub-projects' Cartfiles list their dependencies.  We have CI set up in every sub-project's repo to ensure that each can be built individually.

jasonboyle:
Well, the checkouts for non-binary dependencies already have a symlink to the Build folder.  I would like to symlink the checkouts for binary dependencies as well because our secondary workspace builds them itself.  So for example, `Carthage/Checkouts/BinaryDependency/Carthage/Build` would be a symlink to `Carthage/Checkouts/Build`.  Normally this symlink would be created as part of a Carthage build, but since Carthage doesn't need to build binary dependencies, it is never created.

I think it makes sense for Carthage to leave the checkouts of both binary and non-binary dependencies in the same state after bootstrapping.  That might mean removing the Build folder symlinks for non-binary dependencies instead (although that wouldn't be helpful for me, it would be consistent).

mdiep:
Thank you for your patience on this one! This is very confusing to talk about. But what you have here makes sense. 👍

bwhiteley:
Thanks! We need this enhancement also for the same reasons @jasonboyle outlines. Now I can delete a task to add this functionality to our Carthage helper tool. :)

Issue #2536, opened 2018-07-25
Amphorae - The Carthage Compatibility Tests Project


blender:
In order to catch compatibility issues faster, I have set up a repository at https://github.com/blender/Amphorae to run bats tests that invoke Carthage.

The repo, if accepted, is to be moved  later under the Carthage Org.

The idea is the following: 
- pull the first 10 pages from https://ios.libhunt.com/projects (or other source) for Carthage compatible project
- write [bats test](https://github.com/bats-core/bats-core) for `carthage update --no-use-binaries --cache-builds`
- run tests in parallel as much as possible
- run tests nightly (https://docs.travis-ci.com/user/cron-jobs/)

Here is a screenshot of what I have achieved so far.
![](https://raw.githubusercontent.com/blender/Amphorae/master/Resources/Images/travis-screenshot.png)

Since many have expressed the desire to help, expanding the project is a good start. The project can't fly without everyone's help.

Comments and discussion about this effort are welcome. Consider everything up for debate and needing improvement.

Issue #2433, opened 2018-04-25
Help description for no-use-binaries


yhkaplan:
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `which carthage`: `/usr/local/bin/carthage`
* `carthage version`: `Carthage 0.29.0`

When looking at `carthage help update`, the description for `--no-use-binaries` says `use downloaded binaries when possible`, but I believe the correct explanation would be the opposite: 
**don't** use downloaded binaries when possible

Please confirm if my understanding is correct. If so, I'll make a PR 😉

blender:
Correct

mdiep:
It's specified that way because `--use-binaries` is the option, but it can be inverted to `--no-use-binaries`

PR #2556, opened 2018-08-15
Updated documentation of the --no-use-binaries option

sgl0v:
The description of the `-no-use-binaries option` doesn't look correct:

```
[--no-use-binaries]
	use downloaded binaries when possible
```

Updated with `don't use downloaded binaries when possible`.

### DIFF
--- a/Source/carthage/Build.swift
+++ b/Source/carthage/Build.swift
@@ -22,7 +22,7 @@ extension BuildOptions: OptionsProtocol {
 			<*> mode <| Option<String?>(key: "toolchain", defaultValue: nil, usage: "the toolchain to build with")
 			<*> mode <| Option<String?>(key: "derived-data", defaultValue: nil, usage: "path to the custom derived data folder")
 			<*> mode <| Option(key: "cache-builds", defaultValue: false, usage: "use cached builds when possible")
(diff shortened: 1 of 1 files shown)


blender:
Hi, thanks for the PR!

Unfortunately the description is correct. See https://github.com/Carthage/Carthage/issues/2433

It tricked me too.

If you want to help consider adding a project here: https://github.com/Carthage/Carthage/issues/2536

sgl0v:
@blender Yeah, I got the idea: there is --use-binaries option, that is inverted to --no-use-binaries.

In any case, a user gets the wrong description while running `carthage help build`: 

```
[--no-use-binaries]
	use downloaded binaries when possible

[--no-skip-current]
	don't skip building the Carthage project (in addition to its dependencies)
```

You might notice that description is specified correctly for `skip-current` option ("don't skip building the Carthage..."), because one is inverted to `--no-skip-current`:

```
 		public static func evaluate(_ mode: CommandMode) -> Result<Options, CommandantError<CarthageError>> {
 			return curry(self.init)
 				<*> BuildOptions.evaluate(mode)
 				<*> mode <| Option(key: "skip-current", defaultValue: true, usage: "don't skip building the Carthage project (in addition to its dependencies)")
```

blender:
@sgl0v I think you make an excellent point. @mdiep ?

mdiep:
That makes sense to me. 👍

PR #2561, opened 2018-08-26
[BREAKING] Fix the issue any packages for watchOS couldn't be built on Xcode 10 beta

giginet:
## Problem

When I Built watchOS packages on Xcode 10 beta 6, It would be failed with following error:

```
$ carthage update --platform watchOS --verbose
```

```
--- xcodebuild: WARNING: Unable to open project file '/Users/giginet/work/Project-watchOS/Carthage/Checkouts/AwesomeLibrary/AwesomeLibrary.playground' in workspace '/Users/giginet/work/Project-watchOS/Carthage/Checkouts/AwesomeLibrary/AwesomeLibrary.xcworkspace'.
xcodebuild: error: Unable to find a destination matching the provided destination specifier:
                { platform:watchOS Simulator, id:893B7FA1-9DB3-4AC1-91E6-BC5E078204CA }

        The requested device could not be found because no available devices matched the request.

```

This would happen because the output of  `xcrun simctl list devices` is like following on Xcode 10.

```
-- watchOS 5.0 --
    Apple Watch - 38mm (893B7FA1-9DB3-4AC1-91E6-BC5E078204CA) (Shutdown) (unavailable, device type not supported by runtime)
    Apple Watch - 42mm (839FA3D1-B5C4-42F7-AE2F-93555970D317) (Shutdown) (unavailable, device type not supported by runtime)
    Apple Watch Series 2 - 38mm (159876D0-DB0C-4472-988C-626BA14947B6) (Shutdown)
    Apple Watch Series 2 - 42mm (1F7CC7D7-6EBD-4052-9637-363DC79D8E7E) (Shutdown)
    Apple Watch Series 3 - 38mm (4A0FCBCD-EBFB-4E37-A787-FCA20FB77C07) (Shutdown)
    Apple Watch Series 3 - 42mm (A6D5F20D-6879-4FD1-A7B1-FF8381104A9E) (Shutdown)
```

[Current selection logic](https://github.com/Carthage/Carthage/blob/9503e2c07d9f839dfee3601be8ef0a9f471d2d5f/Source/CarthageKit/Xcode.swift#L648) never consider about availability, So Carthage selects the first device. However it is unavailable.

So I improved the selection logic for build destinations.

## Description

Use `simctl list devices --json` instead to parse simulators. Thanks to this change, we can parse simulators easily.
However, according to the comment, this change drops backward-compatibility below Xcode 6.x. How about this?


Source/CarthageKit/Xcode.swift
```diff
@@ -630,26 +630,23 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
...
+			let destinationLookup = Task("/usr/bin/xcrun", arguments: [ "simctl", "list", "devices", "--json" ])
 			return destinationLookup.launch()
 				.ignoreTaskData()
 				.map { data in
-					let string = String(data: data, encoding: .utf8)!
-					// The output as of Xcode 6.4 is structured text so we
-					// parse it using regex. The destination will be omitted
-					// altogether if parsing fails. Xcode 7.0 beta 4 added a
-					// JSON output option as `xcrun simctl list devices --json`
-					// so this can be switched once 7.0 becomes a requirement.
```
giginet: I think it doesn't need to support Xcode 6.x yet. So I dropped regexp version.
How do you think?

Source/CarthageKit/Xcode.swift
```diff
@@ -630,26 +630,32 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
...
+					func sortedByVersion(_ osNames: [String]) -> [String] {
+						return osNames.sorted { lhs, rhs in
+							guard let lhsVersion = lhs.split(separator: " ").last.flatMap(Float.init),
+								let rhsVersion = rhs.split(separator: " ").last.flatMap(Float.init) else {
+									return lhs < rhs
+							}
+							return lhsVersion < rhsVersion
+						}
+					}
+					let latestOSName = sortedByVersion(Array(allTargetSimulators.keys)).last!
```
giginet: Serialized objects have no order. So we have to sort OS names by their versions.

Source/CarthageKit/Simulator.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+internal struct Simulator: Decodable {
+	enum Availavility: String, Decodable {
```
mdiep: _Availability_ is misspelled here

ikesyo:
Marking as BREAKING for next release operation.

Source/CarthageKit/Xcode.swift
ikesyo: IIUC this would crash if there is no simulator for the specified platform (as opposed to the previous implementation).

Source/CarthageKit/Simulator.swift
giginet: nice catch 👍

Source/CarthageKit/Xcode.swift
giginet: Sure. However if there is no available simulators, previous implementation selects something and attempt to build with that by xcodebuild.
However, it would be failed. 
So both cases are not able to build correctly.

I think we should add assertion to check any simulator is available.

Source/CarthageKit/Xcode.swift
```diff
@@ -630,26 +630,34 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
...
+						return osNames.sorted { lhs, rhs in
+							guard let lhsVersion = lhs.split(separator: " ").last.flatMap(Float.init),
+								let rhsVersion = rhs.split(separator: " ").last.flatMap(Float.init) else {
+									return lhs < rhs
+							}
+							return lhsVersion < rhsVersion
+						}
+					}
+					guard let latestOSName = sortedByVersion(Array(allTargetSimulators.keys)).last else {
+						fatalError("No available simulators found to build this package for \(platformName)")
```
mdiep: This should return an error instead of calling `fatalError`.

That would also let you write a test for it.

Source/CarthageKit/Xcode.swift
```diff
@@ -630,26 +630,34 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
...
+						.filter { $0.key.hasPrefix(platformName) }
+					func sortedByVersion(_ osNames: [String]) -> [String] {
+						return osNames.sorted { lhs, rhs in
+							guard let lhsVersion = lhs.split(separator: " ").last.flatMap(Float.init),
+								let rhsVersion = rhs.split(separator: " ").last.flatMap(Float.init) else {
+									return lhs < rhs
+							}
+							return lhsVersion < rhsVersion
+						}
+					}
```
mdiep: I don't see any tests for this logic. Please add a test that has multiple OS versions.

giginet:
@mdiep Hi! I implemented an error handling and added specs.

https://github.com/giginet/Carthage/compare/2662f7f94842236e6b038eb41ade850cef4fceb3...04ac9326b7f937f1f1022eb9465546cb16772632

How about this?

giginet:
test this please

CI is still failed but it seems not to caused by my changes.

Source/CarthageKit/Simulator.swift
```diff
@@ -0,0 +1,56 @@
...
+/// If there are multiple OSs for the SDK, the latest one would be selected.
+internal func selectAvailableSimulator(of sdk: SDK, from data: Data) -> Simulator? {
+	let decoder = JSONDecoder()
+	// simctl returns following JSON:
+	// {"devices": {"iOS 12.0": [<simulators...>]}]
+	guard let jsonObject = try? decoder.decode([String: [String: [Simulator]]].self, from: data) else {
+		return nil
+	}
+	let platformName = sdk.platform.rawValue
+	let devices = jsonObject["devices"]!
```
giginet: I should not use IUO.

Source/CarthageKit/Errors.swift
```diff
@@ -42,6 +42,9 @@ public enum CarthageError: Error {
 	/// Failed to write a file or directory at the given URL.
 	case writeFailed(URL, NSError?)
 
+	/// No available simulators could be found
+	case noAvailableSimulators
```
ikesyo: This should be tied to a SDK and output its platform name in a error message (e.g. `Could not find any available watchOS simulators`).

giginet: Thank you @ikesyo 

I added `platformName` to `availableSimulators` error.

giginet:
I discussed with @ikesyo. I'll change minor changes before merging.

- Use `SemanticVersion` to compare OS versions
- Refactor some test cases 

Please do  not merge before this.

Source/CarthageKit/Simulator.swift
```diff
@@ -0,0 +1,56 @@
...
+	guard let jsonObject = try? decoder.decode([String: [String: [Simulator]]].self, from: data),
+		let devices = jsonObject["devices"] else {
+		return nil
+	}
+	let platformName = sdk.platform.rawValue
+	let allTargetSimulators = devices
+		.filter { $0.key.hasPrefix(platformName) }
+	func sortedByVersion(_ osNames: [String]) -> [String] {
+		return osNames.sorted { lhs, rhs in
+			guard let lhsVersion = SemanticVersion.from(PinnedVersion(lhs)).value,
```
giginet: It is bit strange to use `PinnedVersion` here. But I think it is better to use `Float` to parse OS versions.

giginet:
I fixed minor points and tried to install some packages.
Please merge this if there are no problems.

jshier:
@ikesyo Can Carthage do a release with this fix? Xcode 10 is out and many users are hitting this bug.

mdiep:
cc @blender

blender:
We can patch `0.30.1` to `0.30.2` that includes this fix and not the disputed change "build" -> "archive" action used to discover schemes.

I would release everything as [Amphorae](https://github.com/blender/Amphorae) has been looking good for some time now with occasional failures

PR #2563, opened 2018-08-31
Update README.md

martnst:
Change last two steps in Quick Start to be bullets points, as they are actually sub-steps of adding New Run Script Phases and not extra steps on their own.

I was confused where to find "Input Files" in step 9, since I did not consider it being in the context of step 8.


mdiep:
That makes sense! Thanks for opening a PR!

PR #2569, opened 2018-09-07
[BREAKING][Cartfile] Reject `git` or `ssh` protocol as a `github` origin as it does not make sense

### LINKED ISSUE #2379: Carthage doesn't support SSH authentication with a custom username in GHE URLs
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `carthage version`: 0.28.0
* `xcodebuild -version`: Xcode 9.2, Build version 9C40b
* Are you using `--cache-builds`? no

**Cartfile**
```
github "ssh://git@some_host/some_owner/some_repo.git" "master"

# Testing tools
github "Quick/Nimble" ~> 7.0
github "Quick/Quick" ~> 1.2
```

**Carthage Output**
```
*** Cloning some_repo
*** Fetching Quick
A shell task (/usr/bin/env git clone --bare --quiet ssh://some_host/some_owner/some_repo.git /Users/petter/Library/Caches/org.carthage.CarthageKit/dependencies/some_repo) failed with exit code 128:
petter@some_host: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

**Actual outcome**
Carthage did not use the user name specified in the URL.

**Expected outcome**
Carthage should use that username (`git`), meaning that somewhere around https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/GitHub.swift#L50-L55, it should also extract the user from the URL and if specified, use that in the URL later used for the clone command.

ikesyo:
Ref: https://github.com/Carthage/Carthage/issues/2379#issuecomment-374428450 and https://github.com/Carthage/Carthage/issues/2379#issuecomment-374456138

Closes #2379.

### DIFF
--- a/Carthage.xcodeproj/project.pbxproj
+++ b/Carthage.xcodeproj/project.pbxproj
@@ -57,6 +57,7 @@
 		BF7A1A0B1E3A7AE9008CBCC5 /* BinaryProject.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF7A1A091E3A7188008CBCC5 /* BinaryProject.swift */; };
 		BFFA7A631E3AAFD200CB95A7 /* BinaryProjectSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFFA7A621E3AAFD200CB95A7 /* BinaryProjectSpec.swift */; };
 		CD09E28B1F0E3FFB00A0378A /* FakeSwift.framework in Resources */ = {isa = PBXBuildFile; fileRef = CD09E2881F0E3F1400A0378A /* FakeSwift.framework */; };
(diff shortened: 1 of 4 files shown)


blender:
100% With you. This however is a breaking change.

I think I did some work (already merged) to so that in the resolved they would appear as `github` anyways.

blender:
@ikesyo did you have the time to check my comment?

mdiep:
This is technically a breaking change. But it should be rare in practice.

ikesyo:
I'm sorry for the late response. Yes that should be a breaking change.

> I think I did some work (already merged) to so that in the resolved they would appear as github anyways.

Looks like the work is a bit different from this, converting a `git` origin with a GitHub host to a `github` origin is still valid.

blender:
LGTM. This will definitely help reduce misuse.

PR #2574, opened 2018-09-11
Refactor `buildScheme` a bit

ikesyo:
Replacing `collect()` and `isEmpty` check with `take(first: 1)`.


ikesyo:
Hmm this (and original code) may not be correct; removing `.collect()` may be the right way.

mdiep:
This looks fine as a refactoring of the original code.

ikesyo:
Thanks!

Issue #2281, opened 2017-12-06
Toolchain regular expression doesn't seem to allow development versions of Swift


tonyarnold:
https://github.com/Carthage/Carthage/blob/d3aef068e1600061c491640ee73292ddb24b5baa/Source/CarthageKit/Xcode.swift#L39

I've been pulling my hair out trying to get my dependencies building using a toolchain downloaded from swift.org, and I think I've found the issue. Here's the output of the command Carthage is trying to run:

```
xcrun --toolchain org.swift.4120171204a swift --version
Apple Swift version 4.1-dev (LLVM cdc9836644, Clang f0e38d2044, Swift 7a0598c7dd)
Target: x86_64-apple-darwin17.3.0
```

Currently, the regular expression for this is:

```
Apple Swift version ([0-9.]+) .*\\((.+)\\)
```

It should probably be:

```
Apple Swift version ([^\\s]+) .*\\((.+)\\)
```

Or something more generic to allow arbitrary toolchain version suffixes.

PR #2585, opened 2018-09-20
Add Xcode version information to VersionFile

giginet:
## Motivation & Context

I want to use [`carthage-verify`](https://github.com/Carthage/workflows/blob/master/carthage-verify) on workflows to skip duplicated builds.

However, current `.Package.version` file doesn't contains Xcode version information.
So we can't detect which Xcode version is used to build.

I added `xcodeVersion` to VersionFile.
Thanks to this PR, we can detect whether Xcode version is changed or not.

```json
{
  "commitish" : "3.2.0",
  "iOS" : [
    {
      "name" : "Himotoki",
      "hash" : "1ab62a1d2a04eb8fdc00d9ed2a843d4a944789eeda0b4ac8250462995754c825",
      "swiftVersion" : "4.2 (swiftlang-1000.11.37.1 clang-1000.11.45.1)"
    }
  ]
}
```

If this value is different with current `xcodebuild --version`, `carthage-verify` have to try to rebuild the package.

## Description

I added `XcodeVersion` to `VersionFile`.

You can try this feature like following:

```console
$ carthage build --platform iOS
$ cat Carthage/Build/.Himotoki.version
```

`xcodeVersion` is optional. So backward-compatibility is well.

### DIFF
--- a/Source/CarthageKit/Errors.swift
+++ b/Source/CarthageKit/Errors.swift
@@ -85,6 +85,8 @@ public enum CarthageError: Error {
 
 	case buildFailed(TaskError, log: URL?)
 
+	case unknownLocalSwiftVersionError
+
 	/// An error occurred while shelling out.
 	case taskError(TaskError)
 
@@ -337,6 +339,9 @@ extension CarthageError: CustomStringConvertible {
 			}
 
 			return message
+
+		case .unknownLocalSwiftVersionError:
+			return SwiftVersionError.unknownLocalSwiftVersion.description
 
 		case let .taskError(taskError):
 			return taskError.description
--- a/Source/CarthageKit/Errors.swift
+++ b/Source/CarthageKit/Errors.swift
@@ -85,7 +85,7 @@ public enum CarthageError: Error {
 
 	case buildFailed(TaskError, log: URL?)
 
-	case unknownLocalSwiftVersionError
+	case unknownLocalSwiftVersion
 
 	/// An error occurred while shelling out.
 	case taskError(TaskError)
@@ -340,7 +340,7 @@ extension CarthageError: CustomStringConvertible {
 
 			return message
 
-		case .unknownLocalSwiftVersionError:
+		case .unknownLocalSwiftVersion:
 			return SwiftVersionError.unknownLocalSwiftVersion.description
 
 		case let .taskError(taskError):
--- a/Source/CarthageKit/Errors.swift
+++ b/Source/CarthageKit/Errors.swift
@@ -85,7 +85,7 @@ public enum CarthageError: Error {
 
 	case buildFailed(TaskError, log: URL?)
 
-	case unknownLocalSwiftVersion
+	case unknownFrameworkSwiftVersion
 
 	/// An error occurred while shelling out.
 	case taskError(TaskError)
@@ -340,8 +340,8 @@ extension CarthageError: CustomStringConvertible {
 
 			return message
 
-		case .unknownLocalSwiftVersion:
-			return SwiftVersionError.unknownLocalSwiftVersion.description
+		case .unknownFrameworkSwiftVersion:
+			return SwiftVersionError.unknownFrameworkSwiftVersion.description
 
 		case let .taskError(taskError):
 			return taskError.description
--- a/Source/CarthageKit/Errors.swift
+++ b/Source/CarthageKit/Errors.swift
@@ -85,7 +85,7 @@ public enum CarthageError: Error {
 
 	case buildFailed(TaskError, log: URL?)
 
-	case unknownFrameworkSwiftVersion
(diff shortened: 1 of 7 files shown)


giginet:
> Why the Xcode version?

That's why parsers for version of Xcode are already implemented.
Carthage depends on xcodebuild. I think there are no problem to use Xcode version.

blender:
Sorry I don't think I understand your answer.

If you want to make sure that the binaries are compatible with the version of xcode currently selected I believe `xcodebuild -version` is not a good measure.

I think `xcrun swift --version` and `xcrun clang --version` are more relevant here.

Anyways, the idea seems valuable to me. 👍

ikesyo:
Considering the use of custom toolchain (e.g. #2281), we should use compiler version over Xcode version.

giginet:
Sorry I’m working on. Could you reopen this issue?

giginet:
@blender @ikesyo Hi!

I add `swiftVersion` column to each frameworks onto VersionFile instead of `xcodeVersion`.
Could you review it?

Sorry about waiting you.

Source/CarthageKit/VersionFile.swift
```diff
@@ -303,21 +304,35 @@ public func createVersionFileForCommitish(
...
 	if !buildProducts.isEmpty {
 		return SignalProducer<URL, CarthageError>(buildProducts)
-			.flatMap(.merge) { url -> SignalProducer<(String, (String, String)), CarthageError> in
+			.flatMap(.merge) { url -> SignalProducer<(String, FrameworkDetail), CarthageError> in
 				let frameworkName = url.deletingPathExtension().lastPathComponent
 				let platformName = url.deletingLastPathComponent().lastPathComponent
-				let frameworkURL = url.appendingPathComponent(frameworkName, isDirectory: false)
-				let details = SignalProducer<(String, String), CarthageError>(value: (platformName, frameworkName))
-				return SignalProducer.zip(hashForFileAtURL(frameworkURL), details)
+				return frameworkSwiftVersion(url)
```
blender: probably this will change: see #2622 

Don't worry just yet

Source/CarthageKit/VersionFile.swift
```diff
@@ -303,21 +304,35 @@ public func createVersionFileForCommitish(
 		platformCaches[platform.rawValue] = []
 	}
 
+	struct FrameworkDetail {
```
blender: private?

Source/CarthageKit/Errors.swift
```diff
@@ -85,6 +85,8 @@ public enum CarthageError: Error {
 
 	case buildFailed(TaskError, log: URL?)
 
+	case unknownFrameworkSwiftVersion
```
blender: probably this will change: see #2622 

Don't worry just yet

Source/CarthageKit/Errors.swift
```diff
@@ -338,6 +340,9 @@ extension CarthageError: CustomStringConvertible {
 
 			return message
 
+		case .unknownFrameworkSwiftVersion:
```
blender: probably this will change: see #2622 

Don't worry just yet

Source/CarthageKit/VersionFile.swift
```diff
@@ -8,10 +8,12 @@ struct CachedFramework: Codable {
 	enum CodingKeys: String, CodingKey {
 		case name = "name"
 		case hash = "hash"
+		case swiftVersion = "swiftVersion"
```
blender: not sure this is the most appropriate name, maybe it should be toolchain version? @ikesyo

blender:
@giginet thanks! ❤️

Source/CarthageKit/VersionFile.swift
giginet: This struct is defined in a method. So it is not possible to add access control.

Source/CarthageKit/VersionFile.swift
giginet: It should be `toolchainVersion` rather than `swiftVersion`?

Source/CarthageKit/VersionFile.swift
blender: Didn't see that. Thanks!

Source/CarthageKit/VersionFile.swift
blender: Lets see what others think

giginet:
Hey, is there any progress?

blender:
@giginet one change and this should be good to go.

giginet:
Hi! I'll fix in this weekend.
Please wait moment 🙏

giginet:
@blender Hi! I renamed `swiftToolchainVersion`.

blender:
Thanks! Awesome job! 🎉

PR #2586, opened 2018-09-20
Strict to execute SwiftLint on CI

giginet:
# Context

SwiftLint is already introduced. However no linters are executed on CI.
Currently, some codes raise warnings.

# Description

I added `make lint` task and execute on each CI jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,13 @@ matrix:
     - os: osx
       language: objective-c
       osx_image: xcode9.4
+      before_script: make lint
       script: make test
       env: JOB=CI_TEST
     - os: osx
       language: objective-c
       osx_image: xcode9.4
+      before_script: make lint
       script:
         - make swiftpm_test
       env: JOB=CI_TEST_SWIFTPM
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -380,7 +380,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 		tryCheckoutDirectory: Bool
 	) -> SignalProducer<CompatibilityInfo.Requirements, CarthageError> {
 		return SignalProducer(resolvedCartfile.dependencies)
-			.flatMap(.concurrent(limit: 4)) { (dependency, pinnedVersion) -> SignalProducer<(Dependency, (Dependency, VersionSpecifier)), CarthageError> in
+			.flatMap(.concurrent(limit: 4)) { dependency, pinnedVersion -> SignalProducer<(Dependency, (Dependency, VersionSpecifier)), CarthageError> in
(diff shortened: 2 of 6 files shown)


blender:
LGTM 👍

blender:
However the results of the linting are not committed or ... did I misunderstand?

ikesyo:
Setting HoundCI up will be another option.

giginet:
Maybe our SwiftLint versions are mismatched.
We should use latest version.

I’ll check it tomorrow.

giginet:
```console
$ swiftlint version
0.27.0
```

My SwiftLint version is latest.

giginet:
We're better to execute on `install` step.

```console
$ brew update && brew upgrade swiftlint
```

ikesyo:
@mdiep Could you active HoundCI integration to this repo?

https://houndci.com/repos

mdiep:
Done!

.travis.yml
```diff
@@ -13,11 +13,13 @@ matrix:
     - os: osx
       language: objective-c
       osx_image: xcode9.4
+      before_script: make lint
```
ikesyo: @giginet So could you remove these and let HoundCI do the job instead?

.hound.yml
```diff
@@ -0,0 +1,2 @@
+swiftlint:
+  config_file: .swiftlint.yml
```
giginet: `.swiftlint.yml` seems to be detected automatically...

ikesyo: Looks like HoundCI uses their default config if it is not specified explicitly. So 👍 on this.

Source/CarthageKit/Project.swift
```diff
@@ -380,7 +380,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 		tryCheckoutDirectory: Bool
 	) -> SignalProducer<CompatibilityInfo.Requirements, CarthageError> {
 		return SignalProducer(resolvedCartfile.dependencies)
-			.flatMap(.concurrent(limit: 4)) { (dependency, pinnedVersion) -> SignalProducer<(Dependency, (Dependency, VersionSpecifier)), CarthageError> in
+			.flatMap(.concurrent(limit: 4)) { dependency, pinnedVersion -> SignalProducer<(Dependency, (Dependency, VersionSpecifier)), CarthageError> in
```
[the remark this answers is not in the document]
ikesyo: This is odd: https://github.com/Carthage/Carthage/blob/733e9a89f75da6d6900aa3469bbe8fb064d7999a/.swiftlint.yml#L33

giginet:
Hi. I enabled Hound CI.

However, HoundCI doesn't support `--strict` option so warnings are passed.

Issue #2400, opened 2018-04-01
Carthage doesn't cache build across projects


kenji21:
* carthage install method: [ ] .pkg, [X] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.29.0
* `xcodebuild -version`:  Xcode 9.3 Build version 9E145
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? yes
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? yes
* Are you using `--new-resolver`? no

**Cartfile**
```
github "SVProgressHUD/SVProgressHUD" ~> 2.0
```

**Carthage Output**
```
10:35:30 kenji@macboo:~/Desktop/CarthageCacheTest
cat Cartfile
github "SVProgressHUD/SVProgressHUD" ~> 2.0

10:35:33 kenji@macboo:~/Desktop/CarthageCacheTest
cat Cartfile.resolved
github "SVProgressHUD/SVProgressHUD" "2.2.4"

10:35:34 kenji@macboo:~/Desktop/CarthageCacheTest
rm -Rf Carthage ; time carthage bootstrap --no-use-binaries --cache-builds --platform iOS
*** Checking out SVProgressHUD at "2.2.4"
*** No cache found for SVProgressHUD, building with all downstream dependencies
*** xcodebuild output can be found in /var/folders/7v/dfg0p2vj1z58n65d99l2n1lh0000gp/T/carthage-xcodebuild.5CyxXz.log
*** Building scheme "SVProgressHUD-Framework" in SVProgressHUD.xcodeproj

real	0m25.314s
user	0m24.218s
sys	0m7.705s

10:36:21 kenji@macboo:~/Desktop/CarthageCacheTest
ls -l /Users/kenji/Library/Caches/org.carthage.CarthageKit/DerivedData/9.3_9E145/SVProgressHUD/
total 0
drwxrwxr-x@ 6 kenji  staff  192 Apr  1 10:36 2.2.4
drwxrwxr-x@ 6 kenji  staff  192 Apr  1 10:19 2.2.5

10:36:28 kenji@macboo:~/Desktop/CarthageCacheTest
rm -Rf Carthage ; time carthage bootstrap --no-use-binaries --cache-builds --platform iOS
*** Checking out SVProgressHUD at "2.2.4"
*** No cache found for SVProgressHUD, building with all downstream dependencies
[...]

Issue #2120, opened 2017-08-16
Carthage Concurrent Build Fail.


zorture:
* `carthage version`: 0.20.1 & 0.24.0
* `xcodebuild -version`: 8.3.2
* Are you using `--no-build` No
* Are you using `--no-use-binaries` No
* Are you using `--use-submodules` No
* Are you using `--cache-builds` No

** Instructions **
I am running **carthage update --platform ios** from Folder A and B simultaneously. The command which has been triggered first fails. 
Conclusion carthage update fails for concurrent update.

**Cartfile**

[A.zip](https://github.com/Carthage/Carthage/files/1226882/A.zip)
[B.zip](https://github.com/Carthage/Carthage/files/1226884/B.zip)

**Carthage Output**

<img width="1107" alt="carthage concurrency issue" src="https://user-images.githubusercontent.com/6214007/29344555-b55491e8-8206-11e7-8214-110d5fdbc8a7.png">

** Error Log **

<img width="820" alt="build failed log" src="https://user-images.githubusercontent.com/6214007/29344560-bb60fdc4-8206-11e7-9f43-1b8eeff233bd.png">

** Requirement **
I make 4 - 5 builds daily with different carthage tags. Since simultaneous build fails, i am not able to completely automate the CI.

** Issue Root Cause **
I noticed the root cause of the issue is the shared cache.

Is there any way to have a private cache for carthage update so i can achieve concurrency?

ikesyo:
I don't think this is an expected usage of Carthage.

Anyway, could you please check https://github.com/Carthage/Carthage/pull/2112#issuecomment-322286205.

> You can specify a custom derived data if you'd like to do that. `carthage bootstrap --derived-data <path to the custom derived data folder>`
>
> Note that you probably want to run `carthage bootstrap` on CI, not `carthage update`.

mdiep:
This should work (but apparently isn't) if you pass a custom derived-data directory for each.

bwhiteley:
`--derived-data` is insufficient because the shared cache contention often occurs under `~/Library/Caches/org.carthage.CarthageKit/dependencies` (git .lock files, for example).

blender:
@bwhiteley  are you experiencing this problem? https://docs.microsoft.com/en-us/vsts/git/git-index-lock ? 

Or can you provide sample Cartfiles? Can you add the error output?

bitbytedog:
[...]

PR #2608, opened 2018-10-08
Issue #2400 Shared cache for built dependencies

kenji21:
See discussion in issue https://github.com/Carthage/Carthage/issues/2400

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -269,16 +269,18 @@ private func parseConfigEntries(_ contents: String, keyPrefix: String = "", keyS
 	}
 }
 
-/// Git’s representation of file system objects at a path relative to the repository root.
+/// Lists the contents of a git tree object at a given path relative to the repository root.
+/// It can be used in a bare repository.
 ///
-/// - parameter path: Path separators at the end of `path` have significance of outputting directory contents.
-///                   Thankfully, multiple contiguous path separators seem to have no adverse effects.
 /// - note: Previously, `path` was recursed through — now, just iterated.
 internal func list(treeish: String, atPath path: String, inRepository repositoryURL: URL) -> SignalProducer<String, CarthageError> {
+	// git ls-tree treats "dir/" and "dir" differently. We make sure the path has an ending slash here.
+	// Thankfully, multiple successive slashes are considered to be the same as one slash.
+	let directoryPath = path.appending("/")
 	return launchGitTask(
 			// `ls-tree`, because `ls-files` returns no output (for all instances I’ve seen) on bare repos.
 			// flag “-z” enables output separated by the nul character (`\0`).
-			[ "ls-tree", "-z", "--full-name", "--name-only", treeish, path ],
+			[ "ls-tree", "-z", "--full-name", "--name-only", treeish, directoryPath ],
 			repositoryFileURL: repositoryURL
 		)
 		.flatMap(.merge) { (output: String) -> SignalProducer<String, CarthageError> in


Source/CarthageKit/Xcode.swift
```diff
@@ -6,6 +6,17 @@ import ReactiveSwift
...
 import XCDBLD
 
+/// Gets the shared cache file path URL for dependency at pinned version
+/// ~/Library/Caches/org.carthage.CarthageKit/SharedCache/XcodeVersion/DependencyName/Version, exemple :
+/// ~/Library/Caches/org.carthage.CarthageKit/SharedCache/9.4.1_9F2000/SwiftyUserDefaults/3.0.1
+internal func sharedCacheURLByXcodeVersionFor(dependency: Dependency, pinnedVersion: PinnedVersion) -> URL {
+	let sharedCacheURL = Constants.userCachesURL.appendingPathComponent("SharedCache")
+	let sharedCachePerXcode = sharedCacheURL.appendingPathComponent(XcodeVersion.makeString(), isDirectory: true)
+	let sharedCachePerDependency = sharedCachePerXcode.appendingPathComponent(dependency.name, isDirectory: true)
+	let sharedCacheVersioned = sharedCachePerDependency.appendingPathComponent(pinnedVersion.commitish, isDirectory: true)
```
blender: Will this cover the case of the the commitish being a tag?

Source/CarthageKit/Xcode.swift
```diff
@@ -821,12 +832,20 @@ public func buildInDirectory( // swiftlint:disable:this function_body_length
...
-
-				return createVersionFile(
-					for: dependency.dependency, version: dependency.version,
-					platforms: options.platforms, buildProducts: urls, rootDirectoryURL: rootDirectoryURL
-				)
-				.flatMapError { _ in .empty }
+				var copyToSharedCache = SignalProducer<(URL), CarthageError>.empty
+				if options.cacheBuilds {
+					let sharedCacheURLForDependency = sharedCacheURLByXcodeVersionFor(dependency: dependency.dependency, pinnedVersion: dependency.version)
+					copyToSharedCache = copyFrameworksAndCreateVersionFile(urls, from: rootDirectoryURL, to: sharedCacheURLForDependency,
```
blender: I would split this function in 2. One that copies the frameworks and one that creates the version file. Is there any practical reason why they are combined?

Source/XCDBLD/XcodeVersion.swift
```diff
@@ -42,4 +42,8 @@ public struct XcodeVersion {
 			}
 			.single()?.value
 	}
+
+	public static func makeString() -> String {
```
blender: I would rather return and error here. Was there any practical reason for this silently defaulting to Unknown?

blender:
@kenji21 Thanks for the PR 🎉 

Can you please comment on the chunks that you have move around and signal them to me? This would make it easier for me to review.

Source/CarthageKit/Xcode.swift
kenji21: It does, here are some tags in my local cache:
ls -1 ~/Library/Caches/org.carthage.CarthageKit/SharedCache/10.0_10A255/Alamofire/
4.7.1
4.7.3

I've built the path the same way as: https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/Project.swift#L1102-L1107

Source/CarthageKit/Xcode.swift
kenji21: My refactoring make me using it twice, in https://github.com/openium/Carthage/blob/master/Source/CarthageKit/Xcode.swift#L838 and https://github.com/openium/Carthage/blob/master/Source/CarthageKit/Project.swift#L1400
So to not duplicate code as now copy can be from DerivedData to Carthage/Build and then from CarthageBuild to ~/.../SharedCache (see paths in https://github.com/Carthage/Carthage/issues/2400#issuecomment-427343658)

Source/XCDBLD/XcodeVersion.swift
kenji21: The defaulting was in https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/Project.swift#L131
but I need it in Xcode.swift, so I moved it to XcodeVersion.swift

kenji21: Precisely at:  https://github.com/openium/Carthage/blob/master/Source/CarthageKit/Xcode.swift#L14

kenji21:
@blender no problems, just done it, and of course will make all changes you ask for as you know better the project than me

Source/CarthageKit/Xcode.swift
blender: ok

Source/XCDBLD/XcodeVersion.swift
blender: ok, cool

Source/CarthageKit/Xcode.swift
kenji21: You see an issue with this ? maybe I should add a dash and commit hash to version in the folder to manage forks (Alamofire 4.7.3 from my fork will be in the same path as the "original", but if I fork it... it might be to fix something, so I'll point to my master/fix-branch) ? In the case of a commit:
ls -1 ~/Library/Caches/org.carthage.CarthageKit/SharedCache/10.0_10A255/openium-lib.ios:
84bb9432df2efdf5bf79b108837fb4416ccc865f

blender: ok

kenji21:
Do you want me to rebase my commit on top of current master ? 

(currently: This branch is 1 commit ahead, 10 commits behind Carthage:master.)

mdiep:
I'm not sure we should add a shared cache like this that's not opt-in. I was thinking there'd be a flag to turn this on. Since we never clean this out, I'm hesitant to start storing additional data that might fill up users' drives.

kenji21:
@blender  was thinking that adding more and more "command options" was not recommended, so I reused the existing "cache" flag: https://github.com/Carthage/Carthage/issues/2400#issuecomment-402439357

We have deployed it, and migrating from an older Xcode to a new one (and swift) with this patch makes us being able to update / bootstrap with the new xcode/swiftc version very quickly

mdiep:
Hmm… I'll take another look

GabKlein:
Hello,

Is this MR would solve this issue as well https://github.com/Carthage/Carthage/issues/2120 ?
It seems that introducing a new environment variable to change the value of `userCachesURL` would solve it.

kenji21:
Just force-pushed to fix the conflit (arguments of a method I moved were reformatted to split them one per line)

acecilia:
What is preventing this to move forward?

kenji21:
this one: https://github.com/Carthage/Carthage/issues/2677 that makes my last "update to master" CI build failing

kenji21:
Ok, I moved my commit to a branch https://github.com/openium/Carthage/tree/feature-2400-shared-cache waiting for #2677 to be solved, so it closed this issue, I'll PR again up to date with current master once #2677 will be fixed

JWShroyer:
This closed PR doesn't seem to be linked to any other PR, but based off what I was reading in #2677 that code should be implemented in 0.33 by the author, @kenji21. Is this work being worked on elsewhere?

kenji21:
@JWShroyer: The updated PR is this one : https://github.com/Carthage/Carthage/pull/2716

JWShroyer:
Thanks! 🙇‍♂️

PR #2617, opened 2018-10-18
Fix an ls-tree issue

CosynPa:
The `list(treeish:atPath:inRepository:)` function was not working as expected. But this hasn't caused any actual problem since there is a further check (the `continue` statement in the `symlinkCheckoutPaths` method) afterward.

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -269,16 +269,18 @@ private func parseConfigEntries(_ contents: String, keyPrefix: String = "", keyS
 	}
 }
 
-/// Git’s representation of file system objects at a path relative to the repository root.
+/// Lists the contents of a git tree object at a given path relative to the repository root.
+/// It can be used in a bare repository.
 ///
-/// - parameter path: Path separators at the end of `path` have significance of outputting directory contents.
-///                   Thankfully, multiple contiguous path separators seem to have no adverse effects.
 /// - note: Previously, `path` was recursed through — now, just iterated.
 internal func list(treeish: String, atPath path: String, inRepository repositoryURL: URL) -> SignalProducer<String, CarthageError> {
+	// git ls-tree treats "dir/" and "dir" differently. We make sure the path has an ending slash here.
+	// Thankfully, multiple successive slashes are considered to be the same as one slash.
+	let directoryPath = path.appending("/")
 	return launchGitTask(
 			// `ls-tree`, because `ls-files` returns no output (for all instances I’ve seen) on bare repos.
 			// flag “-z” enables output separated by the nul character (`\0`).
-			[ "ls-tree", "-z", "--full-name", "--name-only", treeish, path ],
+			[ "ls-tree", "-z", "--full-name", "--name-only", treeish, directoryPath ],
 			repositoryFileURL: repositoryURL
 		)
 		.flatMap(.merge) { (output: String) -> SignalProducer<String, CarthageError> in


Source/CarthageKit/Git.swift
```diff
@@ -269,16 +269,16 @@ private func parseConfigEntries(_ contents: String, keyPrefix: String = "", keyS
 	}
 }
 
-/// Git’s representation of file system objects at a path relative to the repository root.
+/// Lists the contents of the directory at a given path relative to the repository root.
 ///
-/// - parameter path: Path separators at the end of `path` have significance of outputting directory contents.
```
blender: Double checking: this comment is not relevant anymore because of the normalization done by

```
let directoryPath = URL(fileURLWithPath: path, isDirectory: true).relativeString
```

?

Why `relativeString`? There is no baseURL, so this will default to `absoluteString`

blender:
Thanks for opening a PR 🙇 

Just one question from my side

Source/CarthageKit/Git.swift
CosynPa: That's why I deleted the comment about the path parameter.

Running a `URL(fileURLWithPath: "Carthage/Checkouts", isDirectory: true).absoluteString` in a playground will give you some weird thing like `"file:///private/var/folders/3h/7fj4dcs97lqf0yc46r8c64yc0000gn/T/com.apple.dt.Xcode.pg/containers/com.apple.dt.playground.stub.iOS_Simulator.MyPlayground3-33380BCE-E588-450F-8170-5C1FADA1EF1E/Carthage/Checkouts/"`. I think it is actually relative to the current directory, so we can't use `absoluteString`.

Dopebiz6Dime: I’m wondering if I have something to add .. or forward . Let me know thanks

blender: @CosynPa what is the output with `relativeString` instead?

CosynPa: "Carthage/Checkouts/". The key point here is to make sure there is a slash at the end.

blender: ok! Thanks!

Source/CarthageKit/Git.swift
```diff
@@ -269,16 +269,16 @@ private func parseConfigEntries(_ contents: String, keyPrefix: String = "", keyS
...
-///                   Thankfully, multiple contiguous path separators seem to have no adverse effects.
 /// - note: Previously, `path` was recursed through — now, just iterated.
 internal func list(treeish: String, atPath path: String, inRepository repositoryURL: URL) -> SignalProducer<String, CarthageError> {
+	// git ls-tree treats "dir/" and "dir" differently. This makes sure the path is a directory
+	let directoryPath = URL(fileURLWithPath: path, isDirectory: true).relativeString
```
jdhealy: This is a tricky corner where we’re passing paths to check git’s storage in anticipation of writing symlinks at those paths (or filtering those paths out).

Every choice here works well enough for [the single call-site](https://github.com/Carthage/Carthage/blob/0f561ee/Source/CarthageKit/Project.swift#L951) where we invoke the function, but I’d prefer to leave it dual purpose: listing single files as well as contents of directories.

> This comment could definitely be improved — the dir and dir/ examples are good visual indicators.

In that light, we should avoid NSURL here when appending that trailing path separator we’re looking for — just `path + "/"`¹ at [the call-site](https://github.com/Carthage/Carthage/blob/0f561ee/Source/CarthageKit/Project.swift#L951).

<details>

<summary>
And, there reason to actively avoid NSURL here: <a href='https://en.wikipedia.org/wiki/Null_character'>Null characters</a> with `URL(fileURLWithPath:)`
</summary>

```
  7> URL(fileURLWithPath: "etc\u{0000}etc", isDirectory: true).relativeString
$R5: String = ""
  8> URL(fileURLWithPath: "etc\u{0000}etc", isDirectory: true).scheme
$R6: String? = nil
```
</details>

<br>

Unless there’s something I’ve overlooked, what do you think?

⑴ `path` being `carthageProjectCheckoutsPath` here.

jdhealy:
Oh, and thanks so much for opening this PR ☀️ 🎉

Source/CarthageKit/Git.swift
CosynPa: I think the git ls-tree behavior of treating `dir` and `dir/` differently is somewhat surprising, so maybe avoiding it is better than exposing it to the user. The main purpose of ls-tree is "list the contents of a tree object" as the git document states. Passing a file path or a directory path without an ending `/` such as `Carthage/Checkouts`  just gives you the path `Carthage/Checkouts` back, which does not make much sense, unless you want to check if a file or a directory exists. But in that case, I think it's better to use another function for checking the existence of a file since this is naturally a different usage of ls-tree.

I don't think it's a valid path string with any null character in it. This kind of error should be solved before passing the arguments to this function.

Source/CarthageKit/Git.swift
```diff
@@ -269,16 +269,16 @@ private func parseConfigEntries(_ contents: String, keyPrefix: String = "", keyS
 	}
 }
 
-/// Git’s representation of file system objects at a path relative to the repository root.
+/// Lists the contents of the directory at a given path relative to the repository root.
```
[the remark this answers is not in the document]
CosynPa: @jdhealy I don't think it very clear about what is git's representation and what is non-git's representation. How about "list the contents of a git tree object" with additional notes mentioning that it is irrelevant of the files in the working directory and can be used in a bare repository?

CosynPa:
Updated the comment.

Source/CarthageKit/Git.swift
mdiep: I think roundtripping through `URL` is unnecessarily complicated here. The Path<->URL conversions in particular are not very well documented, nor do I think they're very well understood. (I definitely don't have a mental model to understand the safety of this operation.)

Given this comment:

> Thankfully, multiple contiguous path separators seem to have no adverse effects.

I think the safest thing to do is to append a `/` to the incoming `String`. Since that's the behavior we want anyway, let's just do that.

CosynPa:
I found the equivalence between multiple successive slashes and a single slash is documented.
[How does Linux handle multiple consecutive path separators (/home////username///file)?](https://unix.stackexchange.com/a/1919)
So, this would be safe. I've changed the implementation to use string manipulation directly.

mdiep:
Looks great. Thank you for all your work on this!

Issue #2619, opened 2018-10-18
--cache-builds sometimes fails to recognize a cached dependency


blender:
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.31.1
* `xcodebuild -version`: Xcode 9.2 - 9C40b
* Are you using `--no-build`?  no
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? yes
* Are you using `--new-resolver`?  no

**Cartfile**
```
github "typelift/Swiftz" ~> 0.6.3
github "OpenShelter/Tyro" ~> 0.2.0
github "Alamofire/Alamofire" ~> 4.4.0
```

**Cartfile.resolved**
```
github "Alamofire/Alamofire" "4.7.3"
github "OpenShelter/Tyro" "v0.2.0"
github "swisspol/GCDWebServer" "3.4.2"
github "typelift/Swiftz" "0.6.3"
```

**Carthage Output**
```bash
20:03:48 Tommaso@mars ~/Code/<redacted>
$ carthage update --platform iOS --cache-builds
*** Fetching Tyro
*** Fetching Swiftz
*** Fetching GCDWebServer
*** Fetching Alamofire
*** Checking out Alamofire at "4.7.3"
*** Checking out GCDWebServer at "3.4.2"
*** Checking out Swiftz at "0.6.3"
*** Checking out Tyro at "v0.2.0"
*** No cache found for Alamofire, building with all downstream dependencies
*** No cache found for GCDWebServer, building with all downstream dependencies

*** No cache found for Swiftz, building with all downstream dependencies  # <- ok

*** xcodebuild output can be found in /var/folders/b4/qz7m1r9x3mj1bm3nfsjjf64m0000gp/T/carthage-xcodebuild.TRCyO8.log
*** Building scheme "Alamofire iOS" in Alamofire.xcworkspace
*** Building scheme "GCDWebServers (iOS)" in GCDWebServer.xcodeproj
*** Building scheme "Swiftz-iOS" in Swiftz.xcodeproj
*** Building scheme "Tyro-iOS" in Tyro.xcodeproj
20:06:46 Tommaso@mars ~/Code/<redacted>
$ carthage update --platform iOS --cache-builds
*** Fetching Tyro
*** Fetching Swiftz
*** Fetching GCDWebServer
*** Fetching Alamofire
[...]

PR #2622, opened 2018-10-22
Fall back to get swiftlang & clang version from dSYM 

blender:
See https://github.com/Carthage/Carthage/issues/2619 for reference.

This PR only implement the first fallback.

### DIFF
--- a/Source/CarthageKit/VersionFile.swift
+++ b/Source/CarthageKit/VersionFile.swift
@@ -148,6 +148,7 @@ struct VersionFile: Codable {
 					return SignalProducer(value: true)
 				} else {
 					return frameworkSwiftVersion(frameworkURL)
+						.flatMapError { _ in dSYMSwiftVersion(frameworkURL.appendingPathExtension("dSYM")) }
 						.map { swiftVersion -> Bool in
 							return swiftVersion == localSwiftVersion
 						}
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -36,7 +36,7 @@ private func compilerVersionArguments(usingToolchain toolchain: String?) -> [Str
 private func parseSwiftVersionCommand(output: String?) -> String? {
 	guard
 		let output = output,
-		let regex = try? NSRegularExpression(pattern: "Apple Swift version ([^\\s]+) .*\\((.+)\\)", options: []),
+		let regex = try? NSRegularExpression(pattern: "Apple Swift version ([^\\s]+) .*\\((.[^\\)]+)\\)", options: []),
 		let match = regex.firstMatch(in: output, options: [], range: NSRange(output.startIndex..., in: output))
 		else
 	{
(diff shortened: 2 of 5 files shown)


Source/CarthageKit/VersionFile.swift
```diff
@@ -148,10 +148,11 @@ struct VersionFile: Codable {
 					return SignalProducer(value: true)
 				} else {
 					return frameworkSwiftVersion(frameworkURL)
+						.flatMapError { _ in dSYMSwiftVersion(frameworkURL.appendingPathExtension("dSYM")) }
 						.map { swiftVersion -> Bool in
 							return swiftVersion == localSwiftVersion
 						}
-						.flatMapError { _ in SignalProducer<Bool, CarthageError>(value: false) }
+						.flatMapError { _ in SignalProducer<Bool, CarthageError>(value: false)}
```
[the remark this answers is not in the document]
blender: Yup

Source/CarthageKit/Xcode.swift
```diff
@@ -36,7 +36,7 @@ private func compilerVersionArguments(usingToolchain toolchain: String?) -> [Str
 private func parseSwiftVersionCommand(output: String?) -> String? {
 	guard
 		let output = output,
-		let regex = try? NSRegularExpression(pattern: "Apple Swift version ([^\\s]+) .*\\((.+)\\)", options: []),
+		let regex = try? NSRegularExpression(pattern: "Apple Swift version ([^\\s]+) .*\\((.[^\\)]+)\\)", options: []),
```
blender: Before the change this regex would match the second capture group differently in case of no new line:

`Apple Swift version 4.1.2 effective-3.3.2 (swiftlang-902.0.54 clang-902.0.39.2) -emit-object` -> `swiftlang-902.0.54 clang-902.0.39.2) -emit-object` (Garbage at the end)

But (notice new line char)

`Apple Swift version 4.1.2 effective-3.3.2 (swiftlang-902.0.54 clang-902.0.39.2) \n` ->  `swiftlang-902.0.54 clang-902.0.39.2

What we want is `swiftlang-902.0.54 clang-902.0.39.2`

The regex worked fine as it was here https://regex101.com/ but not with NSRegularExpression . Had to tell it to exclude `)`.

Source/CarthageKit/Xcode.swift
```diff
@@ -52,18 +52,69 @@ private func parseSwiftVersionCommand(output: String?) -> String? {
...
 
+internal func dSYMSwiftVersion(_ dSYMURL: URL) -> SignalProducer<String, SwiftVersionError> {
+
+	// Pick one architecture
+	guard let arch = architecturesInPackage(dSYMURL).first()?.value else {
+		return SignalProducer(error: .unknownFrameworkSwiftVersion(message: "No architectures found in dSYM."))
+	}
+
+	// Check the .debug_info section left from the compiler in the dSYM.
+	let task = Task("/usr/bin/xcrun", arguments: ["dwarfdump", "--arch=\(arch)", "--debug-info", dSYMURL.path])
```
blender: This can take quite some time to execute, so to speed things up I decided to just execute it for one architecture. Which one doesn't matter.

mdiep:
Looks good! 👏

PR #2625, opened 2018-10-25
Update function and error for copy frameworks to reflect file lists support

artemnovichkov:
@kishikawakatsumi added support of file lists to copy frameworks script. I have updated wording to reflect these changes.


mdiep:
Thank you!

Issue #2634, opened 2018-11-01
.version file is not created for the current project with --no-skip-current


blender:
* carthage install method: [ ] .pkg, [ ] homebrew, [ ] source
* `which carthage`: `/usr/local/bin/carthage`
* `carthage version`:  0.31.2
* `xcodebuild -version`: `Build version 9F1027a`
* Are you using `--no-build`? No
* Are you using `--no-use-binaries`?  No
* Are you using `--use-submodules`? No
* Are you using `--cache-builds`? Yes
* Are you using `--new-resolver`?  Yes

**Cartfile**
n/a
**Carthage Output**
n/a

**Actual outcome**

When running `carthage build --no-skip-current` .version files are generated for the dependencies but not for the current project

**Expected outcome**
.version file is generated for the current project

**Implementation suggestiong**

To extract the `version` all that needs to be done is:
```
git describe --exact-match `git rev-parse HEAD`
```

In case of no match just take the `HEAD` commitish

blender:
Would appreciate if anyone could give their opinion @mdiep @ikesyo  @jdhealy 👍

mdiep:
Sounds like a reasonable thing to support. 👍

ikesyo:
Looks like this is dupe of #1846🙃

willemfrishert:
@blender could you tell me what the status is of this ticket? The work seems to have been merged and I've been trying to test it out on the latest release but it does not seem to create the version file for me.

My setup:
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `which carthage`: `/usr/local/bin/carthage`
* `carthage version`:  0.32.0
* `xcodebuild -version`: `N/A`
* Are you using `--no-build`? No
* Are you using `--no-use-binaries`?  No
* Are you using `--use-submodules`? No
* Are you using `--cache-builds`? Yes
* Are you using `--new-resolver`?  No

**Cartfile**
n/a
**Carthage Output**
n/a

**Steps**
* git clone https://github.com/ReactiveX/RxSwift.git

**Actual outcome**
* When running `carthage build --no-skip-current --platform iOS --cache-build` the `Carthage/Build/` directory is created but the content of it shows no version file. I only see the `iOS` sub-directory with the content of the RxSwift artefacts. Where should one be able to locate the version?
[...]

PR #2636, opened 2018-11-05
Produce .version file for current project

blender:
This PR addresses #2634 

~~The core method assumes that __the last path component of the current directory is the the name of the current project__ as it would appear in the Cartfile.~~

The core method derives the name of the current project as it would appear in the Cartfile. form `git remote get-url origin`

I'm open to feedback and to discuss other ideas.

### DIFF
--- a/Source/CarthageKit/VersionFile.swift
+++ b/Source/CarthageKit/VersionFile.swift
@@ -252,9 +252,38 @@ public func createVersionFileForCurrentProject(
 	buildProducts: [URL],
 	rootDirectoryURL: URL
 ) -> SignalProducer<(), CarthageError> {
-	let currentProjectAsDependecy = launchGitTask(["remote", "get-url", "origin"])
+
+	// List all remotes known for this repository
+	let allRemotes = launchGitTask(["remote"])
+		.flatMap(.concat) { $0.linesProducer }
 		.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
-		.map { Dependency.git(GitURL($0)) }
+
+	// For each remote, get the `get-url`
+	let allRemoteURLs = allRemotes
+		.flatMap(.concat) { launchGitTask(["remote", "get-url", $0]) }
+		.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+		.collect()
+
+	let currentProjectAsDependecy = allRemoteURLs
+		// Assess the popularity of each remote url
+		.map { $0.reduce([String: Int]()) { map, remoteURL in
+			var remoteURLPopularityMap = map
+			remoteURLPopularityMap[remoteURL] = (map[remoteURL] ?? 0) + 1
+			return remoteURLPopularityMap
+			}
+		}
+		// Sort remotes by popularity or alphabetically in case of a draw
+		.map { (remoteURLPopularityMap: [String: Int]) -> (url: String, popularity: Int) in
+			remoteURLPopularityMap.sorted { lhs, rhs in
+				if lhs.value == rhs.value {
+					return lhs.key < rhs.key
+				}
+				return lhs.value > rhs.value
+			}
+		    // This should never happen... but use a fallback anyways
+			.first ?? ("_Current", 0)
+		}
+		.map { Dependency.git(GitURL($0.url)) }
 
 	let currentGitTagOrCommitish = launchGitTask(["rev-parse", "HEAD"])
 		.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
--- a/Source/CarthageKit/VersionFile.swift
+++ b/Source/CarthageKit/VersionFile.swift
@@ -148,6 +148,7 @@ struct VersionFile: Codable {
 					return SignalProducer(value: true)
 				} else {
 					return frameworkSwiftVersion(frameworkURL)
+						.flatMapError { _ in dSYMSwiftVersion(frameworkURL.appendingPathExtension("dSYM")) }
 						.map { swiftVersion -> Bool in
 							return swiftVersion == localSwiftVersion
 						}
--- a/Source/CarthageKit/VersionFile.swift
+++ b/Source/CarthageKit/VersionFile.swift
@@ -253,37 +253,66 @@ public func createVersionFileForCurrentProject(
 	rootDirectoryURL: URL
 ) -> SignalProducer<(), CarthageError> {
 
-	// List all remotes known for this repository
-	let allRemotes = launchGitTask(["remote"])
-		.flatMap(.concat) { $0.linesProducer }
-		.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+	/*
+	List all remotes known for this repository
+	and keep only the "fetch" urls by which the current repository
+	would be known for the purpose of fetching anyways.
 
-	// For each remote, get the `get-url`
-	let allRemoteURLs = allRemotes
-		.flatMap(.concat) { launchGitTask(["remote", "get-url", $0]) }
-		.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+	Example of well-formed output:
+
+		$ git remote -v
+		origin   https://github.com/blender/Carthage.git (fetch)
+		origin   https://github.com/blender/Carthage.git (push)
+		upstream https://github.com/Carthage/Carthage.git (fetch)
+		upstream https://github.com/Carthage/Carthage.git (push)
+
+	Example of ill-formed output where upstream does not have a url:
+
+		$ git remote -v
+		origin   https://github.com/blender/Carthage.git (fetch)
+		origin   https://github.com/blender/Carthage.git (push)
+		upstream
+	*/
+	let allRemoteURLs = launchGitTask(["remote", "-v"])
+		.flatMap(.concat) { $0.linesProducer }
+		.map { $0.components(separatedBy: .whitespacesAndNewlines) }
+		.filter { $0.count >= 3 && $0.last == "(fetch)" } // Discard ill-formed output as of example
+		.map { ($0[0], $0[1]) }
 		.collect()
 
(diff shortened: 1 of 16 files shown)


blender:
I didn't add test. Any ideas on how to test this?

Source/CarthageKit/VersionFile.swift
```diff
@@ -237,6 +237,40 @@ struct VersionFile: Codable {
...
+) -> SignalProducer<(), CarthageError> {
+	let currentGitTagOrCommitish = launchGitTask(["rev-parse", "HEAD"])
+		.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+		.flatMap(.merge) { headCommitish in
+			return launchGitTask(["describe", "--tags", "--exact-match", headCommitish])
+				.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+				.flatMapError { _  in SignalProducer(value: headCommitish) }
+		}
+
+	let dependecyName = !rootDirectoryURL.lastPathComponent.isEmpty ? rootDirectoryURL.lastPathComponent : "Current-Unknown"
```
mdiep: _dependency_ is misspelled here

mdiep:
> The core method assumes that **the last path component of the current directory is the the name of the current project** as it would appear in the Cartfile.

Would it make more sense to make this as `.version` or `._.version`?

blender:
The idea is to generate the file compatible with `--cache-builds` with no extra steps. `--cache-builds` assumes `<name-as-in-cartfile>.version`. 

`.version` or `._.version` would be a hidden files by default and I can already see the complaints incoming. 
I didn't go with  current `Current.version` because... I can already see someone having a dependency called `Current`.

I though that defaulting to the name of the current directory was less or a wild guess.

I am open to change it to a predictable name so that it can be renamed with a script in simple steps.

mdiep:
I think it'd be better _not_ to use the name of the current directory. That way you can copy the build directory around without it depending on the last component of the project directory. There are lots of reasons you might want to do that.

blender:
How about using the name of the current project or workspace? So

```
.<projectOrWorkspaceName>.version
```

mdiep:
That seems like it could lead to conflicts. 🤔 

Maybe we could use the primary remote (if it exists) and calculate the name just like we'd do for a dependency? Falling back to something like what you have now?

blender:
I’ll investigate and ping back.

blender:
@mdiep relevant bit: https://github.com/Carthage/Carthage/pull/2636/commits/5627347d38eb6b15ac917b37bf2dd1f16e2a3379#diff-6206d1bf8ed3195b3b646bc3d2b6553dR253

I assumed 1 remote (origin) is enough. If we get complaints we can check all remotes and pick the most common name.

Also squashed commits

blender:
@mdiep ping

Source/CarthageKit/VersionFile.swift
```diff
@@ -237,6 +237,43 @@ struct VersionFile: Codable {
...
+///
+/// Returns a signal that succeeds once the file has been created.
+public func createVersionFileForCurrentProject(
+	platforms: Set<Platform>,
+	buildProducts: [URL],
+	rootDirectoryURL: URL
+) -> SignalProducer<(), CarthageError> {
+	let currentProjectAsDependecy = launchGitTask(["remote", "get-url", "origin"])
+		.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+		.map { Dependency.git(GitURL($0)) }
```
mdiep: There's not always a remote named `origin`, so this should probably catch errors and default to another value.

Source/CarthageKit/Xcode.swift
```diff
@@ -818,15 +825,24 @@ public func buildInDirectory( // swiftlint:disable:this function_body_length
...
-				.flatMapError { _ in .empty }
+				if isCurrentProject {
+					return createVersionFileForCurrentProject(platforms: options.platforms,
+							buildProducts: urls,
+							rootDirectoryURL: rootDirectoryURL
+						)
+						.flatMapError { _ in .empty }
+				} else {
+					return createVersionFile(
+						for: dependency!.dependency,
```
mdiep: Instead of checking `== nil` above and then using `!` here, I'd prefer to use `if let` or `guard let`.

mdiep:
Looks good overall. 👍 Just a couple things to clean up.

Source/CarthageKit/VersionFile.swift
blender: ok so I will go the full way.

mdiep: Otherwise you could fallback to a string like "CurrentProject" or something

blender:
So yeah, it might have gotten a little out of hand. https://github.com/Carthage/Carthage/pull/2636/commits/3d80665891950f582f7fa6e307ced88a99d111c0

Anyways I am not fully happy because I would also notify the user that we're defaulting to `_Current.version` (which I picked semi arbitrarily). However this requires some rethinking of the logging.

mdiep:
I'm a little behind from Thanksgiving last week, but I'll try to review this tomorrow or at least sometime this week.

Source/CarthageKit/VersionFile.swift
```diff
@@ -237,6 +237,72 @@ struct VersionFile: Codable {
...
+	// List all remotes known for this repository
+	let allRemotes = launchGitTask(["remote"])
+		.flatMap(.concat) { $0.linesProducer }
+		.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+
+	// For each remote, get the `get-url`
+	let allRemoteURLs = allRemotes
+		.flatMap(.concat) { launchGitTask(["remote", "get-url", $0]) }
+		.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+		.collect()
```
mdiep: You could probably use `git remote -v` to get the URLs instead.

Source/CarthageKit/VersionFile.swift
```diff
@@ -237,6 +237,72 @@ struct VersionFile: Codable {
...
+		.map { (remoteURLPopularityMap: [String: Int]) -> (url: String, popularity: Int) in
+			remoteURLPopularityMap.sorted { lhs, rhs in
+				if lhs.value == rhs.value {
+					return lhs.key < rhs.key
+				}
+				return lhs.value > rhs.value
+			}
+		    // This should never happen... but use a fallback anyways
+			.first ?? ("_Current", 0)
+		}
```
mdiep: I think it'd be nice to default to `origin` if it exists.

* Most people will only have 1 remote.
* Most people with >1 remote will have an `origin`.
* If you don't have either, who knows what you'll have?

But choosing `origin` over the "popular" or alphabetical option seems like a better choice.

Source/CarthageKit/VersionFile.swift
```diff
@@ -237,6 +237,72 @@ struct VersionFile: Codable {
...
+		// Sort remotes by popularity or alphabetically in case of a draw
+		.map { (remoteURLPopularityMap: [String: Int]) -> (url: String, popularity: Int) in
+			remoteURLPopularityMap.sorted { lhs, rhs in
+				if lhs.value == rhs.value {
+					return lhs.key < rhs.key
+				}
+				return lhs.value > rhs.value
+			}
+		    // This should never happen... but use a fallback anyways
```
mdiep: Probably someone will try doing this before pushing their repo anywhere.

Source/CarthageKit/VersionFile.swift
blender: ah, nice. definitely.

Source/CarthageKit/VersionFile.swift
blender: ok

Source/CarthageKit/VersionFile.swift
blender: You bet. I will explain that. Should I keep `_Current` as fallback from `origin` ?

Source/CarthageKit/Xcode.swift
blender: Yes, sorry. I missed it.

Source/CarthageKit/VersionFile.swift
blender: Done

Source/CarthageKit/VersionFile.swift
blender: Done

blender:
@mdiep  I have made the changes. I have tests with:
- malformed remotes ✅
- no remotes ✅

Source/CarthageKit/VersionFile.swift
```diff
@@ -239,6 +239,101 @@ struct VersionFile: Codable {
...
+/// Derives the current project name from `git remote get-url origin`
+///
+/// Returns a signal that succeeds once the file has been created.
+public func createVersionFileForCurrentProject(
+	platforms: Set<Platform>,
+	buildProducts: [URL],
+	rootDirectoryURL: URL
+) -> SignalProducer<(), CarthageError> {
+
+	/*
```
blender: @mdiep cases I have covered

Source/CarthageKit/VersionFile.swift
```diff
@@ -239,6 +239,101 @@ struct VersionFile: Codable {
...
+			guard let origin = remotePopularityMap["origin"] else {
+				let urlOfMostPopularRemote = remotePopularityMap.sorted { lhs, rhs in
+					if lhs.value.popularity == rhs.value.popularity {
+						return lhs.key < rhs.key
+					}
+					return lhs.value.popularity > rhs.value.popularity
+				}
+				.first?.value.remoteNameAndURL.url
+
+				// If the reposiroty is not pushed to any remote
```
blender: @mdiep  Not a pushed repo like we discussed

mdiep:
Looks good! Thanks for all your work on this one!

PR #2643, opened 2018-11-15
Sync submodules from the submodule directory

mdiep:
Otherwise each submodule syncs all submodules. For projects using submodules with large numbers of dependencies, which may themselves have a large number of submodules, this can save quite a bit of time.

Tested on a project locally:

```
» time carthage checkout --use-submodules    
*** Checking out ReactiveCocoa at "8.0.0"
*** Checking out ReactiveSwift at "4.0.0"
*** Checking out Schemata at "0.2.0"
*** Checking out PersistDB at "ab564f1f91964c5e4f02bd790a454ce950ccb5cb"
*** Checking out Result at "4.0.0"
*** Checking out Differ at "1.3.0"
*** Checking out Anchorage at "4.3"
*** Checking out Standards at "89c30330fcac972a84a4876125064001b9c89f0f"
carthage checkout --use-submodules  2.49s user 2.87s system 99% cpu 5.386 total
» time carthage-dev checkout --use-submodules
*** Checking out Anchorage at "4.3"
*** Checking out ReactiveSwift at "4.0.0"
*** Checking out Standards at "89c30330fcac972a84a4876125064001b9c89f0f"
*** Checking out Result at "4.0.0"
*** Checking out Differ at "1.3.0"
*** Checking out ReactiveCocoa at "8.0.0"
*** Checking out Schemata at "0.2.0"
*** Checking out PersistDB at "ab564f1f91964c5e4f02bd790a454ce950ccb5cb"
carthage-dev checkout --use-submodules  1.76s user 1.79s system 103% cpu 3.420 total
```

(`carthage-dev` includes the changes in this PR, but is the Debug configuration)

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -493,7 +493,7 @@ public func addSubmoduleToRepository(_ repositoryFileURL: URL, _ submodule: Subm
 							repositoryFileURL: repositoryFileURL
 						)
 					)
-					.then(launchGitTask([ "submodule", "--quiet", "sync", "--recursive" ], repositoryFileURL: repositoryFileURL))
+					.then(launchGitTask([ "submodule", "--quiet", "sync", "--recursive" ], repositoryFileURL: submoduleDirectoryURL))
 					.then(checkoutSubmodule(submodule, submoduleDirectoryURL))
 					.then(launchGitTask([ "add", "--force", submodule.path ], repositoryFileURL: repositoryFileURL))
 					.then(SignalProducer<(), CarthageError>.empty)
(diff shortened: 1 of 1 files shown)


blender:
@mdiep I am not a submodule heavy-user so just to clarify this change will sync each depedency's submodule separately?

mdiep:
Yup!

Let's say you have 2 submodules: A and B.

Before, this would run `git submodule sync` twice, which would mean that `A` and `B` would both be synced twice. After, this runs `git submodule sync A` once and `git submodule sync B`, so each only gets synced once.

As the number of submodules goes up, you get more and more savings from removing the repetition.

Issue #2616, opened 2018-10-17
Dependencies from different forks


CosynPa:
* carthage install method: [] .pkg, [*] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.31.1
* `xcodebuild -version`: Xcode 10.0 Build version 10A255
* Are you using `--no-build`? Yes
* Are you using `--no-use-binaries`? No
* Are you using `--use-submodules`? No
* Are you using `--cache-builds`? No
* Are you using `--new-resolver`? No

**Cartfile**
```
github "ReactiveCocoa/ReactiveSwift" # This repository depends on antitypical/Result
github "CosynPa/Result" # This is just a dummy repository 
```

**Carthage Output**
```
*** Fetching Result
*** Fetching ReactiveSwift
*** Fetching Result
*** Checking out Result at "4.0.0"
*** Checking out ReactiveSwift at "4.0.0"
*** Checking out Result at "0.0.1"
A shell task (/usr/bin/env git checkout --quiet --force 4.0.0 (launched in /Users/xxx/Library/Caches/org.carthage.CarthageKit/dependencies/Result)) failed with exit code 128:
fatal: Unable to create '/Users/xxx/Library/Caches/org.carthage.CarthageKit/dependencies/Result/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.
```

**Actual outcome**
A fatal error

Cartfile.resolved is
```
github "CosynPa/Result" "0.0.1"
github "ReactiveCocoa/ReactiveSwift" "4.0.0"
github "antitypical/Result" "4.0.0"
```

**Expected outcome**
Carthage should report an error when resolving dependencies, or resolve to a single Result repository.

Although the Cartfile shown above is a contrived one, using a different fork from the standard one is somewhat common practice as in #1987. The dependency resolver should find a compatible version in one fork or report an error if that is impossible. Currently, Carthage resolves out two repositories, but since they have the same name, Carthage tries to check out the same cached repository to 0.0.1 and 4.0.0 simultaneously, that's why a race condition occurs.

blender:
Thanks for opening an issue.

This is well known as repositories are not namespaced by organization or owner.

PR #2644, opened 2018-11-16
Continue to fix cache failing to resolve dependencies with name collision

### LINKED ISSUE #2150: Cache failing to resolve dependencies with name collision

CosynPa:
This continues #2286. In this pull request, any repository "owner/project" is cloned in the directory "project/owner/". So if two repositories have the same name but different owners, they will be cloned in different directories without conflicting.

But I don't think it actually fixes #2150. I'm not sure if that can even be fixed. In #2150 they want different libraries with the same name to exist simultaneously. But when you write "import library" in Xcode, how can Xcode know which library you want?

Another maybe relevant issue is #2616. This pull request doesn't fix it either. As I mentioned in that issue, the root cause is the resolver. Although when this pull request is applied, the fatal error in that issue will disappear, carthage will actually check out the two different repositories into the same working directory (Carthage/Checkouts/Result) leaving mixed contents of the two repositories in the working directory, which may not be what we want.

So what does this pull request solve or improve anyway? It can improve fetching time a little. When people switch forks and then switch back, they don't have to redownload all the branches in the old fork again.


CosynPa:
I fixed trailing whitespace and other warnings.

CosynPa:
Anyone review this? Without this, carthage can easily get confused when there are dependencies of different forks. With this pull request, in situations like #2616, at least carthage won't quit with error, so the user can check out the right dependency manually. This is a huge improvement.

Issue #2651, opened 2018-11-28
Is it possible to avoid remove and clone submodule again every time we checking out?


ainopara:
* carthage install method: [ ] .pkg, [✓] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.31.2
* `xcodebuild -version`: Xcode 10.1 Build version 10B61
* Are you using `--no-build`?  no
* Are you using `--no-use-binaries`?  yes
* Are you using `--use-submodules`?  no
* Are you using `--cache-builds`?  yes
* Are you using `--new-resolver`?  no

**Cartfile**
```
github "SnapKit/SnapKit" ~> 4.0
github "ReactiveX/RxSwift" ~> 4.0
github "RxSwiftCommunity/RxGesture" ~> 1.1.1
github "RxSwiftCommunity/RxDataSources" ~> 3.0
```

**Carthage Output**
```
N/A
```
** Command **
```
carthage bootstrap --platform iOS --no-use-binaries --cache-builds
```
**Actual outcome**
```
*** Checking out RxGesture at "1.2.1"
*** Checking out RxSwift at "4.4.0"
*** Checking out RxDataSources at "3.1.0"
*** Checking out SnapKit at "4.2.0"
<- Waiting here to clone RxSwift twice.
...... (Omitting other output not related to this issue)
```

**Expected outcome**
Carthage should complete check out phase immediately since nothing changed (Cartfile.resolved not changed) here. But in this example, both RxGesture and RxDataSources have a RxSwift submodule.
In CarthageKit/Git.swift we have:
```swift
/// Clones the given submodule into the working directory of its parent
/// repository, but without any Git metadata.
public func cloneSubmoduleInWorkingDirectory(_ submodule: Submodule, _ workingDirectoryURL: URL) -> SignalProducer<(), CarthageError> {
	let submoduleDirectoryURL = workingDirectoryURL.appendingPathComponent(submodule.path, isDirectory: true)

	func repositoryCheck<T>(_ description: String, attempt closure: () throws -> T) -> Result<T, CarthageError> {
		do {
            print("succeed: \(description)")
			return .success(try closure())
		} catch let error as NSError {
			let reason = "could not \(description)"
			return .failure(
				.repositoryCheckoutFailed(workingDirectoryURL: submoduleDirectoryURL, reason: reason, underlyingError: error)
			)
		}
	}

	let purgeGitDirectories = FileManager.default.reactive
[...]

PR #2655, opened 2018-12-02
Improve submodule operation performance

### LINKED ISSUE #1833: Checkout spends a lot of time cloning submodules
* `carthage version`: 0.20.1
* `xcodebuild -version`: 8.2.1(8C1002)
* Are you using `--no-build`? YES
* Are you using `--no-use-binaries`? YES
* Are you using `--use-submodules`? NO
* Are you using `--cache-builds`? NO

**Cartfile**
```
```

I'm running `carthage checkout --no-use-binaries`, it is slow.

I insert some log in the `cloneSubmoduleInWorkingDirectory` signal producer:

screen shot 2017-03-15 at 10 02 58 am

This is the output:

```

[34;1m***[0m Checking out [1mResult[0m at [32m"3.2.1"[0m
[34;1m***[0m Checking out [1mReactiveSwift[0m at [32m"1.1.0"[0m
2017-03-14 13:53:58.433 carthage[57473:1990312] start clone submodule Carthage/Checkouts/Nimble @ 8e2703fabde578a6d28537c73844f84d48e71200
2017-03-14 13:53:58.458 carthage[57473:1990312] start clone submodule Carthage/Checkouts/Quick @ e4fa1e85c0305ba4e0866f25812d3fa398f3a048
2017-03-14 13:53:58.472 carthage[57473:1990312] start clone submodule Carthage/Checkouts/xcconfigs @ 3d9d99634cae6d586e272543d527681283b33eb0
2017-03-14 13:53:58.489 carthage[57473:1990307] start clone submodule Carthage/Checkouts/Result @ 2dd1a0c7bca4cc44082f72020fe8b14491056f88
2017-03-14 13:54:06.657 carthage[57473:1990308] complete clone submodule Carthage/Checkouts/xcconfigs @ 3d9d99634cae6d586e272543d527681283b33eb0
2017-03-14 13:54:23.253 carthage[57473:1990308] complete clone submodule Carthage/Checkouts/Result @ 2dd1a0c7bca4cc44082f72020fe8b14491056f88
2017-03-14 13:56:19.237 carthage[57473:1993541] complete clone submodule Carthage/Checkouts/Nimble @ 8e2703fabde578a6d28537c73844f84d48e71200
2017-03-14 14:00:21.046 carthage[57473:2009707] complete clone submodule Carthage/Checkouts/Quick @ e4fa1e85c0305ba4e0866f25812d3fa398f3a048
[34;1m***[0m Checking out [1mReactiveCocoa[0m at [32m"5.0.1"[0m
2017-03-14 14:00:21.164 carthage[57473:2009731] start clone submodule Carthage/Checkouts/Nimble @ 3720e6b0f6de6d6435f79f8a174fb4bb92df5dc4
2017-03-14 14:00:21.204 carthage[57473:2009755] start clone submodule Carthage/Checkouts/Quick @ e4fa1e85c0305ba4e0866f25812d3fa398f3a048
2017-03-14 14:00:21.276 carthage[57473:2009755] start clone submodule Carthage/Checkouts/xcconfigs @ 3d9d99634cae6d586e272543d527681283b33eb0
2017-03-14 14:00:21.315 carthage[57473:2009749] start clone submodule Carthage/Checkouts/Result @ 2dd1a0c7bca4cc44082f72020fe8b14491056f88
2017-03-14 14:00:21.349 carthage[57473:2009749] start clone submodule Carthage/Checkouts/ReactiveSwift @ c83de32f1adbcef89618cde4222f69a8bcecd7c2
2017-03-14 14:00:35.112 carthage[57473:2010340] complete clone submodule Carthage/Checkouts/xcconfigs @ 3d9d99634cae6d586e272543d527681283b33eb0
2017-03-14 14:00:43.678 carthage[57473:2010338] complete clone submodule Carthage/Checkouts/Result @ 2dd1a0c7bca4cc44082f72020fe8b14491056f88
2017-03-14 14:03:50.218 carthage[57473:2017892] complete clone submodule Carthage/Checkouts/Nimble @ 3720e6b0f6de6d6435f79f8a174fb4bb92df5dc4
2017-03-14 14:11:40.794 carthage[57473:2028920] complete clone submodule Carthage/Checkouts/Quick @ e4fa1e85c0305ba4e0866f25812d3fa398f3a048
2017-03-14 15:01:01.880 carthage[57473:2180717] complete clone submodule Carthage/Checkouts/ReactiveSwift @ c83de32f1adbcef89618cde4222f69a8bcecd7c2

```

We can see most of the time is spent cloning submodules, and the largest submodule is ReactiveSwift. But at the second line of log checking out ReactiveSwift itself only takes a few seconds. And also some submodules are duplicated, both ReactiveSwift and ReactiveCocoa have Nimble, Quick, Result, and xcconfigs as their submodules.

Many libraries are using submodules for managing dependencies, thus many of the submodules are already fetched in the Carthage cache. I believe at least cloning of these submodules can be optimized.

CosynPa:
Fixes #1833 and #2651.

Better wait after #2654 is merged, then rebase on it.


blender:
You can rebase

CosynPa:
Rebased.

blender:
I don't work with submodules enough to feel qualified to review the change. I think @mdiep should be reviewing here.

@CosynPa Ad high level description of the problem and the solution would certainly sleep up the review.

CosynPa:
The problem is, as is pointed out in the two issues, the carthage checkout operation will try cloning submodules of dependencies, which takes a lot of time. 

There are two functions that will clone or fetch submodules, one is `cloneRepository` which will be called when the `--use-submodules` option is not used, the other is `checkoutSubmodule` which will be called both with or without the `--use-submodules` option.

My solution is that first clone or fetch submodules to the local cache directory. Then redirect the slow network operations to the local cache. For `checkoutSubmodule`, which will do a `git submodule update --init --recursive`, the solution is a little bit tricky. I first change the .gitmodules file, so that the URLs of the submodules are pointing to the local cache repository, then do the `git submodule update --init`, and change the URLs back.

I also serialized heavy repository write operations, so cloning the same repository twice can checking out the same repository twice can be avoided.

mdiep:
I'll try to get to this ASAP. I'm on vacation right now, but I'll see if I can find some time to review this.

Source/CarthageKit/Git.swift
```diff
@@ -153,6 +153,7 @@ public func contentsOfFileInRepository(_ repositoryFileURL: URL, _ path: String,
 public func checkoutRepositoryToDirectory(
 	_ repositoryFileURL: URL,
 	_ workingDirectoryURL: URL,
+	force: Bool,
```
mdiep: Why was this added/needed?

Source/CarthageKit/Project.swift
```diff
@@ -1165,6 +1183,173 @@ public final class Project { // swiftlint:disable:this type_body_length
 				return incompatibilities.isEmpty ? .init(value: ()) : .init(error: .invalidResolvedCartfile(incompatibilities))
 			}
 	}
+
+	/// Clones the given submodule into the working directory of its parent
+	/// repository, but without any Git metadata.
+	public func cloneSubmoduleInWorkingDirectory(_ submodule: Submodule, _ workingDirectoryURL: URL) -> SignalProducer<(), CarthageError> {
```
mdiep: I don't think these functions should move here:

1. These are generic functions that shouldn't really be tied to the `Project`
2. Because these were moved, the diff for this PR is really large and almost impossible to review

Can you move them back to the Git file?

Source/CarthageKit/Project.swift
```diff
@@ -79,6 +79,24 @@ extension ProjectEvent: Equatable {
...
+			if let queue = queues[url] {
+				return queue
+			} else {
+				let newQueue = SerialProducerQueue(name: "org.carthage.Constants.Project.cloneOrFetchQueue-\(url.absoluteString)")
+				queues[url] = newQueue
+				return newQueue
+			}
+		}
+	}
+}
```
mdiep: This global state makes me very uncomfortable. I don't think it's a good idea to introduce a `static var` with this. It leads to code that's hard to follow and modify.

We should use another method to enforce exclusivity. I think it makes sense to do something like this:

1. Fetch all the submodules from the repository cache, in parallel but with limited concurrency.
2. Perform the actual checkouts, passing in a queue or doing them one at a time

Source/CarthageKit/Project.swift
```diff
@@ -1165,6 +1183,173 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+						.then(addSubmodule)
+						.then(launchGitTask([ "submodule", "--quiet", "init", "--", submodule.path ], repositoryFileURL: repositoryFileURL))
+						.startOnQueue(RepositoryOperationQueues.queue(forLocalRepositoryURL: submoduleDirectoryURL))
+						.then(self.updateSubmodules(submoduleDirectoryURL))
+				}
+			}
+	}
+
+	public func updateSubmodules(_ superprojectDirectoryURL: URL) -> SignalProducer<(), CarthageError> {
+		func changeGitmodulesUrls(toCacheURL: Bool) -> SignalProducer<(), CarthageError> {
```
mdiep: I don't think this is necessary. The original code in `addSubmoduleToRepository` already handled doing the equivalent of `git submodule update --init` from the repository cache.

If your intent is to make that logic recursive, I think you can do that following the existing method without modifying `.gitmodules`.

mdiep:
First off: Thanks for the PR!

I think the overall idea—make sure we're using the repository cache for the submodules of dependencies—is very good. That would be a very nice contribution.

But I'm not comfortable with the current approach. I think it needs to be reworked to (1) be more reviewable and (2) avoid changing `.gitmodules`.

> I also serialized heavy repository write operations, so cloning the same repository twice and checking out the same repository twice can be avoided.

Is this necessary for the primary change described above? If not, let's leave it for a followup PR. Bundling changes together makes PRs harder, and thus slower, to review—which is worse experience for everyone.

Source/CarthageKit/Git.swift
CosynPa: The original `checkoutRepositoryToDirectory` function was a forced checkout, and in `checkoutSubmodule` there was a non-forced checkout. So to reuse the `checkoutRepositoryToDirectory` function, I added the force parameter.

Source/CarthageKit/Project.swift
CosynPa: These functions were moved because the new implementation depends on `repositoryFileURL` which is a private function in Project.swift.

Source/CarthageKit/Project.swift
CosynPa: I see there are already a lot of queues, but they are instance properties of a Project. So I just need to move the global variable to the Project class?

Source/CarthageKit/Project.swift
mdiep: Then that should be passed in as a parameter

CosynPa: What about `cloneOrFetchDependency`? These functions actually rely on `cloneOrFetchDependency` which uses `cloneOrFetchQueue` and `_projectEventsObserver` which are `Project` properties. I think some kind of queue is definitely needed. So also pass the queue in as a parameter?

CosynPa:
I'd like to separate this PR into two. The first is #2795.

CosynPa:
I've separated this PR into three: #2795, #2832 and #2842.

CosynPa:
I've separated this PR into three: #2795, #2832 and #2842.

1ec5:
Looks like both #2795 and #2832 have been merged, but #2842 got closed as stale.

PR #2664, opened 2018-12-22
Build with SwiftPM without specifying the deployment target

mdiep:
I want to start building **only** with SwiftPM. This makes it a little easier to do so.


Source/CarthageKit/Dependency.swift
```diff
@@ -148,7 +148,10 @@ extension Dependency: Scannable {
 					if url.scheme == "https" || url.scheme == "file" {
 						return .success(self.binary(BinaryURL(url: url, resolvedDescription: url.description)))
 					} else if url.scheme == nil {
-						let absoluteURL = URL(fileURLWithPath: url.relativePath, isDirectory: false, relativeTo: base).standardizedFileURL
+						// This can use URL.init(fileURLWithPath:isDirectory:relativeTo:) once we can target 10.11+
+						let absoluteURL = url.relativePath
+							.withCString { URL(fileURLWithFileSystemRepresentation: $0, isDirectory: false, relativeTo: base) }
+							.standardizedFileURL
```
mdiep: This should be equivalent. https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURL.swift#L359-L363

Carthage.xcodeproj/project.pbxproj
```diff
@@ -1091,7 +1091,7 @@
 				CLANG_WARN_STRICT_PROTOTYPES = YES;
 				CLANG_WARN_SUSPICIOUS_MOVE = YES;
 				GCC_NO_COMMON_BLOCKS = YES;
-				MACOSX_DEPLOYMENT_TARGET = 10.11;
+				MACOSX_DEPLOYMENT_TARGET = 10.10;
```
blender: don't we have a dependency on Xcode 9.4? This requires macOS 10.13.2+

Carthage.xcodeproj/project.pbxproj
```diff
@@ -815,7 +815,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/bash;
-			shellScript = ". script/update-cli-version";
+			shellScript = ". script/update-cli-version\n";
```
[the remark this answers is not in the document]
mdiep: This is Xcode 10's doing. These get added when you edit the project file with Xcode 10 for the first time.

Carthage.xcodeproj/project.pbxproj
mdiep: Yes, but this is the _deployment_ target. You need macOS 10.13.2+ to build, but you can use the built binary on older platforms.

SwiftPM doesn't currently have a way to specify a deployment target and defaults to 10.10. This just aligns the Xcode project with the SwiftPM default.

PR #2665, opened 2018-12-24
Only build with SwiftPM

mdiep:
Includes the changes from #2664.

This removes the Xcode project, Xcode workspace, `Cartfile`s, submodules, etc—building **only** with SwiftPM.

If we want to [use parts of SwiftPM in Carthage](https://forums.swift.org/t/using-swiftpm-from-carthage/18912), it will be easiest to include SwiftPM as a dependency _with_ SwiftPM.


mdiep:
cc @Carthage/carthage because this will affect development

NachoSoto:
🤩

CosynPa:
Can we still debug carthage in Xcode after merging this?

mdiep:
Yes, you can generate an Xcode project and debug with it.

Makefile
```diff
@@ -23,85 +15,55 @@ ZSH_COMMAND := ZDOTDIR='/var/empty' zsh -o NO_GLOBAL_RCS -c
...
-	xcodebuild $(XCODEFLAGS) -configuration Release ENABLE_TESTABILITY=YES test
+all: installables
 
 clean:
-	$(RM) "$(INTERNAL_PACKAGE)"
-	$(RM) "$(OUTPUT_PACKAGE)"
-	$(RM) "$(OUTPUT_FRAMEWORK_ZIP)"
-	$(RM_SAFELY) "$(CARTHAGE_TEMPORARY_FOLDER)"
-	xcodebuild $(XCODEFLAGS) clean
+	$(RM_SAFELY) "./.build/"
```
ikesyo: I think `swift package clean` will be good.

mdiep:
Good catch! I amended and force-pushed so I could remove `CarthageApp.pkg` from history

PR #2668, opened 2018-12-27
Show error when input file list is not readable

ferranpujolcamins:
When one of the files specified in the `SCRIPT_INPUT_FILE_LIST_@` envvars is not readable, the `copy-frameworks` command fails with the following unhelpful error:
```
Could not cast value of type 'NSError' (...) to 'Result.NoError' (...).
```

This PR fixes this issue and now Carthage displays the `readFailed` error.


blender:
@ferranpujolcamins thanks 👏 !  well done!

PR #2670, opened 2018-12-30
Use Version type from SwiftPM

mdiep:
Depends on #2665.

This is the first step in building more off of SwiftPM. I think if we do this, then use their equivalent of `VersionSpecifier`, we can start reading `Package.resolved` and/or using the Swift PM resolver. Using common types will make those things easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,19 +12,13 @@ matrix:
   include:
     - os: osx
       language: objective-c
-      osx_image: xcode9.4
-      script:
-        - make test
-      env: JOB=CI_TEST_9_4
-    - os: osx
-      language: objective-c
       osx_image: xcode10.1
       script:
         - make test
       env: JOB=CI_TEST_10_1
     - os: osx
       language: objective-c
-      osx_image: xcode9.4
+      osx_image: xcode10.1
       script:
         - brew uninstall carthage
         - HOMEBREW_NO_AUTO_UPDATE=1 brew install bats
--- a/Source/CarthageKit/BinaryProject.swift
+++ b/Source/CarthageKit/BinaryProject.swift
@@ -1,5 +1,8 @@
 import Foundation
 import Result
+import Utility
+
+import struct Foundation.URL
 
 /// Represents a binary dependency 
 public struct BinaryProject: Equatable {
@@ -15,7 +18,7 @@ public struct BinaryProject: Equatable {
 
 				for (key, value) in json {
 					let pinnedVersion: PinnedVersion
-					switch SemanticVersion.from(Scanner(string: key)) {
+					switch Version.from(Scanner(string: key)) {
 					case .success:
 						pinnedVersion = PinnedVersion(key)
 					case let .failure(error):
--- a/IntegrationTests/update.bats
+++ b/IntegrationTests/update.bats
@@ -4,8 +4,8 @@ setup() {
     cd $BATS_TMPDIR
     rm -rf UpdateTest
     mkdir UpdateTest && cd UpdateTest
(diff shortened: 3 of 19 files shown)


.gitignore
```diff
@@ -22,4 +22,4 @@ DerivedData
 
 Carthage.pkg
 CarthageApp.pkg
-Carthage.xcodeproj
+Carthage.xcodeproj/
```
mdiep: This was incorrectly specified in my previous PR, so the generated project was accidentally checked in. I've removed it below.

Source/CarthageKit/BinaryProject.swift
```diff
@@ -1,5 +1,8 @@
 import Foundation
 import Result
+import Utility
+
+import struct Foundation.URL
```
mdiep: This disambiguation is necessary because `Utility` defines its own `URL` type.

mdiep:
I'm not sure why this is failing on CI. The tests always pass locally and I did get [one green build](https://travis-ci.org/Carthage/Carthage/jobs/475954169). 😕

.travis.yml
```diff
@@ -24,7 +18,7 @@ matrix:
       env: JOB=CI_TEST_10_1
     - os: osx
       language: objective-c
-      osx_image: xcode9.4
+      osx_image: xcode10.1
```
blender: I am a little confused now as what version of Xcode we required to build.

Requiring 10.1 cuts out High Sierra homebrew builds

mdiep: This PR would raise the requirement from 9.4 to either 10.0 or 10.1. BUT our deployment target is still 10.10, so you only need a newer OS to _build_. (And I think you could build on an older OS if you installed a newer `swift` first.)

SwiftPM's `master` requires Swift 4, which is the reason for this change.

We're using `master` because I had to submit a PR to add a static library version of SwiftPM. I could do that on a fork of SwiftPM against an older release, but I think we'll want to get more changes from SwiftPM `master`.

I think that building off of SwiftPM is a big enough advantage that dropping support for building on 10.11 isn't a big deal. But I'm open to other opinions here. @Carthage/carthage

blender: >  dropping support for building on 10.11 isn't a big deal.

We don't have statistics so it's hard to say. I for one just updated last week since in my organization's IT department wouldn't let us.

Once this is released we effectively cut out whoever is relying on homebrew but on 10.10. 

We could release a one time alternative formula of Carthage (cathage@0.31.2) at the current version which would still allow homebrew builds of 0.31.2.

mdiep: Shouldn't people be able to install older versions of Carthage already?

If you're on an older OS, you have to use older toolchains anyway, so there shouldn't be a pressing need to upgrade your Carthage. And since you could install through the `pkg` installer, I'm not bothered by it—you can still upgrade if you need to.

I know this is potentially inconvenient, but I think the wins from building off of SwiftPM are too great not to move forward with this.

blender: I'm not suggesting to hold back at all :) 

Sure the .pkg works but I think we can still keep the convenience offered by homebrew by adopting the practice of releasing an `end-of-support` formula every time a version of Xcode requires an OS update (which is itself annoying) .

blender: Postgres for example has:

- https://github.com/Homebrew/homebrew-core/blob/master/Formula/postgresql%409.4.rb
- https://github.com/Homebrew/homebrew-core/blob/master/Formula/postgresql%409.5.rb 
- https://github.com/Homebrew/homebrew-core/blob/master/Formula/postgresql.rb (latest and greatest)

PR #2681, opened 2019-01-12
remove wrong xcscheme file existence check

omochi:
Xcode 10.1 (10B61) creates no xcscheme when xcodeproj is created.
But it treats that there is shared framework scheme.
Carthage wrongly say `Dependency "car" has no shared framework schemes` because of it assume that xcscheme file exists always when there is shared scheme in xcodeproj.
So this PR remove this check.

This is xcodeproj files dump.

``` 
$ ls -R
car           car.xcodeproj

./car:
Car.swift  Info.plist car.h

./car.xcodeproj:
project.pbxproj     project.xcworkspace xcuserdata

./car.xcodeproj/project.xcworkspace:
contents.xcworkspacedata xcshareddata             xcuserdata

./car.xcodeproj/project.xcworkspace/xcshareddata:
IDEWorkspaceChecks.plist

./car.xcodeproj/project.xcworkspace/xcuserdata:
omochi.xcuserdatad

./car.xcodeproj/project.xcworkspace/xcuserdata/omochi.xcuserdatad:
UserInterfaceState.xcuserstate

./car.xcodeproj/xcuserdata:
omochi.xcuserdatad

./car.xcodeproj/xcuserdata/omochi.xcuserdatad:
xcschemes

./car.xcodeproj/xcuserdata/omochi.xcuserdatad/xcschemes:
xcschememanagement.plist
```

and this is xcodebuild `-list` output.

```
$ xcodebuild -project car.xcodeproj -list
Information about project "car":
    Targets:
        car

    Build Configurations:
        Debug
        Release

    If no build configuration is specified and -scheme is not passed then "Release" is used.

    Schemes:
        car

```

and this is screenshot.

ss

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -230,14 +230,6 @@ public func buildableSchemesInDirectory( // swiftlint:disable:this function_body
 /// the project.
 public func schemesInProjects(_ projects: [(ProjectLocator, [Scheme])]) -> SignalProducer<[(Scheme, ProjectLocator)], CarthageError> {
 	return SignalProducer<(ProjectLocator, [Scheme]), CarthageError>(projects)
-		.map { (project: ProjectLocator, schemes: [Scheme]) in
-			// Only look for schemes that actually reside in the project
-			let containedSchemes = schemes.filter { scheme -> Bool in
-				let schemePath = project.fileURL.appendingPathComponent("xcshareddata/xcschemes/\(scheme).xcscheme").path
-				return FileManager.default.fileExists(atPath: schemePath)
-			}
-			return (project, containedSchemes)
-		}
 		.filter { (project: ProjectLocator, schemes: [Scheme]) in
 			switch project {
 			case .projectFile where !schemes.isEmpty:


blender:
@omochi  Test are failing on this PR.

mdiep:
This is an Xcode bug. If you uncheck and re-check the _Shared_ checkbox, a shared scheme should actually appear on disk.

omochi:
@blender

I checked testing failure.
It comes from `NoSharedSchemesTest.zip` resource.
In my opinion, test code here is wrong.
I will fix test if basic idea of this PR is accepted.

@mdiep

Why do you think this is bug?

Even if `.xcscheme` file is not exists,
Xcode shows that there is shared schemes in scheme management dialog.
It can build this scheme ordinally.
`xcodebuild -list` command prints schemes ordinally.

And this happens super easy just you do below.

- Open Xcode 10.1
- `File` > `New` > `Project...` > `iOS` > `Cocoa Touch Framework`
  `Next` > fill `Product Name:` > `Next` > Select any directory  
- In this point, you get shread schemes without `.xcscheme`

I think that the wrong is assumption about `.xcscheme` file existance of carthage team.

I actually struggled this behavior when I make my library supporting carthage.
Because the reason of error which is absent of `xcscheme` is completely invisible from me.
I checked `xcodebuild -list` and confused.
Finally I build carthage by myself and use debugger to track how it fails.

mdiep:
> * Open Xcode 10.1
> * `File` > `New` > `Project...` > `iOS` > `Cocoa Touch Framework`
>   `Next` > fill `Product Name:` > `Next` > Select any directory
> * In this point, you get shread schemes without `.xcscheme`

At this point, `xcodebuild -list` will return the scheme. That's because _Autocreate Schemes_ is turned on. If you turn it off, then `xcodebuild -list` doesn't return the scheme.

_Maybe_ that'd be okay. But there have been plenty of bugs (and hangs) with that in the past, so I'd prefer to not rely on it.

Also, if you uncheck _Shared_ and check it again, then it _does_ create the `xcscheme` file.

Finally, if you:

1. Follow your steps
2. Turn off _Autocreate schemes_
3. Delete the scheme that was generated
4. Click _Autocreate Schemes Now_

Then you'll see that the Shared checkbox is checked but `xcodebuild -list` doesn't return the scheme.

This is definitely an Xcode bug.

mdiep:
I filed rdar://47274589

ikesyo:
This is related to #488 and #655. If we remove the logic, it leads to that non-shared, user-specific schemes will be built those reside in `xcuserdata` directory.

omochi:
@mdiep @ikesyo Thanks to explain details.

PR #2689, opened 2019-01-23
add link to binary specification JSON

DivineDominion:
I wondered where the JSON contents are explained at first, so I think a link might help those who don't just scroll all the way down when puzzled :)


mdiep:
This is great, thanks! 👏🏻 These are some of the best enhancements.

Issue #2691, opened 2019-01-25
Unable to find iOS Simulators (Post Xcode 10.2 install)


msaps:
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `which carthage`: `/usr/local/bin/carthage`
* `carthage version`: 0.31.2
* `xcodebuild -version`: 10.1

After installing Xcode 10.2, Carthage fails to build any project due to `Could not find any available simulators for iOS`. I've tried a number of different things to resolve it:

- Build a project in Xcode 10.2 on a simulator (suggested in #2690)
- Use `Xcode.app` and `Xcode-beta.app`for `xcode-select`.
- Reset all my simulators etc.

I think the problem is the keys that are in the output of `simctl`. Looking through [Simulator.swift](https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/Simulator.swift) in CarthageKit I can see that it expects to find a platform prefix for the key (i.e. `iOS`) - my keys however are formatted like this: `com.apple.CoreSimulator.SimRuntime.iOS-12-1`.

Is this a problem with my installation? Or is it something that has changed in Xcode 10.2 and should be supported in Carthage?

Example `xcrun simctl list devices --json` output:

```
"com.apple.CoreSimulator.SimRuntime.iOS-12-1" : [
      {
        "availability" : "(available)",
        "state" : "Shutdown",
        "isAvailable" : true,
        "name" : "iPhone 5s",
        "udid" : "9AC9332D-7DB0-4F2E-8615-61EE86E363CA",
        "availabilityError" : ""
      },
      {
        "availability" : "(available)",
        "state" : "Shutdown",
        "isAvailable" : true,
        "name" : "iPhone 6",
        "udid" : "B1DA5CFF-D918-467B-8E49-04B6172B9456",
        "availabilityError" : ""
      }
]
```

blender:
Are you installing Xcode via MAS or developer portal?

I don't think it's and Xcode 10.2 specific issue but something related to your installation. I have no idea however why something like this would happen.

msaps:
@blender I run the stable version of Xcode through the MAS - I'll try nuking everything and see what if that resolves it. Also maybe avoid the MAS 🤣

blender:
Yeah, I personally don't use MAS and I have never so far encountered issues like this. I don't want to correlated the two but...

mattgallagher:
[...]

PR #2693, opened 2019-01-27
Add Xcode 10.2 beta simulator support

kmcbride:
re: https://github.com/Carthage/Carthage/issues/2691

This implementation could probably use some finesse, but hopefully it's of some use.

### DIFF
--- a/Source/CarthageKit/Simulator.swift
+++ b/Source/CarthageKit/Simulator.swift
@@ -68,13 +68,13 @@ internal func selectAvailableSimulator(of sdk: SDK, from data: Data) -> Simulato
 
 /// Parses a matching platform and version from a given identifier.
 internal func parsePlatformVersion(for platformName: String, from identifier: String) -> String? {
-    guard let platformRange = identifier.range(of: platformName) else { return nil }
+	guard let platformRange = identifier.range(of: platformName) else { return nil }
 
-    let nonDigitCharacters = CharacterSet.decimalDigits.inverted
-    let version = identifier
-        .suffix(from: platformRange.upperBound)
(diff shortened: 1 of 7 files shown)


Tests/CarthageKitTests/Resources/Simulators/availables-xcode101-beta.json
```diff
@@ -12,7 +12,7 @@
         "isAvailable" : "NO",
         "name" : "iPhone 6",
         "udid" : "ABDA7BC1-DB72-4332-90C2-C3D9AA8A5003"
-      },
+      }
```
giginet: Please do not get rid of trailing commas.

Source/CarthageKit/Simulator.swift
```diff
@@ -45,8 +45,15 @@ internal func selectAvailableSimulator(of sdk: SDK, from data: Data) -> Simulato
 		return nil
 	}
 	let platformName = sdk.platform.rawValue
-	let allTargetSimulators = devices
-		.filter { $0.key.hasPrefix(platformName) }
+	func reducePlatformNames(_ result: inout [String: [Simulator]], _ entry: (key: String, value: [Simulator])) {
+		guard let range = entry.key.range(of: platformName) else { return }
+		let platform = entry.key
```
giginet: This implementation is bit complex to read.

I suggest implementing a function to parse OS version from its identifiers.
And you should add tests for that function.

```swift
func parsePlatformVersion(from identifier: String) -> String?
```

kmcbride:
@giginet, thanks for the suggestion. I've refactored the platform parsing to (hopefully) be more understandable.

PR #2696, opened 2019-01-30
Fix issue where Carthage cannot find simulators when using Xcode 10.1 after installing 10.2 beta

DavidBrunow:
The implementation may need to be tweaked but this fixes an issue where the latest iOS version found by Carthage is not the latest version with available simulators. For example, when running Carthage on the command line using `xcodebuild` version 10.1 no simulators can be found because the latest OS version is 12.2 but there are no available simulators with that version.


giginet:
Caused by this changes?
https://github.com/Carthage/Carthage/pull/2693

ikesyo:
Could you please add a test case for this?

Source/CarthageKit/Simulator.swift
```diff
@@ -49,7 +49,9 @@ internal func selectAvailableSimulator(of sdk: SDK, from data: Data) -> Simulato
 		guard let platformVersion = parsePlatformVersion(for: platformName, from: entry.key) else { return }
 		result[platformVersion] = entry.value
 	}
-	let allTargetSimulators = devices.reduce(into: [:], reducePlatformNames)
+	let allTargetSimulators = devices.filter({ _, devices in 
+		return devices.filter({ $0.isAvailable }).count > 0
```
ikesyo: This could be written as `devices.contains { $0.isAvailable }` which is more efficient.

DavidBrunow:
> Caused by this changes?
> #2693

I believe so.

DavidBrunow:
Will add a test case when I get a chance.

kmcbride:
> > Caused by this changes?
> > #2693
> 
> I believe so.

The issue isn't with those changes, as the filtering done for `allTargetSimulators` is the same. The issue also exists in the original code when, as you said, there are no simulators for the latest OS available to you.

You could combine the filtering operations if you wanted to potentially save a few iterations.
```swift
func filterAvailableSimulators(_ result: inout [String: [Simulator]], _ entry: (key: String, value: [Simulator])) {
	guard let platformVersion = parsePlatformVersion(for: platformName, from: entry.key) else { return }
	guard entry.value.contains { $0.isAvailable } else { return }
	result[platformVersion] = entry.value
}
let allTargetSimulators = devices.reduce(into: [:], filterAvailableSimulators)
```

giginet:
>  For example, when running Carthage on the command line using xcodebuild version 10.1 no simulators can be found because the latest OS version is 12.2 but there are no available simulators with that version.

Sorry I haven't tried yet. So I can't imagine JSON values these versions are exported.
Could you describe changes between each version?

I'll try current master and your patch on some Xcode version tomorrow.

DavidBrunow:
> > > Caused by this changes?
> > > #2693
> > 
> > 
> > I believe so.
> 
> The issue isn't with those changes, as the filtering done for `allTargetSimulators` is the same. The issue also exists in the original code when, as you said, there are no simulators for the latest OS available to you.
> 
> You could combine the filtering operations if you wanted to potentially save a few iterations.
> 
> ```swift
> func filterAvailableSimulators(_ result: inout [String: [Simulator]], _ entry: (key: String, value: [Simulator])) {
> 	guard let platformVersion = parsePlatformVersion(for: platformName, from: entry.key) else { return }
> 	guard entry.value.contains { $0.isAvailable } else { return }
> 	result[platformVersion] = entry.value
> }
> let allTargetSimulators = devices.reduce(into: [:], filterAvailableSimulators)
> ```

Sorry to mislead about that PR being the cause.

I've made a change similar to what you suggested @kmcbride although the compiler didn't like that exact syntax. In addition to saving cycles I think the intent is clearer.

DavidBrunow:
> > For example, when running Carthage on the command line using xcodebuild version 10.1 no simulators can be found because the latest OS version is 12.2 but there are no available simulators with that version.
> 
> Sorry I haven't tried yet. So I can't imagine JSON values these versions export.
> Could you describe changes between each version?
> 
> I'll try current master and your patch on some Xcode version tomorrow.

The JSON values are the same as the ones returned for Xcode 10.2. The only problem is that the 12.2 simulators are in that JSON list but are not available so that when Carthage finds the highest OS version to be 12.2 no simulators are available.

DavidBrunow:
The build failure seems unrelated to the changes I made -- is it a known issue?

Tests/CarthageKitTests/SimulatorSpec.swift
```diff
@@ -146,6 +146,26 @@ class SimulatorSpec: QuickSpec {
 					}
 				}
 
+				context("Xcode 10.1 after installing 10.2 beta") {
```
giginet: [IMO] I think this context is not good.

Because this name confuses us as if behaviors change with install states of 10.1.

This context should describe this condition.
Like `with unavailable newer simulators`.

Tests/CarthageKitTests/SimulatorSpec.swift
```diff
@@ -146,6 +146,26 @@ class SimulatorSpec: QuickSpec {
...
+					it("should return the first simulator of the latest version") {
+						let data = loadJSON(for: "Simulators/availables-xcode101-after-installing-102-beta")
```
giginet: The same holds for this fixture name.

Like `with-unavailable-newer-simulators`

Sorry, I have no better idea. 😅

DavidBrunow: I think your idea is much better than mine was :)

Tests/CarthageKitTests/SimulatorSpec.swift
DavidBrunow: I think I addressed this the way you suggested, let me know if I interpreted you incorrectly.

DavidBrunow: OK I think I understand now -- I was getting stuck on the "should" that prefixed the surrounding context labels. Made another change.

giginet: No. I meant about sentence in `context`.
`Xcode 10.1 after installing 10.2 beta` is not properly named because by described reasons.

```suggestion
				context("when the latest installed simulator is unavailable") {
```

Sorry to confuse you.

giginet:
I'm going to approve and merge this PR after fixing this suggestion. https://github.com/Carthage/Carthage/pull/2696#discussion_r252773560

@Carthage/carthage I want another reviews. 👀

DavidBrunow:
Sorry to be so dense, I’m trying to rush too much. I think I’ve got it right now.

jdhealy:
@DavidBrunow appreciate all you’re doing, thanks! ✨

PR #2704, opened 2019-02-12
Mention MacPorts in README

FranklinYu:
MacPorts also provides [a package](https://www.macports.org/ports.php?by=library&substr=carthage).


mdiep:
Thanks!

Issue #1885, opened 2017-04-14
Can carthage support HTTP besides HTTPS?


wqzhang123:
My organization uses Artifactory with HTTP, I can't use carthage to distribute pre-compiled frameworks on internal Artifactory. Can carthage make an option like "--use-http" for update?

mdiep:
We could. We've decided not to because of the potential for security issues.

While it's safe to use http internally, Carthage can't know whether you're hitting an intranet server. So we've required https to safeguard other users.

The correct thing to do here is to use https.

wqzhang123:
I totally agree https is the correct way, but let carthage's user have an alternative. For example, the default is https, but user must pass the "--use-http" option to the update command, which means the user knows what he is doing, using http instead of https.

So for me right now, until my organization enables Artifactory to https, I can't use carthage to distribute pre-compiled frameworks. Anyway, maybe most organization use https, only mine is http, but I wonder if other people have requested http as an option before.

jacobjiangwei:
yes, why not let user take his own risks?

hlineholm:
I totally agree with @wqzhang123 and I have the exact same problem. I've created a #2706 as a proposal for a solution to this.

PR #2706, opened 2019-02-16
Add Allow HTTP Flag

hlineholm:
This is a proposal for a solution to the problem described in #1885. I have the exact same problem and as I think Carthage should safeguard its users, there should be an option available to use HTTP at your own risk. (As I worked with the changes in this PR, I discovered that http is allowed for enterprise GitHub URLs but the validation for this seems to be pretty fragile as it only relies on a certain combination of path components in the url.) 

This issue is so big for us that we cannot use Carthage unless this is implemented.

### DIFF
--- a/Documentation/Artifacts.md
+++ b/Documentation/Artifacts.md
@@ -145,7 +145,7 @@ For dependencies that do not have source code available, a binary project specif
 * The JSON specification file name **should** have the same name as the framework and **not** be named **Carthage.json**, (example: MyFramework.json).
 * The JSON structure is a top-level dictionary with the key-value pairs of version / location.
 * The version **must** be a semantic version.  Git branches, tags and commits are not valid.
-* The location **must** be an `https` url.
+* The location **must** be an `https` url unless you use the `--allow-http` flag.
 
 #### Example binary project specification
 
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -4,6 +4,7 @@
 * `xcodebuild -version`: 
 * Are you using `--no-build`? 
 * Are you using `--no-use-binaries`? 
+* Are you using `--allow-http`?
 * Are you using `--use-submodules`? 
 * Are you using `--cache-builds`? 
 * Are you using `--new-resolver`? 
--- a/README.md
+++ b/README.md
@@ -206,6 +206,12 @@ If the framework you want to add to your project has dependencies explicitly lis
 
 If the embedded framework in your project has dependencies to other frameworks you must  **link them to application target** (even if application target does not have dependency to that frameworks and never uses them).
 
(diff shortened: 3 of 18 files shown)


blender:
The code looks good to me. This however doesn't mean I agree with the change since it's promoting a bad practice that can quickly spread.

You can issue your own TLS certificates and even create your own CA an trust it. So adding encryption to your internal hosts seems doable to me.

hlineholm-flir:
@blender yes it is easy to do if you own your host but I don't. It could take months (or worse before) they who own it decides to or has the time to fix it.

Not sure why [#3244.1](https://travis-ci.org/Carthage/Carthage/jobs/494357636) fails. It didn't fail when I ran `make tests`.

mdiep:
> there should be an option available to use HTTP at your own risk.

The issue is that you're not just putting yourself at risk: the users of your app are also at risk.

> http is allowed for enterprise GitHub URLs

We should address this as well.

---

There's no reason to use HTTP. It's dangerous. I don't think we should support it because teams don't want to generate a free SSL certificate and deploy it to their backends.

hlineholm:
The risks are well known and the use of *http* should be limited. But I think to totally forbid it is outside the scope of Carthage. By making it default to not allow *http* (similar to *curl*), Carthage does what it can to make the world safer. 

Does [CocoaPods](http://cocoapods.org) enforce *https*? [Gradle](https://gradle.org) has a [plugin](https://http-builder-ng.github.io/gradle-http-plugin/) which makes it possible to run *http* requests. [ATS](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html#//apple_ref/doc/uid/TP40009251-SW33) by default doesn't allow *http* but have a *NSExceptionAllowsInsecureHTTPLoads* option for specified domains.

blender:
@hlineholm I understand that you need this patch.

Why not just keep your fork up to date and distribute a .pkg from there?

hlineholm:
Yes @blender that is what I'm going to do. So I'll manage and it should be possible for others who needs to do the same thing to use [hlineholm/allow-http](
https://github.com/hlineholm/Carthage/tree/hlineholm/allow-http). 

But I wanted to evaluate this with you guys first, I mean for not having to keep this fork up to date 🙂

hlineholm:
So I discovered that when parsing a *framework.json* containing *http* schemes, the BuildCommand needs to know if *http* should be allowed. This gave me a build error from Argo:

> "Build.swift:65:5: The compiler is unable to type-check this expression in reasonable time; try breaking up the expression into distinct sub-expressions"

hlineholm:
@ikesyo seems to be VSCode that messes it up by using spaces instead of tabs. I'll fix it...

Issue #2677, opened 2019-01-10
No version file create for objc only framework


kenji21:
* carthage install method: [ ] .pkg, [ ] homebrew, [X] source
* `which carthage`: /usr/local/bin/cartage
* `carthage version`: current master (https://github.com/Carthage/Carthage/commit/c31b3f5261d9ac1927ad26cb0543f9202bec1a39)
* `xcodebuild -version`: Xcode 10.1
Build version 10B61
* Are you using `--no-build`? n
* Are you using `--no-use-binaries`? y
* Are you using `--use-submodules`? y
* Are you using `--cache-builds`? y
* Are you using `--new-resolver`? n

**Cartfile**
```
github "Openium/SwiftiumKit" ~> 1.3
github "github/Archimedes" ~> 1.0
```

**Carthage Output**
```
carthage update --platform iOS --no-use-binaries --use-submodules
*** Fetching Archimedes
*** Fetching SwiftiumKit
*** Checking out SwiftiumKit at "v1.4.0"
*** Checking out Archimedes at "1.1.5"
*** xcodebuild output can be found in /var/folders/7v/dfg0p2vj1z58n65d99l2n1lh0000gp/T/carthage-xcodebuild.vPJM7m.log
*** Building scheme "Archimedes iOS" in Archimedes.xcworkspace
*** Building scheme "SwiftiumKit" in SwiftiumKit.xcodeproj
```

**Actual outcome**
Carthage did not produce the `.Archimedes.version` file in Carthage/Build:

```
$ ls -a Carthage/Build/
.                    ..                   .SwiftiumKit.version iOS
```

I think it is related to https://github.com/Carthage/Carthage/pull/2585, which adds `swiftToolchainVersion` to the version file, by looking for swift version in `Dep.framework/Headers/Dep-Swift.h`
And thus, make CI job (https://travis-ci.org/Carthage/Carthage/jobs/477319552) of the https://github.com/Carthage/Carthage/issues/2400 PR I made on "shared cache" fail after last update to current master

The origin is there is no `Dep-Swift.h` header generated for "ObjC only" frameworks (as `github/Archimedes` is)

**Expected outcome**
Carthage should generate the `.Archimedes.version` file

blender:
Carthage also looks at dSYMs to generate the version file.

kenji21:
You're right:

https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/VersionFile.swift#L151-L152

[...]

PR #2707, opened 2019-02-17
Make swiftToolchainVersion optional for ObjC only frameworks

kenji21:
Fix issue #2677

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -51,6 +51,14 @@ private func parseSwiftVersionCommand(output: String?) -> String? {
 }
 
 /// Determines the Swift version of a framework at a given `URL`.
+internal func frameworkSwiftVersionIfIsSwiftFramework(_ frameworkURL: URL) -> SignalProducer<String?, SwiftVersionError> {
+	guard isSwiftFramework(frameworkURL) else {
+		return SignalProducer(value: nil)
+	}
+	return frameworkSwiftVersion(frameworkURL).map(Optional.some)
+}
+
+/// Determines the Swift version of a framework at a given `URL`.
 internal func frameworkSwiftVersion(_ frameworkURL: URL) -> SignalProducer<String, SwiftVersionError> {
 
 	guard
--- a/Source/CarthageKit/VersionFile.swift
+++ b/Source/CarthageKit/VersionFile.swift
@@ -13,7 +13,10 @@ struct CachedFramework: Codable {
 
 	let name: String
 	let hash: String
-	let swiftToolchainVersion: String
+	let swiftToolchainVersion: String?
+	var isSwiftFramework: Bool {
+		return swiftToolchainVersion != nil
+	}
 }
 
 struct VersionFile: Codable {
@@ -403,7 +406,7 @@ public func createVersionFileForCommitish(
 	struct FrameworkDetail {
 		let platformName: String
 		let frameworkName: String
-		let frameworkSwiftVersion: String
+		let frameworkSwiftVersion: String?
 	}
 
 	if !buildProducts.isEmpty {
@@ -411,7 +414,7 @@ public func createVersionFileForCommitish(
 			.flatMap(.merge) { url -> SignalProducer<(String, FrameworkDetail), CarthageError> in
 				let frameworkName = url.deletingPathExtension().lastPathComponent
 				let platformName = url.deletingLastPathComponent().lastPathComponent
-				return frameworkSwiftVersion(url)
+				return frameworkSwiftVersionIfIsSwiftFramework(url)
 					.mapError { swiftVersionError -> CarthageError in .unknownFrameworkSwiftVersion(swiftVersionError.description) }
 					.flatMap(.merge) { frameworkSwiftVersion -> SignalProducer<(String, FrameworkDetail), CarthageError> in
 					let frameworkDetail: FrameworkDetail = .init(platformName: platformName,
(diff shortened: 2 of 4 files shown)


blender:
Code looks good to me. Can you show me how this is change affects the decision to rebuild (or not) the dependencies?

kenji21:
replace spaces by tabs (IDEs should auto detect indentation characters used...)

kenji21:
The case "not a swift framework" is already managed : https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/VersionFile.swift#L149

If an objc only framework is updated to use swift (or converted entirely to), I assume a new (major) version will be created

I've added a test : 
https://github.com/Carthage/Carthage/pull/2707/files#diff-75f7a281ef0b6a7f62a9b105850dda7bR168

hlineholm:
I've created an alternative (#2708) which does make use of [Xcode.swift:isSwiftFramework](https://github.com/Carthage/Carthage/blob/0aee76db5e5172edb1c618e4d303acfd1af741c2/Source/CarthageKit/Xcode.swift#L119)

blender:
I think #2708 is a little cleaner. I would merge this one first and then @hlineholm 's patch.

So in the end we're not adding a property to the .version file? We're just relying on a missing property?

kenji21:
> So in the end we're not adding a property to the .version file? We're just relying on a missing property?

Yes, when I tried to add a boolean, Codable needs a value when creating VersionFile from a file, here is a "playgroundable" code:

```
struct CachedFramework: Codable {
	enum CodingKeys: String, CodingKey {
		case name = "name"
		case hash = "hash"
		case swiftToolchainVersion = "swiftToolchainVersion"
		case isSwiftFramework = "isSwiftFramework"
	}
	
	let name: String
	let hash: String
	let swiftToolchainVersion: String?
	var isSwiftFramework: Bool
}

let json = "{\"name\":\"ObjcFramework\",\"hash\":\"Hash\",\"swiftToolchainVersion\"}"
let jsonData = json.data(using: .utf8)!
let cachedFramework = try? JSONDecoder().decode(CachedFramework.self, from: jsonData)

if cachedFramework == nil {
	print("not backward compatible")
}
```

which prints `not backward compatible` and I don't know how to nicely fix this (specifying a default value for the isSwiftFramework to true didn't help)

kenji21:
Tried with the Cartfile/commands of issue #2677:

```
$ carthage version
0.32.0
$ carthage update --platform iOS --no-use-binaries --use-submodules
*** Fetching SwiftiumKit
*** Fetching Archimedes
*** Checking out Archimedes at "1.1.5"
*** Checking out SwiftiumKit at "v1.4.0"
*** xcodebuild output can be found in /var/folders/7v/dfg0p2vj1z58n65d99l2n1lh0000gp/T/carthage-xcodebuild.JJ2yn1.log
*** Building scheme "Archimedes iOS" in Archimedes.xcworkspace
$ ls -a Carthage/Build/
.                    ..                   .Archimedes.version  .SwiftiumKit.version iOS
$ cat Carthage/Build/.Archimedes.version
{
  "commitish" : "1.1.5",
  "iOS" : [
    {
      "name" : "Archimedes",
      "hash" : "d2cdb749ef6bcf813c7973ce8ba93e3edd2ac2e18da6f5315c4bf76fedec6e98"
    }
  ]
}
$ cat Carthage/Build/.SwiftiumKit.version
{
  "commitish" : "v1.4.0",
  "iOS" : [
    {
      "name" : "SwiftiumKit",
      "hash" : "4f7a7bb95c913c7549e2847fa6f335898ea24aabd4c10a4f52ad1b58ad704fb1"
    }
  ]
}
```

Swift framework doesn't have the `swiftToolchainVersion` json key
Now let's bootstrap on top of 0.32.0 cache:
```
~/github/Carthage/.build/x86_64-apple-macosx10.10/release/carthage bootstrap --platform iOS --no-use-binaries --use-submodules --cache-builds
*** Checking out SwiftiumKit at "v1.4.0"
*** Checking out Archimedes at "1.1.5"
*** xcodebuild output can be found in /var/folders/7v/dfg0p2vj1z58n65d99l2n1lh0000gp/T/carthage-xcodebuild.y48rty.log
*** Valid cache found for Archimedes, skipping build
*** Valid cache found for SwiftiumKit, skipping build
```

The cache is correct so no dependencies recompilation needed.

Source/CarthageKit/Xcode.swift
```diff
@@ -50,6 +50,16 @@ private func parseSwiftVersionCommand(output: String?) -> String? {
...
 }
 
+/// Determines the Swift version of a framework at a given `URL`.
+internal func frameworkSwiftVersionIfIsSwiftFramework(_ frameworkURL: URL) -> SignalProducer<String?, SwiftVersionError> {
+	guard isSwiftFramework(frameworkURL) else {
+		return SignalProducer(value: nil)
+	}
+	return frameworkSwiftVersion(frameworkURL).flatMap(.merge, { (swiftVersion) -> SignalProducer<String?, SwiftVersionError> in
+		return SignalProducer(value: swiftVersion)
+	})
```
mdiep: This can be `frameworkSwiftVersion(frameworkURL).map(Optional.some)`

mdiep:
This looks good to me. I'd just like to see 1 minor change: using `map` instead of `flatMap`. I think that's a lot clearer.

Source/CarthageKit/Xcode.swift
kenji21: Ammended my commit to simplify code : https://github.com/Carthage/Carthage/pull/2707/commits/602e99905e95284fde8d1f472bf6fbd622c1cfcf#diff-f66644e8cfbcad3c3826395a690ddfbcR58

Thanks

mdiep:
Thank you for the PR!

Issue #2530, opened 2018-07-24
Carthage builds my framework for x86_64 while i have linked arm64 framework


don1ck:
* carthage install method: homebrew
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.30.1
* `xcodebuild -version`: Xcode 9.4.1
* Are you using `--no-build`? 
No
* Are you using `--no-use-binaries`? 
No
* Are you using `--use-submodules`? 
No
* Are you using `--cache-builds`? 
No
* Are you using `--new-resolver`? 
No

**Cartfile**
No

**Carthage Output**
```
victor$ carthage build BanubaCore --no-skip-current
*** xcodebuild output can be found in /var/folders/6s/9_8ytpgd13s0v2d5b8_h69940000gq/T/carthage-xcodebuild.sYIV6A.log
*** Building scheme "BanubaCore" in BanubaCore.xcworkspace
Build Failed
	Task failed with exit code 65:
	/usr/bin/xcrun xcodebuild -workspace /Users/victor/Projects/banubacore/BanubaCore/BanubaCore.xcworkspace -scheme BanubaCore -configuration Release -sdk iphonesimulator -destination platform=iOS\ Simulator,id=1F48929E-E833-472E-8A5E-E27F7CD22D4C -destination-timeout 3 ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES build (launched in /Users/victor/Projects/banubacore/BanubaCore)

This usually indicates that project itself failed to compile. Please check the xcodebuild log for more details: /var/folders/6s/9_8ytpgd13s0v2d5b8_h69940000gq/T/carthage-xcodebuild.sYIV6A.log
```

**Actual outcome**
```
ld: warning: ignoring file .../effect_player_wrap_ios.framework/effect_player_wrap_ios, file was built for arm64 which is not the architecture being linked (x86_64): .../effect_player_wrap_ios.framework/effect_player_wrap_ios
```

**Expected outcome**
A Valid build

Hello!
I'm developing a framework, which is requires some other framework
My valid architecture is arm64 only. Other framework is builded for arm64 as well.
xcodebuild has no issues with building it. But looks like carthage builds for ios simulator?

I have tried cathage build --platform iOS  but it doesn't helped.

Any advice? 

Thanks!

blender:
Something is probably wrong with your project setup. 

[...]

PR #2712, opened 2019-02-20
Add "skip-simulators" option

RustyKnight:
# Background
I now have two (specifically iOS) projects which make use of third party libraries (`lib*.a`) which do not support simulator architecture slices and therefore can not be built against simulators.

I've attempted to produce wrapper frameworks around these libraries to make their use simpler in another projects, but they can't be managed through Carthage, as Carthage wants to build them against both "Generic" and "Simulator" based platforms.

In an attempt to reduce further issues, I'd like also take advantage of Carthage's "binary only" frameworks support, but, again, I can't use my existing work flow to use Carthage to generate these results (yes, I could do it manually, but I've built a work flow around Carthage).

# Solution
Skip "sdk"s which are backed by simulators.

The intended solution is too, when requested by the user, skip building `SDK`s which report `isSimulator` as `true`.

The solution would provide a opt-in parameter `--skip-simulators` which, when used, would instruct the build process to skip any `SDK` which `isSimulator`

## Specifying device ID 
I did research using `destination` to specify the physical device to build against, but choose not to go this route due to the complexity of collecting this information and deciding when the override should be applied

# Considerations
When building a project/archive in this way, all dependencies built by Carthage will also be built in the same way (no simulator slices).

Since any resulting project won't be able to use the simulator(s) anyway, this shouldn't be a major issue anyway

# Notes
The change was made against the release branch as the current master kept failing, as it was unable to determine the Swift version

### DIFF
--- a/Source/CarthageKit/BuildOptions.swift
+++ b/Source/CarthageKit/BuildOptions.swift
@@ -15,7 +15,7 @@ public struct BuildOptions {
 	/// Whether to use downloaded binaries if possible.
 	public var useBinaries: Bool
 	/// Whether to use platform simulators or not
-	public var noSimulators: Bool
+	public var skipSimulators: Bool
 
 	public init(
 		configuration: String,
@@ -24,7 +24,7 @@ public struct BuildOptions {
 		derivedDataPath: String? = nil,
 		cacheBuilds: Bool = true,
 		useBinaries: Bool = true,
-		noSimulators: Bool = false
+		skipSimulators: Bool = false
 	) {
 		self.configuration = configuration
 		self.platforms = platforms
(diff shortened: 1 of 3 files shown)


ikesyo:
Thanks for the pull request! However, I don't think this is a right direction.

A framework's supporting platform can be configured by `SUPPORTED_PLATFORMS` build setting. So you should be able to create a wrapper framework which only supports device builds by removing `iphonesimulator` from `SUPPORTED_PLATFORMS`.

ikesyo:
See #2530 for example.

RustyKnight:
> #2530

Well, that's not an obscure, out-of-the-way, don't look and me, hard to find setting 🙄😓😅

I've verified that I can change the setting in my framework source and Carthage will build it and since my intention is to generate a "binary only" release, my target project shouldn't care when updating the dependencies.

Thank, I've learnt a lot of the last few days and I appreciate your input

I accept that this pull request can be rejected and closed 😉

Issue #2698, opened 2019-02-04
Speeding up `carthage outdated` build phases


ctxppc:
Hi everybody,

This is my first issue on this repository and I'm not familiar with the Carthage codebase (yet) to make a PR. I use Carthage to manage my apps' dependencies, and I also use the `outdated` subcommand as part of the build process (as an Xcode build phase, with `--xcode-warnings`) to be notified of updates to dependencies. Both updating and checking for updates work as intended. However, in my quest for ever tighter debug turnaround times, I'm trying to decrease my debug build times as much as possible. And one part of the build turnaround is 1 to 3 seconds spent by Carthage to fetch my dependencies' repositories. It doesn't seem Xcode can parallelise this so it's a blocking second or three, on a fast connection.

Unless I'm missing a command line option or implementation detail, I'm proposing that `carthage outdated` should cache its result (basically skip fetching if it has been done in the past few minutes or hours), possibly as an opt-in (or opt-out) option. It should consistently report outdated repositories regardless of whether it actually fetched anything in this run. I don't think `carthage update` should do any such caching since it's expected that it fetches all updates that are published as of _right now_.

Any thoughts on this? Thanks for all the work! 😄

# Configuration

* carthage install method: [ ] .pkg, [X] homebrew, [ ] source
* `which carthage`: `/usr/local/bin/carthage`
* `carthage version`: 0.31.2
* `xcodebuild -version`: 10.1 (10B61)
* Are you using `--no-build`? No.
* Are you using `--no-use-binaries`? No.
* Are you using `--use-submodules`? No.
* Are you using `--cache-builds`? Yes.
* Are you using `--new-resolver`? No.

**Cartfile**
```
github "ReactiveX/RxSwift" ~> 4.0
github "roberthein/TinyConstraints"
```

**Carthage Output**
(Probably) N/A — updating works as intended

**Actual outcome**
Carthage fetches all dependencies' repositories and build phase takes 1 to 3 seconds to finish, regardless of previous builds preceding it.

**Expected outcome**
Carthage should skip fetching repositories in the second build soon after the first and the build phase should complete within 0.1 second.

blender:
Is there a reason why you think carthage should do this as opposed of wrapping carthage commands in a script of your making that does exactly what you propose?

ctxppc:
[...]

PR #2714, opened 2019-02-24
Add fetch-interval to `outdated` command to skip fetching the dependencies

mollyIV:
### Motivation and Context

`carthage outdated --xcode-warnings` (Xcode build phase) command fetches the dependencies each time when executed. In order to improve the performance of the command, we'd like to introduce a new parameter `fetch-interval`. This parameter specifies how often the dependencies should be fetched from the remote location. Example usage:

```bash
carthage outdated --xcode-warning --fetch-interval 30
```

`fetch-interval` - the minimum amount of time that must elapse between dependencies fetch operation (provided in ~~minutes~~ seconds)

https://github.com/Carthage/Carthage/issues/2698

### Description

~~This PR is draft one [WIP]. Kindly ask you to check the review comments~~  🙏
EDIT: PR is ready for review 🙏 

_It's my first Carthage contribution_ 🎉

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -22,13 +22,12 @@ public struct FetchCache {
 		lastFetchTimes.removeAll()
 	}
 
-	internal static func needsFetch(forURL url: GitURL) -> Bool {
-		guard let lastFetch = lastFetchTimes[url] else {
+    internal static func needsFetch(forURL url: GitURL) -> Bool {
+		guard let lastFetch = lastFetchTimes[url] ?? repositoriesLastFetchTime()?.timeIntervalSince1970 else {
 			return true
 		}
 
 		let difference = Date().timeIntervalSince1970 - lastFetch
-
 		return !(0...fetchCacheInterval).contains(difference)
 	}
 
@@ -37,6 +36,19 @@ public struct FetchCache {
 			lastFetchTimes[url] = Date().timeIntervalSince1970
 		}
 	}
+
+    fileprivate static func repositoriesLastFetchTime() -> Date? {
+        do {
+            let attr = try FileManager.default.attributesOfItem(atPath: Constants.Dependency.repositoriesURL.path)
+            return attr[FileAttributeKey.modificationDate] as? Date
+        } catch {
+            return nil
+        }
+    }
+
+    public static func updateRepositoriesLastFetchTime() {
+        try? FileManager.default.setAttributes([FileAttributeKey.modificationDate: Date()], ofItemAtPath: Constants.Dependency.repositoriesURL.path)
+    }
 }
 
 /// Shells out to `git` with the given arguments, optionally in the directory
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -23,7 +23,7 @@ public struct FetchCache {
 	}
 
     internal static func needsFetch(forURL url: GitURL) -> Bool {
-		guard let lastFetch = lastFetchTimes[url] ?? repositoriesLastFetchTime()?.timeIntervalSince1970 else {
+		guard let lastFetch = lastFetchTimes[url] ?? repositoriesLastFetchTime() else {
 			return true
 		}
 
@@ -37,17 +37,50 @@ public struct FetchCache {
 		}
 	}
 
-    fileprivate static func repositoriesLastFetchTime() -> Date? {
-        do {
-            let attr = try FileManager.default.attributesOfItem(atPath: Constants.Dependency.repositoriesURL.path)
-            return attr[FileAttributeKey.modificationDate] as? Date
-        } catch {
+    fileprivate static func repositoriesLastFetchTime() -> TimeInterval? {
+        guard let data = try? Constants.Dependency.repositoriesURL.extendedAttribute(for: "\(Constants.bundleIdentifier).repositories.last.fetch.time") else {
             return nil
         }
+        return data.withUnsafeBytes { $0.pointee as TimeInterval }
     }
 
     public static func updateRepositoriesLastFetchTime() {
-        try? FileManager.default.setAttributes([FileAttributeKey.modificationDate: Date()], ofItemAtPath: Constants.Dependency.repositoriesURL.path)
+        var now = Date().timeIntervalSince1970
+        let data = Data(bytes: &now, count: MemoryLayout<TimeInterval>.size)
+        try? Constants.Dependency.repositoriesURL.setExtendedAttribute(for: "\(Constants.bundleIdentifier).repositories.last.fetch.time", data: data)
+    }
+}
+
+/// URL extension to support operations on extended attributes of a file
+fileprivate extension URL {
+
+    enum ExtendedAttributesError: Error {
+        case cannotReadAttribute
+        case cannotSetAttribute
+    }
+
+    func extendedAttribute(for name: String) throws -> Data  {
+        let data = try withUnsafeFileSystemRepresentation { path -> Data in
+            let length = getxattr(path, name, nil, 0, 0, 0)
+            guard length >= 0 else { throw ExtendedAttributesError.cannotReadAttribute }
+
+            var data = Data(count: length)
+            let result =  data.withUnsafeMutableBytes { [count = data.count] in
+                getxattr(path, name, $0.baseAddress, count, 0, 0)
+            }
+            guard result >= 0 else { throw ExtendedAttributesError.cannotReadAttribute }
+            return data
+        }
+        return data
+    }
+
+    func setExtendedAttribute(for name: String, data: Data) throws {
+        try withUnsafeFileSystemRepresentation { path in
+            let result = data.withUnsafeBytes {
+                setxattr(path, name, $0, data.count, 0, 0)
+            }
+            guard result >= 0 else { throw ExtendedAttributesError.cannotSetAttribute }
+        }
     }
 }
 
--- a/Source/carthage/Outdated.swift
+++ b/Source/carthage/Outdated.swift
@@ -51,6 +51,7 @@ public struct OutdatedCommand: CommandProtocol {
 		public let useSSH: Bool
 		public let isVerbose: Bool
 		public let outputXcodeWarnings: Bool
+        public let fetchInterval: Int
 		public let colorOptions: ColorOptions
 		public let directoryPath: String
 
@@ -65,7 +66,8 @@ public struct OutdatedCommand: CommandProtocol {
 				<*> mode <| Option(key: "use-ssh", defaultValue: false, usage: "use SSH for downloading GitHub repositories")
 				<*> mode <| Option(key: "verbose", defaultValue: false, usage: "include nested dependencies")
 				<*> mode <| Option(key: "xcode-warnings", defaultValue: false, usage: "output Xcode compatible warning messages")
-				<*> ColorOptions.evaluate(mode, additionalUsage: UpdateType.legend)
+                <*> mode <| Option(key: "fetch-interval", defaultValue: 0, usage: "the minimum amount of time that must elapse between dependencies fetch operation")
+                <*> ColorOptions.evaluate(mode, additionalUsage: UpdateType.legend)
 				<*> mode <| projectDirectoryOption
 		}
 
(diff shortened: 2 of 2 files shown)


Source/carthage/Outdated.swift
```diff
@@ -112,3 +124,26 @@ public struct OutdatedCommand: CommandProtocol {
...
+    }
+
+    // If the fetch interval time has not been set before, start measuring time from the current date.
+    guard let lastFetchDate = UserDefaults.standard.value(forKey: key) as? Date else {
+        UserDefaults.standard.set(Date(), forKey: key)
+        return SignalProducer<Bool, CarthageError>({ false })
+    }
+
+    // Check if the fetch interval is expired.
+    let fetchIntervalExpired = Date() > lastFetchDate.addingTimeInterval(fetchInterval)
```
mollyIV: Even though the pull request description says `fetch-interval` represents the number of minutes, the implementation is still in seconds. Gonna fix that ⚒

Source/carthage/Outdated.swift
```diff
@@ -112,3 +124,26 @@ public struct OutdatedCommand: CommandProtocol {
...
+
+    // When the fetch interval is not provided, clear the storage and fetch the dependencies.
+    guard fetchInterval > 0 else {
+        UserDefaults.standard.removeObject(forKey: key)
+        return SignalProducer<Bool, CarthageError>({ false })
+    }
+
+    // If the fetch interval time has not been set before, start measuring time from the current date.
+    guard let lastFetchDate = UserDefaults.standard.value(forKey: key) as? Date else {
+        UserDefaults.standard.set(Date(), forKey: key)
```
mollyIV: As a storage, I decided to give a go for a `UserDefaults`. Is there any storage in Carthage code base which I should  have used? 🤔

Source/carthage/Outdated.swift
```diff
@@ -87,8 +89,18 @@ public struct OutdatedCommand: CommandProtocol {
...
 	public func run(_ options: Options) -> Result<(), CarthageError> {
+        let shouldIgnoreFetching = shouldIgnoreFetchingDependencies(fetchInterval: TimeInterval(options.fetchInterval))
 		return options.loadProject()
-			.flatMap(.merge) { $0.outdatedDependencies(options.isVerbose) }
+            .combineLatest(with: shouldIgnoreFetching)
+            .flatMap(.merge) { project, shouldIgnoreFetching -> SignalProducer<[Project.OutdatedDependency], CarthageError> in
+                if shouldIgnoreFetching {
+                    // TODO: Non-fetching changes required.
+                    carthage.println("The dependencies will not be fetched due to non-expired fetch interval.")
+                    return project.outdatedDependencies(options.isVerbose)
```
mollyIV: I noticed that we do download the remote dependencies when we initialise `ResolverProtocol` type object:
https://github.com/Carthage/Carthage/blob/1c3e23a3dffc7f19ccd6dd8073454bb39fdfafa5/Source/CarthageKit/Project.swift#L505-L509

Wondering if modifying the resolver initialisation is a way to go, i.e. passing `[:]` to resolver's versions 🤔 

Could you guys please give me a hint what would be the best way to ignore fetching the remote dependencies? 🙏

Source/carthage/Outdated.swift
```diff
@@ -51,6 +51,7 @@ public struct OutdatedCommand: CommandProtocol {
 		public let useSSH: Bool
 		public let isVerbose: Bool
 		public let outputXcodeWarnings: Bool
+        public let fetchInterval: Int
```
mdiep: Indentation is off here. Please be sure to fix all these up.

Source/carthage/Outdated.swift
mdiep: I think you can just set `FetchCache.fetchCacheInterval` to the value that's passed in. That should be _all_ you need to do?

mollyIV: 👋 @mdiep Thanks for the comment and your help 🙇 Just to double check: `FetchCache.fetchCacheInterval` is just a static property without the persistent storage behind it, so we need the `UserDefaults` logic I introduced in this PR anyway, right? 

With the suggested changes we are going to have:

```Swift
...
if shouldIgnoreFetching {
    carthage.println("The dependencies will not be fetched due to non-expired fetch interval.")
    FetchCache.fetchCacheInterval = TimeInterval(options.fetchInterval)
    return project.outdatedDependencies(options.isVerbose)
} 
....
```

Could you please just confirm we are on the same page? 🙏

mdiep: > so we need the UserDefaults logic I introduced in this PR anyway, right?

Why?

mollyIV: Please correct me if I am wrong @mdiep , but I though `carthage` (CLI application) process is killed as soon as the command execution ends. It means `carthage` does not hold in memory state (the static properties) between the command executions. That's why I introduced `UserDefaults` - to keep the state of the application between the command executions.

mdiep: Oh, right. Sorry, I wasn't thinking.

I'm not sure how we'd want to store that. [This](https://stackoverflow.com/a/9229377) might help?

mollyIV: Thanks for the suggestion 🙇I tested the command `stat -f '%m' .git/FETCH_HEAD`, unfortunately **`FETCH_HEAD` object is not updated when `carthage` fetches just the dependencies**:

screenshot 2019-03-04 at 22 13 08

The `FETCH_HEAD` gets updated when we do fetch the current project:

image

It seems this idea won't work for us.

mollyIV: I was testing `FetchCache` structure by running the same `carthage` command `carthage outdated --xcode-warnings` a few times in a row and `lastFetchTimes` dictionary was empty each time. The abstraction seems to be broken. Maybe there's some context I am missing here? 🤔 

I'd suggest to fix it by introducing the persistent storage, i.e. `UserDefaults` (not sure how that might affect the performance) or finding another way how to keep a state between the command executions. By making this change indeed we could just set the `fetchCacheInterval` to the value passed by `--fetch-interval` parameter.

mdiep: The `FetchCache` is only used for in-process caching right now; it's not persistent.

I don't love using `UserDefaults`. I think we should either infer it from the dependencies directory or store it somewhere on disk there.

mollyIV: I see 👍 I will try to find a way to infer the last modification date (last fetch) from the dependencies directory. However I'd like to know why you prefer storing `fetch-interval` value somewhere on disk (in a file/plist probably) instead of using `UserDefaults`? 🤔 

`UserDefaults` are pretty fast and easy to do - the simplest key-value storage available.

mdiep: It's easier to manual trigger. It seems more straightforward if you delete the cache entirely. Even though it's a little harder to implement, I think it's the simpler solution.

mollyIV: > It seems more straightforward if you delete the cache entirely.

Could you please explain me a bit more what do you mean by ☝️ 🙏  

I am a bit confused now and dunno in what direction to push this PR.

**First option**

Check if we are able to detect when a last fetch happened based on the `Dependencies` catalog. In that way we wouldn't need to store `fetch-interval` value anywhere, but read it on the runtime. 

**Second option** 

Please correct me if I am wrong - you'd like to keep `fetch-interval` value in the file on the disk and if this param wasn't provided, we would just delete the whole file? 🤔 

🙇 🙇

mdiep: I prefer option one.

But failing that, I think we should keep the last fetched time in a file on disk in the dependencies directory. Users sometimes delete the entire cache to get back to a clean state. I'd prefer to keep all the cache info in that directory instead of putting some of it in user defaults.

mollyIV: 👋 @mdiep 

I was investigating option one and unfortunately couldn't find a reasonable solution. 

I was playing with with the option number two - file solution. I'd suggest to modify `FetchCache` component and convert it from in-process caching to persistent caching. We could store last fetch timestamp per dependency (`../Library/Caches/org.carthage.CarthageKit/dependencies`): 

image

Please tell me what you think about this solution 🙇

mdiep: What if _we_ set the last modified date of the directory for the dependency?

mollyIV: Good idea, thanks! 👍 

Tested on just three dependencies, but already noticed a great improvement (a few seconds):

Screenshot 2019-04-01 at 08 03 12

Screenshot 2019-04-01 at 08 03 27

I implemented the changes and kindly ask you for re-review @mdiep  🙇 🙇

Source/carthage/Outdated.swift
mollyIV: After second thoughts I'd suggest to leave it as the number of seconds.

mollyIV:
👋 @jdhealy 

Thanks for adding yourself as a reviewer 🙇 Please let me know if you have any questions about the implementation 👍

Source/carthage/Outdated.swift
mollyIV: 👋 Hey @mdiep 🙂 Just wanted to kindly ask if you did get a chance to have a look at the newest changes? 🙇 🤔

jdhealy:
> It's my first Carthage contribution 🎉

_High five over the internet_ Thanks, so much!

I've got time for one or two notes before I head out the door this morning.

Source/CarthageKit/Git.swift
```diff
@@ -37,6 +36,19 @@ public struct FetchCache {
 			lastFetchTimes[url] = Date().timeIntervalSince1970
 		}
 	}
+
+    fileprivate static func repositoriesLastFetchTime() -> Date? {
+        do {
+            let attr = try FileManager.default.attributesOfItem(atPath: Constants.Dependency.repositoriesURL.path)
+            return attr[FileAttributeKey.modificationDate] as? Date
```
jdhealy: For us relying on file system modification date here — there's far too many antivirus scanners and cloud storage sync tools that continually just modify large swaths of files for me to be comfortable with this 😕

Maybe our own [xattr (extended attribute)](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/setxattr.2.html), could work? I'll try to give some thought to this later…

mollyIV: Thank you for your suggestion @jdhealy 🙇 I updated the code - now we are using custom extended attribute instead of modification date attribute 💪 

Screenshot 2019-04-27 at 17 36 54

Could you please have a 👀 🙏

blender:
@mollyIV It doesn't look like this PR is building. Do you have time to follow up?

mollyIV:
👋 @blender , 
It might be a global issue https://github.com/Carthage/Carthage/issues/2780 - I saw many PRs with failing tests. 

Will do the `master` rebase 👍

mollyIV:
Hey @blender, 
I did the rebase however the tests are still failing. I compered the `CI_INTEGRATION_TESTS` from [this MR](https://github.com/Carthage/Carthage/pull/2789)

image

and the error log seems to be the same. 

Any hint what that could relate to? 
Shall we create a new issue about fixing those? (seems a global issue)

🤔

mollyIV:
Hey guys 👋 

Any news about a CI issue? 🤔 Could anyone try to re-run it? 🙏

mollyIV:
👋 **stale**, still in review.

mollyIV:
👋@stale, still in review.

mollyIV:
ping stale 🤖

mollyIV:
ping stale

Source/CarthageKit/Git.swift
```diff
@@ -22,13 +22,12 @@ public struct FetchCache {
 		lastFetchTimes.removeAll()
 	}
 
-	internal static func needsFetch(forURL url: GitURL) -> Bool {
-		guard let lastFetch = lastFetchTimes[url] else {
+    internal static func needsFetch(forURL url: GitURL) -> Bool {
```
DavidBrunow: Carthage uses tabs instead of spaces for indentation -- could you please reformat your changes to do the same?

Source/CarthageKit/Git.swift
```diff
@@ -22,13 +22,12 @@ public struct FetchCache {
...
+		guard let lastFetch = lastFetchTimes[url] ?? repositoriesLastFetchTime() else {
 			return true
 		}
 
 		let difference = Date().timeIntervalSince1970 - lastFetch
-
```
DavidBrunow: Could you please revert this change?

Source/CarthageKit/Git.swift
```diff
@@ -37,6 +36,52 @@ public struct FetchCache {
...
+/// URL extension to support operations on extended attributes of a file
+fileprivate extension URL {
+
+    enum ExtendedAttributesError: Error {
+        case cannotReadAttribute
+        case cannotSetAttribute
+    }
+
+    func extendedAttribute(for name: String) throws -> Data  {
+        let data = try withUnsafeFileSystemRepresentation { path -> Data in
```
DavidBrunow: Curious -- what are your thoughts on using this approach rather than using the methods provided by `NSFileManager`? I've never used either personally.

DavidBrunow:
Ping stale?

PR #1990, opened 2017-06-15
Exclude specific schemes via Cartfile.ignore

Dschee:
Closes #1227, #437 and maybe more.

### Goal

My **goal** ist to ensure that with a **Cartfile** that looks like this:

```
github "Moya/Moya"
```

Which would normally build `Moya`, `RxMoya`, `ReactiveMoya`, `Alamofire`, `Result`, `RxSwift` and `ReactiveSwift`. To introduce **a new file named `Cartfile.ignore`** with something like this content:

```
# Ignore RxSwift variant of Moya
scheme "RxMoya"
scheme "RxSwift"

# Ignore ReactiveSwift variant of Moya
scheme "ReactiveMoya"
scheme "ReactiveSwift"
```

Which leads to Carthage only building `Moya`, `Alamofire` and `Result`.

In detail the structure of the `Cartfile.ignore` is highly inspired by the structure of the current existing `Cartfile`. Instead keys like `github` and `git` the `.ignore` file should only support `scheme`.

### `scheme`

This is the first step I want to reach – to ensure all schemes with the name specified after a `scheme` entry within the `Cartfile.ignore` are not built when running Carthage tasks. Note that the user is being notified about this with something like `*** Ignoring scheme "RxMoya" in Moya.xcworkspace`.

> **UPDATE NOTE:** Originally I also had the suggestion to introduce a `project` keyword which would then even prevent from downloading the dependency. See the section "Where to go from here" below where I consider an alternative for that which can be added later on, in a subsequent PR.

---
Please also note, that since this change is purely additive, there should be no side effects whatsoever to any Carthage users who don't specifically create a `Cartfile.ignore` to use this feature.

## Use the Cartfile.ignore right now via Carthage-Pro

I created a fork of Carthage until this gets merged (which might take years) which includes this feature already. Please read [this comment](https://github.com/Carthage/Carthage/pull/1990#issuecomment-386763916) further below for additional details.

# Where to go from here
Here are a few "Next Steps" possible after this PR is merged to go further into this direction:
- [ ] Add the possibility to ignore a dependency even during dependency graph resolution by adding support for entries like `github "ReactiveCocoa/ReactiveSwift"`
- [ ] Consider the `Cartfile.ignore` in dependencies during dependency graph resolution so framework authors can improve the build time for their Carthage users without extra-work for them

PR #2716, opened 2019-03-01
Issue #2400 Shared cache for built dependencies

kenji21:
see discussion in issue #2400


nikolaeu:
Nice one @kenji21! How about some details in `readme.md`?

blender:
LGTM.

My only concern is that this would be enabled by default.

How does @Carthage/carthage  feel about it?

killobatt:
Hello. Merging of this PR can reduce our clean/build time from 22minutes to 5-6 minutes, our team really waits for it. 
It there something we can do to help it be merged faster?

joshuapoq:
For clarity, as (I understand it) this allows Carthage to share already built dependencies across projects.

We have a workspace that has `Y` many targets across many projects that all depend downwards on `X` number of common lower level targets. Am I correct in thinking this will now only build all `X` of these common frameworks once instead of `X * Y` times?

kenji21:
> For clarity, (as I understand it) this allows Carthage to share already built dependencies across projects.
> 
> We have a workspace that has `Y` many targets across many projects that all depend downwards on `X` number of common lower level targets. Am I correct in thinking this will now only build all `X` of these common frameworks once instead of `X * Y` times?

That's the point, here is how @blender nicely recap it : https://github.com/Carthage/Carthage/issues/2400#issuecomment-402437688

Dschee:
Anybody coming across this issue nowadays, please checkout my comment [here](https://github.com/Carthage/Carthage/pull/1990#issuecomment-479442640).

blender:
@mdiep ping for review :)

nikolaeu:
@mdiep double ping ;)

Source/CarthageKit/Project.swift
```diff
@@ -1378,6 +1357,104 @@ internal func frameworksInDirectory(_ directoryURL: URL) -> SignalProducer<URL,
...
+internal func copyFrameworksAndCreateVersionFile(
+	_ frameworkURLs: [URL],
+	from directoryURL: URL,
+	to destinationURL: URL,
+	projectName: String,
+	pinnedVersion: PinnedVersion,
+	toolchain: String?
+) -> SignalProducer<URL, CarthageError> {
+	return SignalProducer(value: frameworkURLs)
+		.flatten()
```
mdiep: `SignalProducer(value: frameworkURLs).flatten()` can be written as `SignalProducer(frameworkURLs)`

Source/CarthageKit/Project.swift
```diff
@@ -1064,6 +997,21 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+					.flatMap(.concat) { dependency, version -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
+						self._projectEventsObserver.send(value: .skippedBuildingCopyFromSharedCache(dependency))
+						return SignalProducer(value: (dependency, version))
+					}
+					.collect()
+					.map { copiedFromSharedCacheDependencies -> [(Dependency, PinnedVersion)] in
+						// Filters out dependencies that we've copied from shared cache
+						// but preserves the build order
+						return dependencies.filter { dependency -> Bool in
+							!copiedFromSharedCacheDependencies.contains { $0 == dependency }
```
mdiep: This could be `!copiedFromSharedCacheDependencies.contains(dependency)`

Source/CarthageKit/Project.swift
```diff
@@ -1139,6 +1087,37 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+
+				return SignalProducer.combineLatest(
+					SignalProducer(value: dependency),
+					SignalProducer(value: version),
+					SignalProducer(value: sharedCacheURLForDependency),
+					versionFileMatches(dependency, version: version, platforms: options.platforms,
+									   rootDirectoryURL: sharedCacheURLForDependency, toolchain: options.toolchain))
+			}
+			.flatMap(.concat) { dependency, version, sharedCacheVersioned, sharedCacheMatches -> SignalProducer<(Dependency, PinnedVersion), CarthageError> in
+				if let sharedCacheVersionFileMatches = sharedCacheMatches, sharedCacheVersionFileMatches == true {
```
mdiep: `== true` is unnecessary

Source/CarthageKit/Xcode.swift
```diff
@@ -893,15 +904,23 @@ public func buildInDirectory( // swiftlint:disable:this function_body_length
...
-					platforms: options.platforms,
-					buildProducts: urls,
-					rootDirectoryURL: rootDirectoryURL
-					)
-					.flatMapError { _ in .empty }
+				var copyToSharedCache = SignalProducer<(URL), CarthageError>.empty
+				if options.cacheBuilds {
+					let sharedCacheURLForDependency = sharedCacheURLByXcodeVersionFor(dependency: dependency.dependency, pinnedVersion: dependency.version)
+					copyToSharedCache = copyFrameworksAndCreateVersionFile(urls, from: rootDirectoryURL, to: sharedCacheURLForDependency,
+																		   projectName: dependency.dependency.name, pinnedVersion: dependency.version, toolchain: options.toolchain)
```
mdiep: This should ignore errors because:

1. If copying to the shared cache fails, the build should still succeed
2. If copying to the shared cache fails, we'll still want to create the version file

Source/CarthageKit/Project.swift
```diff
@@ -1064,6 +997,21 @@ public final class Project { // swiftlint:disable:this type_body_length
 					return dependenciesIncludingNext
 				}
 			}
+			.flatMap(.concat) { (dependencies: [(Dependency, PinnedVersion)]) -> SignalProducer<[(Dependency, PinnedVersion)], CarthageError> in
+				return self.copyBuiltFrameworksFromSharedCacheIfVersionMatches(for: dependencies, options: options)
```
mdiep: I think copying to/from the shared cached should be hidden behind another flag. We already have the in-repo cache behind a flag and this is inherently riskier.

mdiep:
Looks pretty good overall, but I think we need a couple more items.

Source/CarthageKit/Project.swift
kenji21: Impossible because tuple (Dependency, PinnedVersion) doesn't conforms to Equatable (same filter is done few lines below when downloading dependencies (Filters out dependencies that we've downloaded binaries for))

Source/CarthageKit/Project.swift
kenji21: Simplified, I'll push sooner

Source/CarthageKit/Project.swift
kenji21: removed

Source/CarthageKit/Xcode.swift
kenji21: so I added a `.flatMapError { _ in .empty }` right after `copyFrameworksAndCreateVersionFile` method call

Source/CarthageKit/Project.swift
kenji21: added a `if options.cacheBuilds {` block to prevent copy from shared cache without `--cache-builds` being set

kenji21:
Hi, I updated code responding to reviews, and rebased my feature branch on top of current master, after a `make install` on my machine: 

3m1.871s for initial `carthage bootstrap`, 0m15.297s for the second one

Full log:
```
time carthage bootstrap --no-use-binaries --cache-builds --platform iOS
*** Checking out Archimedes at "1.1.5"
*** Checking out AlamofireObjectMapper at "5.2.0"
*** Checking out ObjectMapper at "3.4.2"
*** Checking out GCDWebServer at "3.5.2"
*** Checking out SwiftiumKit at "v1.4.0"
*** Checking out Alamofire at "4.8.1"
*** No cache found for Alamofire, building with all downstream dependencies
*** No cache found for Archimedes, building with all downstream dependencies
*** No cache found for GCDWebServer, building with all downstream dependencies
*** No cache found for ObjectMapper, building with all downstream dependencies
*** No cache found for SwiftiumKit, building with all downstream dependencies
*** xcodebuild output can be found in /var/folders/7v/dfg0p2vj1z58n65d99l2n1lh0000gp/T/carthage-xcodebuild.9Ea0OC.log
*** Building scheme "Alamofire iOS" in Alamofire.xcworkspace
*** Building scheme "Archimedes iOS" in Archimedes.xcworkspace
*** Building scheme "GCDWebServers (iOS)" in GCDWebServer.xcodeproj
*** Building scheme "ObjectMapper-iOS" in ObjectMapper.xcworkspace
*** Building scheme "AlamofireObjectMapper iOS" in AlamofireObjectMapper.xcworkspace

real	3m1.871s
user	1m48.996s
sys	0m32.073s
kenji@macboo:~/Desktop/cart-issue-2400 [master] $
rm -Rf Carthage/Build/
kenji@macboo:~/Desktop/cart-issue-2400 [master] $
time carthage bootstrap --no-use-binaries --cache-builds --platform iOS
*** Checking out Alamofire at "4.8.1"
*** Checking out GCDWebServer at "3.5.2"
*** Checking out Archimedes at "1.1.5"
*** Checking out SwiftiumKit at "v1.4.0"
*** Checking out AlamofireObjectMapper at "5.2.0"
*** Checking out ObjectMapper at "3.4.2"
*** No cache found for Alamofire, building with all downstream dependencies
*** No cache found for Archimedes, building with all downstream dependencies
*** No cache found for GCDWebServer, building with all downstream dependencies
*** No cache found for ObjectMapper, building with all downstream dependencies
*** No cache found for SwiftiumKit, building with all downstream dependencies
*** Copying from shared cache for Alamofire, skipping build
*** Copying from shared cache for Archimedes, skipping build
*** Copying from shared cache for GCDWebServer, skipping build
*** xcodebuild output can be found in /var/folders/7v/dfg0p2vj1z58n65d99l2n1lh0000gp/T/carthage-xcodebuild.dMgJ11.log
*** Copying from shared cache for ObjectMapper, skipping build
*** Copying from shared cache for AlamofireObjectMapper, skipping build
*** Copying from shared cache for SwiftiumKit, skipping build

real	0m15.297s
user	0m4.562s
sys	0m3.595s
```

Source/CarthageKit/Xcode.swift
```diff
@@ -947,15 +958,24 @@ public func buildInDirectory( // swiftlint:disable:this function_body_length
...
+							.flatMapError { _ in .empty }
+				}
+				return copyToSharedCache
+					.flatMap(.merge) { _ -> SignalProducer<(), CarthageError> in
```
ikesyo: > `var copyToSharedCache = SignalProducer<(URL), CarthageError>.empty`

and

> `.flatMapError { _ in .empty }`

flatMapping to `.empty` SignalProducer will not work because there are no values to map over. `then` operator should be appropriate here.

kenji21: I'm not so comfortable with ReactiveSwift, replacing `.flatMapError { _ in .empty }` by `.then(SignalProducer<(URL), CarthageError>.empty)` fails when running `make test` :

```
$ make test
[...]
Test Case '-[CarthageKitTests.ProjectSpec buildCheckedOutDependenciesWithOptions, createAndCheckVersionFiles, should rebuild cached frameworks (and dependencies) whose version does not match the version file]' started.
2019-04-12 12:07:20.497 xctest[57360:18611164] Building scheme "TestFramework3_Mac" in TestFramework3.xcodeproj
2019-04-12 12:07:31.388 xctest[57360:18611225] Building scheme "TestFramework2_Mac" in TestFramework2.xcodeproj
2019-04-12 12:07:42.378 xctest[57360:18611164] Building scheme "TestFramework1_Mac" in TestFramework1.xcodeproj
Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=NSCocoaErrorDomain Code=260 "The file “.TestFramework3.version” couldn’t be opened because there is no such file." UserInfo={NSFilePath=.build/x86_64-apple-macosx10.10/debug/CarthagePackageTests.xctest/Contents/Resources/DependencyTest/Carthage/Build/.TestFramework3.version, NSUnderlyingError=0x7f83abc83440 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}: file /BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang_Fall2018/swiftlang_Fall2018-1000.11.42/src/swift/stdlib/public/core/ErrorType.swift, line 184
Exited with signal code 4
make: *** [test] Error 1
```

ikesyo: No, I'm not saying about replacing `.flatMapError { _ in .empty }` with `.then(SignalProducer<(URL), CarthageError>.empty)`.

```swift
return copyToSharedCache.then(
  createVersionFile(
  	for: dependency.dependency,
  	version: dependency.version,
  	platforms: options.platforms,
  	buildProducts: urls,
  	rootDirectoryURL: rootDirectoryURL
  )
  .flatMapError { _ in .empty }
)
```

is what I mean.

kenji21: oh, i misread "flatMapping", sorry, I updated to change this simpler way, simpler, thanks for your review

kenji21:
Oups, didn't saw that CI was failing, I will update my commit again

Source/CarthageKit/Xcode.swift
```diff
@@ -6,6 +6,17 @@ import ReactiveSwift
 import ReactiveTask
 import XCDBLD
 
+/// Gets the shared cache file path URL for dependency at pinned version
+/// ~/Library/Caches/org.carthage.CarthageKit/SharedCache/XcodeVersion/DependencyName/Version, exemple :
+/// ~/Library/Caches/org.carthage.CarthageKit/SharedCache/9.4.1_9F2000/SwiftyUserDefaults/3.0.1
+internal func sharedCacheURLByXcodeVersionFor(dependency: Dependency, pinnedVersion: PinnedVersion) -> URL {
+	let sharedCacheURL = Constants.userCachesURL.appendingPathComponent("SharedCache")
+	let sharedCachePerXcode = sharedCacheURL.appendingPathComponent(XcodeVersion.makeString(), isDirectory: true)
```
werner77: I think you'll want to use the swift toolchain version, not the xcode version. You could switch toolchains within xcode leading to incompatible builds.

kenji21: you're right, i've replaced Xcode version with the toolchain one if it is set (kept XcodeVersion if no toolchain set in BuildOptions)

kenji21: new version visible here : https://github.com/Carthage/Carthage/commit/23d433f835d985d34ee6d8589230e478416268a2#diff-f66644e8cfbcad3c3826395a690ddfbcR12

kenji21:
So two commits behind master, current tests are fine
	 Executed 236 tests, with 0 failures (0 unexpected) in 672.691 (672.723) seconds

kenji21:
rebased to current master

Source/CarthageKit/Xcode.swift
werner77: Why not just use the xcrun swift --version like other places in the code? Even if not set explicitly (default) it will yield the correct swift toolchain version used.

werner77:
I looked at the code, but I think there is a much more elegant solution which does not involve creating a new cache and all the new code that goes with it: the existing binaries cache can just be reused by doing an archive step at the end of buildInDirectory and storing the archive in the binaries cache folder. The installBinaries step will just find the cached archive there and will use it instead of building from source.
This archive can be created by triggering the same code as the carthage archive action does (with concrete specified framework names).
The directory layout of the binaries cache does need to be modified to support different swift versions and/or build configurations though (which I did in my fork already).

kenji21:
The reason I don't want to zip / unzip is that APFS makes copy almost instantly, so when a dependency at is compiled at version X.Y with swift version Y.Z for one project, it is available for another project using the same dependency (let's say Alamofire, which can be in many project for example)

The other aim of the "shared cache" is to speed up Continuous Integration "clean" builds, by not having to recompile again and again dependencies that are not changing (see results in comment : https://github.com/Carthage/Carthage/issues/2400#issuecomment-466442108)

we discussed this in issue #2400 specifically well defined in this comment: https://github.com/Carthage/Carthage/issues/2400#issuecomment-402437688

blender:
@jdhealy do you want to give an opinion? I'll be reviewing again ASAP.

DerFlash:
This PR fails to build against official Carthage master because it's been updated for Swift 5.1 already.

```
.../Source/CarthageKit/Xcode.swift:972:12: error: ambiguous use of 'then'
                                return copyToSharedCache.then(
                                       ^
ReactiveSwift.SignalProducer<τ_0_0, τ_0_1>:5:17: note: found this candidate
    public func then<U>(_ replacement: ReactiveSwift.SignalProducer<U, Result.NoError>) -> ReactiveSwift.SignalProducer<U, Error>
                ^
ReactiveSwift.SignalProducer<τ_0_0, τ_0_1>:6:17: note: found this candidate
    public func then<Replacement>(_ replacement: Replacement) -> ReactiveSwift.SignalProducer<Replacement.Value, Error> where Replacement : SignalProducerConvertible, Replacement.Error == Result.NoError
                ^
ReactiveSwift.SignalProducer<τ_0_0, τ_0_1>:7:17: note: found this candidate
    public func then<U>(_ replacement: ReactiveSwift.SignalProducer<U, Error>) -> ReactiveSwift.SignalProducer<U, Error>
                ^
ReactiveSwift.SignalProducer<τ_0_0, τ_0_1>:8:17: note: found this candidate
    public func then<Replacement>(_ replacement: Replacement) -> ReactiveSwift.SignalProducer<Replacement.Value, Error> where Error == Replacement.Error, Replacement : SignalProducerConvertible
                ^

make: *** [installables] Error 1
```

Anyway, thanks for the good work. Build against Xcode 10.1 straight from the PR branch. Works like a charm!

Regards
Björn

kenji21:
it seems that using Xcode 10.2.1 wasn't a good idea... CI is down...

kenji21:
updated to current master, running `SWIFTPM_BOOTSTRAP=true make test` on my macbook :

```
	 Executed 236 tests, with 0 failures (0 unexpected) in 896.945 (896.963) seconds
```

DerFlash:
PR seems to be removed from the roadmap again. Any reason why?

joshuapoq:
Any chance of a rereview @mdiep @ikesyo? I’ve been watching this PR for years now as it will save weeks if not months of CI and developer time for our SDK.

Issue #2718, opened 2019-03-05
Swift header file not created correctly when building with Xcode 10.2


DavidBrunow:
* carthage install method: [ ] .pkg, [ ] homebrew, [x] source
* `which carthage`: Carthage binary in our path
* `carthage version`: 0.32.0
* `xcodebuild -version`: Xcode 10.2
Build version 10P99q
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? no
* Are you using `--new-resolver`? no

**Cartfile**
```
internal frameworks I cannot share
```

**Issue**

Swift 5 seems to have changed the format of the generated Swift header file by adding the following conditionals dependent on the architecture:

```bash
#if 0
#elif defined(__arm64__) && __arm64__
...
#elif defined(__ARM_ARCH_7A__) && __ARM_ARCH_7A__
...
```

When building the iOS platform with Carthage the generated Swift header file only has `__arm64__` and `__ARM_ARCH_7A__` but does not have any conditionals for the simulator architectures and therefore the Swift parts of the framework can't be used by Objective C code in the app.

I'm reviewing the Carthage code to find a way to resolve this issue but would appreciate any help pointing me towards the right files.

DavidBrunow:
Thinking about this more I can think of two different approaches that could be taken:

* Append the contents of the simulator `FrameworkName-Swift.h` to the end of the device `FrameworkName-Swift.h` file
* Create separate `Build` folders for simulators and devices

blender:
Carthage doesn’t touch any file in your project. This is a problem with your project’s settings.

DavidBrunow:
It may be true that it is a problem with my project's settings but I don't believe that to be the case because this dependency builds and can be used with no issues using Xcode 10.1. I understand that Carthage doesn't touch any of the files in the project but I believe that the way that the Swift compiler works for Swift 5 has changed the way that the generated `FrameworkName-Swift.h` files are created.

I'll create a small sample project to demonstrate the behavior with code that can be shared.

DavidBrunow:
I created a Small Sample Project and committed the `Carthage/Build` folder here: https://github.com/DavidBrunow/Small-Sample-Project

[...]

PR #2723, opened 2019-03-09
Add functionality to merge Swift header files

DavidBrunow:
This is meant to resolve an issue with Xcode 10.2 using the new build system where different architectures have conditional blocks in the Swift headers.

Issue: https://github.com/Carthage/Carthage/issues/2718

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -6,6 +6,8 @@ import ReactiveSwift
 import ReactiveTask
 import XCDBLD
 
+import struct Utility.Version
+
 /// Emits the currect Swift version
 internal func swiftVersion(usingToolchain toolchain: String? = nil) -> SignalProducer<String, SwiftVersionError> {
 	return determineSwiftVersion(usingToolchain: toolchain).replayLazily(upTo: 1)
@@ -355,6 +357,65 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
 		.then(SignalProducer<(), CarthageError>.empty)
 }
 
+private func findSwiftHeaderFiles(fromExecutableURLs executableURLs: [URL]) -> [URL] {
+	return executableURLs.map { 
+		let executableName = $0.lastPathComponent
+		return $0.deletingLastPathComponent().appendingPathComponent("Headers/\(executableName)-Swift.h") 
+	}
+}
+
+private func findSwiftHeaderOutputURL(fromURL url: URL) -> URL {
+	let executableName = url.lastPathComponent
+	
+	return url.deletingLastPathComponent().appendingPathComponent("Headers/\(executableName)-Swift.h")
+}
+
+private func mergeSwiftHeaderFiles(_ executableURLs: [URL], _ executableOutputURL: URL) -> SignalProducer<(), CarthageError> {
+	precondition(executableOutputURL.isFileURL)
+	precondition(executableURLs.count > 1)
+
+	let headerURLs = findSwiftHeaderFiles(fromExecutableURLs: executableURLs)
+	let outputURL = findSwiftHeaderOutputURL(fromURL: executableOutputURL)
+    let conditionalPrefix = "#if 0\n#elif (defined(__x86_64__) && __x86_64__) || (defined(__i386__) && __i386__)\n"
+    let conditionalSuffix = "\n#endif"
+    
+    guard let conditionalPrefixContents = conditionalPrefix.data(using: .utf8) else { return .empty }
+    guard let conditionalSuffixContents = conditionalSuffix.data(using: .utf8) else { return .empty }
+	
+	var fileContents = Data()
+
+	for url in headerURLs {
+		guard let contents = FileManager.default.contents(atPath: url.path) else { continue }
+
+        // Need to work around an inconsistency in the way that the Swift
+        // header files are created on simulators on different platforms. On
+        // iOS simulators the Swift header file has a structure that matches
+        // the Swift header file for devices because they both include
+        // conditionals for the supported platforms. This is not true for
+        // watchOS simulators (any maybe others?) at this time so we're going to
+        // add a conditional around those files
+        var needToAddConditionals = false
+        
+        if String(data: contents, encoding: .utf8)?.starts(with: "#if 0") == false {
+            needToAddConditionals = true
+        }
+        
+        if needToAddConditionals {
+            fileContents.append(conditionalPrefixContents)
+        }
+        
+		fileContents.append(contents)
+        
+        if needToAddConditionals {
+            fileContents.append(conditionalSuffixContents)
+        }
+	}
+
+	FileManager.default.createFile(atPath: outputURL.path, contents: fileContents, attributes: nil)
+	
+	return .empty
+}
+
 /// If the given source URL represents an LLVM module, copies its contents into
 /// the destination module.
 ///
@@ -474,6 +535,33 @@ private func mergeBuildProducts(
 					)
 				}
 
+			let mergeProductSwiftHeaderFilesIfNeeded = SignalProducer(result: executableURLs.fanout(outputURL))
+				.flatMap(.concat) { (executableURLs: [URL], outputURL:  URL) -> SignalProducer<(), CarthageError> in 
+					guard isSwiftFramework(productURL) else { return .empty }
+					
+					return swiftVersion()
+						.mapError { error in CarthageError.internalError(description: error.description) }
+						.flatMap(.concat) { localSwiftVersion -> SignalProducer<(), CarthageError> in
+							guard let versionSubstring = localSwiftVersion.split(separator: " ").first else { return .empty }
+							let scanner = Scanner(string: String(versionSubstring))
+                            guard let semanticVersion = Version.from(scanner).value else { return .empty }
+
+                            // Would rather not compare to a major Swift version
+                            // here and instead rely upon the structure of the
+                            // Swift header files to determine whether they get
+                            // merged together but as described in the
+                            // `mergeSwiftHeaderFiles(,)` implementation that
+                            // behavior does not appear to be consistent across
+                            // platforms at this time.
+							guard semanticVersion.major >= 5 else { return .empty }
+
+                            return mergeSwiftHeaderFiles(
+								executableURLs.map { $0.resolvingSymlinksInPath() },
+								outputURL.resolvingSymlinksInPath()
+							)
+					}
+				}
+
 			let sourceModulesURL = SignalProducer(result: simulatorBuildSettings.relativeModulesPath.fanout(simulatorBuildSettings.builtProductsDirectoryURL))
 				.filter { $0.0 != nil }
 				.map { modulesPath, productsURL in
@@ -492,6 +580,7 @@ private func mergeBuildProducts(
 				}
 
 			return mergeProductBinaries
+				.then(mergeProductSwiftHeaderFilesIfNeeded)
 				.then(mergeProductModules)
 				.then(copyBCSymbolMapsForBuildProductIntoDirectory(destinationFolderURL, simulatorBuildSettings))
 				.then(SignalProducer<URL, CarthageError>(value: productURL))
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -384,9 +384,10 @@ private func mergeSwiftHeaderFiles(_ simulatorExecutableURL: URL,
 	fileContents.append(deviceHeaderContents)
 	fileContents.append(conditionalSuffixContents)
 	
-	FileManager.default.createFile(atPath: outputURL.path, contents: fileContents, attributes: nil)
-	
-	return .empty
+	switch FileManager.default.createFile(atPath: outputURL.path, contents: fileContents) {
+	case false: return .init(error: .writeFailed(outputURL, nil))
+	case true: return .empty
+	}
 }
 
 /// If the given source URL represents an LLVM module, copies its contents into
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -357,34 +357,24 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
 		.then(SignalProducer<(), CarthageError>.empty)
 }
 
-private func findSwiftHeaderFiles(fromExecutableURLs executableURLs: [URL]) -> [URL] {
-	return executableURLs.map { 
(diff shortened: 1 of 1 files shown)


DavidBrunow:
I have no prior experience with ReactiveSwift so please let me know if I'm doing something wrong there or if there are any other improvements I could make.

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +357,65 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
...
+    
+    guard let conditionalPrefixContents = conditionalPrefix.data(using: .utf8) else { return .empty }
+    guard let conditionalSuffixContents = conditionalSuffix.data(using: .utf8) else { return .empty }
+	
+	var fileContents = Data()
+
+	for url in headerURLs {
+		guard let contents = FileManager.default.contents(atPath: url.path) else { continue }
+
+        // Need to work around an inconsistency in the way that the Swift
```
DavidBrunow: Created radar 48743840 to try to get this behavior to be consistent across platforms.

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +357,65 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
...
 }
 
+private func findSwiftHeaderFiles(fromExecutableURLs executableURLs: [URL]) -> [URL] {
+	return executableURLs.map { 
+		let executableName = $0.lastPathComponent
+		return $0.deletingLastPathComponent().appendingPathComponent("Headers/\(executableName)-Swift.h") 
+	}
+}
+
+private func findSwiftHeaderOutputURL(fromURL url: URL) -> URL {
```
blender: What do you think of renaming this function to `swiftHeaderURL(fromURL url: URL) -> URL`

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +357,65 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
 		.then(SignalProducer<(), CarthageError>.empty)
 }
 
+private func findSwiftHeaderFiles(fromExecutableURLs executableURLs: [URL]) -> [URL] {
```
blender: What do you think of deleting this function and just mapping over the array using the function below it

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +357,65 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
...
+	let executableName = url.lastPathComponent
+	
+	return url.deletingLastPathComponent().appendingPathComponent("Headers/\(executableName)-Swift.h")
+}
+
+private func mergeSwiftHeaderFiles(_ executableURLs: [URL], _ executableOutputURL: URL) -> SignalProducer<(), CarthageError> {
+	precondition(executableOutputURL.isFileURL)
+	precondition(executableURLs.count > 1)
+
+	let headerURLs = findSwiftHeaderFiles(fromExecutableURLs: executableURLs)
```
blender: `executableURLs.map(swiftHeaderURL(fromURL:))`

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +357,65 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
...
+	let outputURL = findSwiftHeaderOutputURL(fromURL: executableOutputURL)
```
blender: `swiftHeaderURL(fromURL: executableOutputURL)`

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +357,65 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
...
+        // Need to work around an inconsistency in the way that the Swift
+        // header files are created on simulators on different platforms. On
+        // iOS simulators the Swift header file has a structure that matches
+        // the Swift header file for devices because they both include
+        // conditionals for the supported platforms. This is not true for
+        // watchOS simulators (any maybe others?) at this time so we're going to
+        // add a conditional around those files
+        var needToAddConditionals = false
+        
+        if String(data: contents, encoding: .utf8)?.starts(with: "#if 0") == false {
```
blender: Have you considered just loading a sensible portion of the data and not the entire file?

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +357,65 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
...
+
+private func findSwiftHeaderOutputURL(fromURL url: URL) -> URL {
+	let executableName = url.lastPathComponent
+	
+	return url.deletingLastPathComponent().appendingPathComponent("Headers/\(executableName)-Swift.h")
+}
+
+private func mergeSwiftHeaderFiles(_ executableURLs: [URL], _ executableOutputURL: URL) -> SignalProducer<(), CarthageError> {
+	precondition(executableOutputURL.isFileURL)
+	precondition(executableURLs.count > 1)
```
blender: does it need not be the empty array?

Source/CarthageKit/Xcode.swift
```diff
@@ -474,6 +535,33 @@ private func mergeBuildProducts(
 					)
 				}
 
+			let mergeProductSwiftHeaderFilesIfNeeded = SignalProducer(result: executableURLs.fanout(outputURL))
```
ikesyo: Please use tabs for indentation.

Source/CarthageKit/Xcode.swift
ikesyo: We should be able to use the existing `URL.swiftHeaderURL()` extension instead of adding this new func.

Source/CarthageKit/Xcode.swift
DavidBrunow: Don't think so, removed.

Source/CarthageKit/Xcode.swift
DavidBrunow: Good idea -- I've changed it to not convert the data into a string at all and just compare data to data.

Source/CarthageKit/Xcode.swift
DavidBrunow: Used `URL.swiftHeaderURL()`.

Source/CarthageKit/Xcode.swift
DavidBrunow: Deleted this function and used existing extension to find the header file.

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +357,55 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
...
+private func mergeSwiftHeaderFiles(_ executableURLs: [URL], _ executableOutputURL: URL) -> SignalProducer<(), CarthageError> {
+	precondition(executableOutputURL.isFileURL)
+	
+	let conditionalsCheck = "#if 0"
+	let conditionalPrefix = "#if 0\n#elif (defined(__x86_64__) && __x86_64__) || (defined(__i386__) && __i386__)\n"
+	let conditionalSuffix = "\n#endif"
+	
+	guard let conditionalsCheckContents = conditionalsCheck.data(using: .utf8) else { return .empty }
+	guard let conditionalPrefixContents = conditionalPrefix.data(using: .utf8) else { return .empty }
+	guard let conditionalSuffixContents = conditionalSuffix.data(using: .utf8) else { return .empty }
```
mdiep: I think these should just use `!`. Since the contents are static, this absolutely a programmer error that we want to catch. Crashing will make it obvious; returning an empty producer will hide the error.

Source/CarthageKit/Xcode.swift
```diff
@@ -6,6 +6,8 @@ import ReactiveSwift
 import ReactiveTask
 import XCDBLD
 
+import struct Utility.Version
```
mdiep: You'll need to merge `master` because I merged #2720.

Source/CarthageKit/Xcode.swift
DavidBrunow: Good point, will do.

Source/CarthageKit/Xcode.swift
DavidBrunow: Will do.

dalemyers:
Is this good to go? Xcode 10.2 beta 4 came out yesterday, so we are likely only a few days away from the actual release. It would be good to have a Carthage update out with this fix before that point.

blender:
@DavidBrunow  I was under the impression that this behavior is only exhibited with the new build system, correct?

What do we do with the legacy system?

dalemyers:
The bug has come to light with the new build system, but it has always been an issue if anyone has simulator/device specific code in generated headers. Doing the same fix for the older versions would be a no-op for the most part though.

DavidBrunow:
> Is this good to go? Xcode 10.2 beta 4 came out yesterday, so we are likely only a few days away from the actual release. It would be good to have a Carthage update out with this fix before that point.

Beta 4 came out a couple of weeks ago, don't know if there will be another beta or not before what I am assuming is an Xcode release near the March 25th event.

I need to figure out the root cause of this other issue (https://github.com/Carthage/Carthage/issues/2724) before I could use a new Carthage release.

dalemyers:
Sorry, my bad. I don't know where I got it in my head that a new beta came out yesterday. Either way, I know for sure that the actual release is coming very soon.

blender:
This LGTM

mdiep:
We'll give @jdhealy one more day to review.

Source/CarthageKit/Xcode.swift
```diff
@@ -474,6 +508,17 @@ private func mergeBuildProducts(
 					)
 				}
 
+			let mergeProductSwiftHeaderFilesIfNeeded = SignalProducer.zip(simulatorBuildSettings.executableURL, deviceBuildSettings.executableURL, outputURL)
+				.flatMap(.concat) { (simulatorURL: URL, deviceURL: URL, outputURL:  URL) -> SignalProducer<(), CarthageError> in
+					guard isSwiftFramework(productURL) else { return .empty }
```
jdhealy: I'm onboard with the current solution of only header-merging `swiftmodule`-containing frameworks, but we might consider widening this to all frameworks once we have more time to gather data on if that's mostly safe or less-so.

We've [kinda been given indications](https://twitter.com/UINT_MIN/status/1100085577158815755) that differing of headers might also happen under some non-Swift-framework criteria via the new build system.

I've not actually seen (or seen indications) that these non-Swift-but-still-mostly-vanilla frameworks that produce differing headers might already be out there — but if we get reports of them, we should put out another release that drops this `isSwiftFramework` (in some form) and accommodates them.

Source/CarthageKit/Xcode.swift
```diff
@@ -355,6 +355,40 @@ private func mergeExecutables(_ executableURLs: [URL], _ outputURL: URL) -> Sign
...
+	
+	var fileContents = Data()
+	
+	fileContents.append(conditionalPrefixContents)
+	fileContents.append(simulatorHeaderContents)
+	fileContents.append(conditionalElseContents)
+	fileContents.append(deviceHeaderContents)
+	fileContents.append(conditionalSuffixContents)
+	
+	FileManager.default.createFile(atPath: outputURL.path, contents: fileContents, attributes: nil)
```
jdhealy: In the case where overwriting this file doesn't succeed, `FileManager.createFile` will output `false` — we should check for that — erroring out, rather than letting a potential non-overwrite of that header go unaddressed:

```swift
        switch FileManager.default.createFile(atPath: outputURL.path, contents: fileContents) {
        case false: return .init(error: .writeFailed(outputURL, nil))
        case true: return .empty
        }
```

jdhealy:
@DavidBrunow Most of all, thanks so much for your diagnosing of this and working through this PR ✨

Source/CarthageKit/Xcode.swift
DavidBrunow: Good catch, should be resolved now.

jdhealy: Wow, fantastic, very fast — thank you!

mdiep:
Awesome, thanks @DavidBrunow! Really appreciate this contribution. 👏🏻

bvirlet:
Thanks for implementing this! 
Do you know when we can expect a release of Carthage with this fix?

mdiep:
We're planning to release sometime this week

tonnihyldgaard:
Hi @mdiep, do you have an ETA on when this will be released? :)

srdanrasic:
Thanks guys for on working on this! I would really love to see this released, we have multiple projects suffering from the issue.

mdiep:
I'm hoping to release this today or tomorrow. We've started working on a release, but there are a couple things we need to work through.

JetForMe:
I downloaded the current master and tried using that Carthage to update my dependencies. It made a combined header file, but it seems `TARGET_OS_SIMULATOR` isn't being set by Xcode when I build my project, so it still doesn't see the header contents. I could just be doing something wrong. The attached project has some `#error` directives in the generated `Valet-Swift.h` file. Even thought I'm building for simulator, the directive in the `#else` section of `#if TARGET_OS_SIMULATOR` is what's triggered on build.

[TestValetObjC.zip](https://github.com/Carthage/Carthage/files/3020571/TestValetObjC.zip)

nighthawk:
@JetForMe sounds like this PR is also me is also needed then: https://github.com/Carthage/Carthage/pull/2748

JetForMe:
Ah yes, that seems like it would address the issue. Here I was blaming Xcode for not pre-setting it, but I think that PR is the right solution.

quangDecember:
Hi, which version will this fix delivered? I updated my Carthage to 0.32 via HomeBrew already, still suffers the problem

intoxicated:
I'm also having this issue which I can't build react native project (objective-c) with swift frameworks built by carthage. I'm also using 0.32..

mdiep:
0.33.0 has been released

JetForMe:
Yep, updated a bit ago. First run with it segfaulted, but that didn't happen again on a subsequent run. Now I'm running into an issue building PromiseKit with it that I haven't had time to run down. PromiseKit couldn't find its Swift bridging header:

```
…/Carthage/Checkouts/PromiseKit/Sources/AnyPromise.m:4:13: fatal error: 'PromiseKit/PromiseKit-Swift.h' file not found
    #import <PromiseKit/PromiseKit-Swift.h>
```

Not sure if that's related to Xcode 10.2 or something else entirely.

The other problem I think we're going to run into is Carthage projects not being configured to build with Swift 5.

Issue #2724, opened 2019-03-09
`carthage build` hangs when `Carthage/Build` folder does not exist


DavidBrunow:
* carthage install method: [ ] .pkg, [ ] homebrew, [X] source
* `which carthage`: binary from master
* `carthage version`: master
* `xcodebuild -version`: Xcode 10.1
Build version 10B61
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? no
* Are you using `--new-resolver`? no

**Cartfile**
```
github "ReactiveX/RxSwift" ~> 4.0
```

**Carthage Output**
```
$ carthage build
Please update to the latest Carthage version: 0.32.0. You currently are on 0.31.2
*** xcodebuild output can be found in /var/folders/jx/jckhr36n021bb_pwd07nrvdc0000gn/T/carthage-xcodebuild.wO1JkA.log
```

**Actual outcome**
Carthage did or did not download the binary for my dependency when they were already downloaded to my `Carthage/Checkouts` folder and the `Carthage/Build` folder did not exist (conditions used in CI environment). Instead it did not respond for minutes and I eventually killed it.

**Expected outcome**
Carthage should download the binaries for my dependencies based upon the files in the `Carthage/Checkouts` folder.

I've identified the recent change to add the `GitHubURLSessionDelegate` to be the source of the issue: https://github.com/Carthage/Carthage/commit/444224eaac53c37a9549733b434f3de181f3d836

DavidBrunow:
Updated with what I've learned about this applying to GitHub as well as GitHub Enterprise.

PR #2713, opened 2019-02-23
Pass the Authorization header across to the redirected request

NxSoftware:
Should resolve #2711

Issue #2711, opened 2019-02-19
GitHub Access Token not being used causing Rate Limit to be hit


NxSoftware:
* carthage install method: Pre-installed on Azure Pipelines Hosted macOS agent
* `which carthage`: 
* `carthage version`: 0.29.0
* `xcodebuild -version`: Xcode 10.1 Build version 10B61
* Are you using `--no-build`?  No
* Are you using `--no-use-binaries`? No
* Are you using `--use-submodules`? No
* Are you using `--cache-builds`? No
* Are you using `--new-resolver`? No

**Cartfile**
```
github "aws/aws-sdk-ios" == 2.6.31
github "daltoniam/Starscream" ~> 3.0
github "socketio/socket.io-client-swift" == 13.3.1
github "kishikawakatsumi/KeychainAccess" == 3.1.2
```

**Carthage Output**
```
2019-02-19T08:21:37.4263280Z *** No Cartfile.resolved found, updating dependencies
2019-02-19T08:21:37.4628150Z *** Fetching Starscream
2019-02-19T08:21:38.4336330Z *** Fetching socket.io-client-swift
2019-02-19T08:21:39.0465680Z *** Fetching KeychainAccess
2019-02-19T08:21:39.6142120Z *** Cloning aws-sdk-ios
2019-02-19T08:21:52.5121650Z *** Checking out aws-sdk-ios at "2.6.31"
2019-02-19T08:21:52.5157120Z *** Checking out Starscream at "3.0.6"
2019-02-19T08:21:52.5158360Z *** Checking out socket.io-client-swift at "v13.3.1"
2019-02-19T08:21:52.5158740Z *** Checking out KeychainAccess at "v3.1.2"
2019-02-19T08:21:53.1632780Z *** xcodebuild output can be found in /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/carthage-xcodebuild.B9zfOW.log
2019-02-19T08:21:53.6859200Z *** Skipped downloading aws-sdk-ios.framework binary due to the error:
2019-02-19T08:21:53.6860170Z 	"API rate limit exceeded for A.B.C.D. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"
```

**Actual outcome**
Carthage did not successfully download the frameworks from GitHub as it appears to have ignored the supplied GITHUB_ACCESS_TOKEN.

**Expected outcome**
Carthage should use the access token to make use of higher rate limits.

### More Info
[...]

PR #2725, opened 2019-03-09
Fix issue where `carthage build` does not download binaries for dependencies in a specific situation

DavidBrunow:
This PR does not work at this time. When it was originally created I believed that the source of the problem was due to the differences between GitHub and GitHub Enterprise. Since that time I have been able to reproduce the issue on GitHub as well so this fix is not valid. 

Steps to reproduce: https://github.com/Carthage/Carthage/pull/2725#issuecomment-475957108

Fix for issue: https://github.com/Carthage/Carthage/issues/2724

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -148,7 +148,13 @@ extension Client {
 			Client.userAgent = gitHubUserAgent()
 		}
 
-		let urlSession = URLSession.proxiedSession
+        var urlSession: URLSession
+        
+        if server.url.host?.contains("github.com") == true {
+            urlSession = URLSession.proxiedSessionWithDelegate
+        } else {
+            urlSession = URLSession.proxiedSession
+        }
 
 		if !isAuthenticated {
 			self.init(server, urlSession: urlSession)
@@ -163,7 +169,14 @@ extension Client {
 }
 
 extension URLSession {
-	public static var proxiedSession: URLSession {
+    public static var proxiedSession: URLSession {
+        let configuration = URLSessionConfiguration.default
+        configuration.connectionProxyDictionary = Proxy.default.connectionProxyDictionary
+        
+        return URLSession(configuration: configuration)
+    }
+    
+	public static var proxiedSessionWithDelegate: URLSession {
 		let configuration = URLSessionConfiguration.default
 		configuration.connectionProxyDictionary = Proxy.default.connectionProxyDictionary
 
--- a/IntegrationTests/build.bats
+++ b/IntegrationTests/build.bats
@@ -17,3 +17,14 @@ EOF
     [ "$status" -eq 0 ]
     [ -e Carthage/Build/iOS/MMMarkdown.framework ]
 }
+
+@test "carthage build works right IN MY SPECIAL SCENARIO WHICH NEEDS TO BE DESCRIBED HERE" {
+cat >| Cartfile <<-EOF
+github "ReactiveX/RxSwift" ~> 4.0
+EOF
+run carthage update
+rm -rf Carthage/Build
+run carthage build
+[ "$status" -eq 0 ]
+[ -e Carthage/Build/Mac/RxSwift.framework ]
+}


Source/CarthageKit/GitHub.swift
```diff
@@ -148,7 +148,13 @@ extension Client {
...
 		}
 
-		let urlSession = URLSession.proxiedSession
+        var urlSession: URLSession
+        
+        if server.url.host?.contains("github.com") == true {
+            urlSession = URLSession.proxiedSessionWithDelegate
+        } else {
+            urlSession = URLSession.proxiedSession
+        }
```
mdiep: Any idea why this is needed? The delegate implementation below seems like it should be correct—even for enterprise repositories.

My preference would be to fix the delegate if possible instead of introducing a second proxy.

DavidBrunow: Unfortunately I haven't been able to figure that out yet. I tried implementing all the delegate methods to see if the session was getting caught up in one of them and none of them seemed to get hit. The other strange thing is that a `carthage update` command works normally while the `carthage build` does not. I have not dug into the differences in the implementations yet to figure out why that may be.

DavidBrunow: I have some time to look into this more today and I think I was wrong in thinking that the issue was with GitHub Enterprise. I think the issue is with downloading binaries and might affect GitHub dependencies as well. Still in the process of tracking down the issue.

DavidBrunow: I've tracked the way that Carthage "hangs" into ReactiveSwift in FoundationExtensions.swift. Given this code:

```swift
extension Reactive where Base: URLSession {
	/// Returns a SignalProducer which performs the work associated with an
	/// `NSURLSession`
	///
	/// - parameters:
	///   - request: A request that will be performed when the producer is
	///              started
	///
	/// - returns: A producer that will execute the given request once for each
	///            invocation of `start()`.
	///
	/// - note: This method will not send an error event in the case of a server
	///         side error (i.e. when a response with status code other than
	///         200...299 is received).
	public func data(with request: URLRequest) -> SignalProducer<(Data, URLResponse), AnyError> {
		return SignalProducer { [base = self.base] observer, lifetime in
			print("here in data with request: \(request)")
			
			let task = base.dataTask(with: request) { data, response, error in
				print("here in task completion handler: \(error)")
				if let data = data, let response = response {
					observer.send(value: (data, response))
					observer.sendCompleted()
				} else {
					observer.send(error: AnyError(error ?? defaultSessionError))
				}
			}

			lifetime.observeEnded(task.cancel)
			task.resume()
		}
	}
}
```

The first print statement gets hit 4 times (which makes sense because the concurrent downloads are limited to 4) but the second print statement never gets hit.

I'm out of ideas and energy at this point.

mdiep: I'm able to download a binary from GitHub with `master`. 🤷🏻‍♂️

DavidBrunow:
The steps to reproduce might be different, here is the situation I have:

* Carthage/Checkouts is committed to git
* Carthage/Build does not exist
* Running carthage build

If Carthage/Build has the .version files in it then the binaries get downloaded properly.

Can you easily test that scenario? Or is that what you already did? I’m planning on testing that if you don’t get a chance but it will be a couple days.

mdiep:
That's very odd. Can you try to put together a reproducible test case?

DavidBrunow:
Yep, will do but might be a couple of days.

DavidBrunow:
Here is a reproducible test case: https://github.com/DavidBrunow/Small-Sample-Project

I've committed the `Carthage/Checkouts` folder but not the `Carthage/Build` folder. If you run `carthage build` with a Carthage build off of master it should hang after outputting `*** xcodebuild output can be found in...`.

I'm going to look into how to add this scenario to the Carthage test suite but any help pointing me in the right direction would be appreciated.

DavidBrunow:
Added an attempt at a test scenario in the `IntegrationTests` folder. Unfortunately since Carthage hangs this test never completes. The test scenario also needs some cleanup that I'll implement once I know that this is the right direction to go with the test.

NxSoftware:
I suspect this is due to the fact the [delegate queue in the reverted code](https://github.com/Carthage/Carthage/pull/2741/files#diff-f180fa8591016e7668e628c72dd98d31L171) was set to `OperationQueue.current`. While the original implementation was not specifying a queue, thus a new serial operation queue would be used.

I have changed my original fix  ([PR 2713](https://github.com/Carthage/Carthage/pull/2713) for issue #2711) to no longer provide a delegate queue, see https://github.com/NxSoftware/Carthage/blob/redirect-auth/Source/CarthageKit/GitHub.swift#L171.

@DavidBrunow, unfortunately I've been unable to verify the described hang using the [sample repo](https://github.com/DavidBrunow/Small-Sample-Project) with `carthage bootstrap` so would appreciate if you could try it out on my fork's branch "[redirect-auth](https://github.com/NxSoftware/Carthage/tree/redirect-auth)" and let me know how it goes before I create another PR.

DavidBrunow:
@DavidBrunow Will verify but it will be a few days before I can get to it.

DavidBrunow:
@NxSoftware Will verify but it will be a few days before I can get to it.

PR #2731, opened 2019-03-13
Automate copy-frameworks

### LINKED ISSUE #2605: Automate `copy-frameworks` command?
Hello Carthage team! 

As you know in order to get things done we need to invoke `copy-frameworks` at the Build Phase. This also involves specifying a list of input files for each linked framework. This seems to be a good subject for an automation. Currently it might be a little annoying to keep this list in sync and to manually fill it in. This is the main motivation. 

From my attempt to make a draft of such functionality here is what I found:
Initially we need to grab list of the linked frameworks that was built by Carthage. Turns out that they are stored in the `Frameworks Build Phase`. Unfortunately built-in tool `xcodebuild` doesn't allow us to read such information, therefore we used to read contents of the `.xcodeproj` itself. There are well-known tools such as gem [xcodeproj](https://rubygems.org/gems/xcodeproj/versions/1.6.0) and its Swift counterpart [xcodeproj](https://github.com/tuist/xcodeproj). 

First of all we're reading contents of the Frameworks Build Phase. The rest of the process is very simple: select only those frameworks that are located in `Carthage/Build` but not in the `Static` subfolder. 
Then we simply iterating over found frameworks and filling in env variables like `SCRIPT_INPUT_FILE_` and so on. 

Below you can find working draft in ruby:
```ruby
#!/usr/bin/env ruby

require 'xcodeproj'

# We need to find linked frameworks that are located in the `Carthage/Build/*` folder
# but also doesn't appear to be static.
# Focus on the linked frameworks prevents us from accidental copying of the
# frameworks that belong to a different target (e.g. UnitTests).
def find_linked_frameworks_refs
  # These env variables passed by Xcode
  # during invocation of the Run Script Build Phase
  # Therefore running script on its own will have no effect.
  project_file_path = ENV["PROJECT_FILE_PATH"]
  target_name = ENV["TARGET_NAME"]

  return [] if project_file_path.nil? || target_name.nil?

  project = Xcodeproj::Project.open(project_file_path)
  target = project.targets.detect { |target| target.name == target_name }

  return [] if target.nil?

  refs = target.frameworks_build_phases.files_references

  # Selecting only Dynamic Framworks built by Carthage. 
  # Static frameworks also appears in this list but we don't need to process them.
  regexp = %r{^Carthage\/Build\/((?!\/Static\/).)*$}
  refs.select do |ref|
    regexp.match(ref.path) != nil
  end
end

def export_io_vars_for_refs(refs)
  # The same: env vars exported by Xcode. Manual invocation will have no effect.
  srcroot = ENV["SRCROOT"]
  build_products_directory = ENV["BUILT_PRODUCTS_DIR"]
  frameworks_folder_path = ENV["FRAMEWORKS_FOLDER_PATH"]

  return if srcroot.nil? || build_products_directory.nil? || frameworks_folder_path.nil?

  refs.each_with_index do |ref, index|
    # Ref has a relative path. We assume that Carthage folder located in the $SRCROOT.
    input_path = File.join(srcroot, ref.path)
    # Exporting variables for `carthage copy-frameworks`.
    ENV["SCRIPT_INPUT_FILE_#{index}"] = input_path

    # Specify output files to speed up execution.
    # My concern at this point is that I'm not sure whether specifying these
    # vars within script itself actually have any impact on Xcode.
    # It might be the case when Xcode evaluates equality of the input/output
    # files prior to run of the script.
    output_path = File.join(build_products_directory, frameworks_folder_path, File.basename(ref.path))
    ENV["SCRIPT_OUTPUT_FILE_#{index}"] = output_path
  end

  ENV["SCRIPT_INPUT_FILE_COUNT"] = refs.count.to_s
  ENV["SCRIPT_OUTPUT_FILE_COUNT"] = refs.count.to_s
end

refs = find_linked_frameworks_refs
if refs.count > 0 then
  export_io_vars_for_refs(refs)
  exec "/usr/local/bin/carthage copy-frameworks"
end
```

Simply put it to the new build phase and invoke as follow:
```
ruby ${SRCROOT}/copy-frameworks.rb
```

**Important:** script requires `Xcodeproj` gem. Therefore is case you're managing ruby via `rbenv`, please use following script:
```
export PATH=~/.rbenv/shims:$PATH
ruby ${SRCROOT}/copy-frameworks.rb
```

The only downside at this point is that this automation would require from Carthage to add one more dependency ([ruby gem](https://rubygems.org/gems/xcodeproj/versions/1.6.0) or [swift counterpart](https://github.com/tuist/xcodeproj)).

UPD:
Alternatives considered: 
Simply copy contents of the `Carthage/Build/...` to the destination. 
Pros: removes extra dependency
Cons: copies frameworks from other targets (if any) to the destination target.

dimazen:
This PR resolves https://github.com/Carthage/Carthage/issues/2605. 
Tested on a test project as well as two production projects - works as expected. However, I need your advise on: 
- How to better write tests for this case? For integration testing we would need a project with couple of frameworks and an executable to run `otool -L`. I don't know for sure, but MachO format seems to be quite stable, so no changes from otool expected.
- I'm not sure whether check for static frameworks is reliable or not. `frameworksInDirectory` returns both static and dynamic frameworks. Attempt to filter them out by checking their MachHeader.fileType doesn't help, because both return `MH_EXECUTE`. I end up checking whether framework nested in the `Static` folder or not but is that ok?


blender:
> How to better write tests for this case? For integration testing we would need a project with couple of frameworks and an executable to run otool -L. I don't know for sure, but MachO format seems to be quite stable, so no changes from otool expected. 

Using otool is perfectly fine. You can add integration tests, check `bats` files.

> I'm not sure whether check for static frameworks is reliable or not. frameworksInDirectory returns both static and dynamic frameworks. Attempt to filter them out by checking their MachHeader.fileType doesn't help, because both return MH_EXECUTE. I end up checking whether framework nested in the Static folder or not but is that ok?

`MH_EXECUTE` means they are executable files, not frameworks. `MH_OBJECT` is for static libraries or object files and MH_DYLIB for dynamically linked libraries.

Looking into the `Static` directory is ok because carthage has already separated static from dynamic libraries before hand.

dimazen:
@blender when I'm checking a Static Library (for Bolts in particular) file type, it reports Executable (surprisingly!) and not the MH_OBJECT which I was expecting. But I'll recheck. However, if `Static` component check is fine, then lets leave it as is.

I'll review `bats` files and ping you if needed.

Source/CarthageKit/Project.swift
```diff
@@ -1521,3 +1559,44 @@ public func cloneOrFetch(
...
+
+/// Stripping linked shared frameworks from
+/// @rpath/Alamofire.framework/Alamofire (compatibility version 1.0.0, current version 1.0.0)
+/// to Alamofire as well as filtering out system frameworks and various dylibs.
+/// Static frameworks and libraries won't show up here, so we can ignore them.
+///
+/// - Parameter input: Output of the otool -L
+/// - Returns: Array of Shared Framework IDs.
+internal func sharedLinkedFrameworkIDs(from input: String) -> [String] {
+    guard let regex = try? NSRegularExpression(pattern: "@rpath.*\\.framework\\/([\\w]+)") else {
```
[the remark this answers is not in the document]
dimazen: @blender we can rely on the `[\\w]+\\.framework\\/([\\w]+)` instead. This will give us also system frameworks like Foundation, however I don't see any performance penalty because of this. Will update regex later today.

bwhiteley:
Will this search for frameworks in `FRAMEWORK_SEARCH_PATHS` or only under `Carthage/Build/<platform>` ?

dimazen:
@bwhiteley it only looks at the `Carthage/Build/<platform>`, since that's the default way of the Carthage usage. Or there might be some issues by not evaluating `FRAMEWORK_SEARCH_PATH`?

typfel:
While working on changes to a framework I often drag in the framework in question as a subproject in the main project. I then modify the copy phase to take the dependency from the  `BUILT_PRODUCTS_DIR` instead of `Carthage/Build`, this enables me to make changes to the framework and have them immediately available in the main project. 

Would such a workflow also be possible when using this automated copy phase?

blender:
The manually specified list will still be there @typfel. 

Maybe there should be a set union based on the framework name, where the manual list wins over the inferred list.

dimazen:
@typfel you shouldn't worry. This PR omits need to specify standard paths such as `Carthage/Build/iOS/Alamofire.framework`. It doesn't replace this functionality but automates whenever possible. 
That's a perfect catch, however, to give a higher priority for any manually specified values. I'll incorporate those changes on the weekends, @blender.

blender:
@dimazen I would also consider adding a flag to enable this behavior, making it opt-in rather than default.

bwhiteley:
Here is how I would hope this feature would work...

1. Add first level dependencies to Linked Frameworks and Libraries.
2. One-liner build phase for copy frameworks with no input/output file lists. 
3. Copy frameworks phase builds the dependency graph recursively with `otool -L`. 
4. Frameworks are searched for in `FRAMEWORK_SEARCH_PATH`. This could be enabled with a command-line switch with the default behavior only looking in `Carthage/Build/<platform>`.
5. Copy frameworks phase checks timestamps to skip frameworks that are already up-to-date. 

(This is how our Carthage wrapper works, mentioned in #2605 . When it was mentioned in #2605 it was reading the project file to get the framework list, but has since changed to use `otool -L`)

The reason for #4 is similar to what @typfel said. We have a lot of internal frameworks. Often we work on the framework individually but sometimes we work on the framework in the context of a consuming app (or another framework). To work on a framework in the context of a consuming app, simply create a temporary workspace containing the app project and the framework project. Due to #4, if the framework project is in the workspace the built framework in `BUILT_PRODUCTS_DIR` is preferred to the Carthage-built framework. To switch back to the Carthage framework just close the workspace and open the app project instead of the workspace (since `BUILT_PRODUCTS_DIR` by default is different depending on whether you open the workspace or the app project). The nice thing about this approach is that no changes to the project are necessary to switch a framework into or out of "development mode" (by opening a workspace vs. the project). You don't have to adjust  framework paths; it just automatically does what you would expect when the framework project is in a workspace . You can have multiple workspaces with different subsets of frameworks. Or you can quickly change which frameworks are in "development mode" by adding or removing projects to a workspace. Again, no project modifications necessary. 

It would be great if the Carthage team found this behavior desirable because then we could use carthage directly in our copy frameworks build phase and remove a bunch of functionality from our wrapper tool. :) 

I've seen at least three separate implementation of a tool to flip frameworks into "development mode" in a carthage environment, so I know we aren't the only group interested in this functionality. 

Perhaps if #4 isn't part of this PR it could come later...

blender:
@bwhiteley are you sure that if you use a workspace `copy-frameworks` is needed?

dimazen:
@bwhiteley
3. For the recursively linked frameworks I'm going to consult with Cartfile.resolved to ignore frameworks that are out of scope. However, does development dependencies somehow reflected in the `Cartfile.resolved` or `Cartfile`? We definitely don't want to start linking automatically all of the frameworks we'll be able to find :) 
4. As above I need to ensure, that linked frameworks listed within Cartfile or Cartfile.resolved (preferable). 
5. That's definitely needs to be done as a separate PR :) 

I gonna play around with development projects to see how they affect `xcodebuild` build settings to see what we can do with it. If you have a sample project - please attach it here :) 

@blender what do you think about automatic linking of the development dependencies? Should we add separate option for this? 
- `--auto` will utilise automatic linked frameworks inference and copying
- `--prefer-development-builds` will be used to indicate that we wanna copy dependency from the build dir (or whatever env it'll be). 

Option names are just an example (especially second one).

bwhiteley:
> @bwhiteley are you sure that if you use a workspace `copy-frameworks` is needed?

If you always use a workspace (meaning you probably aren't using carthage to build the frameworks, only to resolve/fetch them) then you probably have the frameworks listed in Emedded Binaries, and you don't need a carthage copy-frameworks build phase. However, if you normally build the frameworks with carthage and open the project file, but sometimes use a workspace to temporarily have a framework in "development" mode, then the carthage copy-frameworks phase is needed to avoid needing to temporarily modify the project (such as temporarily adding the frameworks to Emedded Binaries with correct paths).

bwhiteley:
> 3. For the recursively linked frameworks I'm going to consult with Cartfile.resolved to ignore frameworks that are out of scope. However, does development dependencies somehow reflected in the `Cartfile.resolved` or `Cartfile`? We definitely don't want to start linking automatically all of the frameworks we'll be able to find :)
> 4. As above I need to ensure, that linked frameworks listed within Cartfile or Cartfile.resolved (preferable).

But `Cartfile` and `Cartfile.resolved` list repos, not frameworks. It is common for a repo to contain multiple frameworks or for the framework to have a different name from the repo. As a result, `Cartfile.resolved` can't be used to filter the frameworks. 

A dependency graph recursively built with `otool -L` rooted at `EXECUTABLE_PATH` is sufficient to get all the necessary frameworks without pulling in extraneous frameworks.

bwhiteley:
It will also be useful to print the input files that were found to satisfy the graph with their full paths, as well as indicating any that were skipped because the corresponding output was already up-to-date. This could be enabled with a verbose switch, or just do it by default.

dimazen:
Here is what I found so far:
- Automatic copying of the nested dependencies is possible. When I removed FBSDKCoreKit and Bolts from the Linked Frameworks Phase but copied them over to the app it was fine. What I'm not clear about is how `canImport` works in this case, because it is STILL works even if framework is not linked explicitly. Also, `NSClassFromString` returns me correct classes. I would assume that everything is still fine, because `dyld` was able to load `FBSDKCoreKit` for `FacebookCore` but looking up at the @rpath/Frameworks/FBSDKCoreKit.framework/FBSDKCoreKit`. 
My suggestion about `Cartfile` and `Cartfile.resolved` was just assumption. `otool` works great.  

- Regarding `FRAMEWORK_SEARCH_PATH` usage: it's a bit more tricky than I was thinking initially. What if user have specified multiple locations that contain a framework we're looking for? Which directory should we pick? I don't see for now how to make it explicit for user in which order we're going to traverse directories. What if one of the directories contains framework copy (different build/version) that we actually don't want to copy?
However, relying on the manually specified script input file works fine, because it takes precednse over `./Carthage/Build/<platform>/...`. Any suggestions on this ambiguity? 

Also, can you provide me some sample project with development dependency you're talking about please? I don't seem to understand how it works (never worked like that before).

bwhiteley:
> * Regarding `FRAMEWORK_SEARCH_PATH` usage: it's a bit more tricky than I was thinking initially. What if user have specified multiple locations that contain a framework we're looking for? Which directory should we pick? I don't see for now how to make it explicit for user in which order we're going to traverse directories. What if one of the directories contains framework copy (different build/version) that we actually don't want to copy?
>   However, relying on the manually specified script input file works fine, because it takes precednse over `./Carthage/Build/<platform>/...`. Any suggestions on this ambiguity?

`FRAMEWORK_SEARCH_PATH` is an array of directories in priority order. If two directories contain the same framework, the directory listed first wins. This ensures that the bundled frameworks match those used by Xcode while compiling/linking (which also honors the priority order). 

Using `FRAMEWORK_SEARCH_PATH` might simplify the algorithm because you don't need special cases for `BUILT_PRODUCTS_DIR` or `Carthage/Build/<platform>`. Those are both in `FRAMEWORK_SEARCH_PATH` in the correct order. 

> 
> Also, can you provide me some sample project with development dependency you're talking about please? I don't seem to understand how it works (never worked like that before).

Yeah, I can put something together.

dimazen:
@bwhiteley ok, I got what you mean about `FRAMEWORK_SEARCH_PATH`. And I'm afraid it won't work. Here is why: right know I'm traversing linked frameworks from the main executable and filtering out all them that are out of the `Carthage` build directory. This acts as a constrain to prevent copying frameworks that are managed by user / other package manager. In case of `FRAMEWORK_SEARCH_PATH` there are no more such constraints which mean that we'll end up copying everything that we was linked and located in the specified directories. Hence we may copy frameworks from the `Embed Frameworks` phase and other frameworks that are not handled by the Carthage at all. Therefore we can start even interfering with CocoaPods :)

bwhiteley:
> @bwhiteley ok, I got what you mean about `FRAMEWORK_SEARCH_PATH`. And I'm afraid it won't work. Here is why: right know I'm traversing linked frameworks from the main executable and filtering out all of them that are out of the `Carthage` build directory. This acts as a constrain to prevent copying frameworks that are managed by user / other package manager. In case of `FRAMEWORK_SEARCH_PATH` there are no more such constraints which mean that we'll end up copying everything that was linked and located in the specified directories. Hence we may copy frameworks from the `Embed Frameworks` phase and other frameworks that are not handled by the Carthage at all. Therefore we can start even interfering with CocoaPods :)

The timestamp/up-to-date check takes care of this. The only constraint is that the carthage copy-frameworks build phase come after the `Embed Frameworks` build phase and any cocoapods build phase. Some of our app projects have "internal" framework targets with the frameworks listed in `Embedded Binaries`. These frameworks are skipped because nothing needs to be done.

dimazen:
@bwhiteley here is what I end up with at this point:
- Timestamp check will be added as a separate PR. Out of scope for now.
- Recursive copying is there.
- I've limited amount of frameworks that I'm allowed to copy only to two scopes: frameworks that are located at `./Carthage/Build/<platform>/` and to the manually specified. Inspection and automatic copying of all of the linked frameworks should be prohibited at least at this point because of the following reason: 
Imaging you have some Umbrella framework that copies some nested frameworks on it's own (has separate call to `copy-frameworks`). This framework linked against your main app that in turn has it's own Carthage dependencies. By recursive inspection we will find both target's and umbrella's dependencies and gonna copy then to the main target's frameworks directory. It mean that in the final `Sample.app` frameworks from the Umbrella will be copied twice: to the `Sample.app/Frameworks` and to the `Sample.app/Frameworks/Umbrella.framework/Frameworks`. And this is something we can not handle easily (or at least I have no clue at this point) without some addition `xcodeproj` parsing. 

--

@blender I need you suggestion :) I've separated whole logic to the [InputFilesInferrer](
https://github.com/Carthage/Carthage/pull/2731/files#diff-f7daae01bb94ae19222b4cc8fb2541f4) alongside with [InputFilesInferrerSpec](
https://github.com/Carthage/Carthage/pull/2731/files#diff-651c8319349ad65441c059e907bfa1ac). Spec is running without touching any disk or so, therefore bat tests I left out of PR. Will this be enough?

One more thing I'm struggling with is: I can't run all of the tests from the Xcode - they're failing in the `ArchiveSpec.swift:10`
```
let archiveURL = Bundle(for: type(of: self)).url(forResource: "CartfilePrivateOnly", withExtension: "zip")!
> Thread 1: Fatal error: Unexpectedly found nil while unwrapping an Optional value
```

However, I haven't touched anything related to the tests nor have local uncommitted changes. This is the reason tests check on GitHub is failing as well (from what I can see). Any suggestions on how to fix that?

werner77:
In our fork (https://github.com/nsoperations/Carthage) I got rid of Quick for testing which seems to solve this random Unexpectedly found nil in test cases. Also it makes running individual tests from Xcode possible and swift test --filter now works as expected.

dimazen:
Hello, @werner77!  I'll try it. My main issue happens when I'm running Carthage's tests from the Xcode (10.1 at this moment). Weird thing is that CI seems to be able to run tests successfully randomly - some of my commits succeed while other - failed. 

Anyway, thanks for getting back with your suggestion!

blender:
@dimazen  sorry for the delay. I'll try to answer your questions ASAP.

Source/CarthageKit/InputFilesInferrer.swift
```diff
@@ -0,0 +1,148 @@
...
+    private let builtFrameworks: SignalProducer<URL, CarthageError>
+    private let linkedFrameworksResolver: LinkedFrameworksResolver
+    
+    init(builtFrameworks: SignalProducer<URL, CarthageError>, linkedFrameworksResolver: @escaping LinkedFrameworksResolver) {
+        self.builtFrameworks = builtFrameworks
+        self.linkedFrameworksResolver = linkedFrameworksResolver
+    }
+
+    public convenience init(projectDirectory: URL, platform: Platform) {
+        let searchDirectory = projectDirectory.appendingPathComponent(platform.relativePath, isDirectory: true)
```
blender: Can you remind me what URL does this resolves to?

tmspzz (commented):
So we're not filtering by frameworks contained in Carthage/Build ?

dimazen:
@blender we are filtering by only frameworks that are located at the `Carthage/Build/<platform>/`. However, user can specify input file with any arbitrary location which is not subject for filtering.

blender:
> Spec is running without touching any disk or so, therefore bat tests I left out of PR. Will this be enough?

That is ok

Source/CarthageKit/InputFilesInferrer.swift
dimazen: @blender this one uses path to the `.xcodeproj` file and appends to it relative path for the given platform: i.e. for iOS - `Carthage/Build/iOS`

blender:
> @blender we are filtering by only frameworks that are located at the `Carthage/Build/<platform>/`. However, user can specify input file with any arbitrary location which is not subject for filtering.

is this the URL obtained from the line where I left the comment?   https://github.com/Carthage/Carthage/pull/2731#discussion_r272496954

dimazen:
@blender exactly. `CopyFrameworks` invokes this class with the project directory and platform. Init forms a final Carthage's artifacts path.

blender:
> By recursive inspection we will find both target's and umbrella's dependencies and gonna copy then to the main target's frameworks directory. It mean that in the final Sample.app frameworks from the Umbrella will be copied twice: to the Sample.app/Frameworks and to the Sample.app/Frameworks/Umbrella.framework/Frameworks. And this is something we can not handle easily (or at least I have no clue at this point) without some addition xcodeproj parsing.

but the load dyld load path should be different. That is how you can tell where to copy maybe?

So to recap, so far:

- List linked frameworks from the executable
- Copy only the ones located in Carthage/Build/<platform>

It's a bit of a bummer to leave out the solution with `FRAMEWORK_SEARCH_PATH`

dimazen:
@blender I would assume that I've missed dyld path difference - doesn't have that much knowledge :)
Will play around with it on the weekend. 
@bwhiteley if that's still the case can you provide, please, some example with `FRAMEWORK_SEARCH_PATH`?

blender:
Also, this case with Umbrella frameworks is __only relevant for macOS__

dimazen:
@blender, here is just a small addition to your comment
> So to recap, so far:
> * List linked frameworks from the executable
> * Copy only the ones located in Carthage/Build/<platform>

Just to add: copy transitively linked dependencies automatically even if not linked directly to the executable. It means that linking `FacebookCore` (Swift version) only will automatically copy `FBSDKCore` (ObjC dependency) as well as `Bolts` - dependency of the `FBSDKCore`.

@bwhiteley 
Turns out that I've missed part where `FRAMEWORKS_SEARCH_PATH` lookup enabled by a separate flag. I went through your implementation in the [carttool](https://github.com/Ancestry/CartTool/blob/master/Sources/CartToolCore/XcodeProj.swift#L144) and now it is totally clear what you've meant initially. I'm sorry for misunderstanding. 
In this case we gonna add additional flag (`--use-framework-search-path` or maybe `--development-dependencies`? any suggestions?) as well as build time check to prevent copying of the wrong frameworks. In addition in a readme we will specify how to use this option to prevent any impact on the `CocoaPods` or `Embed Frameworks`. 
Have I understood your https://github.com/Carthage/Carthage/pull/2731#issuecomment-473154376 correctly? 
Again, sorry for misunderstanding. I was thinking that we should force `FRAMEWORKS_SEARCH_PATH` to be the default option rather than opt-in behaviour.

blender:
`--use-framework-search-path` seems good to me. 
Alternatively I would just say `--use-paths $FRAMWORKS_SEARCH_PATH` but I'm a bit worried if this is "too much flexible" and would cause confusion to the end user.

bwhiteley:
In `carttool` we always search `FRAMEWORKS_SEARCH_PATH` for the frameworks to embed and haven't had a problem with that. But if you think that is not generally safe, enabling that behavior with a command-line switch is fine. Since `FRAMEWORKS_SEARCH_PATH` contains both `BUILT_PRODUCTS_DIR` and `Carthage/Build/<platform>` in the correct order, it simplifies the implementation. 

As far as the name of the option, `--use-framework-search-path` is fine.

dimazen:
@blender code updated:
- add `--use-framework-search-paths` flag that will utilise `FRAMEWORK_SEARCH_PATHS` environment variable to search for the frameworks. In case single frameworks presented in multiple directories than first wins. Recursive paths (ends with `**`) handled as well. During recursive frameworks expansion Carthage will ignore various Xcode-specific directories such as .xcodeproj (full list [here](https://github.com/Carthage/Carthage/pull/2731/files#diff-f7daae01bb94ae19222b4cc8fb2541f4R148)) as well as hidden directories like `.git`. In case `FRAMEWORK_SEARCH_PATHS` contains a path that do not represent any real path - Carthage gonna ignore such paths.
- add check for `contentModificationDate` to skip copying of the outdated framework. Therefore in case product at the target URL has a higher modification date than a product at the source URL - Carthage gonna skip copying. This flag enabled only in case `--auto` flag is presented. Therefore for existing users it won't introduce any difference in behaviour.
- Update static frameworks filtering: instead of checking for `Static` folder component in path it is not checking for the mach-o file type to be of `MH_DYLIB` kind.

From the performance implications: `--auto` doesn't seem to add too much extra time. `skipIfOutdated` optimisation even speeds things up a little bit but not that much. 
However, using `--use-framework-search-paths` with large amount of directories increases this nearly twice.
For example on my production project there are 20 Carthage dependencies. It takes around 1.4-1.5 seconds for Carthage to run `copy-frameworks` with manual input files or with `--auto`. When I'm adding a `--use-framework-search-paths` option, time raises up to 2.9 seconds with 45 dynamic frameworks being found. 
Running Time Profiler doesn't give much sense because stack trace polluted by ReactiveCocoa calls :D
Any ideas on optimisation or it is not that critical?

blender:
@dimazen Were do we stand with this PR?

dimazen:
@blender it is ready for testing / merge. (Testing went fine for me)

blender:
I'm struggling to understand what `skipIfOutdated` does. What are we skipping?

dimazen:
@blender 
@bwhiteley wrote above:
> Copy frameworks phase checks timestamps to skip frameworks that are already up-to-date.

Basically, since copy-frameworks may start copying foreign frameworks (in case `--use-framework-search-paths` passed). To prevent unnecessary work it is best to skip copying them. 
Also, it improves performance since we're not doing

Source/CarthageKit/InputFilesInferrer.swift
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+/// Invokes otool -L for a given executable URL.
+///
+/// - Parameter executableURL: URL to a valid executable.
+/// - Returns: Array of the Shared Library ID that are linked against given executable (`Alamofire`, `Realm`, etc).
+/// System libraries and dylibs are omited.
+internal func linkedFrameworks(for executable: URL) -> Result<[String], CarthageError> {
+    return Task("/usr/bin/xcrun", arguments: ["otool", "-L", executable.path.spm_shellEscaped()])
```
blender: this line doesn't work if the path contains spaces

blender:
So after a bit of testing this seems to work pretty well, apart from the line that I have pointed out. Also, the CLI switch should reflect the name of the env variable "FRAMEWORK_SEARCH_PATHS". Make sure that also in the docs the name of the variable is correct.

However at this point I'm not even sure (backtracking on my own opinion) if we need the CLI switch `--use-framework-search-paths`. Let me explain.

I'm reading this comment:
```
// Carthage's default framework search path should always be presented. Under rare circumstances
// framework located at the non-default path can be linked against Carthage's framework.
// Since we're allowing user to specify only first-level frameworks, App might not link nested framework,
// therefore FRAMEWORKS_SEARCH_PATHS won't contain Carthage default search path.
// To prevent such failure we're appending default path at the end.
```

I'm not sure why we're adding "Carthage/Build/<platform>" to the search paths. If it's not there then probably none of the linked frameworks come from that path.

If there are frameworks pulled in from Carthage's build dir, then the FRAMEWORK_SEARCH_PATHS will contain it. So just always use FRAMEWORK_SEARCH_PATHS + manual override from input file/input file list.

@dimazen @bwhiteley  @typfel what do you guys think?

bwhiteley:
> If there are frameworks pulled in from Carthage's build dir, then the FRAMEWORK_SEARCH_PATHS will contain it. So just always use FRAMEWORK_SEARCH_PATHS + manual override from input file/input file list.

That works for me. This is the way our Carthage wrapper works that we've been using for a year or two. This way ensures that the same frameworks used to build/link are bundled with the app. If you deviate from FRAMEWORK_SEARCH_PATHS there is a risk of bundling a different version of a framework from what was used to compile/link.

bwhiteley:
I'm trying to build  dimazen:master with `make install` and I get a bunch of missing swift symbols. Am I doing something wrong?

blender:
@bwhiteley maybe you don't have xcode 10.2 selected?

README.md
```diff
@@ -109,6 +113,18 @@ Additionally, you'll need to copy debug symbols for debugging and crash reportin
 1. On your application targets’ _General_ settings tab, in the “Linked Frameworks and Libraries” section, drag and drop each framework you want to use from the [Carthage/Build][] folder on disk.
 1. On your application targets’ _Build Phases_ settings tab, click the _+_ icon and choose _New Run Script Phase_. Create a Run Script in which you specify your shell (ex: `/bin/sh`), add the following contents to the script area below the shell:
 
+###### Automatic
+
+    ```sh
+    /usr/local/bin/carthage copy-frameworks --auto
+    ```
+
+From this point Carthage will infer and copy all Carthage's frameworks that are linked against target. It also capable to copy frameworks that are nested. For example, you have linked to your app `SocialSDK-Swift` that links internally `SocialSDK-ObjC` which in turns uses utilitary dependency `SocialTools`. In this case you don't need nested dependencies it should be enough to link against your target only `SocialSDK-Swift`. Nested dependencies will be resolved and copied automatically to your app.
```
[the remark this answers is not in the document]
dimazen: "Nested" replaced by "transitive".

README.md
```diff
@@ -137,6 +153,15 @@ With the debug information copied into the built products directory, Xcode will
...
 
+###### Combining Automatic and Manual copying
+
+Note that you can combine both automatic and manual ways to copy frameworks, however manually specified frameworks always take precedence over automatically inferred. Therefore in case you have `SomeFramework.framework` located anywhere as well as `SomeFramework.framework` located at `./Carthage/Build/<platform>/`, Carthage will pick manually specified framework. This is useful when you're working with development frameworks and want to copy your version of the framework instead of default one. 
+Important to undestand, that Carthage won't resolve nested dependencies for your custom framework unless they either located at `./Carthage/Build/<platform>/` or specified manually in “Input Files".
+
+###### Automatic depencencies copying FRAMEWORK_SEARCH_PATHS
+
+If you're working on a development dependencies and would like to utilize `--auto` flag to automate copying of the build artifacts you also can be interested in using `--use-framework-search-paths` flag. This will intstruct Carthage to search for a linked dependcies and copy them using `FRAMEWORK_SEARCH_PATHS` environment variable.
```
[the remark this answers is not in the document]
dimazen: "Nested" replaced by "transitive".

README.md
```diff
@@ -137,6 +153,15 @@ With the debug information copied into the built products directory, Xcode will
 
 When archiving your application for submission to the App Store or TestFlight, Xcode will also copy these files into the dSYMs subdirectory of your application’s `.xcarchive` bundle.
 
+###### Combining Automatic and Manual copying
+
+Note that you can combine both automatic and manual ways to copy frameworks, however manually specified frameworks always take precedence over automatically inferred. Therefore in case you have `SomeFramework.framework` located anywhere as well as `SomeFramework.framework` located at `./Carthage/Build/<platform>/`, Carthage will pick manually specified framework. This is useful when you're working with development frameworks and want to copy your version of the framework instead of default one. 
+Important to undestand, that Carthage won't resolve nested dependencies for your custom framework unless they either located at `./Carthage/Build/<platform>/` or specified manually in “Input Files".
```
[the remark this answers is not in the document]
dimazen: "Nested" replaced by "transitive".

Source/CarthageKit/InputFilesInferrer.swift
```diff
@@ -0,0 +1,238 @@
...
+            throw error
+        }
+    }
+    
+    // MARK: - Utility
+    
+    static func allFrameworkSearchPaths(forProjectIn directory: URL, platform: Platform, frameworkSearchPaths: [URL]) -> [URL] {
+        // Carthage's default framework search path should always be presented. Under rare circumstances
+        // framework located at the non-default path can be linked against Carthage's framework.
+        // Since we're allowing user to specify only first-level frameworks, App might not link nested framework,
```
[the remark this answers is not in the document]
dimazen: "Nested" replaced by "transitive".

bwhiteley:
> @bwhiteley maybe you don't have xcode 10.2 selected?

I'm using Xcode 10.2.1. 
I was able to build with this change: 
```
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ test:
        swift test --skip-build
 
 installables:
-       swift build -c release -Xswiftc -static-stdlib
+       swift build -c release
 
 
```

dimazen:
@blender 
> However at this point I'm not even sure (backtracking on my own opinion) if we need the CLI switch --use-framework-search-paths. Let me explain.

It might be not clear from the description so let me provide you some sample use case where it can be useful:

First of all, automatic inference uses (recursively) `otool` to find all linked dependencies including transient. 
Once found and located on disk, Carthage copies them to the target directory.
There are few key points in this process: 1) we don't want to copy all of the linked frameworks that we have found (to not interfere CocoaPods, Embed Frameworks, etc). In the end `copy-frameworks` looks like a tool for copying only a Carthage-specific frameworks and by default I as a User wouldn't expect from Carthage to act as a `Embed Frameworks Build Phase` - this would be an unintended side-effect. 2) evaluating foreign frameworks (i.e. by enumerating all of the FRAMEWORK_SEARCH_PATHS values) increases time because we have to go thought all of the directories, find dynamic frameworks, etc. By limiting scope to the `./Carthage/Build/<platform>/` we provide same performance as we have now with Input Files, therefore users won't notice any difference in compile time. 

`--use-framework-search-paths` is done solely to enable copying of the development dependencies just as @bwhiteley does with his `carttool`. This option has a side-effect that leads to a copying of all of the linked and found frameworks, because there is 0 knowledge about whether it is a Carthage's framework or not. It didn't introduce any issues on my production project (but I don't have a CocoaPods to play around with) except increased (+2 seconds) build time. 
So my idea was to leave this switch as an extra option for those who is working on a development dependencies and don't want to add any workarounds (see https://github.com/Carthage/Carthage/pull/2731#issuecomment-473536822 for the usage scenario) and accepts increased build time as well as potential interference with other tools like CocoaPods. 

Long story short: `--use-framework-search-paths` is for developers that are working on the Development Dependencies (not sure whether that's a correct term, sorry). For the day-to-day usage simple `--auto` will provide same behaviour as with the Input Files.

dimazen:
@bwhiteley sorry, my branch seems to be outdated. I'm building using Xcode 10.1. 
@blender should I merge current master to my branch first? 
Fix for the space in the path will be addressed tomorrow :)

Source/CarthageKit/InputFilesInferrer.swift
werner77: A common way to avoid the process of escaping characters is to execute: /bin/sh -c "<full command>". 
In this case:

`Task("/bin/sh", arguments: ["-c", "/usr/bin/xcrun otool -L \"\(executable.path)\""]`

blender:
@dimazen I see. About that, can we express an "exclusion" list as well? Like 
`--use-framework-search-paths --exclude name,name,name` 

Maybe this can be added at another point in time if needed? Opinions? @bwhiteley

dimazen:
@blender you mean to exclude specific folders? I would rely on the @bwhiteley opinion because of his experience with a similar workflow. 

Implementation of the `exclude` can be addressed right after this PR or as part of this PR - I'm totally fine with both options and will implement on the weekends if needed.

Source/CarthageKit/InputFilesInferrer.swift
dimazen: Fixed. Thanks for your suggestion, @werner77!

bwhiteley:
I tried this fork with some projects. It works for some but fails on one because it tries to process a framework that has already been taken care of in Embed Frameworks (it's not a carthage framework, and apparently `carthage copy-frameworks` is not happy with it: "lipo: -remove's specified would result in an empty fat file"). Our current solution works because it checks timestamps and skips frameworks that are already up-to-date. It would also be possible to fix this with an exclusion list. 

This particular project has 22 frameworks copied in the Embed phase. Most of these are local framework targets in the same project. All of these can/should be skipped, ideally automatically with an up-to-date check. Adding 22 items to an exclusion list would be less desirable, particularly as this PR is trying to avoid the need to list specific files.

blender:
@dimazen let's skip the exclude for now.

dimazen:
@bwhiteley thanks for the feedback. Issue you've describes can be handled by putting check earlier in code. Basically we have to check whether framework needs an update prior to any lipo and other things and not in the copy function as I did it. 

Will update code accordingly.

dimazen:
@bwhiteley good day! May I ask you to test updated version on your project? 
Updates: 
- "is outdated" check perform prior to any framework processing (stripping, etc).
- updated logging. passing `--verbose` now correctly prints both framework that is going to be copied and the one that is going to be skipped by Carthage.

bwhiteley:
@dimazen This looks good. The problem from before is fixed. 

One nitpick: In the verbose output, the order of source/destination files is opposite if a framework is skipped vs. if it is copied, which makes the output a little harder to read.

bwhiteley:
I should point out, that I'm unable to do a `carthage build` with this fork. 

How to reproduce: 
Use this Cartfile: 
```
github "Molbie/Outlaw" ~> 3.0
```

Then, 
```
carthage bootstrap --no-build
carthage build --cache-builds --platform iOS,watchOS --configuration Debug Outlaw
```

Observed results: 
Seems like a deadlock. 

Carthage 0.33.0 from homebrew doesn't have this problem.

dimazen:
@bwhiteley thanks for your feedback! You are really helping me a lot! 

Regarding your test repo: 
It seems to fall into a deadlock in my version even without any changes (i.e. on clean master that I've forked initially).

Finally merged latest master from Carthage repo into my fork to not hassle with 10.1 / 10.2 versions and various build issues that you've been experiencing for a while.

Now everything is ok.

Also here is a output for the latest verbose logging (want to know your thoughts @bwhiteley and @blender):

```
Skipping Bolts.framework copying because the framework at target path has equal or higher modification date than the framework at source path:
source: "/Users/.../Carthage/Build/iOS/Bolts.framework"
target: "/Users/.../Build/Products/.../Frameworks/Bolts.framework"

Skipping CHIPageControl.framework copying because the framework at target path has equal or higher modification date than the framework at source path:
source: "/Users/.../Carthage/Build/iOS/CHIPageControl.framework"
target: "/Users/.../Build/Products/.../Frameworks/CHIPageControl.framework"

Copying "FacebookCore.framework" automatically:
source: "/Users/.../Carthage/Build/iOS/FacebookCore.framework"
target: "/Users/.../Build/Products/.../Frameworks/FacebookCore.framework"
```

dimazen:
@blender any updates on that? :)

bwhiteley:
I like the new verbose logging.

blender:
@dimazen I'm overloaded and I cannot follow atm. I will do a quick pass now.

@bwhiteley can you review instead? I cannot add you to the reviewers for some reason. Please write in the comments

README.md
```diff
@@ -55,9 +55,13 @@ Carthage builds your dependencies and provides you with binary frameworks, but y
 1. On your application targets’ _Build Phases_ settings tab, click the _+_ icon and choose _New Run Script Phase_. Create a Run Script in which you specify your shell (ex: `/bin/sh`), add the following contents to the script area below the shell:
 
     ```sh
-    /usr/local/bin/carthage copy-frameworks
```
blender: @dimazen  let's but this as *experimental* and leave the documentation for both

I'm suggesting this in case people start depending on some behavior we have finalized yet.

Source/CarthageKit/FoundationExtensions.swift
```diff
@@ -0,0 +1,48 @@
...
+        }
+        
+        let options: FileManager.DirectoryEnumerationOptions = [.skipsHiddenFiles]
+        let keys: [URLResourceKey] = [.isDirectoryKey]
+        guard
+            directoryURL.isFileURL,
+            isDirectory(at: directoryURL),
+            let enumerator = self.enumerator(at: directoryURL, includingPropertiesForKeys: keys, options: options)
+        else
+        {
```
blender: Not sure about the bracket style here. (I know, great comment ;D )

Source/carthage/CopyFrameworks.swift
```diff
@@ -3,15 +3,29 @@ import Commandant
...
 /// Type that encapsulates the configuration and evaluation of the `copy-frameworks` subcommand.
 public struct CopyFrameworksCommand: CommandProtocol {
+    public struct Options: OptionsProtocol {
+        public let automatic: Bool
+        public let useFrameworkSearchPaths: Bool
+        public let isVerbose: Bool
+        
+        public static func evaluate(_ mode: CommandMode) -> Result<Options, CommandantError<CarthageError>> {
+            return curry(self.init)
+                <*> mode <| Option(key: "auto", defaultValue: false, usage: "infers and copies linked frameworks automatically")
```
blender: add *experimental* notice here.

I'm suggesting this in case people start depending on some behavior we have finalized yet.

tmspzz (approved):

Source/CarthageKit/FoundationExtensions.swift
dimazen: Fixed to match existing style

Source/carthage/CopyFrameworks.swift
dimazen: Added `(experimental)` note.

README.md
dimazen: Update documentation to match latest master. Appended `auto` section below. Also reordered auto section for `If you're building for iOS, tvOS, or watchOS` section to be lower than manual as well as prepend it with the "Experimental" prefix.

blender:
@typfel maybe you might want to try this out as well?

blender:
@typfel @jdhealy  any feedback?

typfel:
So I finally took the time to test this out on https://github.com/wireapp/wire-ios

I ran the following without any input/output files defined:
```
carthage copy-frameworks --auto --verbose
```

It failed run because it didn't copy the `ziphy.framework`. After renaming the framework in the Carthage build directory to `Ziphy.framework ` it picked it up and it all worked. I guess something got confused by the fact the the scheme and the framework file had different casing, but maybe that's expected and we should just fix this inconsistency on our side.

blender:
@typfel if I understand correctly, that's also the care for Alamofire. The schemes are called "Alamofire iOS" but the product is just "Alamofire.framework"

typfel:
@blender and for Alamofire it works or what was your question?

bwhiteley:
I don't think this is related to scheme names, but maybe target names. I think the issue here is that APFS is case-insensitive, case-preserving by default, while this new framework resolver (not to be confused with the dependency resolver) in Carthage that maps otool output to frameworks found on the filesystem is case-sensitive. in the wireapp example, It looks like the app is linked against `Ziphy`. The wireapp build would likely fail on a Mac where APFS had been set up to be case-sensitive (not common).

bwhiteley:
From the Wire-iOS project.pbxproj: 
```
		5EF4A6E720F497E300A9EF22 /* Ziphy.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = Ziphy.framework; path = Carthage/Build/iOS/Ziphy.framework; sourceTree = "<group>"; };

```

Tests/CarthageKitTests/InputFilesInferrerSpec.swift
```diff
@@ -0,0 +1,488 @@
...
+                                default:
+                                    return .success([])
+                                }
+                            }
+                        )
+                        sut.executableResolver = executableResolver
+                        sut.builtFrameworkFilter = builtFrameworkFilter
+                    }
+
+                    it("should resolve dependncies at a enumerator's path") {
```
blender: typo `dependncies`

Tests/CarthageKitTests/InputFilesInferrerSpec.swift
```diff
@@ -0,0 +1,488 @@
...
+                                default:
+                                    return .success([])
+                                }
+                        }
+                        )
+                        sut.executableResolver = executableResolver
+                        sut.builtFrameworkFilter = builtFrameworkFilter
+                    }
+                    
+                    it("should resolve dependncies at a first enumerator's path") {
```
blender: typo `dependncies`

Tests/CarthageKitTests/InputFilesInferrerSpec.swift
```diff
@@ -0,0 +1,488 @@
...
+                                default:
+                                    return .success([])
+                                }
+                            }
+                        )
+                        sut.executableResolver = executableResolver
+                        sut.builtFrameworkFilter = builtFrameworkFilter
+                    }
+
+                    it("should resolve dependncies at user input file's path") {
```
blender: typo `dependncies`

Tests/CarthageKitTests/InputFilesInferrerSpec.swift
```diff
@@ -0,0 +1,488 @@
...
+                }
+            }
+        }
+        
+        describe("paths resolving") {
+            let projectDirectory = URL(fileURLWithPath: "/Project/Directory")
+            let platform: Platform = .iOS
+            
+            describe("default framework search path") {
+                it("should combine project ULR and platform-related relative path") {
```
blender: ULR -> URL

tmspzz (changes requested):

README.md
```diff
@@ -137,6 +148,25 @@ With the debug information copied into the built products directory, Xcode will
...
+
+###### Combining Automatic and Manual copying
+
+Note that you can combine both automatic and manual ways to copy frameworks, however manually specified frameworks always take precedence over automatically inferred. Therefore in case you have `SomeFramework.framework` located anywhere as well as `SomeFramework.framework` located at `./Carthage/Build/<platform>/`, Carthage will pick manually specified framework. This is useful when you're working with development frameworks and want to copy your version of the framework instead of default one. 
+Important to undestand, that Carthage won't resolve transient dependencies for your custom framework unless they either located at `./Carthage/Build/<platform>/` or specified manually in “Input Files".
+
+###### Automatic depencencies copying FRAMEWORK_SEARCH_PATHS
+
+If you're working on a development dependencies and would like to utilize `--auto` flag to automate copying of the build artifacts you also can be interested in using `--use-framework-search-paths` flag. This will instruct Carthage to search for a linked dependcies and copy them using `FRAMEWORK_SEARCH_PATHS` environment variable.
```
blender: typo `dependncies`

blender:
@dimazen last effort. Please fix the typos and then let's merge it.

Tests/CarthageKitTests/InputFilesInferrerSpec.swift
dimazen: Fixed

Tests/CarthageKitTests/InputFilesInferrerSpec.swift
dimazen: Fixed

Tests/CarthageKitTests/InputFilesInferrerSpec.swift
dimazen: Fixed

Tests/CarthageKitTests/InputFilesInferrerSpec.swift
dimazen: Fixed

README.md
dimazen: Fixed

dimazen:
@blender  Should we also fix case-sensitive lookup to handle issue described by @typfel?

blender:
I'm a little confused. 

Correct me if I'm wrong:

Wire's executable was probably linked against `Ziphy.framework` (and that's exactly what the product name in `wire-ios-zipy` is. I checked and built, this is the output: `/Users/Tommaso/Library/Developer/Xcode/DerivedData/Ziphy-fehvdjnzdgvnntcgvxedbmecnqow/Build/Products/Debug-iphonesimulator/Ziphy.framework/Ziphy` ) but for some reason @typfel has on disk "ziphy.framework".

So if @typfel had a case sensitive file system, it should not have worked. 
If he had a case-insensitive filesystem, this should have worked.

Since it didn't work, I can only make out he has a case-sensitive file system. That in combination with the name mismatch on Ziphy-ziphy, would have led even the regular copy-framework phase to fail. 

Are we interfering with the file system lookup of the path on disk somewhere? I didn't find any evidence. The framework map comes directly form otool, which outputs a path thus preserving whatever file system preference there was at the time of linking and leaving the re-interpretation of the path to be handled by the file system where `dyld`will be executed.

If I'm right, I don't think there should be any fix.

typfel:
> Wire's executable was probably linked against Ziphy.framework (and that's exactly what the product name in wire-ios-zipy is. I checked and built, this is the output: /Users/Tommaso/Library/Developer/Xcode/DerivedData/Ziphy-fehvdjnzdgvnntcgvxedbmecnqow/Build/Products/Debug-iphonesimulator/Ziphy.framework/Ziphy ) but for some reason @typfel has on disk "ziphy.framework".

This is probably due to some CI script which incorrectly upload the file based on the project name.

> So if @typfel had a case sensitive file system, it should not have worked.
If he had a case-insensitive filesystem, this should have worked.

My file system is the default case insensitive, APFS (Encrypted)

> Since it didn't work, I can only make out he has a case-sensitive file system. That in combination with the name mismatch on Ziphy-ziphy, would have led even the regular copy-framework phase to fail.

This worked because in the regular copy-framework phase the output file was listed as `Ziphy.framework`.

One last observation:

When running `copy-frameworks --auto --verbose` it doesn't copy the `ziphy.framework` unless I rename to  the correct name `Ziphy.framework`. I also tried re-adding the framework as `ziphy.framework` to the project but the `copy-frameworks` still didn't pick it up.

But maybe this behaviour is what we want since the iOS file system is case sensitive?

blender:
@typfel was your error a compile time or a "failed to load library" from dyld?

typfel:
> @typfel was your error a compile time or a "failed to load library" from dyld?

"failed to load library" since it's not copied to the framework directory.

blender:
could you please `otool -L` Wire's executable and paste the output here?

typfel:
```
$ otool -L Wire 
Wire:
	@rpath/Ziphy.framework/Ziphy (compatibility version 13.0.0, current version 1.0.0)
	@rpath/WireDataModel.framework/WireDataModel (compatibility version 185.0.0, current version 185.0.0)
	@rpath/WireCanvas.framework/WireCanvas (compatibility version 10.0.0, current version 10.0.0)
	/System/Library/Frameworks/LocalAuthentication.framework/LocalAuthentication (compatibility version 1.0.0, current version 425.250.11)
	@rpath/FLAnimatedImage.framework/FLAnimatedImage (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Contacts.framework/Contacts (compatibility version 0.0.0, current version 0.0.0)
	@rpath/Down.framework/Down (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Intents.framework/Intents (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/CallKit.framework/CallKit (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Photos.framework/Photos (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/AVKit.framework/AVKit (compatibility version 1.0.0, current version 1.0.0)
	@rpath/WireRequestStrategy.framework/WireRequestStrategy (compatibility version 163.0.0, current version 163.0.0)
	@rpath/DifferenceKit.framework/DifferenceKit (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/VideoToolbox.framework/VideoToolbox (compatibility version 1.0.0, current version 1.0.0)
	@rpath/libcmark.framework/libcmark (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/UIKit.framework/UIKit (compatibility version 1.0.0, current version 61000.0.0)
	@rpath/WireShareEngine.framework/WireShareEngine (compatibility version 151.0.0, current version 151.0.0)
	@rpath/WireCommonComponents.framework/WireCommonComponents (compatibility version 0.0.0, current version 0.0.0)
	@rpath/HockeySDK.framework/HockeySDK (compatibility version 1.0.0, current version 1.0.0)
	@rpath/FormatterKit.framework/FormatterKit (compatibility version 1.0.0, current version 1.0.0)
	@rpath/avs.framework/avs (compatibility version 0.0.0, current version 0.0.0)
	@rpath/HTMLString.framework/HTMLString (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Foundation.framework/Foundation (compatibility version 300.0.0, current version 1570.15.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.250.1)
	/System/Library/Frameworks/AVFoundation.framework/AVFoundation (compatibility version 1.0.0, current version 2.0.0)
	/System/Library/Frameworks/CoreData.framework/CoreData (compatibility version 1.0.0, current version 866.5.0)
	/System/Library/Frameworks/CoreFoundation.framework/CoreFoundation (compatibility version 150.0.0, current version 1570.15.0)
	/System/Library/Frameworks/CoreGraphics.framework/CoreGraphics (compatibility version 64.0.0, current version 1251.12.0)
	/System/Library/Frameworks/CoreImage.framework/CoreImage (compatibility version 1.0.0, current version 5.0.0)
	/System/Library/Frameworks/CoreLocation.framework/CoreLocation (compatibility version 1.0.0, current version 2245.12.30)
	/System/Library/Frameworks/CoreMedia.framework/CoreMedia (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/CoreMotion.framework/CoreMotion (compatibility version 1.0.0, current version 2245.12.30)
	/System/Library/Frameworks/CoreTelephony.framework/CoreTelephony (compatibility version 1.0.0, current version 0.0.0)
	/System/Library/Frameworks/CoreText.framework/CoreText (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/ImageIO.framework/ImageIO (compatibility version 1.0.0, current version 0.0.0)
	/System/Library/Frameworks/MapKit.framework/MapKit (compatibility version 1.0.0, current version 14.0.0)
	/System/Library/Frameworks/MediaPlayer.framework/MediaPlayer (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/MessageUI.framework/MessageUI (compatibility version 1.0.0, current version 3445.104.7)
	/System/Library/Frameworks/MobileCoreServices.framework/MobileCoreServices (compatibility version 1.0.0, current version 943.1.0)
	/System/Library/Frameworks/PassKit.framework/PassKit (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/QuartzCore.framework/QuartzCore (compatibility version 1.2.0, current version 1.11.0)
	/System/Library/Frameworks/SafariServices.framework/SafariServices (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/UserNotifications.framework/UserNotifications (compatibility version 1.0.0, current version 1.0.0)
	@rpath/Cartography.framework/Cartography (compatibility version 1.0.0, current version 1.0.0)
	@rpath/WireCryptobox.framework/WireCryptobox (compatibility version 19.0.0, current version 19.0.0)
	@rpath/WireLinkPreview.framework/WireLinkPreview (compatibility version 23.0.0, current version 23.0.0)
	@rpath/WireProtos.framework/WireProtos (compatibility version 20.0.0, current version 20.2.0)
	@rpath/WireSyncEngine.framework/WireSyncEngine (compatibility version 267.0.0, current version 267.0.0)
	@rpath/WireSystem.framework/WireSystem (compatibility version 28.0.0, current version 28.0.2)
	@rpath/WireTransport.framework/WireTransport (compatibility version 60.0.0, current version 60.0.0)
	@rpath/WireUtilities.framework/WireUtilities (compatibility version 33.0.0, current version 33.0.0)
	@rpath/ZipArchive.framework/ZipArchive (compatibility version 1.0.0, current version 1.0.0)
	@rpath/libswiftAVFoundation.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftAssetsLibrary.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftCallKit.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftContacts.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftCore.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftCoreAudio.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftCoreData.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftCoreFoundation.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftCoreGraphics.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftCoreImage.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftCoreLocation.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftCoreMedia.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftDarwin.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftDispatch.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftFoundation.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftIntents.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftMapKit.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftMediaPlayer.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftMetal.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftObjectiveC.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftPhotos.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftQuartzCore.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftSwiftOnoneSupport.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftUIKit.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftos.dylib (compatibility version 1.0.0, current version 1001.0.82)
	@rpath/libswiftsimd.dylib (compatibility version 1.0.0, current version 1001.0.82)
```

bwhiteley:
> > @typfel was your error a compile time or a "failed to load library" from dyld?
> 
> "failed to load library" since it's not copied to the framework directory.

OK, this is a problem. The current implementation in this PR silently ignores missing frameworks. I verified by renaming a framework in Carthage/Build/iOS by changing its case. The build succeeded but the framework was not bundled. `carthage copy-frameworks --auto --verbose` exited with `0` and shows no complaints about the missing framework.

IMO this needs to be fixed so that a missing library causes `carthage copy-frameworks --auto` to exit with a non-zero exit status and print what went wrong (with or without `--verbose`).

bwhiteley:
@typfel I'm curious... in the situation where your app links against `Ziphy` but `ziphy.framework` is what is bundled, does the app run on a device? I would expect it to run in the iOS simulator but fail to launch on device because APFS is case-insensitive on macOS (by default) but is case-sensitive on iOS.

dimazen:
@bwhiteley 
> OK, this is a problem. The current implementation in this PR silently ignores missing frameworks. I verified by renaming a framework in Carthage/Build/iOS by changing its case. The build succeeded but the framework was not bundled. carthage copy-frameworks --auto --verbose exited with 0 and shows no complaints about the missing framework. This moves what should be a build-time error to run-time.

Let me explain current implementation and it's limitations in the give context: 
it recursively finds all dependencies via `otool` and based on the dependencies list in the default mode (without `--use-framework-search-paths`) performs lookup of the found dependencies in the `Carthage/Build/<platform>/` directory. In case some dependency is missing - we're skipping it because there is zero knowledge about whether this is a Carthage's dependency or a regular third-party dependency. Raising an errors will cause false-positive errors about missing dependencies that are managed by "Embed Frameworks" phase or by another dependency manager. 

Workaround is to ask user to place `carthage copy-frameworks` Build Phase in the end of the list, but this is not reliable.

dimazen:
Just to add: in order to verify that all of the dependencies are in place I'm running a [carthage-verify](https://github.com/Carthage/workflows/blob/master/carthage-verify) script. Would be good to make it a built-in tool. Anyway, verification is needed.

bwhiteley:
Fair enough. In that case I vote to merge.

blender:
```
@rpath/Ziphy.framework/Ziphy (compatibility version 13.0.0, current version 1.0.0)
```

I think we can all agree that the actual casing on the file system should be `Ziphy.framework` and this PR is good to merge as is.

@dimazen @typfel @bwhiteley  @jdhealy

dimazen:
Any update on this?

tmspzz:
@dimazen  yep @jdhealy wanted to review

dimazen:
Any updates about review? :|

dimazen:
Sooo, is there anything else left to add to merge this PR finally? :)

Issue #2733, opened 2019-03-14
carthage hangs when git prompt has no where to go


potrebic:
* carthage install method: [ ] .pkg, [ X ] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.32.0
* `xcodebuild -version`:  Xcode 10.1 Build version 10B61
* Are you using `--no-build`? No
* Are you using `--no-use-binaries`? No
* Are you using `--use-submodules`?  Yes
* Are you using `--cache-builds`? No
* Are you using `--new-resolver`? No

**Cartfile**
```
-- SNIP --
git "git@gitenterprise.inside-box.net:Team/some-repo.git" "master"
-- SNIP --
```

**Carthage Output**
```
<Carthage hangs>
```

**Actual outcome**
We use GitHub Enterprise at Box. For a convoluted reason my known_hosts file got into a stale state. And this was causing carthage to silently hang on a git prompt of the the user (me)

Once I figured out the issue in ~/ssh/known_hosts file it was easy to see that git-prompt
```
$ git fetch     # when in the local repo that Carthage was attempting to manage
Warning: the RSA host key for 'gitenterprise.inside-box.net' differs from ' [ -- SNIP --]
Are you sure you want to continue connecting (yes/no)? no
```
Once I fixed the known_hosts file carthage worked fine again.

**Expected outcome**
Could carthage, given that these prompts are hidden, instead gracefully fail. I investigated solutions like `GIT_TERMINAL_PROMPT=0`, but that did NOT this prompt. So some other solution would be needed.

I looked at similar Issues like Issue 1617 and 1705. Those were looking for solutions to avoid prompts. Here I'm saying... if git never the less issues a prompt given that there's no way for a user to respond... carthage will hang. Wouldn't it be better to gracefully fail, giving the user a better picture as to the issue at hand. Thanks!

ikesyo:
- `GIT_TERMINAL_PROMPT=0` is already specified: https://github.com/Carthage/Carthage/blob/0c4c64916f9505fd7fbcded6001e83915c572ecf/Source/CarthageKit/Git.swift#L52
- https://github.com/Carthage/Carthage/issues/1705#issuecomment-435270220 will be a possible solution.

jlawton:
Yeah. It might have been unclear, but the above was saying that `GIT_TERMINAL_PROMPT` doesn't stop the prompt for `known_hosts`. I don't know of an option to prevent it, either.

[...]

Issue #1705, opened 2017-01-11
Carthage stuck on updating if SSH agent isn't active


iSevenDays:
* `carthage version`: 
0.18.1
* `xcodebuild -version`: 
Xcode 8.2
Build version 8C38
* Are you using `--no-build`? NO
* Are you using `--no-use-binaries`? NO
* Are you using `--use-submodules`? NO

**Cartfile**
```
github "facebook/pop"
```

**Carthage Output**
```
sevens:project seven$ carthage update --verbose --platform iOS
*** Cloning pop
^C

**STUCK**
```

1. Restart MAC
2. run carthage update --verbose --platform iOS
3. See nothing happens
4. execute ssh-add -k ~/.ssh/id_rsa
5. Type password for id_rsa
6. Run carthage update --verbose --platform iOS
7. Executed successfully

15458434:
Same here.

mdiep:
How can Carthage prevent this? Is there a flag we can add somewhere to not prompt?

jliu218-ford:
yes.
https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/

ikesyo:
Closing this due to inactivity.

ikesyo:
Looks like we can replicate this: https://github.com/apple/swift-package-manager/pull/1839

```
env["GIT_SSH_COMMAND"] = "ssh -oBatchMode=yes"
```

hakanyildizay:
Same here too. 
Thank you so much @iSevenDays

ikesyo:
Addressed by #2734.

PR #2734, opened 2019-03-15
Use ssh batch mode so git doesn't attempt to prompt the user and hang.

jlawton:
Fix for #1705 and #2733.


jlawton:
I don't know if we'd want to make this conditional on `GIT_SSH_COMMAND` not already being set in the environment.

mdiep:
Looks good to me. @ikesyo ?

mdiep:
Proceeding with this based on [the usage in SwiftPM](https://github.com/apple/swift-package-manager/blob/79f527df94b689058078055a012121bd41ed7bf3/Sources/SPMUtility/Git.swift#L88) and [the documentation](https://linux.die.net/man/5/ssh_config) they link to.

PR #2748, opened 2019-03-28
Fix merging Swift header files for Xcode 10.2

VictorNouvellet:
This is meant to fix merging Swift header files for Xcode 10.2 by adding missing include directive for `TARGET_OS_SIMULATOR` definition.

- Xcode doesn't set `TARGET_OS_SIMULATOR` value by default and shows `undeclared identifier` errors at compilation for the symbols in generated Swift header file.
- Including `TargetConditionals.h` to the header file solves the issue.

Generated Swift header file should now look like this:
```
#include <TargetConditionals.h>
#if TARGET_OS_SIMULATOR
<simulator Swift.h>
#else
<device Swift.h>
#endif
```
Coauthored with @buranmert

Related: https://github.com/Carthage/Carthage/pull/2723

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -364,7 +364,7 @@ private func mergeSwiftHeaderFiles(_ simulatorExecutableURL: URL,
 	
 	let conditionalPrefix = "#if TARGET_OS_SIMULATOR\n"
 	let conditionalElse = "\n#else\n"
-	let conditionalSuffix = "\n#endif"
+	let conditionalSuffix = "\n#endif\n"
 	
 	let conditionalPrefixContents = conditionalPrefix.data(using: .utf8)!
 	let conditionalElseContents = conditionalElse.data(using: .utf8)!
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -362,10 +362,16 @@ private func mergeSwiftHeaderFiles(_ simulatorExecutableURL: URL,
 	precondition(deviceExecutableURL.isFileURL)
 	precondition(executableOutputURL.isFileURL)
 	
+    let includeTargetConditionals = """
+                                    #ifndef TARGET_OS_SIMULATOR
+                                    #include <TargetConditionals.h>
+                                    #endif\n
+                                    """
 	let conditionalPrefix = "#if TARGET_OS_SIMULATOR\n"
 	let conditionalElse = "\n#else\n"
 	let conditionalSuffix = "\n#endif\n"
 	
+	let includeTargetConditionalsContents = includeTargetConditionals.data(using: .utf8)!
 	let conditionalPrefixContents = conditionalPrefix.data(using: .utf8)!
 	let conditionalElseContents = conditionalElse.data(using: .utf8)!
 	let conditionalSuffixContents = conditionalSuffix.data(using: .utf8)!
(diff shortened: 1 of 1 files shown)


VictorNouvellet:
I am getting the same weird tests failure @dimazen described [here](https://github.com/Carthage/Carthage/pull/2731#issuecomment-476393892). If I run `make test` locally on the `master` branch, I get the same error.
@mdiep Do you have any idea what I can do about it? Should we just ignore it?

buranmert:
TL;DR: CI looks fine but we need to update [TestFramework1,2,3](https://github.com/Carthage/Carthage/blob/master/.travis.yml#L15) to newer `Swift` versions when we are gonna start using `Swift 5`

@mdiep @jdhealy 
do you have a plan to migrate to `Xcode 10.2/Swift 5` soon? if so, i will create an issue about updating `TestFramework1,2,3`

@VictorNouvellet 
1. you shouldn't run tests from Xcode as it runs the tests from `DerivedData` in which case `bundle` has a different path than expected. `Bundle` should have the path [here](https://github.com/Carthage/Carthage/blob/master/Makefile#L42). it works _almost_ fine when you run `make test` from `Terminal`.
2. your problem with running `make test` is another one: it fails because you probably are using `Xcode 10.2/Swift 5` and there are `TestFrameworks` in test target's resources which are `Swift 3` and they are not supported, so they fail at compilation and this leads your `ProjectSpec`s to fail. what you should do is using `Xcode 10.1` then it will work fine. [you can see here CI is still using it](https://github.com/Carthage/Carthage/blob/master/.travis.yml#L15)
3. i didn't really understand why travis failed at your commit (`RemoteVersionSpec` failed 🤔), it seems green now although it's failed to send back build status to github somehow

DanielAsher:
I have been unable to build my project that has a dependency on [realm-cocoa](https://github.com/realm/realm-cocoahttps://github.com/realm/realm-cocoa) for the simulator using carthage as I get compiler errors of the following kind:
`Error: Cannot find interface declaration of 'RealmSwiftObject', superclass of 'MyRealmDataObject'`

Compiling for the Generic iOS Device does work.

I'm including this as comment as it may help others who may receive this error and not understand that this is a carthage bug. A note on README.md that Xcode 10.2 is not support could save hundreds of developer hours.

VictorNouvellet:
You are right, it seems that Travis build failed on `RemoteVersionSpec.swift:21` test and not on `ArchiveSpec.swift:10`.

> If I run make test locally on the master branch, I get the same error

It was using `Xcode 10.2/Swift 5`, it is working fine with `Xcode 10.1/Swift 4.2` locally.

Nevertheless, it looks like none of the changes made here are related to `RemoteVersionSpec` test failure. This PR does not require Carthage to be migrated to `Xcode 10.2/Swift 5`.

VictorNouvellet:
I squashed @buranmert commit and mine. And now tests pass 💯

tonnihyldgaard:
Hi @VictorNouvellet love the work that you guys are doing! I have a few projects that are affected by this, and was wondering when you expect to release this fix? 🥇 👍

buranmert:
Thanks @tonnihyldgaard!

Can you please do us a favor and test this fix on your affected projects and let us know?

As discussed [here](https://github.com/Carthage/Carthage/issues/2718#issuecomment-478024001), we are still trying to get more insight on the issue although we've already made sure that this fix works for us.

tonnihyldgaard:
@buranmert Will do! Cloning the repo now 👍

tonnihyldgaard:
Hey guys! @buranmert @mdiep @VictorNouvellet @jdhealy 
it works! 🎉

3a4oT:
Can we expect release anytime soon?

mdiep:
We're working on a release, but Swift 5's ABI stability has changed how some of the command-line tool building works, so we have to do a little extra work for this release.

Urri25:
54321

PR #2761, opened 2019-04-07
Avoid warnings on Xcode 10.2 in CarthageKit

giginet:
Use `func hash(into:)` to avoid warnings on Xcode 10.2.

This PR make all warnings from CarthageKit were suppressed.


Source/CarthageKit/Xcode.swift
```diff
@@ -361,7 +361,7 @@ private func mergeSwiftHeaderFiles(_ simulatorExecutableURL: URL,
 	precondition(simulatorExecutableURL.isFileURL)
 	precondition(deviceExecutableURL.isFileURL)
 	precondition(executableOutputURL.isFileURL)
-	
+
```
giginet: Just apply `swiftlint autocorrect`

jdhealy:
Thanks, @giginet ✨

PR #2770, opened 2019-04-19
Installing module manually only

alpascual:
#### Background
When installing some repo from Github, users normally get 3 options, cocoapods, carthage and manually. The last one, the developers just explains to copy the `Source` directory to their project and compile it. Most of the github repos include instructions just to copy the source code instead of using cocoapods or carthage. Most developers like the source code to modify without recompile. Helps in debugging, learning and without project modifications.

#### Context
A way to make carthage simpler and easier to download the code you need. 

Example usage:
```
carthage update --manually
```

#### Output
carthage won't compile the produced projects to a framework, instead will copy the Source directory to the project in a new directory called `Carthage-{PROJECTNAME}` where the user can compile inside their project without having to reference any framework. Of course, those are for source code without collisions in classes.

#### Assumptions
- The developer gave instructions the source code can be installed manually by a simple copy
- The developer added all the source code into a directory called `Sources`, `Source`, `Manual`

#### Improvements
- Allow the developer to create a `.manually` file in their repo when the default directory is different than the default name. But just trying to keep it very simple at this time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,11 @@ With the debug information copied into the built products directory, Xcode will 
 
 When archiving your application for submission to the App Store or TestFlight, Xcode will also copy these files into the dSYMs subdirectory of your application’s `.xcarchive` bundle.
 
+##### If you're building for iOS, tvOS, or watchOS and the there is an option for manual installation
+1. Like above. Create a [Cartfile][] that lists the frameworks you’d like to use in your project.
+1. Run `carthage update --manually`. This will fetch dependencies into a [Carthage/Checkouts][] folder, but the projects won't be build, instead, the source code will be copied to your project directory on subdirectories called `Carthage-PROJECTNAME`.
+1. Just add those directories into your project using `Add Files`.
+
 ##### For both platforms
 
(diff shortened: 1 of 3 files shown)


mdiep:
Why not use the source directly from `Carthage/Checkouts/{dependency}`?

alpascual:
> Why not use the source directly from `Carthage/Checkouts/{dependency}`?

That's how I normally used the code, but then I started moving them to their own folder to keep the `Carthage/Checkouts` clean in case I deleted `plist` and `md` file that I didn't need. Also easier to add them and manage them with the new git. Eventually if I need to distribute, I compile the framework and add it.
Some developers clean their `Carthage` directories and update it frequently, the idea of manually installing is to make your modifications to the code, a copy that cannot be overridden unless  is by you, its always a great idea. 
Trying to keep the workflow of the `Manual` install found on repos around github.

mdiep:
Sorry, but I don't think this makes sense to include in Carthage.

I would recommend maintaining these changes on a fork instead.

PR #2776, opened 2019-04-29
Copy dSYMs that are incomplete

hlineholm:
When downloading a binary archive that also contains a dSYM, Carthage doesn't copy that dSYM to _Carthage/Build_ unless it covers all of the architectures in the binary. So if it covers 2/4 architectures, it isn't copied.

With this PR, all dSYMs covering at least one of the architectures in the binary, are copied.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -1383,8 +1383,9 @@ internal func dSYMsInDirectory(_ directoryURL: URL) -> SignalProducer<URL, Carth
 	return filesInDirectory(directoryURL, "com.apple.xcode.dsym")
 }
 
-/// Sends the URL of the dSYM whose UUIDs match those of the given framework, or
-/// errors if there was an error parsing a dSYM contained within the directory.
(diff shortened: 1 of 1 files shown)


blender:
Please add a comment to that line.

LGTM

Source/CarthageKit/Project.swift
```diff
@@ -1392,7 +1392,7 @@ private func dSYMForFramework(_ frameworkURL: URL, inDirectoryURL directoryURL:
 				.flatMap(.merge) { dSYMURL in
 					return UUIDsForDSYM(dSYMURL)
 						.filter { (dSYMUUIDs: Set<UUID>) in
-							return dSYMUUIDs == frameworkUUIDs
+                            return !dSYMUUIDs.isDisjoint(with: frameworkUUIDs)
```
ikesyo: Looks like the indentation is wrong (tabs vs spaces).

hlineholm: Should be fixed now!

mdiep:
Thanks for the PR!

hlineholm:
np

PR #2778, opened 2019-04-30
Readme: "If you're building for OS X"

gubikmic:
Shouldn't the command there be `carthage update --platform macOS`?
Otherwise it will build the frameworks for other platforms as well.

Also, changes 'OS X' to macOS


blender:
Thanks! Much appreciated!

PR #2765, opened 2019-04-11
Reject archives containing duplicated framework names

blender:
See #2764 in which the archive contains "TealiumIOS.framework" multiple times.

PR #2784, opened 2019-05-13
[SwiftLint] Add a custom rule to detect possible Carthage typos

ikesyo:
https://github.com/Carthage/Carthage/pull/2765#issuecomment-491802610 😉


blender:
*hem*... 🙈  I would also suggest "famework" instead of "framework"

mdiep:
👏🏻

PR #2785, opened 2019-05-13
Address SwiftLint violations

ikesyo:
Please see each commits for the details.

### DIFF
--- a/Source/CarthageKit/BinaryProject.swift
+++ b/Source/CarthageKit/BinaryProject.swift
@@ -1,8 +1,6 @@
 import Foundation
 import Result
-import SPMUtility
-
-import struct Foundation.URL // swiftlint:disable:this duplicate_imports
+import struct SPMUtility.Version
 
 /// Represents a binary dependency 
 public struct BinaryProject: Equatable {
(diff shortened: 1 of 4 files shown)


Source/CarthageKit/BinaryProject.swift
```diff
@@ -2,8 +2,6 @@ import Foundation
 import Result
 import SPMUtility
 
-import struct Foundation.URL
```
ikesyo: If per-type imports are preferred, we should add `Foundation.JSONDecoder` and `Foundation.Scanner`.

Source/CarthageKit/Project.swift
```diff
@@ -8,9 +8,6 @@ import XCDBLD
 import ReactiveTask
 import SPMUtility
 
-import struct Foundation.URL
```
ikesyo: If per-type imports are preferred, we should add more Foundation types here (I don't check what types are required in this file).

Source/CarthageKit/BinaryProject.swift
ikesyo: I understand that `Foundation.URL` is required to disambiguate against `SPMUtility.URL`.

Issue #2788, opened 2019-05-14
Carthage copy-frameworks lipo failures


kalkwarf:
With a very complex build, where multiple application targets share Carthage output, there is a concurrency issue if two build phases running in parallel try to work on the same framework.

```
A shell task (/usr/bin/xcrun lipo -remove armv7 -output [the project]/derived_data/Build/Products/Debug-iphonesimulator/OptimizelySDKiOS.framework.dSYM/Contents/Resources/DWARF/OptimizelySDKiOS [the project]/derived_data/Build/Products/Debug-iphonesimulator/OptimizelySDKiOS.framework.dSYM/Contents/Resources/DWARF/OptimizelySDKiOS) failed with exit code 1:
fatal error: /Applications/Xcode-10.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: -remove armv7 specified but fat file: [the project]/derived_data/Build/Products/Debug-iphonesimulator/OptimizelySDKiOS.framework.dSYM/Contents/Resources/DWARF/OptimizelySDKiOS does not contain that architecture
```

I added some logging to carthage, so I could trace the events. The first column is a timestamp, the second column is the carthage pid, then the file, line, function, and some arguments.

```
 1 732701638227897:1094:CarthageKit/Xcode.swift:1029:copyProduct(_:_:):OptimizelySDKiOS.framework.dSYM
 2 732701702823325:1094:CarthageKit/Xcode.swift:1129:stripArchitecture(_:_:):armv7:[the project]/derived_data/Build/Products/Debug-iphonesimulator/OptimizelySDKiOS.framework.dSYM
 3 732704610600709:1094:CarthageKit/Xcode.swift:1129:stripArchitecture(_:_:):arm64:[the project]/derived_data/Build/Products/Debug-iphonesimulator/OptimizelySDKiOS.framework.dSYM
 4 732716549186062:1684:CarthageKit/Xcode.swift:1029:copyProduct(_:_:):OptimizelySDKiOS.framework.dSYM
 5 732716595167774:1684:CarthageKit/Xcode.swift:1129:stripArchitecture(_:_:):armv7:[the project]/derived_data/Build/Products/Debug-iphonesimulator/OptimizelySDKiOS.framework.dSYM
 6 732717714194111:1916:CarthageKit/Xcode.swift:1029:copyProduct(_:_:):OptimizelySDKiOS.framework.dSYM
[...]

PR #2789, opened 2019-05-14
Modify packages atomically

kalkwarf:
This change is to address https://github.com/Carthage/Carthage/issues/2788

When executing `stripBinary` we need to be careful about other processes modifying the same framework we're working on. To do so, we copy the source into a temporary directory, modify it in place, then copy it to the final resting place.

While there is still a chance another build phase might do the exact same thing, the end result of N copy-then-strip operations should ultimately be the same.

### DIFF
--- a/Source/CarthageKit/FrameworkExtensions.swift
+++ b/Source/CarthageKit/FrameworkExtensions.swift
@@ -408,6 +408,35 @@ extension Reactive where Base: FileManager {
 		}
 		.map { URL(fileURLWithPath: $0, isDirectory: true) }
 	}
+  
+  public func copyItem(at: URL, to: URL) -> SignalProducer<URL, CarthageError> {
+    do {
+      try self.base.copyItem(at: at, to: to, avoiding·rdar·32984063: true)
+      return SignalProducer(value: to)
+    } catch {
+      return SignalProducer(error: .internalError(description: "copyItem failed: \(error)"))
+    }
+  }
+  
+  public func removeItem(at itemURL: URL) -> SignalProducer<(), CarthageError> {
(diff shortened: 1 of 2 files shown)


mdiep:
Looks good to me, but some tests are failing. 🙈

werner77:
This problem not only exists for stripping binaries, but also for concurrent git operations, concurrent caching operations (read/write from shared cache directories), concurrent operations on a shared derived data folder and so on.
A more fruitful approach is to lock these files/directories to protect them from concurrent access in the first place. Locking can be done by writing an associated lock file using /usr/bin/shlock.

mdiep:
> This problem not only exists for stripping binaries, but also for concurrent git operations, concurrent caching operations (read/write from shared cache directories), concurrent operations on a shared derived data folder and so on.

Within the context of `copy-frameworks`, those are all irrelevant.

> A more fruitful approach is to lock these files/directories to protect them from concurrent access in the first place. Locking can be done by writing an associated lock file using /usr/bin/shlock.

That wouldn't be helpful here. That would protect within a `carthage` invocation, but it would still leave temporarily inconsistent states that would be visible to other tools. An atomic/idempotent approach like the one taken here sidesteps these problems entirely.

werner77:
I do see your point, but then you also need to ensure that the stripDirectory and stripDebugSymbols operations are atomic. A better approach would then maybe be to copy the source framework/bcsymbolmaps/dsyms to a temp directory, perform all the operations there and only when everything is done, move the resulting files to the final directory (within derived data).

kalkwarf:
Needs work. Back shortly.

Issue #2777, opened 2019-04-29
Carthage update fails due to framework "couldn't be removed because you don't have permission to access it"


gbrhaz:
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `which carthage`: 0.33.0
* `carthage version`:  0.33.0
* `xcodebuild -version`:  10.2
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`?  yes
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? yes
* Are you using `--new-resolver`? yes

**Cartfile**
```
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseRemoteConfigBinary.json" == 5.20.0
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json" == 5.20.0
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseABTestingBinary.json" == 5.20.0
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebasePerformanceBinary.json" == 5.20.0
```

**Carthage Output**
```
*** Downloading binary-only framework FirebaseABTestingBinary at "https://dl.google.com/dl/firebase/ios/carthage/FirebaseABTestingBinary.json"
*** Downloading binary-only framework FirebasePerformanceBinary at "https://dl.google.com/dl/firebase/ios/carthage/FirebasePerformanceBinary.json"
*** Downloading binary-only framework FirebaseAnalyticsBinary at "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json"
*** Downloading binary-only framework FirebaseRemoteConfigBinary at "https://dl.google.com/dl/firebase/ios/carthage/FirebaseRemoteConfigBinary.json"
*** Downloading binary-only framework FirebaseABTestingBinary at "https://dl.google.com/dl/firebase/ios/carthage/FirebaseABTestingBinary.json"
*** xcodebuild output can be found in /var/folders/ny/mwt5zhw17cbdxg3d13rf22hrqtll8q/T/carthage-xcodebuild.77Rtd9.log
*** Downloading binary-only framework FirebaseAnalyticsBinary at "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json"
[...]

PR #2791, opened 2019-05-17
Reject invalid archives - Fix

blender:
Fixes inverted parameters in source and destination. Noticed by @ffittschen

### DIFF
--- a/IntegrationTests/binary.bats
+++ b/IntegrationTests/binary.bats
@@ -4,7 +4,7 @@ setup() {
     cd $BATS_TMPDIR
     rm -rf BinaryTest
     mkdir BinaryTest && cd BinaryTest
-    echo 'binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json"' > Carfile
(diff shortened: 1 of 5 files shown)


IntegrationTests/binary.bats
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bats
+
+setup() {
+    cd $BATS_TMPDIR
+    rm -rf BinaryTest
+    mkdir BinaryTest && cd BinaryTest
+    echo 'binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json"' > Carfile
```
ffittschen:
```suggestion
    echo 'binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json"' > Cartfile
```

blender:
I am not sure this solves #2777 I tried to run Carthage several times in a row and I don't get #2777 anymore

ffittschen:
It definitely doesn't resolve #2777. Only the intermediate issue I reported in https://github.com/Carthage/Carthage/pull/2765#issuecomment-493384319

I added steps to reproduce to the original issue: https://github.com/Carthage/Carthage/issues/2777#issuecomment-493464781

PR #2795, opened 2019-05-26
Improve submodule performance when --use-submodules is not used

CosynPa:
By using the repository cache instead of directly cloning the submodules from remote.


Source/CarthageKit/Git.swift
```diff
@@ -217,14 +227,24 @@ public func cloneSubmoduleInWorkingDirectory(_ submodule: Submodule, _ workingDi
...
 		}
 
-	return SignalProducer<(), CarthageError> { () -> Result<(), CarthageError> in
-		repositoryCheck("remove submodule checkout") {
-			try FileManager.default.removeItem(at: submoduleDirectoryURL)
+	let dependencyForSubmodule = Dependency.git(submodule.url)
+
+	return cloneOrFetchOnComplete(dependency: dependencyForSubmodule, preferHTTPS: preferHTTPS, commitish: submodule.sha)
+		.flatMap(.merge) { repositoryURL -> SignalProducer<String, CarthageError> in
+			return SignalProducer<(), CarthageError> { () -> Result<(), CarthageError> in
```
mdiep: It's not necessary to use `flatMap` with a custom `SignalProducer`. You can use `attemptMap` here instead.

Source/CarthageKit/Git.swift
```diff
@@ -217,14 +227,24 @@ public func cloneSubmoduleInWorkingDirectory(_ submodule: Submodule, _ workingDi
 				}
 		}
 
-	return SignalProducer<(), CarthageError> { () -> Result<(), CarthageError> in
-		repositoryCheck("remove submodule checkout") {
-			try FileManager.default.removeItem(at: submoduleDirectoryURL)
+	let dependencyForSubmodule = Dependency.git(submodule.url)
+
+	return cloneOrFetchOnComplete(dependency: dependencyForSubmodule, preferHTTPS: preferHTTPS, commitish: submodule.sha)
```
mdiep: `Git.swift` is meant to be independent of the Carthage-specifics, so I don't think we should call `cloneOrFetch` here. If we need to do something like this here, then we should add a `referenceRepository: URL?` parameter to this function to keep it generic.

Source/CarthageKit/Git.swift
```diff
@@ -217,14 +227,24 @@ public func cloneSubmoduleInWorkingDirectory(_ submodule: Submodule, _ workingDi
...
+					try FileManager.default.removeItem(at: submoduleDirectoryURL)
+				}
+			}
+			.then(cloneRepository(GitURL(repositoryURL.path), submoduleDirectoryURL, isBare: false))
 		}
-	}
-	.then(cloneRepository(submodule.url, workingDirectoryURL.appendingPathComponent(submodule.path), isBare: false))
-	.then(checkoutSubmodule(submodule, submoduleDirectoryURL))
-	.then(purgeGitDirectories)
+		.then(checkoutRepositoryToDirectory(submoduleDirectoryURL, submoduleDirectoryURL, force: false, revision: submodule.sha))
```
mdiep: Why is it important to have `force: false` here?

CosynPa: I'm just trying to keep the original logic. I don't know why sometimes the checkout is forced sometimes is not.

Source/CarthageKit/Git.swift
CosynPa: How can it be possible when this `cloneSubmoduleInWorkingDirectory` function deals with all submodules recursively? Pass a map that maps any submodule to its cache path?

CosynPa:
To make `Git.swift` independent of the Carthage-specifics, I've moved git related procedure of `cloneOrFetch` to `Git.swift`. The movement is in one commit, and the modification of `cloneOrFetch` in another, so reviewing individual commits should be easier than reviewing the whole PR.

I also added a `cacheURLMap` parameter to `cloneSubmoduleInWorkingDirectory`, so it is able to know how a git URL is mapped to a cache repository path without introducing any Carthage-specifics into `Git.swift`.

tmspzz:
@CosynPa 

Do I understand correctly that you want to check if there is already a clone containing the branch & commit of the repository in Carthage's cache? If so then skip the cloning from remote?

CosynPa:
Yes, and even if the commit does not exist in the cache repository, we only need to fetch in the cache repository, no need to clone the whole every time.

CosynPa:
I removed the last commit which tried to solve the race condition problem because I found that commit did not solve the problem completely. The complete solution would be a little complex, so I think it's better to put it in another pull request.

tmspzz:
@CosynPa please update the branch from master to fix the CI

CosynPa:
Rebased onto master, but a test failed. Any idea?

tmspzz:
Yes, it's a test trying to get the remote version. It's unrelated to you PR so for me this is good.

tmspzz:
Thanks! 🎉

PR #2796, opened 2019-05-27
Bugfix 2788

kalkwarf:
When stripping frameworks, copy the source into a temporary directory, process it there, and then move it back.

This is a rework of my previous PR, which didn't actually do what I thought it did.

### DIFF
--- a/Source/CarthageKit/FrameworkExtensions.swift
+++ b/Source/CarthageKit/FrameworkExtensions.swift
@@ -409,30 +409,22 @@ extension Reactive where Base: FileManager {
 		.map { URL(fileURLWithPath: $0, isDirectory: true) }
 	}
   
-  public func copyItem(at: URL, to: URL) -> SignalProducer<URL, CarthageError> {
+  public func copyItem(_ source: URL, into: URL) -> SignalProducer<URL, CarthageError> {
+    let destination = into.appendingPathComponent(source.lastPathComponent)
     do {
-      try self.base.copyItem(at: at, to: to, avoiding·rdar·32984063: true)
-      return SignalProducer(value: to)
+      try self.base.copyItem(at: source, to: destination, avoiding·rdar·32984063: true)
+      return SignalProducer(value: destination)
     } catch {
       return SignalProducer(error: .internalError(description: "copyItem failed: \(error)"))
     }
   }
   
-  public func removeItem(at itemURL: URL) -> SignalProducer<(), CarthageError> {
+  public func replaceItem(at originalItemURL: URL, withItemAt newItemURL: URL) -> SignalProducer<(), CarthageError> {
     do {
-      try FileManager().removeItem(at: itemURL)
-      return SignalProducer(.empty)
-    } catch {
-      return SignalProducer(error: .internalError(description: "removeItem failed: \(error)"))
-    }
-  }
-  
-  public func replaceItem(at originalItemURL: URL, withItemAt newItemURL: URL) -> SignalProducer<URL, CarthageError> {
-    do {
-      guard let url = try FileManager().replaceItemAt(originalItemURL, withItemAt: newItemURL, backupItemName: nil, options: .usingNewMetadataOnly) else {
+      guard (try FileManager().replaceItemAt(originalItemURL, withItemAt: newItemURL, backupItemName: nil, options: .usingNewMetadataOnly)) != nil else {
         return SignalProducer(error: .internalError(description: "replaceItem succeeded, but returned nil"))
       }
-      return SignalProducer(value: url)
+      return SignalProducer(.empty)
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -1012,9 +1012,35 @@ public func stripDSYM(_ dSYMURL: URL, keepingArchitectures: [String]) -> SignalP
...
+  
+  return createTempDir
+    .flatMap(.merge) { temp in
+      copyItem(binaryURL, temp)
+        .flatMap(.merge) { workspace in
+          strip(workspace, keepingArchitectures)
+            .flatMap(.merge) { workspace in
+              replace(binaryURL, workspace)
+          }
+      }
```
mdiep: You should be able to unnest these:

```swift
  return createTempDir
     .flatMap(.merge) { temp in
       copyItem(binaryURL, temp)
     }
     .flatMap(.merge) { workspace in
       strip(workspace, keepingArchitectures)
     }
     .flatMap(.merge) { workspace in
       replace(binaryURL, workspace)
     }
```

Source/CarthageKit/Xcode.swift
```diff
@@ -1012,9 +1012,35 @@ public func stripDSYM(_ dSYMURL: URL, keepingArchitectures: [String]) -> SignalP
 
 /// Strips a universal file from unexpected architectures.
 private func stripBinary(_ binaryURL: URL, keepingArchitectures: [String]) -> SignalProducer<(), CarthageError> {
-	return architecturesInPackage(binaryURL)
-		.filter { !keepingArchitectures.contains($0) }
-		.flatMap(.concat) { stripArchitecture(binaryURL, $0) }
+  let fileManager = FileManager().reactive
```
mdiep: Should this use `FileManager.default`?

kalkwarf: I don't remember the specific details, but there is (was?) a concurrency issue with the default FileManager.

Ah, here it is:

https://useyourloaf.com/blog/nsfilemanager-defaultmanager-is-not-thread-safe/

and once again, I know things that are no longer true. :-) I'll update it.

mdiep: TIL!

Source/CarthageKit/Xcode.swift
kalkwarf: Will do.

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -408,6 +408,27 @@ extension Reactive where Base: FileManager {
...
+      try self.base.copyItem(at: source, to: destination, avoiding·rdar·32984063: true)
+      return SignalProducer(value: destination)
+    } catch {
+      return SignalProducer(error: .internalError(description: "copyItem failed: \(error)"))
+    }
+  }
+  
+  public func replaceItem(at originalItemURL: URL, withItemAt newItemURL: URL) -> SignalProducer<(), CarthageError> {
+    do {
+      guard (try FileManager().replaceItemAt(originalItemURL, withItemAt: newItemURL, backupItemName: nil, options: .usingNewMetadataOnly)) != nil else {
```
mdiep: I missed this before, but this should be using `self.base`, not `FileManager()`

Source/CarthageKit/Xcode.swift
```diff
@@ -1012,9 +1012,35 @@ public func stripDSYM(_ dSYMURL: URL, keepingArchitectures: [String]) -> SignalP
...
+      .filter { !keepingArchitectures.contains($0) }
+      .flatMap(.concat) { stripArchitecture(workspace, $0) }
+      .then(SignalProducer(value: workspace))
+  }
+  
+  let replace: (URL, URL) -> SignalProducer<(), CarthageError> = { original, modified in
+    return fileManager.replaceItem(at: original, withItemAt: modified)
+  }
+  
+  return createTempDir
```
mdiep: Would you mind adding a comment here that explains why we're going through the trouble of copying/stripping/replacing instead of just stripping?

Source/CarthageKit/FrameworkExtensions.swift
kalkwarf: Ah. You're right. Stand by…

Source/CarthageKit/Xcode.swift
kalkwarf: Done!

Source/CarthageKit/FrameworkExtensions.swift
kalkwarf: Done!

mdiep:
Awesome! This is great, thanks @kalkwarf. ✨

PR #2798, opened 2019-06-03
Add .xcfilelist documentation

sticksen:
This PR adds documentation about the `xcfilelist`s that were introduced during WWDC18.
I think it´s easier to modify or automate the modification of these dependency declaration if they are in a dedicated file.


blender:
looks goo to me! Thanks!

Issue #2799, opened 2019-06-03
Catalyst support


NachoSoto:
Apparently in order for this to work Carthage will need to build `.xcframework`s now: https://twitter.com/NeoNacho/status/1135691439734575104.

ikesyo:
`.xcframework` can be made by `xcodebuild -create-xcframework` which takes `-framework` and `-out` options. So that's means that we should run the command instead of `lipo`ing.

blender:
example:
```
$ xcodebuild archive -workspace Alamofire.xcworkspace -scheme "Alamofire iOS" -sdk iphoneos13.0 OBJROOT=build/iOS
$ xcodebuild archive -workspace Alamofire.xcworkspace -scheme "Alamofire iOS" -sdk iphonesimulator13.0 OBJROOT=build/simulator
$ xcodebuild -create-xcframework -framework build/iOS/UninstalledProducts/iphoneos/Alamofire.framework -framework build/simulator/UninstalledProducts/iphonesimulator/Alamofire.framework -output build/Alamofire.xcframework
```

blender:
Can anyone actually build Carthage on Xcode 11 Swift 5.1?

alexaubry:
@blender I just tried, it seems unhappy with the currying in the `OptionsProtocol` objects evaluation (e.g. in `CheckoutCommand.Options.evaluate`):

![Error in ](https://user-images.githubusercontent.com/16192914/58887547-1d149900-86e6-11e9-9dbf-dec6c5926c13.png)

Using `Options.init` seems to work, though.

blender:
Yeah, I was trying to go around this by specifying a lower version of swift in the Makefile but I'm hitting a wall.

ikesyo:
Looks like building Carthage itself with Xcode 11 (Swift 5.1) is a separate issue.

blender:
I’m on this btw ;)

blender:
https://github.com/Carthage/Carthage/pull/2801 this is in progress but can already build xcframeworks with `carthage  build --platform iOS --no-skip-current --create-xcframework`

I'm not sure what we're supposed to do with dSYMs

haithngn:
> Can anyone actually build Carthage on Xcode 11 Swift 5.1?

yes I did but no luck,
It seems the framework will be build in swift 4.1.2, that is very strange.

blender:
[...]

PR #2801, opened 2019-06-05
Support xcframeworks

tmspzz:
Implement support for #2799 

This feature involves: 
- [x] supporting creation of xcframeworks in `build`
- [x] supporting creation of archives with xcframeworks
- [x] supporting binary xcframeworks
- [ ] avoiding xcframeworks in copy-frameworks
- [ ] dSYMs ?

### DIFF
--- a/Source/CarthageKit/BuildOptions.swift
+++ b/Source/CarthageKit/BuildOptions.swift
@@ -15,13 +15,17 @@ public struct BuildOptions {
 	/// Whether to use downloaded binaries if possible.
 	public var useBinaries: Bool
 
+	/// Whether to use xcframeworks or not
+	public var useXCFrameworks: Bool
+
 	public init(
 		configuration: String,
 		platforms: Set<Platform> = [],
 		toolchain: String? = nil,
 		derivedDataPath: String? = nil,
 		cacheBuilds: Bool = true,
-		useBinaries: Bool = true
+		useBinaries: Bool = true,
+		useXCFrameworks: Bool = false
 	) {
 		self.configuration = configuration
 		self.platforms = platforms
@@ -29,5 +33,6 @@ public struct BuildOptions {
 		self.derivedDataPath = derivedDataPath
 		self.cacheBuilds = cacheBuilds
 		self.useBinaries = useBinaries
+		self.useXCFrameworks = useXCFrameworks
 	}
 }
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -216,7 +216,9 @@ public struct BuildSettings {
 			pathComponent = "\(configuration)\(effectivePlatformName)"
 		}
 
-		let path = (basePath as NSString).appendingPathComponent(pathComponent)
+		// TODO: could use self["variant"] maybe, is SDK is passed with variant
+		let isUIKitForMac = (self.supportsUIKitForMac.value ?? false) && (self["SDK_NAME"].map { $0.contains("macos") }.value ?? false)
+		let path = (basePath as NSString).appendingPathComponent(isUIKitForMac ? "\(pathComponent)-uikitformac" : pathComponent)
 		return .success(path)
 	}
 
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -305,7 +305,7 @@ public struct BuildSettings {
 
 	/// Attepts to determine if UIKit for Mac is supported
 	public var supportsUIKitForMac: Result<Bool, CarthageError> {
-		self["SUPPORTS_UIKITFORMAC"].map { $0 == "YES" }
+		return self["SUPPORTS_UIKITFORMAC"].map { $0 == "YES" }
 	}
 
 	/// Add subdirectory path if it's not possible to paste product to destination path
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -43,7 +43,9 @@ public struct BuildSettings {
 	///
 	/// Upon .success, sends one BuildSettings value for each target included in
 	/// the referenced scheme.
-	public static func load(with arguments: BuildArguments, for action: BuildArguments.Action? = nil) -> SignalProducer<BuildSettings, CarthageError> {
+	public static func load(with arguments: BuildArguments,
+		useRawArguments: Bool = false,
+		for action: BuildArguments.Action? = nil) -> SignalProducer<BuildSettings, CarthageError> {
 		// xcodebuild (in Xcode 8.0) has a bug where xcodebuild -showBuildSettings
 		// can hang indefinitely on projects that contain core data models.
 		// rdar://27052195
@@ -52,7 +54,9 @@ public struct BuildSettings {
 		//
 		// "archive" also works around the issue above so use it to determine if
 		// it is configured for the archive action.
-		let task = xcodebuildTask(["archive", "-showBuildSettings", "-skipUnavailableActions"], arguments)
+		let task = xcodebuildTask(["archive", "-showBuildSettings", "-skipUnavailableActions"],
+								  arguments,
+								  useRawArguments: useRawArguments)
 
 		return task.launch()
 			.ignoreTaskData()
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -297,6 +297,11 @@ public struct BuildSettings {
 		return self["TARGET_BUILD_DIR"]
 	}
 
+	/// Attepts to determine if UIKit for Mac is supported
+	public var supportsUIKitForMac: Result<Bool, CarthageError> {
+		self["SUPPORTS_UIKITFORMAC"].map { $0 == "YES" }
+	}
+
 	/// Add subdirectory path if it's not possible to paste product to destination path
 	public func productDestinationPath(in destinationURL: URL) -> URL {
 		let directoryURL: URL
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -241,6 +241,11 @@ public struct BuildSettings {
 		return self["WRAPPER_NAME"]
 	}
 
+	/// Attempts to determine the name of the built product's wrapper bundle replacing "framework" with "xcframework".
+	public var xcFrameworkWrapperName: Result<String, CarthageError> {
+		return self["WRAPPER_NAME"].map { $0.spm_dropSuffix(".framework").appending(".xcframework") }
+	}
+
 	/// Attempts to determine the URL to the built product's wrapper, corresponding
 	/// to its xcodebuild action.
 	public var wrapperURL: Result<URL, CarthageError> {
--- a/Source/CarthageKit/FrameworkExtensions.swift
+++ b/Source/CarthageKit/FrameworkExtensions.swift
@@ -1,6 +1,7 @@
 import Foundation
 import Result
 import ReactiveSwift
+import XCDBLD
 
 extension String {
 	/// Returns a producer that will enumerate each line of the receiver, then
@@ -295,14 +296,50 @@ extension URL {
 
 	/// Returns the first `URL` to match `<self>/Headers/*-Swift.h`. Otherwise `nil`.
 	internal func swiftHeaderURL() -> URL? {
-		let headersURL = self.appendingPathComponent("Headers", isDirectory: true).resolvingSymlinksInPath()
+
+		var libraryIdentifier: String = ""
+		var libraryPath: String = ""
+		if self.pathExtension == "xcframework" {
+			let xcFrameworkInfo = Bundle(url: self)?
+				.infoDictionary
+				.flatMap(XCFrameworkInfo.init)
+			let firstLibrary =  xcFrameworkInfo?
+				.availableLibraries
+				.first
+			libraryIdentifier = firstLibrary?
+				.identifier ?? ""
+			libraryPath = firstLibrary?.path ?? ""
+		}
+
+		let headersURL = self.appendingPathComponent(libraryIdentifier)
+			.appendingPathComponent(libraryPath)
+			.appendingPathComponent("Headers", isDirectory: true)
+			.resolvingSymlinksInPath()
 		let dirContents = try? FileManager.default.contentsOfDirectory(at: headersURL, includingPropertiesForKeys: [], options: [])
 		return dirContents?.first { $0.absoluteString.contains("-Swift.h") }
 	}
 
 	/// Returns the first `URL` to match `<self>/Modules/*.swiftmodule`. Otherwise `nil`.
 	internal func swiftmoduleURL() -> URL? {
-		let headersURL = self.appendingPathComponent("Modules", isDirectory: true).resolvingSymlinksInPath()
+
+		var libraryIdentifier: String = ""
+		var libraryPath: String = ""
+		if self.pathExtension == "xcframework" {
+			let xcFrameworkInfo = Bundle(url: self)?
+				.infoDictionary
+				.flatMap(XCFrameworkInfo.init)
+			let firstLibrary =  xcFrameworkInfo?
+				.availableLibraries
+				.first
+			libraryIdentifier = firstLibrary?
+				.identifier ?? ""
+			libraryPath = firstLibrary?.path ?? ""
+		}
+
+		let headersURL = self.appendingPathComponent(libraryIdentifier)
+			.appendingPathComponent(libraryPath)
+			.appendingPathComponent("Modules", isDirectory: true)
+			.resolvingSymlinksInPath()
 		let dirContents = try? FileManager.default.contentsOfDirectory(at: headersURL, includingPropertiesForKeys: [], options: [])
 		return dirContents?.first { $0.absoluteString.contains("swiftmodule") }
 	}
--- a/Source/CarthageKit/FrameworkExtensions.swift
+++ b/Source/CarthageKit/FrameworkExtensions.swift
@@ -297,22 +297,9 @@ extension URL {
 	/// Returns the first `URL` to match `<self>/Headers/*-Swift.h`. Otherwise `nil`.
 	internal func swiftHeaderURL() -> URL? {
 
-		var libraryIdentifier: String = ""
-		var libraryPath: String = ""
-		if self.pathExtension == "xcframework" {
-			let xcFrameworkInfo = Bundle(url: self)?
-				.infoDictionary
-				.flatMap(XCFrameworkInfo.init)
-			let firstLibrary =  xcFrameworkInfo?
-				.availableLibraries
-				.first
-			libraryIdentifier = firstLibrary?
-				.identifier ?? ""
-			libraryPath = firstLibrary?.path ?? ""
-		}
-
-		let headersURL = self.appendingPathComponent(libraryIdentifier)
-			.appendingPathComponent(libraryPath)
+		let xcFrameworkLibraryIdAndPath: (identifier: String, path: String)? = self.xcFrameworkLibraryIdentifiedAndPath()
+		let headersURL = self.appendingPathComponent(xcFrameworkLibraryIdAndPath?.identifier ?? "" )
+			.appendingPathComponent(xcFrameworkLibraryIdAndPath?.path ?? "")
 			.appendingPathComponent("Headers", isDirectory: true)
 			.resolvingSymlinksInPath()
 		let dirContents = try? FileManager.default.contentsOfDirectory(at: headersURL, includingPropertiesForKeys: [], options: [])
@@ -322,26 +309,39 @@ extension URL {
 	/// Returns the first `URL` to match `<self>/Modules/*.swiftmodule`. Otherwise `nil`.
 	internal func swiftmoduleURL() -> URL? {
 
-		var libraryIdentifier: String = ""
-		var libraryPath: String = ""
+		let xcFrameworkLibraryIdAndPath: (identifier: String, path: String)? = self.xcFrameworkLibraryIdentifiedAndPath()
+		let modulesURL = self.appendingPathComponent(xcFrameworkLibraryIdAndPath?.identifier ?? "" )
+			.appendingPathComponent(xcFrameworkLibraryIdAndPath?.path ?? "")
+			.appendingPathComponent("Modules", isDirectory: true)
+			.resolvingSymlinksInPath()
+		let dirContents = try? FileManager.default.contentsOfDirectory(at: modulesURL, includingPropertiesForKeys: [], options: [])
+		return dirContents?.first { $0.absoluteString.contains("swiftmodule") }
+	}
+
+	func xcFrameworkLibraryIdentifiedAndPath() -> (String, String)? {
+
+		let libraryIdentifier: String
+		let libraryPath: String
 		if self.pathExtension == "xcframework" {
 			let xcFrameworkInfo = Bundle(url: self)?
 				.infoDictionary
 				.flatMap(XCFrameworkInfo.init)
-			let firstLibrary =  xcFrameworkInfo?
-				.availableLibraries
-				.first
-			libraryIdentifier = firstLibrary?
-				.identifier ?? ""
-			libraryPath = firstLibrary?.path ?? ""
-		}
 
-		let headersURL = self.appendingPathComponent(libraryIdentifier)
-			.appendingPathComponent(libraryPath)
-			.appendingPathComponent("Modules", isDirectory: true)
-			.resolvingSymlinksInPath()
-		let dirContents = try? FileManager.default.contentsOfDirectory(at: headersURL, includingPropertiesForKeys: [], options: [])
-		return dirContents?.first { $0.absoluteString.contains("swiftmodule") }
+			guard let firstLibrary = xcFrameworkInfo?
+				.availableLibraries
+				.first else {
+					return nil
+			}
+			libraryIdentifier = firstLibrary
+				.identifier
+			libraryPath = firstLibrary.path
+
+			return (libraryIdentifier, libraryPath)
+
+		}
+		else {
+			return nil
+		}
 	}
 }
 
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -704,18 +704,18 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 						}
 					}
 					.flatMapTaskEvents(.concat) { deviceSettings, simulatorSettings in
-
 						if options.useXCFrameworks {
+							let frameworkURLs = (deviceSettings.wrapperURL.fanout(simulatorSettings.wrapperURL))
+								.map { [ $0, $1 ] }
+							let outputURL = deviceSettings
+								.xcFrameworkWrapperName
+								.map(folderURL.appendingPathComponent)
 
-							let r = folderURL.appendingPathComponent(deviceSettings.wrapperName.value!.spm_dropSuffix("framework")+"xcframework")
-							let rP = SignalProducer<URL, CarthageError>(value: r)
-							return createXCFramework(
-								[deviceSettings.wrapperURL.value!, simulatorSettings.wrapperURL.value!],
-								r
-							).then(rP)
+							return SignalProducer(result: frameworkURLs.fanout(outputURL))
+								.flatMap(.merge, createXCFramework)
+								.then(SignalProducer(result: outputURL))
 						}
 						else {
-
 							return mergeBuildProducts(
 								deviceBuildSettings: deviceSettings,
 								simulatorBuildSettings: simulatorSettings,
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -472,7 +472,9 @@ private func shouldBuildScheme(_ buildArguments: BuildArguments, _ forPlatforms:
 					.flatMapError { _ in .empty }
 			} else {
 				return settings.buildSDKs
-					.filter { forPlatforms.contains($0.platform) }
+					.filter {
+						forPlatforms.contains($0.platform)
+					}
 					.flatMap(.merge) { _ in frameworkType }
 					.flatMapError { _ in .empty }
 			}
@@ -613,7 +615,7 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 	)
 
 	return BuildSettings.SDKsForScheme(scheme, inProject: project)
-		.flatMap(.concat) { sdk -> SignalProducer<SDK, CarthageError> in
+		.flatMap(.concat) { sdk -> SignalProducer<(BuildSettings, SDK), CarthageError> in
 			var argsForLoading = buildArgs
 			argsForLoading.sdk = sdk
 
@@ -625,14 +627,19 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 					// must add a User-Defined setting of ENABLE_BITCODE=NO.
 					return settings.bitcodeEnabled.value == true || ![.tvOS, .watchOS].contains(sdk)
 				}
-				.map { _ in sdk }
+				.map { settings in return (settings, sdk) }
 		}
-		.reduce(into: [:]) { (sdksByPlatform: inout [Platform: Set<SDK>], sdk: SDK) in
+		.reduce(into: [Platform: Set<SDK>]()) { sdksByPlatform, next in
+			let (settings, sdk) = next
 			let platform = sdk.platform
 
 			if var sdks = sdksByPlatform[platform] {
 				sdks.insert(sdk)
 				sdksByPlatform.updateValue(sdks, forKey: platform)
+				if platform == .iOS && settings.supportsUIKitForMac.value == true && options.useXCFrameworks {
+					sdks.insert(.macOSX)
+					sdksByPlatform.updateValue(sdks, forKey: platform)
+				}
 			} else {
 				sdksByPlatform[platform] = [sdk]
 			}
@@ -646,7 +653,9 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 			return SignalProducer(values)
 		}
 		.flatMap(.concat) { platform, sdks -> SignalProducer<(Platform, [SDK]), CarthageError> in
-			let filterResult = sdkFilter(sdks, scheme, options.configuration, project)
+
+			let wantsXCFramworks = platform == .iOS && sdks.contains(.macOSX) && options.useXCFrameworks
+			let filterResult = !wantsXCFramworks ? sdkFilter(sdks, scheme, options.configuration, project) : .success(sdks)
 			return SignalProducer(result: filterResult.map { (platform, $0) })
 		}
 		.filter { _, sdks in
@@ -729,6 +738,11 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 							)
 						}
 					}
+
+			case 3 where platform == .iOS && sdks.contains(.macOSX) && options.useXCFrameworks:
+
+				fatalError()
+
 
 			default:
 				fatalError("SDK count \(sdks.count) in scheme \(scheme) is not supported")
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -583,7 +583,8 @@ private func mergeBuildProducts(
 				}
 
 			return mergeProductBinaries
-				.then(mergeProductSwiftHeaderFilesIfNeeded).then(mergeProductModules)
+				.then(mergeProductSwiftHeaderFilesIfNeeded)
+				.then(mergeProductModules)
 				.then(copyBCSymbolMapsForBuildProductIntoDirectory(destinationFolderURL, simulatorBuildSettings))
 				.then(SignalProducer<URL, CarthageError>(value: productURL))
 		}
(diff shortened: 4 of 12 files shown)


blender:
@ikesyo  @mdiep  @jdhealy  @NachoSoto can anyone please give a first review pass? 

The review path should be: build command  -> archive command > bootstrap with `binary` in Carfile

ikesyo:
I don't take deeper look yet, but I think that the XCFrameworks should not be in `Carthage/Build/$platform` because the format supports bundling frameworks for multiple platforms. Those should be placed directly in `Carthage/Build` or `Carthage/Build/XCFrameworks`.

blender:
The way I implemented it, the .xcframework created by Carthage contains platform+simulator never multiple platforms.

NachoSoto:
> The way I implemented it, the .xcframework created by Carthage contains platform+platformsimulator, never multiple platforms.

I haven’t looked at the code yet, but I would expect the Xcode command to generate these for all platforms? If not, we probably do need to change that, as the main advantage of this is to be able to finally get rid of `copy-frameworks` (http://www.openradar.me/radar?id=6409498411401216) and to support #2799.

blender:
@NachoSoto 

> I haven’t looked at the code yet, but I would expect the Xcode command to generate these for all platforms?

It does, just it doesn't combine tvOS and watchOS in the same .xcframework. It combines tvOS + tvSimualtor in one xcframework bundle where inside there are 2 separate framework bundles each for slices for the correct architecture. They are not fat binaries anymore.

So there radar problem is solved.

blender:
@ikesyo I think that technically you are correct, I imagine the format would support placing all platforms in one .xcframework . If one did that then it wouldn't make sense to go with `Carthage/Build/$platform`.

I didn't go that way since I'm not sure how this is expected to be used. However the platform specific .xcframeworks could be just repacked in another .xcframework in the end if we wanted to do so.

blender:
So this is the current structure on disk, just use your imagination for the other platforms :D

Screenshot 2019-06-06 at 11 08 18

ikesyo:
It will be okay as a first step 😃

blender:
I'm not sure what else is supposed to happen to support Catalyst ? Can anyone point me to more documentation? https://developer.apple.com/documentation/uikit/creating_a_mac_version_of_your_ipad_app

mdiep:
My guess for Catalyst is that the XCFramework needs to contain both iOS and macOS versions so that the correct platform can be chosen.

samrayner:
It's a bit more complicated than that unfortunately. With some help from @owensd on Twitter I managed to rebuild my app's dependencies and get things running.

A `ios-x86_64-uikitformac` build is required, done against the My Mac destination (`-destination generic/platform=macOS`) and included alongside the iphoneos13.0 and iphonesimulator13.0 SDK builds. A native macOS framework in the xcframework won't help with Catalyst. 

I assume this can be done through the command line but I couldn't figure it out so here are my manual Xcode steps. As usual with these things I'm not sure all of them are required but they got things working.

- Open dependency workspace or project from Carthage/Checkouts
- Edit scheme and set Run build configuration to Release
- Edit Build Settings:
    - Base SDK = iOS (in case framework base is macOS)
    - Supported Platforms = iOS (in case framework supports native macOS builds)
    - Build Libraries for Distribution = YES (recommended in Xcode 11 release notes)
    - Supports UIKit for Mac = YES
    - Derive UIKit for Mac Product Bundle Identifier = YES
- Ensure the Mac device checkbox is checked in General > Deployment Info
- Run the against My Mac
- Copy the framework and dSYMs from `Build > Products > Release-uikitformac` in DerivedData
- `xcodebuild -create-xcframework -framework ios.framework -framework sim.framework -framework uikitformac.framework -output output.xcframework`

I'm sure there's a way to maintain a framework's native macOS support while also offering UIKit for Mac support but I'm not sure how. I guess you would need two Mac build destinations?

Hope that helps.

blender:
@mdiep just tried that, didn't work :(

@samrayner thanks for all the information. I'll see I can make of it

owensd:
A few pointers:
  - [What is an XCFramework?](https://help.apple.com/xcode/mac/11.0/index.html?localePath=en.lproj#/dev6f6ac218b)
  - [Introducing iPad Apps for Mac](https://developer.apple.com/wwdc19/205)
  - [Binary Frameworks in Swift](https://developer.apple.com/wwdc19/416)

@NachoSoto Yes, an XCFramework can and *should* have all of the platforms you wish to support for any given framework. It's not intended that you ship multiple versions of a `.xcframework` for different platforms (e.g. macOS, iOS, tvOS, etc...). The intention is that a single `.xcframework` will have all of those in one.

Also, it's recommended to use the output of an archive build when packaging up frameworks for distribution, not just release builds. I didn't actually look at the code changes here to see if that's what's being done, just a general comment.

> Build Libraries for Distribution = YES (recommended in Xcode 11 release notes)

This is not just _recommended_, for Swift, this is **required**. Without this setting, you will not be building Swift in a module stable way; this setting is what enables module stability and generates the stable `.swiftinterface` files. Without it, consumers of your XCFrameworks will be forced to use the same compiler version, just as you do today for Swift.

blender:
There are 2 issues to solve:
- [ ] build iOS schemes with `macosx` sdk, this produces the Release-uikitformac build that we want
- [ ] repack all .xcframeworks in a new one

blender:
example build command:

```
xcrun xcodebuild -workspace /Users/tommaso/Code/lab/Alamofire/Alamofire.xcworkspace -scheme Alamofire\ iOS -configuration Release -sdk macosx ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES archive -archivePath ./here SKIP_INSTALL=YES GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=NO CLANG_ENABLE_CODE_COVERAGE=NO STRIP_INSTALLED_PRODUCT=NO
```

samrayner:
FYI, thread: https://twitter.com/samrayner/timelines/1136688278529290241

blender:
I'm getting closer. 

Current problem: Seems like that 2 builds folders overwrite each others to me. If anyone can figure out why https://github.com/Carthage/Carthage/pull/2801/commits/403afee8fd7c2bcac1c39d47a4a58161c082a614#diff-f66644e8cfbcad3c3826395a690ddfbcR805 iPhoneSimulator and MacOS have the same build path I would be grateful

blender:
🎉 

Screenshot 2019-06-07 at 18 27 51

NachoSoto:
🙌🏻🙌🏻🙌🏻👏🏻👏🏻👏🏻

blender:
Hi all,

Carthage is now able to produce xcframworks with uikitformac support.

As soon as I have time I will:
- Add a detailed explanation of the changes, probably in this thread
- Clean up some types
- Check and add tests

In the mean time, please try with:

```
carthage build --no-skip-current --create-xcframework
```

ikesyo:
Related session video: https://developer.apple.com/videos/play/wwdc2019/416/

Source/XCDBLD/XCFrameworkInfo.swift
```diff
@@ -0,0 +1,101 @@
...
+			case .tvOS:
+				self = isSimulatorVariant ? .tvSimulator : .tvOS
+			case .watchOS:
+				self = isSimulatorVariant ? .watchSimulator : .watchOS
+		}
+	}
+}
+
+/// Represents the information parsed from the Info.plist of an .xcframework bundle
+public struct XCFrameworkInfo {
```
ikesyo: An example plist output:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>AvailableLibraries</key>
    <array>
        <dict>
            <key>LibraryIdentifier</key>
            <string>ios-x86_64-uikitformac</string>
            <key>LibraryPath</key>
            <string>FooBar.framework</string>
            <key>SupportedArchitectures</key>
            <array>
                <string>x86_64</string>
            </array>
            <key>SupportedPlatform</key>
            <string>ios</string>
            <key>SupportedPlatformVariant</key>
            <string>uikitformac</string>
        </dict>
        <dict>
            <key>LibraryIdentifier</key>
            <string>ios-armv7_arm64</string>
            <key>LibraryPath</key>
            <string>FooBar.framework</string>
            <key>SupportedArchitectures</key>
            <array>
                <string>armv7</string>
                <string>arm64</string>
            </array>
            <key>SupportedPlatform</key>
            <string>ios</string>
        </dict>
        <dict>
            <key>LibraryIdentifier</key>
            <string>ios-i386_x86_64-simulator</string>
            <key>LibraryPath</key>
            <string>FooBar.framework</string>
            <key>SupportedArchitectures</key>
            <array>
                <string>i386</string>
                <string>x86_64</string>
            </array>
            <key>SupportedPlatform</key>
            <string>ios</string>
            <key>SupportedPlatformVariant</key>
            <string>simulator</string>
        </dict>
    </array>
    <key>CFBundlePackageType</key>
    <string>XFWK</string>
    <key>XCFrameworkFormatVersion</key>
    <string>1.0</string>
</dict>
</plist>
```

blender: Did I mess it up?

ikesyo: I don't mean that, just as a review note. 😃

min:
not sure if I'm doing something wrong but `carthage build --no-skip-current --create-xcframework` produces an `xcframework` but doesn't generate a uikitformac version

ikesyo:
@min IIUC to build `uikitformac` variant you should make it sure that your framework supports `uikitformac` and you need Catalina beta (since UIKit for Mac is only available in Catalina).

min:
@ikesyo yep, I'm on Catalina and tried on this on Alamofire. I'll keep digging

blender:
@min check @samrayner ‘s comment. I changed Alamofire’s project to get this to work.

blender:
@ikesyo do you we think we should be injecting `BUILD_LIBRARY_FOR_DISTRIBUTION` ?

blender:
Hi all, before I countinue with the PR I would like to spend
a few words explaining the changes, the desing choices and possible
limitations of this implementation. This should also help in reviewing
the changes.

First off, the ability to __bundle__ multiple frameworks into an
.xcframework is done via the flag `--use-xcframeworks` so the behavior
if completely opt-in.

When specified, this flag enables the following:
1. bundling all architectures supported by a target for a specific scheme into one .xcframework
2. building iOS targets with Catalyst support when the target supports it

### 1. Bundling all architectures supported by a target for a specific scheme into one .xcframework

Hoping I haven't missed anything, Carthage's current approach to
building is roughly the following:

1. Find buildable schemes
1. Find which SDKs are supported by the schemes at the previous step
1. Filter SDKs to match `--platform` arguments if any
1. Build all targets in the scheme for a __max__ for 2 SKDs (sim + device)

Build products are then placed into `Carthage/Build/$platform/`

The approach I took for .xcframeworks repects this convention.
That means that it will simply bundle together products of the same
target & scheme into an .xcframework . This means that, Catalyst aside
for now, there will at most __two__ platform included in the .xcframework (sim + device)

While it's true that .xcframework can include all platforms this is
not a concept carthage (and some projects) are built around. Take
Alamofire for example: one scheme, one target, at most two platforms.

If were to implement bundling all platforms together, how would be
correlate between schemes in a setup like Alamofire?

So the limitation here is that carthage won't produce .xcframeworks including all platforms.

### Building iOS targets with Catalyst support when the target supports it

To make this work I have added support to build for 3 SKDs when the target specifies `SUPPORTS_UIKITFORMAC`.

When that is the case carthage will force `macOS` into the SDKs for `iOS` targets, effectlivetly building for UIKit for Mac. 

Unfortunately building twice the same target clears derived data so I had to copy files around to be able to perform 3 builds of the same target.

The documented solution in the WWDC session is to use multiple destinations. I haven't tried as making that change seemed a much bigger effect than the current solution.


### Conclusion

I hope to have expressed the changes and choices clearly and in a
fashion that will make review easier. Feel free to propose any changes,
I'm available for discussion.

CAPIStkidd:
I'm trying to create an example app like yours using Alamofire inclusion as a test sand although I think I've applied the changes to the Alamofire project that @samrayner listed, I'm still either getting no `uikitformac` variant in the `.xcframework` (just the other two), or I'm getting an error - looks like it's building `Alamofire.framework` in the Mac directory, then at some point during the iOS build it deletes that and then throws an error saying it's missing. 

I'm sure I'm missing a setting I'm just not sure what. Could you perhaps post your CatalystApp/Alamofire source somewhere and I could see what the difference is?

min:
@CAPIStkidd worked for me when I made the build settings changes to Alamofire at the project level instead of the target level

blender:
@CAPIStkidd 

<details>
<summary>Alamofire patch</summary>

```diff
diff --git a/Alamofire.xcodeproj/project.pbxproj b/Alamofire.xcodeproj/project.pbxproj
index 79f0d31..eb9e59d 100644
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -1858,10 +1858,12 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				DEFINES_MODULE = YES;
+				DERIVE_UIKITFORMAC_PRODUCT_BUNDLE_IDENTIFIER = YES;
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
+				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
@@ -1869,6 +1871,7 @@
 				);
 				SDKROOT = iphoneos;
 				SKIP_INSTALL = YES;
+				SUPPORTS_UIKITFORMAC = YES;
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
 			name = Debug;
@@ -1877,10 +1880,12 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				DEFINES_MODULE = YES;
+				DERIVE_UIKITFORMAC_PRODUCT_BUNDLE_IDENTIFIER = YES;
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
+				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
@@ -1888,6 +1893,7 @@
 				);
 				SDKROOT = iphoneos;
 				SKIP_INSTALL = YES;
+				SUPPORTS_UIKITFORMAC = YES;
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
 			name = Release;
diff --git a/Alamofire.xcodeproj/xcshareddata/xcschemes/Alamofire iOS.xcscheme b/Alamofire.xcodeproj/xcshareddata/xcschemes/Alamofire iOS.xcscheme
index e64738b..b41743d 100644
--- a/Alamofire.xcodeproj/xcshareddata/xcschemes/Alamofire iOS.xcscheme	
+++ b/Alamofire.xcodeproj/xcshareddata/xcschemes/Alamofire iOS.xcscheme	
@@ -40,31 +40,10 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      shouldUseLaunchSchemeArgsEnv = "NO"
       enableThreadSanitizer = "YES"
       codeCoverageEnabled = "YES"
-      onlyGenerateCoverageForSpecifiedTargets = "YES"
-      shouldUseLaunchSchemeArgsEnv = "NO">
-      <CodeCoverageTargets>
-         <BuildableReference
-            BuildableIdentifier = "primary"
-            BlueprintIdentifier = "F8111E3219A95C8B0040E7D1"
-            BuildableName = "Alamofire.framework"
-            BlueprintName = "Alamofire iOS"
-            ReferencedContainer = "container:Alamofire.xcodeproj">
-         </BuildableReference>
-      </CodeCoverageTargets>
-      <Testables>
-         <TestableReference
-            skipped = "NO">
-            <BuildableReference
-               BuildableIdentifier = "primary"
-               BlueprintIdentifier = "F8111E3D19A95C8B0040E7D1"
-               BuildableName = "Alamofire iOS Tests.xctest"
-               BlueprintName = "Alamofire iOS Tests"
-               ReferencedContainer = "container:Alamofire.xcodeproj">
-            </BuildableReference>
-         </TestableReference>
-      </Testables>
+      onlyGenerateCoverageForSpecifiedTargets = "YES">
       <MacroExpansion>
          <BuildableReference
             BuildableIdentifier = "primary"
@@ -86,8 +65,27 @@
             isEnabled = "YES">
          </EnvironmentVariable>
       </EnvironmentVariables>
-      <AdditionalOptions>
-      </AdditionalOptions>
+      <CodeCoverageTargets>
+         <BuildableReference
+            BuildableIdentifier = "primary"
+            BlueprintIdentifier = "F8111E3219A95C8B0040E7D1"
+            BuildableName = "Alamofire.framework"
+            BlueprintName = "Alamofire iOS"
+            ReferencedContainer = "container:Alamofire.xcodeproj">
+         </BuildableReference>
+      </CodeCoverageTargets>
+      <Testables>
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "F8111E3D19A95C8B0040E7D1"
+               BuildableName = "Alamofire iOS Tests.xctest"
+               BlueprintName = "Alamofire iOS Tests"
+               ReferencedContainer = "container:Alamofire.xcodeproj">
+            </BuildableReference>
+         </TestableReference>
+      </Testables>
    </TestAction>
    <LaunchAction
       buildConfiguration = "Debug"
@@ -108,8 +106,6 @@
             ReferencedContainer = "container:Alamofire.xcodeproj">
          </BuildableReference>
       </MacroExpansion>
-      <AdditionalOptions>
-      </AdditionalOptions>
    </LaunchAction>
    <ProfileAction
       buildConfiguration = "Release"
```

Source/CarthageKit/BuildOptions.swift
```diff
@@ -15,19 +15,24 @@ public struct BuildOptions {
 	/// Whether to use downloaded binaries if possible.
 	public var useBinaries: Bool
 
+	/// Whether to use xcframeworks or not
+	public var useXCFrameworks: Bool
```
mdiep: Can we do this automatically when using Xcode 11+?

Source/CarthageKit/BuildSettings.swift
```diff
@@ -216,7 +216,9 @@ public struct BuildSettings {
 			pathComponent = "\(configuration)\(effectivePlatformName)"
 		}
 
-		let path = (basePath as NSString).appendingPathComponent(pathComponent)
+		// TODO: could use self["variant"] maybe, is SDK is passed with variant
+		let isUIKitForMac = (self.supportsUIKitForMac.value ?? false) && (self["SDK_NAME"].map { $0.contains("macos") }.value ?? false)
+		let path = (basePath as NSString).appendingPathComponent(isUIKitForMac ? "\(pathComponent)-uikitformac" : pathComponent)
```
mdiep: It might be better to split Catalyst support into a separate PR?

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -295,14 +296,50 @@ extension URL {
 
 	/// Returns the first `URL` to match `<self>/Headers/*-Swift.h`. Otherwise `nil`.
 	internal func swiftHeaderURL() -> URL? {
-		let headersURL = self.appendingPathComponent("Headers", isDirectory: true).resolvingSymlinksInPath()
+
+		var libraryIdentifier: String = ""
+		var libraryPath: String = ""
```
mdiep: Let's make these `let`s and move the `""` values into an `else` below.

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -295,14 +296,50 @@ extension URL {
...
+		if self.pathExtension == "xcframework" {
+			let xcFrameworkInfo = Bundle(url: self)?
+				.infoDictionary
+				.flatMap(XCFrameworkInfo.init)
+			let firstLibrary =  xcFrameworkInfo?
```
mdiep: Extra space here after the `=`. 🙈

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -295,14 +296,50 @@ extension URL {
...
+				.availableLibraries
+				.first
+			libraryIdentifier = firstLibrary?
+				.identifier ?? ""
+			libraryPath = firstLibrary?.path ?? ""
+		}
+
+		let headersURL = self.appendingPathComponent(libraryIdentifier)
+			.appendingPathComponent(libraryPath)
```
mdiep: Since this logic to calculate the `headersURL` is repeated here and above, can we extract it to a reusable method?

Source/CarthageKit/Xcode.swift
```diff
@@ -678,18 +719,151 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
...
+							.xcFrameworkWrapperName
+							.map(folderURL.appendingPathComponent)
+
+							let adaptedURLs = Result<[URL], CarthageError>(Array(urls))
+
+							return SignalProducer(outputURL.fanout(adaptedURLs))
+								.flatMap(.merge) { createXCFramework($0.1, $0.0) }
+								.then(SignalProducer(result: outputURL))
+					}
+				}
```
mdiep: This code is hard to read in the diff on GitHub. It's long and has a lot of nesting. 🙈 Can we do something to clean it up?

Source/XCDBLD/BuildArguments.swift
```diff
@@ -74,6 +74,30 @@ public struct BuildArguments {
...
+			// Passing in -sdk macosx appears to break implicit dependency
+			// resolution (see Carthage/Carthage#347).
+			//
+			// Since we wouldn't be trying to build this target unless it were
+			// for macOS already, just let xcodebuild figure out the SDK on its
+			// own.
+			if sdk == .macOSX, let sdkIndex = args.firstIndex(where: { $0 == "-sdk"} ) {
+
+				if sdkIndex + 1 < args.count {
+					args.remove(at: sdkIndex) // remove -skd
```
mdiep: `sdk` typo here

Source/XCDBLD/BuildArguments.swift
```diff
@@ -74,6 +74,30 @@ public struct BuildArguments {
...
+					args.remove(at: sdkIndex) // remove paramter
```
mdiep: `parameter` type here

Source/XCDBLD/XCFrameworkInfo.swift
```diff
@@ -0,0 +1,101 @@
...
+		self.path = path
+		self.supportedArchitectures = supportedArchs
+		self.supportedPlatform = supportedPlatform
+		self.supportedSDK = sdk
+	}
+}
+
+fileprivate extension Platform {
+
+	static var aliases: [String : Platform] {
```
mdiep: This could be a `static let`

blender:
@mdiep thanks for the review. I will address the issues next week. 

There are a few more design choices that I would like to discuss. Will write on that later too.

nikita-leonov:
👋 What is an expected behavior for this implementation? I was trying to build `RxSwift` with using Carthage for this PR in a following way — `carthage build --platform iOS MacOS --create-xcframework`. The result was. Build folder only with iOS subfolder and there iOS specific binaries. No mention of MacOS neither on top level, nor in xcframework. In case of xcframework not used I do get multiple folders for multiple platforms.

nikita-leonov:
Thanks for providing such a detail clarification on how to build xcfraameworks and how to adjust existing frameworks to built with UIKit for mac. I used this branch to build Carthage, and to build multiple open source frameworks. So far I have seven built, and still many to go, but I quite confident that this change does the indented job well. Consider this a some extra validation aka QA effort.

Source/CarthageKit/BuildOptions.swift
blender: Are you suggesting to default automatically to xcframeworks if Xcode >= 11 ?

We can but, for macOS carthage will still produce a .framework (since packing just 1 framework has no effect)

Source/CarthageKit/BuildSettings.swift
blender: I'll check and see how much I have tied the two together

Source/CarthageKit/FrameworkExtensions.swift
blender: Done

ngocdaothanh:
README.md should also be updated to explain how to use this new feature?

Source/XCDBLD/XCFrameworkInfo.swift
```diff
@@ -0,0 +1,101 @@
...
+		["ios" : .iOS,
```
ngocdaothanh: There's error when compiling with Swift version 5.0.1, 10.15 Beta 2, Xcode 11 Beta 2:

```
Source/XCDBLD/XCFrameworkInfo.swift:41:3: error: type of expression is ambiguous without more context
                ["ios" : .iOS,
                ^~~~~~~~~~~~~~

make: *** [installables] Error 1
```

Source/CarthageKit/BuildSettings.swift
```diff
@@ -292,6 +299,11 @@ public struct BuildSettings {
 		return self["TARGET_BUILD_DIR"]
 	}
 
+	/// Attepts to determine if UIKit for Mac is supported
+	public var supportsUIKitForMac: Result<Bool, CarthageError> {
+		self["SUPPORTS_UIKITFORMAC"].map { $0 == "YES" }
+	}
```
ngocdaothanh: There's error when compiling with Swift version 5.0.1, 10.15 Beta 2, Xcode 11 Beta 2:

```
Source/CarthageKit/BuildSettings.swift:305:2: error: missing return in a function expected to return 'Result<Bool, CarthageError>'
        }
        ^

make: *** [installables] Error 1
```

blender:
Sure, but nothing of this if final yet.

ngocdaothanh:
@samrayner 

> xcodebuild -create-xcframework -framework ios.framework -framework sim.framework -framework uikitformac.framework -output output.xcframework

I'm using Xcode 11 beta 2. When I run the command `xcodebuild ...`, it shows:

```
xcodebuild: error: invalid option '-create-xcframework'
```

samrayner:
@ngocdaothanh Make sure your command line tools are set to 11 in Xcode preferences

opentable-echamberlain:
I'm running in an error with this branch.

I've updated AFNetworking and I can manually build an xcframework with uikitformac included.

Why I try using carthage, I get the following error:
`*** Building scheme "AFNetworking iOS" in AFNetworking.xcworkspace
Failed to write to /var/folders/4c/ppg18x1x43x_26_jqms4xfpm0000gn/T/carthage-xcframework-tmp.a9P0Mf/UIKitForMac/AFNetworking.framework: Error Domain=NSCocoaErrorDomain Code=260 "The file “AFNetworking.framework” couldn’t be opened because there is no such file." UserInfo={NSURL=file:///Users/ios/Library/Caches/org.carthage.CarthageKit/DerivedData/11.0_11M337n/AFNetworking/2fdbf3926bf548811b97afd3cadd699d221feb50/Build/Intermediates.noindex/ArchiveIntermediates/AFNetworking%20iOS/BuildProductsPath/Release-iphoneos/AFNetworking.framework, NSFilePath=/Users/ios/Library/Caches/org.carthage.CarthageKit/DerivedData/11.0_11M337n/AFNetworking/2fdbf3926bf548811b97afd3cadd699d221feb50/Build/Intermediates.noindex/ArchiveIntermediates/AFNetworking iOS/BuildProductsPath/Release-iphoneos/AFNetworking.framework, NSUnderlyingError=0x7ffdc24ae9f0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}`

If I look in the tail of the log, I see that the archive for uikitformac succeeded:
```
Touch /Users/ios/Library/Caches/org.carthage.CarthageKit/DerivedData/11.0_11M337n/AFNetworking/2fdbf3926bf548811b97afd3cadd699d221feb50/Build/Intermediates.noindex/ArchiveIntermediates/AFNetworking\ iOS/IntermediateBuildFilesPath/UninstalledProducts/macosx/AFNetworking.framework (in target: AFNetworking iOS)
    cd /Users/ios/Documents/opentable-iphone/Carthage/Checkouts/AFNetworking
    /usr/bin/touch -c /Users/ios/Library/Caches/org.carthage.CarthageKit/DerivedData/11.0_11M337n/AFNetworking/2fdbf3926bf548811b97afd3cadd699d221feb50/Build/Intermediates.noindex/ArchiveIntermediates/AFNetworking\ iOS/IntermediateBuildFilesPath/UninstalledProducts/macosx/AFNetworking.framework

** ARCHIVE SUCCEEDED **
```

I'm not all that familiar with Carthage's internals, any suggestions?

DavidBrunow:
Thanks for the work towards this!

I think it would be valuable to be able to have frameworks for multiple platforms in the same `.xcframework`. My reasoning is that I could have a framework that supports watchOS, tvOS, and iOS that then needs to link against a framework that also supports watchOS, tvOS, and iOS. The straightforward way that I've found to do this linking is through Xcode's user interface through which I cannot add multiple `.xcframework`s of the same name. Maybe I'm trying to swim upstream by having one target that supports multiple platforms?

dissimiral:
Hey guys, I faced with the same problem as @opentable-echamberlain did.
Seems it's a mistake. 
I don't know much about Carthage's internals, but I did a few small(dirty) changes just to make it work.

<details><summary>PATCH</summary>
<p>
diff --git a/Source/CarthageKit/BuildSettings.swift b/Source/CarthageKit/BuildSettings.swift
index d951d885..4335e4b5 100644
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -16,6 +16,8 @@ public struct BuildSettings {
 
 	/// The designated xcodebuild action if present.
 	public let action: BuildArguments.Action?
+    
+    public var isUIKitForMac: Bool = false
 
 	internal init(
 		target: String,
@@ -192,8 +194,13 @@ public struct BuildSettings {
 
 		let basePath = "ArchiveIntermediates/\(schemeOrTarget)/BuildProductsPath"
 		let pathComponent: String
-
-		if
+        
+        let isUIKitForMac = (self.supportsUIKitForMac.value ?? false) && self.isUIKitForMac
+        
+        if isUIKitForMac {
+            let r2 = self["CONFIGURATION"]
+            pathComponent = r2.value!
+        } else if
 			let buildDir = self["BUILD_DIR"].value,
 			let builtProductsDir = self["BUILT_PRODUCTS_DIR"].value,
 			builtProductsDir.hasPrefix(buildDir)
@@ -217,7 +224,6 @@ public struct BuildSettings {
 		}
 
 		// TODO: could use self["variant"] maybe, is SDK is passed with variant
-		let isUIKitForMac = (self.supportsUIKitForMac.value ?? false) && (self["SDK_NAME"].map { $0.contains("macos") }.value ?? false)
 		let path = (basePath as NSString).appendingPathComponent(isUIKitForMac ? "\(pathComponent)-uikitformac" : pathComponent)
 		return .success(path)
 	}
diff --git a/Source/CarthageKit/Xcode.swift b/Source/CarthageKit/Xcode.swift
index 55ed0144..fffabaf1 100644
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -309,14 +309,20 @@ internal enum PackageType: String {
 /// If this built product has any *.bcsymbolmap files they will also be copied.
 ///
 /// Returns a signal that will send the URL after copying upon .success.
-private func copyBuildProductIntoDirectory(_ directoryURL: URL, _ settings: BuildSettings) -> SignalProducer<URL, CarthageError> {
-	let target = settings.wrapperName.map(directoryURL.appendingPathComponent)
-	return SignalProducer(result: target.fanout(settings.wrapperURL))
+private func copyBuildProductIntoDirectory(
+    _ directoryURL: URL,
+    _ settings: BuildSettings,
+    isUIKitForMac: Bool
+    ) -> SignalProducer<URL, CarthageError> {
+    var _settings = settings
+    _settings.isUIKitForMac = isUIKitForMac
+	let target = _settings.wrapperName.map(directoryURL.appendingPathComponent)
+	return SignalProducer(result: target.fanout(_settings.wrapperURL))
 		.flatMap(.merge) { target, source in
 			return copyProduct(source.resolvingSymlinksInPath(), target)
 		}
 		.flatMap(.merge) { url in
-			return copyBCSymbolMapsForBuildProductIntoDirectory(directoryURL, settings)
+			return copyBCSymbolMapsForBuildProductIntoDirectory(directoryURL, _settings)
 				.then(SignalProducer<URL, CarthageError>(value: url))
 		}
 }
@@ -540,7 +546,7 @@ private func mergeBuildProducts(
 	simulatorBuildSettings: BuildSettings,
 	into destinationFolderURL: URL
 ) -> SignalProducer<URL, CarthageError> {
-	return copyBuildProductIntoDirectory(destinationFolderURL, deviceBuildSettings)
+    return copyBuildProductIntoDirectory(destinationFolderURL, deviceBuildSettings, isUIKitForMac: false)
 		.flatMap(.merge) { productURL -> SignalProducer<URL, CarthageError> in
 			let executableURLs = (deviceBuildSettings.executableURL.fanout(simulatorBuildSettings.executableURL)).map { [ $0, $1 ] }
 			let outputURL = deviceBuildSettings.executablePath.map(destinationFolderURL.appendingPathComponent)
@@ -668,7 +674,7 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 			case 1:
 				return build(sdk: sdks[0], with: buildArgs, in: workingDirectoryURL)
 					.flatMapTaskEvents(.merge) { settings in
-						return copyBuildProductIntoDirectory(settings.productDestinationPath(in: folderURL), settings)
+                        return copyBuildProductIntoDirectory(settings.productDestinationPath(in: folderURL), settings, isUIKitForMac: false)
 					}
 
 			case 2:
@@ -771,7 +777,7 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 					case let .success(d):
 						return SignalProducer(d).flatMap(.merge) { target, settings in
 							FileManager.default.reactive.createTemporaryDirectoryWithTemplate(tmpProductDirTemplate).flatMap(.merge) { directoryURL in
-								copyBuildProductIntoDirectory(settings.productDestinationPath(in: directoryURL.appendingPathComponent("iOS")), settings)
+								copyBuildProductIntoDirectory(settings.productDestinationPath(in: directoryURL.appendingPathComponent("iOS")), settings, isUIKitForMac: false)
 								.zip(with: SignalProducer<String, CarthageError>(value: target))
 								.collect()
 								.flatMap(.merge) { iphoneTargetsDestinationULRs in
@@ -785,14 +791,14 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 											return SignalProducer(value: .standardError(data))
 										case let .success(m):
 											return SignalProducer(m).flatMap(.merge) { target, settings in
-												copyBuildProductIntoDirectory(settings.productDestinationPath(in: directoryURL.appendingPathComponent("UIKitForMac")), settings)
+												copyBuildProductIntoDirectory(settings.productDestinationPath(in: directoryURL.appendingPathComponent("UIKitForMac")), settings, isUIKitForMac: true)
 													.zip(with: SignalProducer<String, CarthageError>(value: target))
 												}
 												.collect()
 												.flatMap(.merge) { macOSTargetsDestinationULRs in
 													return iPhoneSimulatorSettingsProducer.flatMapTaskEvents(.concat) { s in
 														return SignalProducer(s).flatMap(.merge) { (target, settings) -> SignalProducer<(URL, String), CarthageError>  in
-															copyBuildProductIntoDirectory(settings.productDestinationPath(in: directoryURL.appendingPathComponent("iOSSimulator")), settings)
+															copyBuildProductIntoDirectory(settings.productDestinationPath(in: directoryURL.appendingPathComponent("iOSSimulator")), settings, isUIKitForMac: false)
 															.zip(with: SignalProducer<String, CarthageError>(value: target))
 															}
 															.collect()

</p>
</details>

khatchik:
Guys, any status on this PR? Are we planning to merge this soon to master?

tmspzz:
I'm working on it when I can. The most difficult part is simplifying this https://github.com/Carthage/Carthage/pull/2801#discussion_r293352339 

I would be glad to spend all my time working on Carthage if I could.

DavidBrunow:
I don't think it makes sense to merge this to master until Xcode 11 is out of beta. The latest beta changed the naming convention of "uikitformac" to "maccatalyst" which breaks this implementation and we don't know what other changes are coming in the future.

nancy-gomez:
Catalina GM and Xcode 11 GM have been out, and it seems quite a few PRs exist to enable the ability to build .xcframeworks for catalyst support. So, curious to know what the hold up is? 🤔😢

boobesh-s:
When this support will be available?

ddaddy:
Is there still no Mac Catalyst support in Carthage?

eltonf:
I've been begrudgingly moving all of my Carthage frameworks to CocoaPods and to SPM when available. I love Carthage but not supporting Mac Catalyst is becoming a non-starter for me.

chrisballinger:
For anyone else who was watching this issue, it looks like xcframework work has continued in this other PR: https://github.com/Carthage/Carthage/pull/2881

PR #2802, opened 2019-06-05
Feature/skip archive and custom cartfile

canius:
Features:
1. pass --skip-archive to skip archive framework during build/bootstrap/update if SDK is Device. Speed up 30% build time when develop.
2. pass --custom XXX to custom Cartfile/Cartfile.private/Cartfile.resolved/Carthage Name. Useful when have multi branch eg: develop and master , avoid merge conflict

### DIFF
--- a/Source/CarthageKit/BuildOptions.swift
+++ b/Source/CarthageKit/BuildOptions.swift
@@ -16,6 +16,8 @@ public struct BuildOptions {
 	public var useBinaries: Bool
     /// Skip archive when build Device SDK
     public var skipArchive: Bool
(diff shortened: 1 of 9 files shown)


blender:
Hi,

Thanks for the pull request. Can you please open an issue to discuss what is that you want to improve/change? 

It’s never a good idea to open a PR first without checking if this is a direction the project wants to take.

ikesyo:
> pass --skip-archive to skip archive framework during build/bootstrap/update if SDK is Device. Speed up 30% build time when develop.

`archive` action is import because of #2056 #2158 (and correct bitcode support).

PR #2820, opened 2019-07-14
WIP - Create single XCFramework for all platforms

DavidBrunow:
This is based upon @blender's PR for adding support for XCFrameworks #2801 

Current known issues that need to be fixed:

- [ ] Current implementation only builds a single XCFramework with all combined platforms and does not currently support building frameworks split by platform
- [ ] Current implementation must have a derived data directory set in order to avoid issues with builds locking files on each other

I made a design decision to output the combined XCFramework to the `Carthage/Build` folder and would appreciate input on that decision.

### DIFF
--- a/Source/carthage/Update.swift
+++ b/Source/carthage/Update.swift
@@ -65,15 +65,18 @@ public struct UpdateCommand: CommandProtocol {
 			let buildDescription = "skip the building of dependencies after updating\n(ignored if --no-checkout option is present)"
 
 			let dependenciesUsage = "the dependency names to update, checkout and build"
+			let defaultValue: String? = nil
 
-			return curry(Options.init)
-				<*> mode <| Option(key: "checkout", defaultValue: true, usage: "skip the checking out of dependencies after updating")
-				<*> mode <| Option(key: "build", defaultValue: true, usage: buildDescription)
-				<*> mode <| Option(key: "verbose", defaultValue: false, usage: "print xcodebuild output inline (ignored if --no-build option is present)")
-				<*> mode <| Option(key: "log-path", defaultValue: nil, usage: "path to the xcode build output. A temporary file is used by default")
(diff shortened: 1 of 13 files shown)


DavidBrunow:
FYI I am still working on this and making it work for my specific use case but won't have bandwidth to generalize the implementation to work for everyone until after iOS 13 comes out.

Source/carthage/Update.swift
```diff
@@ -65,15 +65,18 @@ public struct UpdateCommand: CommandProtocol {
 			let buildDescription = "skip the building of dependencies after updating\n(ignored if --no-checkout option is present)"
 
 			let dependenciesUsage = "the dependency names to update, checkout and build"
+			let defaultValue: String? = nil
```
jdhealy: @DavidBrunow — as you're probably already aware (but hopefully positively expecting), the merging of https://github.com/Carthage/Carthage/pull/2859 into `master` will cause merge conflicts with this hunk.

And a way over due: major thanks for working on this 🙂👍

Aranoledur:
@DavidBrunow @tmspzz Hey, guys! Can we set up goals for this feature? Our company need this feature in release and ok for 1 person working on it fulltime

tmspzz:
@Aranoledur let me take put something together

DavidBrunow:
@tmspzz Who would make the decision that having the ability to create a single XCFramework for all platforms is something that Carthage should do? I want to make sure we have buy-in that that is valuable.

tmspzz:
@DavidBrunow long story short, it's desirable to support both. Before you continue though, please allow me some more time to post a "roadmap" for what the goals are for Carthage in the near future. I need a run it by @Carthage/carthage and then will share it and gather feedback

tmspzz:
@Aranoledur please see this: https://github.com/Carthage/Carthage/issues/2890

DavidBrunow:
Still working on this...

PR #2823, opened 2019-07-14
Fix building with SPM

tmspzz:
# What happened?
[swift-llbuild](https://github.com/apple/swift-llbuild.git)'s Package.swift was updated to Swift 5. This in combination with `swift build` __automatically updating dependencies__ prevents carthage from building with SwiftPM.

# Solution 

First, to stop SwiftPM from updating dependencies, it is necessary to add `--skip-update`.

Unfortunately however that's not enough for carthage to build. SwiftPM itself [adds a very nasty dependency to swift-llbuild master](https://github.com/apple/swift-package-manager/blob/916450f7e8a5aef13630edf9e6fc0b16ed9f5f65/Package.swift#L262)

To prevent that from happening it's enough to set `SWIFTPM_BOOTSTRAP` environment variable, however this completely removes swift-llbuild from the dependency graph.

To keep swift-llbuild in the dependency graph it's necessary to add it explicitly as a dependency.

However, that also is not enough as SwiftPM won't build anymore because of missing dependency on swift-llbuild. To prevent this, it is first necessary to ask SwiftPM to resolve the dependencies at the right version (without updating), then find the checkout directory of swift-llbuild and finally pass the location of it's modulemap to all swiftc commands.

#### Personal remarks

I would have not expected any of this to happen, since builds are supposed to be reproducible. I believe there are serious issue with how SwiftPM behaves, [some clarification is needed](https://forums.swift.org/t/about-reproducible-builds/26957).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
       language: objective-c
       osx_image: xcode10.1
       script:
-        - make test
+        - make clean && SWIFTPM_BOOTSTRAP=true make test
       env: JOB=CI_TEST_10_1
     - os: osx
       language: objective-c
@@ -22,7 +22,7 @@ matrix:
       script:
         - brew uninstall carthage
         - HOMEBREW_NO_AUTO_UPDATE=1 brew install bats
-        - make install
+        - make clean && SWIFTPM_BOOTSTRAP=true make install
         - bats IntegrationTests
       env:
         - JOB=CI_INTEGRATION_TESTS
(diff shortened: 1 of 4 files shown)


mdiep:
This is unfortunate. Am I reading correctly that this can be removed if we upgrade to a semantic version of SwiftPM?

tmspzz:
It seems so, but tagged version 0.4.0 displays another set of issues.

PR #2830, opened 2019-07-23
Remove pipe in carthage outdated step

nteissler:
The snippet in **README.md** to add to the build phase is wrong by one pipe (`|`)

I'm not sure what a pipe does in this scenario, but it is not the right thing. Bash seems to take `echo "foo" | 2>/dev/null` and alias 2 to be a `cd` command. If one wants to redirect `stderr` to **/dev/null**, one only needs `2>/dev/null`. Before this change. all the output from the outdated dependencies step is swallowed and nothing gets added to Xcode. After, the warnings will show up in the console, but no network connection doesn't cause build failures (https://github.com/Carthage/Carthage/commit/70f5b2c51c78fb5440155346063f792347c893d9#diff-04c6e90faac2675aa89e2176d2eec7d8)


tmspzz:
Thanks! 🎉

jdhealy:
Thanks for looking into this! 🎉

PR #2842, opened 2019-08-04
Improve submodule performance when --use-submodule

CosynPa:
Try to clone or fetch submodules from the Carthage cache, instead of simply running a `git submodule update --init --recursive`.

A `git submodule init` operation seems unnecessary when there is a `git submodule add` before it, so I removed it.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -866,7 +866,10 @@ public final class Project { // swiftlint:disable:this type_body_length
 			if let submodule = submodule {
 				// In the presence of `submodule` for `dependency` — before symlinking, (not after) — add submodule and its submodules:
 				// `dependency`, subdependencies that are submodules, and non-Carthage-housed submodules.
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/Project.swift
```diff
@@ -832,8 +832,6 @@ public final class Project { // swiftlint:disable:this type_body_length
 		return createVersionFileForCommitish(commitish, dependencyName: projectName, buildProducts: frameworkURLs, rootDirectoryURL: self.directoryURL)
 	}
 
-	private let gitOperationQueue = SerialProducerQueue(name: "org.carthage.Constants.Project.gitOperationQueue")
```
tmspzz: Why was this queue removed?

CosynPa: No longer need that queue since `checkoutDependency` operations are already serialized.

CosynPa:
Hi

Issue #2839, opened 2019-08-02
Contributing: Can't initialize Xcode project


relikd:
I've cloned this repository and did run `make clean && SWIFTPM_BOOTSTRAP=true make xcodeproj`. All dependencies get resolved and the Xcode project file is created. However, if I open the Xcode project and try to build it, `SPMLLBuild/llbuild.swift` will fail with:

```
@_exported import llbuildSwift         <--- Missing required module 'llbuild'
```

relikd:
Updating the `swift-package-manager` dependency to `swift-4.2-branch` will fix these build issues and the project can be run. However, in doing so, other issues arise when building the release version with `make install`.

olejnjak:
Yep, same issue here, wanted to try to fix #2844 but unable to build the project from `master` branch.

tmspzz:
That's curious. I don't have a solution for this at the moment.

Off the top of my head I would probably add to the search paths the location of `llbuild` (see .travis.yml)

relikd:
Tried five different search paths, no luck.

My workaround so far:

1. Change the branch to 4.2 and update all `SPMUtility` header files (as seen [here](https://github.com/Carthage/Carthage/pull/2838/files)).
2. Use this environment for development (it compiles nicely in Xcode)
3. Copy all changes manually to the `master` branch (without the 4.2 branch modifications)
4. Run `make clean && SWIFTPM_BOOTSTRAP=true make test` 
5. Run `bats IntegrationTests`

manicmaniac:
I'm facing the same issue after posting https://github.com/Carthage/Carthage/pull/2853.

jdhealy:
Fixed in #2863.

Issue #2834, opened 2019-07-29
Use Carthage with Quicklook plugins of type .qlgenerator


relikd:
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.33.0

I was wondering if Carthage does support Quicklook generators (`.qlgenerator`) or only `.framework`?  
I tried to replace all occurences of `.qlgenerator` with `.framework` in the shared scheme. No luck. I always get:

> Dependency 'X' has no shared framework schemes

Similar to Frameworks, I would like to code-sign and copy the result to a subfolder of the target application. You can test with this repository: [QLOPML](https://github.com/relikd/QLOPML/).  
**Note:** the shared scheme is not pushed in the latest commit since I couldn't get it to work. But I've tried with local pushes and Cartfile: `git "file:///.../QLOPML" "master"`.

tmspzz:
Cool project! 👍 

I'm afraid only `.framework` with Mach-O dynamic or static library are supported

relikd:
Is this something that could be added in the future? If its not too time-consuming …

tmspzz:
@relikd you can make PR to add support.

relikd:
Ok, I'll look into that. Any suggestion where I should start looking?

But I can't build the project. `make xcodeproj` will result in `error: dependency graph is unresolvable`. The build directory `.build/repositories` contains 11 repos but `.build/checkout` is empty. Also, the fetched repositories are not git work trees but rather the git index, with folders `branches`, `HEAD`, and `objects`, etc.

kenji21:
same issue (trying to update my PR) :

```
$ rm -rf .build/ ; make
swift package --skip-update resolve
Fetching https://github.com/thoughtbot/Curry.git
Fetching https://github.com/Quick/Quick.git
Fetching https://github.com/ReactiveCocoa/ReactiveSwift.git
Fetching https://github.com/Carthage/Commandant.git
Fetching https://github.com/jdhealy/PrettyColors.git
Fetching https://github.com/antitypical/Result.git
Fetching https://github.com/apple/swift-package-manager.git
Fetching https://github.com/Quick/Nimble.git
Fetching https://github.com/mdiep/Tentacle.git
Fetching https://github.com/Carthage/ReactiveTask.git
Fetching https://github.com/apple/swift-llbuild.git
error: dependency graph is unresolvable; found these conflicting requirements:

Dependencies:
[...]

PR #2854, opened 2019-08-25
Use libSwiftPM 0.4.0 and Xcode 10.2 to avoid build errors

elliottwilliams:
I've been working on Carthage and have been hit by the SPM resolution errors described in #2839, #2834, and #2823. Invoking `make install` the same way .travis.yml does (like @tmspzz suggested in https://github.com/Carthage/Carthage/issues/2834#issuecomment-516455570) works fine, but I wanted to see if this could be avoided altogether.

This change updates Carthage to require SwiftPM version `0.4.0` instead of the previous snapshot tag. AFAICT, 0.4.0 is an official "package dependency" release intended to be used by other packages. The important difference is that 0.4.0 has a [version requirement on llbuild](https://github.com/apple/swift-package-manager/commit/8656a25cb906c1896339f950ac960ee1b4fe8034) instead of pinning to llbuild's `master`.

This means that Carthage can continue to require llbuild without breaking every time Apple pushes a new commit to master, and that checkouts of Carthage's dependencies will always be reproducible.

SPM 0.4.0 requires the Swift 5.1 / Xcode 10.2 toolchain, so I updated travis to use 10.2 images. If there's a reason Carthage needs to support building on Swift 4.2 / Xcode 10.1, then we won't be able to this :(


elliottwilliams:
Hmm, looks like integration test failures will require changes upstream in https://github.com/ikesyo/CarthageCopyFrameworksFixture.git. The project committed is set to `SWIFT_VERSION = 3.0`, which will need to be bumped to a version the Swift 5.1 compiler supports.

elliottwilliams:
In #2863, SwiftPM-the-library was removed altogether to avoid these errors, and @jdhealy clarified that Carthage _has_ to support building on Swift 4.2.x as a consequence of support for brew on High Sierra. Closing this for now, but could easily be reopened later.

PR #2863, opened 2019-09-07
Remove — as a library dependency — `SwiftPM` and `llbuild` and reinstate `struct Carthage.SemanticVersion`

jdhealy:
A [bug in the-tool-SwiftPM's package resolution](https://github.com/apple/swift-package-manager/pull/2197) resulted in disregard for the (crucial) ‘resolved file’ when resolving branch-based dependencies.

Which would put us in the position of needing any commit of `apple/swift-package-manager` that didn't specify its dependency on `apple/swift-llbuild` in the branch-based style.

Also criteria for the above, [working Swift 4.2.X manifests and compilation](https://forums.swift.org/t/about-reproducible-builds/26957/10) _and_ working Swift 5.X manifests and compilation.

No such commit of `apple/swift-package-manager` could be found.

In addition, some confusion around [`llbuild`'s `sqlite3` linkage](https://github.com/apple/swift-llbuild/commit/f1c9ad9a253cdf1aa89a7f5c99c30b4513b06ddb#diff-37ca2dd15ca0f6b1b49e78db084ef5b9L5-L10) made criteria even more confusing.

#### Notes:

- Some initializers and methods on `struct Carthage.SemanticVersion` have differences from the previous incarnation found in v0.33.0.
- Reinstate `struct Carthage.SemanticVersion` in all callsites in codebase and tests, removing `SPMUtility.Version`.
- No longer necessary to `import struct Foundation.URL` with the removed import of SPMUtility.
- Makefile removes complications which previously supported `llbuild`'s `sqlite3` linkage.
- Working Swift 4.2.X compilation is necessary to support building for [macOS High Sierra](https://en.wikipedia.org/wiki/MacOS_High_Sierra) on [Homebrew bottling infrastructure](https://github.com/Homebrew/brew/blob/7dd0466/Library/Homebrew/os/mac/xcode.rb#L18).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ clean:
 test:
 	$(RM_SAFELY) ./.build/debug/CarthagePackageTests.xctest
 	swift package --skip-update resolve
-	swift build --skip-update --build-tests -Xswiftc -suppress-warnings -Xswiftc -Xcc -Xswiftc -fmodule-map-file=$(PWD)/`find .build/checkouts -name "swift-llbuild.git*"`/products/libllbuild/include/module.modulemap
+	swift build --skip-update --build-tests -Xswiftc -suppress-warnings
 	$(CP) -R Tests/CarthageKitTests/Resources ./.build/debug/CarthagePackageTests.xctest/Contents
 	$(CP) Tests/CarthageKitTests/fixtures/CartfilePrivateOnly.zip ./.build/debug/CarthagePackageTests.xctest/Contents/Resources
 	script/copy-fixtures ./.build/debug/CarthagePackageTests.xctest/Contents/Resources
@@ -53,7 +53,7 @@ test:
 
 installables:
 	swift package --skip-update resolve
-	swift build $(SWIFT_BUILD_FLAGS) -Xswiftc -Xcc -Xswiftc -fmodule-map-file=$(PWD)/`find .build/checkouts -name "swift-llbuild.git*"`/products/libllbuild/include/module.modulemap
+	swift build $(SWIFT_BUILD_FLAGS)
 
 package: installables
 	$(MKDIR) "$(CARTHAGE_TEMPORARY_FOLDER)$(BINARIES_FOLDER)"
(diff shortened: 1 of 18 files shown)


elliottwilliams:
FWIW, I've been working on making Carthage work with SwiftPM's resolver as means of improving resolver performance on large dependency graphs. [The branch][1] should be ready for a PR once I write some tests demonstrating its performance characteristics, but if Carthage is intentionally moving away from SwiftPM, it sounds like borrowing their resolver won't be possible in the future.

[1]: https://github.com/Carthage/Carthage/compare/master...elliottwilliams:spm_resolver_xcode_10_2

jdhealy:
@elliottwilliams, exciting to see! Thanks a million times over for starting to run with that! ✨ 

We can definitely revisit reinstating SwiftPM-as-a-library-dependency down the road.

With Homebrew's criteria [necessitating us to build with Swift 4.2.X in order to bottle binary releases for High Sierra](https://github.com/Homebrew/brew/blob/7dd0466/Library/Homebrew/os/mac/xcode.rb#L18) — and High Sierra being just one behind the latest macOS release — I'm not comfortable dropping support for High Sierra until we see some significant numbers from [Catalina](https://en.wikipedia.org/wiki/MacOS_Catalina).

〜 Might be worthy of exploration into additional features via cross-process communication with some companion binary — one that wouldn't necessarily be constrained to compiling under Swift 4.2.X.

〜 Or maybe even a split-off single-binary stripped-down/beefed-up variant of Carthage under a different name, …I'm open to discussion.

jdhealy:
Discussion on the long-term state of this is extremely welcome. Please continue to post here. 🙏 

The approach in this PR can always be adjusted or reverted.

Merged this into `master` in pursuit of getting `brew install --HEAD master` back into working shape before this afternoon.

Please, feel free to continue to post any comments or update here 👍

elliottwilliams:
@jdhealy, sounds good! Definitely understand that this doesn't necessarily affect long-term goals for Carthage + SwiftPM

> I'm not comfortable dropping support for High Sierra until we see some significant numbers from Catalina.

Since Catalina will be out soon, it sounds like this may not impact me that much — I can continue to try to get a PR up which reverts this one and is based off #2854 (which drops support for Swift 4.2.x).

> Or maybe even a split-off single-binary stripped-down/beefed-up variant of Carthage under a different name, …I'm open to discussion.

Exciting! Yeah, happy to discuss with you all if there are are other directions I can help take this ✨

PR #2864, opened 2019-09-10
Swift Package Manager: Suppress `sandbox-exec` under Homebrew.

jdhealy:
Crucial to enable Homebrew installs: due to Homebrew's own use of sandboxing.

Travis: remove `make clean` and `SWIFTPM_BOOTSTRAP` (`var` for certain versions of SwiftPM-as-a-library) — no longer necessary as of Carthage/Carthage@9b009cb408afb08e3b32d234adb7a8ee5f1be627.

Restores previous behavior.

Partial revert of Carthage/Carthage@60aaf683dc4e5d446c2b64704e7c02b80de45ba3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,7 @@ OUTPUT_PACKAGE=Carthage.pkg
 CARTHAGE_EXECUTABLE=./.build/release/carthage
 BINARIES_FOLDER=/usr/local/bin
 
-PWD := $(shell pwd)
-SWIFT_BUILD_FLAGS=--skip-update --configuration release -Xswiftc -suppress-warnings
+SWIFT_BUILD_FLAGS=--configuration release -Xswiftc -suppress-warnings
 
 SWIFTPM_DISABLE_SANDBOX_SHOULD_BE_FLAGGED:=$(shell test -n "$${HOMEBREW_SDKROOT}" && echo should_be_flagged)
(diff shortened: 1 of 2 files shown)


jdhealy:
Sorry to merge this without approval (as of current) — but, it's in pursuit of getting `brew install --HEAD master` back into working shape before this afternoon. (Hopefully that justifies it).

〜 The approach in this PR can always be adjusted or reverted.

jdhealy:
With this merged, `brew install --HEAD master` should be back in working order 🎉

Makefile
```diff
@@ -9,8 +9,7 @@ OUTPUT_PACKAGE=Carthage.pkg
 CARTHAGE_EXECUTABLE=./.build/release/carthage
 BINARIES_FOLDER=/usr/local/bin
 
-PWD := $(shell pwd)
-SWIFT_BUILD_FLAGS=--skip-update --configuration release -Xswiftc -suppress-warnings
+SWIFT_BUILD_FLAGS=--configuration release -Xswiftc -suppress-warnings
```
tmspzz: I would keep --skip-update

jdhealy: Question (if anybody knows): what does that do if a previous resolution doesn’t exist on-disk?

tmspzz: I'm guessing it would kick of a resolution.

Leaving this here for future reference if needed: https://forums.swift.org/t/about-reproducible-builds/26957/5?u=tmpz

PR #2868, opened 2019-09-12
Gaustin/mac catalyst

atlassian-gaustin:
This change creates a new platform-"macCatalyst"-which will build targets using the MacCatalyst SDK. This builds _most_ iOS libraries for Catalyst without modification.

This does revert the intermediate fix #2823 which used non-public SHAs.


nighthawk:
Thank you very much for this, @atlassian-gaustin. I gave this a try and it indeed works mostly. Are there any instructions for what needs to be done to fix up iOS libraries where this doesn't work out-of-the-box?

In my testing, the main error that I faced was "Fatal error: No SDKs found for scheme X" in Source/CarthageKit/Xcode.swift, line 639. In the case of RxSwift I could address this by setting a "Base SDK" in the checkout before calling `carthage build --platform macCatalyst`. For RxDataSources I have the same error, but couldn't yet figure out what to do about it as it does have an SDK set.

atlassian-gaustin:
@nighthawk That has been my experience as well. I suspect that there is a stealth path or option in `xcodebuild` to get *everything* building for macCatalyst that builds for iOS.
I did have to get carthage to ignore serious errors returned from `xcodebuild` when It would hit some embedded demo projects that carthage should have ignored.
I suspect that much of this code change could be folded into the `xcframework` PR to get it working. Xcframeworks would be a far more preferable solution, so you'd only need one path to use the proper framework.

nighthawk:
@atlassian-gaustin I've bumped into another in a second app when using this:

The issue I'm seeing is that the `/usr/local/bin/carthage copy-frameworks` build step doesn't like these frameworks. It always throws errors like "MyFramework.framework: bundle format is ambiguous (could be app or framework)" when trying to copy one of these frameworks from the `$(SRCROOT)/Carthage/Build/macCatalyst/` folder. Did you get this to work?

What's curious about this is that in the first app where I used your fork successfully, I didn't see this issue. I had forgotten to fix up that build step when building for Mac Catalyst. Yet, it still works, even though it's copying the iOS frameworks rather than the macCatalyst frameworks. Suffice to say, I'm very puzzled by this!

nighthawk:
After a few different attempts, I've landed on a process to get this all working pretty reliably for me – included archiving and signing with a Developer ID. If anyone else is struggling with this, I've put the [steps on my blog](https://adrian.schoenig.me/blog/2019/10/21/carthage-and-catalyst/).

atlassian-gaustin:
I've done a bit more research, and it appears that the '-destination' platform really seems to confuse `xcodebuild ... showBuildSettings`

nighthawk:
FWIW, I've been using this for a couple months now and it's working fine as long as you are aware of a few gotchas mentioned above. Thanks, Glenn!

NachoSoto:
@atlassian-gaustin did you consider this approach with the new `XCFramework`s support in Carthage?
I tried rebasing your branch to `master` but unfortunately there's lots of conflicts that aren't trivial to solve.

atlassian-gaustin:
This was done so that we (and others) could get working with Catalyst while the Carthage team decided what they wanted to do. Now that Carthage has XCFramework support, this is no longer needed.

NachoSoto:
But the xcframeworks don't contain a Catalyst slice?

atlassian-gaustin:
My understanding is that Catalyst is an iOS SDK target with an i386 processor target.

PR #2872, opened 2019-09-23
Add `carthage cleanup` command again

sidepelican:
Base issue: Cleanup command #902
Base PR: #2333

I really wanted to use the `cleanup` command, so I implemented the remaining tasks


giginet:
Could you add new subcommand to Shell completions?

https://github.com/Carthage/Carthage/tree/master/Source/Scripts

PR #2877, opened 2019-09-24
Add alternative content type for zip archives

MatkovIvan:
Sometimes GitHub uses `application/x-zip-compressed` content type for zip archives. This type should be allowed as it's just a zip file.

PS: It was really tricky to figure out what was wrong with the release.

### DIFF
--- a/Source/CarthageKit/Constants.swift
+++ b/Source/CarthageKit/Constants.swift
@@ -102,6 +102,6 @@ public struct Constants {
 
 		/// MIME types allowed for GitHub Release assets, for them to be considered as
 		/// binary frameworks.
-		public static let binaryAssetContentTypes = ["application/zip", "application/octet-stream"]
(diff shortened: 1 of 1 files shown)


tmspzz:
`application/octet-stream` is [arbitrary data](https://www.iana.org/assignments/media-types/application/octet-stream). I don't think we want that.

MatkovIvan:
@tmspzz I've added `application/x-zip-compressed`, not `application/octet-stream`.
It was already there.

tmspzz:
You are right, my bad.

tmspzz:
@MatkovIvan thanks for the contribution 🎉

ikesyo:
`octet-stream` was added in #2448.

tmspzz:
Yes, seems I have a split personality. I guess I was much more into the issue then than now.

PR #2878, opened 2019-09-24
Fixup `cleanup` command

sidepelican:
I found some mistakes in #2872.

1.  It causes error when there is a lack of a `.framework` . This situation may often caused when the user uses `carthage build --platform XXX`  

2. Performance issue. `BuildSettings.load` can run in parallel.


tmspzz:
Thanks! 👏

Issue #2879, opened 2019-09-26
git@<ip>:<name> Name of repo is not extracted properly.  Build fails (Xcode 11)


kamzab95:
* carthage install method: [ ] .pkg, [x] homebrew, [x] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`:  0.33.0
* `xcodebuild -version`:  Xcode 11.0 Build version 11A420a
* Are you using --no-build? no
* Are you using --no-use-binaries? no
* Are you using --use-submodules? no
* Are you using --cache-builds? no
* Are you using --new-resolver? no 

**Cartfile**
```
git "git@192.168.1.7:myFramework" == 1.0.0
```

**Carthage Output**
```
Task failed with exit code 65:
```
**Log**
```
error: error reading dependency file '...11.0_11A420a/git@192.168.1.7:myFramework/v2.2.0/Build/...' unexpected character in prerequisites at position 330
```

**Actual outcome**
Carthage did fail to build project.

**Expected outcome**
Carthage should build the project.

I was able to find what is causing this issue. 
When repo is defined as in the example, the whole url is used as the project name, it means that whole "git@192.168.1.7:myFramework" will be placed inside the path.
With the new toolchain there is some issue when it comes to colon ":" sign. If colon exists in the path the build will fail.

PR #2880, opened 2019-09-26
FIX: Colon cause the build to fail #2879

kamzab95:
When extracting the name from the repository url beside "/" treat ":" as separator
Added test for this specific case i.e. 
Cartfile:
git "git@192.168.1.2:myFramework" == 1.0.0

Expected result:
name == "myFramework"

### DIFF
--- a/Source/CarthageKit/GitURL.swift
+++ b/Source/CarthageKit/GitURL.swift
@@ -47,7 +47,7 @@ public struct GitURL {
 
 	/// The name of the repository, if it can be inferred from the URL.
 	public var name: String? {
-		let components = urlString.split(omittingEmptySubsequences: true) { $0 == "/" }
+        let components = urlString.split(omittingEmptySubsequences: true) { $0 == "/" || $0 == ":"}
 
 		return components
 			.last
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/GitURL.swift
```diff
@@ -47,7 +47,7 @@ public struct GitURL {
 
 	/// The name of the repository, if it can be inferred from the URL.
 	public var name: String? {
-		let components = urlString.split(omittingEmptySubsequences: true) { $0 == "/" }
+        let components = urlString.split(omittingEmptySubsequences: true) { $0 == "/" || $0 == ":"}
```
giginet:
```suggestion
        let components = urlString.split(omittingEmptySubsequences: true) { ["/", ":"].contains($0)  }
```

jdhealy:
I have questions about this.

jdhealy:
@kamzab95 · sorry for not reviewing this for so long, in fact, …still don't have a review 🥵

But, just FYI, the URL-style representation of what you have above would be `ssh://git@192.168.1.2/myFramework` · would be put in the Cartfile as `git "ssh://git@192.168.1.2/myFramework" == 1.0.0`.

My starting reference point for these has always been <https://stackoverflow.com/questions/31801271/what-are-the-supported-git-url-formats> — both the topmost list and the source hyperlinks.

PR #2893, opened 2019-10-15
Release for 0.34.0

giginet:
Hi ✋ 

Current latest release(0.33.0) is broken for Swift 5.1 compiler.
[Add workaround for Swift 5.1 compiler crash by michaelmcguire · Pull Request #2859 · Carthage/Carthage](https://github.com/Carthage/Carthage/pull/2859)

It's an inconvenience. I suggest to release a newer version from current master.
How do you think? @Carthage/carthage 

I'll prepare changelogs for this release after agreements.
If you have reasons we should not bump up minor. I'll release 0.33.1 with cherry-picking #2859 .

PR #2900, opened 2019-10-26
Track static frameworks in version files

elliottwilliams:
Hi! I've been leveraging Carthage's `--cache-builds` option and have been bit by a bug that was mentioned in the [0.33.0 release notes][1]:

> Static frameworks under `--cache-builds` mode highly likely to proceed errantly cache-valid ≈ late caught bug —
> - Will be correctly cache-invalid upon a version-specification change (including changing the specified SHA); also, upon a toolchain change for a Swift-containing static framework (including after a different `xcode-select`).
> - The first run of a never-modified-by-Carthage-v0.33.0 `.version file` for a Swift-containing static framework under Carthage v0.33.0 will also cache-invalidate it.
> - `carthage build <framework>` without `--cache-builds` will always attempt rebuilding it and it's dependencies.

[1]: https://github.com/Carthage/Carthage/releases/tag/0.33.0

The code in `VersionFile` seems written to assume that every framework is in a platform directory. When `createVersionFileForCommitish` was given a framework in a static directory (e.g. `Carthage/Build/iOS/Static`), it would think "Static" was the platform name and would fail to find the actual framework when writing a version file. As a result, version files for dependencies which build static frameworks were not validating them.

This change fixes the platform-finding logic in `createVersionFileForCommitish`, and ~causes Carthage to include the "Static/" path in a dependency's version file. For example:~

```json
{
  "commitish": "v1.2.3",
  "iOS": [
    {
      "name": "Static/MyFramework",
      "hash": "88c79677521f0a7dc6cc2fe5861fb3d6c6370fffae9df10af7554e53e108c80c",
      "swiftToolchainVersion": "5.1 (swiftlang-1100.0.270.13 clang-1100.0.33.7)"
    }
  ]
}
```

~refers to a framework at `Carthage/Build/iOS/Static/MyFramework.framework`.~
(added a `linking` parameter to CachedFramework instead, see [below](https://github.com/Carthage/Carthage/pull/2900#issuecomment-557858350))

Let me know if this change makes sense, and if there's anything I can do to help test + improve it.


Source/CarthageKit/VersionFile.swift
```diff
@@ -79,13 +79,13 @@ struct VersionFile: Codable {
...
-        let rootBinariesURL = rootDirectoryURL
+	static func url(for dependency: Dependency, rootDirectoryURL: URL) -> URL {
+		let rootBinariesURL = rootDirectoryURL
 			.appendingPathComponent(Constants.binariesFolderPath, isDirectory: true)
 			.resolvingSymlinksInPath()
-        return rootBinariesURL
+		return rootBinariesURL
 			.appendingPathComponent(".\(dependency.name).\(VersionFile.pathExtension)")
-    }
+	}
```
elliottwilliams: This change is unrelated — I saw some un-tabbed indentation and reindented it. Happy to drop in the name of brevity, though 🙂

elliottwilliams:
Re-ran the [broken integration test](https://github.com/Carthage/Carthage/pull/2900/checks?check_run_id=277943110#step:4:11) locally, and it looks like it's failing because of the `Please update to the latest Carthage version: 0.34.0. You currently are on 0.33.0` upgrade message.

The version bump in #2893 hasn't been merged back to master yet, so it sounds like I should wait for that to happen first!

tmspzz:
I would rather have the linking style of the framework mentioned explicitly. If there is no mention, assume dynamic. Existing Version files could also be migrated.


@Carthage/carthage 

```
{
  "commitish": "v1.2.3",
  "iOS": [
    {
      "name": "MyFramework",
      "linking": "static" // dynamic 
      "hash": "88c79677521f0a7dc6cc2fe5861fb3d6c6370fffae9df10af7554e53e108c80c",
      "swiftToolchainVersion": "5.1 (swiftlang-1100.0.270.13 clang-1100.0.33.7)"
    }
  ]
}
```

jdhealy:
@tmspzz  @elliottwilliams — more to post on this later, the real burden in developing a solution for this misbehavior is making sure JSON changes we make (or choose to put in a different file) don't break the unschema-ed `.version` file handling behavior we _currently_ have. 

Otherwise there will be many sad robots…

Robot crying

jdhealy:
It's a tough position to be in — major apologies for not code reviewing more thoroughly in the past.

tmspzz:
I don't think this will break our existing schema. If the "linking" attribute is missing, assume dynamic. 

Currently as far as I understand from the issue, versions file for static libraries are not working at all.

Am I missing something?

elliottwilliams:
> Currently as far as I understand from the issue, versions file for static libraries are not working at all.

That's what I'm seeing. Version files get created for static frameworks, but are empty. This makes `--cache-builds` potentially false positive for static libs — it skips building even when a static framework is missing or invalid.

> I would rather have the linking style of the framework mentioned explicitly. If there is no mention, assume dynamic. Existing Version files could also be migrated.

@tmspzz makes sense to me, I can try that. What should be migrated here — should Carthage try to invalidate old version files that don't list any static frameworks?

> the real burden in developing a solution for this misbehavior is making sure JSON changes we make (or choose to put in a different file) don't break the unschema-ed .version file handling behavior we currently have.

@jdhealy one thing worth mentioning is that either approach (adding "Static" to the `name` or adding an attribute to the version file) will change behavior for existing carthages. If an older carthage is given a version file produced by a newer carthage, `--cache-builds` will now be _false negative_ for static frameworks, and will _always_ rebuild them. I could imagine folks depending on this buggy behavior to avoid rebuilds. Do you think it's worth coming up with a way to do this that's backwards compatible with the bug? (🙅‍♂️😢🤖)

elliottwilliams:
Implemented @tmspzz's suggestion – version files now have an optional `linking` parameter, i.e. 

```json
{
  "commitish": "v1.2.3",
  "iOS": [
    {
      "name": "MyFramework",
      "linking": "static",
      "hash": "88c79677521f0a7dc6cc2fe5861fb3d6c6370fffae9df10af7554e53e108c80c",
      "swiftToolchainVersion": "5.1 (swiftlang-1100.0.270.13 clang-1100.0.33.7)"
    }
  ]
}
```

Updated `ProjectSpec`'s test fixtures to build a static framework, so there's some verification of the caching behavior for staticlibs.

elliottwilliams:
@tmspzz @jdhealy anything else you'd like to see here?

elliottwilliams:
Bump – is this a change you're still interested in?

elliottwilliams:
@tmspzz sorry to bug you again – is there anything I should be doing on this PR to move it forward?

tmspzz:
This LGTM

PR #2901, opened 2019-11-05
Consider module stability when building frameworks

DavidBrunow:
Don't fail builds because the local Swift version doesn't match the framework Swift version if both are module stable

- [ ] Update tests
- [ ] Bikeshed naming
- [ ] Discuss better approaches with reviewers


DavidBrunow:
I branched off of `release/0.34.0` because that seems to be the latest code -- please let me know if this needs to be branched off of `master` or some other branch instead.

DavidBrunow:
I dirtied this up with a merge from `master` so I restarted in #2902

PR #2902, opened 2019-11-05
Consider module stability when building frameworks

DavidBrunow:
Don't fail builds because the local Swift version doesn't match the framework Swift version if both are module stable

- [x] Consider whether the framework is built for distribution
- [x] Bikeshed naming
- [x] Reformat spaces to tabs
- [x] Discuss better approaches with reviewers
- [x] Update tests


DavidBrunow:
I branched off of `release/0.34.0` because that seems to be the latest code -- please let me know if this needs to be branched off of `master` or some other branch instead.

kenji21:
When compiling with BUILD_LIBRARY_FOR_DISTRIBUTION setting, .swiftinterface files are created within the swiftmodule:

```
echo 'BUILD_LIBRARY_FOR_DISTRIBUTION=YES'>/tmp/config.xcconfig; XCODE_XCCONFIG_FILE=/tmp/config.xcconfig carthage build --no-skip-current ; rm /tmp/config.xcconfig

tree Carthage/Build/iOS/SwiftiumKit.framework/
Carthage/Build/iOS/SwiftiumKit.framework/
├── Headers
│   ├── Crypto.h
│   ├── SwiftiumKit-Swift.h
│   └── SwiftiumKit.h
├── Info.plist
├── Modules
│   ├── SwiftiumKit.swiftmodule
│   │   ├── arm.swiftdoc
│   │   ├── arm.swiftinterface
│   │   ├── arm.swiftmodule
│   │   ├── arm64-apple-ios.swiftdoc
│   │   ├── arm64-apple-ios.swiftinterface
│   │   ├── arm64-apple-ios.swiftmodule
│   │   ├── arm64.swiftdoc
│   │   ├── arm64.swiftinterface
│   │   ├── arm64.swiftmodule
│   │   ├── armv7-apple-ios.swiftdoc
│   │   ├── armv7-apple-ios.swiftinterface
│   │   ├── armv7-apple-ios.swiftmodule
│   │   ├── armv7.swiftdoc
│   │   ├── armv7.swiftinterface
│   │   ├── armv7.swiftmodule
│   │   ├── i386-apple-ios-simulator.swiftdoc
│   │   ├── i386-apple-ios-simulator.swiftinterface
│   │   ├── i386-apple-ios-simulator.swiftmodule
│   │   ├── i386.swiftdoc
│   │   ├── i386.swiftinterface
│   │   ├── i386.swiftmodule
│   │   ├── x86_64-apple-ios-simulator.swiftdoc
│   │   ├── x86_64-apple-ios-simulator.swiftinterface
│   │   ├── x86_64-apple-ios-simulator.swiftmodule
│   │   ├── x86_64.swiftdoc
│   │   ├── x86_64.swiftinterface
│   │   └── x86_64.swiftmodule
│   └── module.modulemap
└── SwiftiumKit

3 directories, 33 files
```

maybe presence of such files must be checked in your `isModuleStableAPI` function

DavidBrunow:
> maybe presence of such files must be checked in your `isModuleStableAPI` function

I would think that the presence of those files should be checked in the `frameworkSwiftVersion` method that determines the version of the framework and that effort would be separate from this one. Happy to rethink that if you disagree.

ikesyo:
IIUC module stability is only effective when `.swiftinterface` exists (when a framework is built with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES`), so I agree that the existence must be checked.

DavidBrunow:
> Module stability is only effective when `.swiftinterface` exists, so I agree that the existence must be checked.

Ahh OK I understand now. Sorry I was a bit dense there. I'll look into that.

DavidBrunow:
I've added some logic to check for the existence of a `.swiftinterface` file but I have not found an open source framework that has pre-built binaries and also supports Swift 5.1 and is built for distribution. Do y'all have any examples that I could test against? All the internal frameworks I use are built as XCFrameworks so I can't use them for this testing.

tmspzz:
@DavidBrunow  build Alamofire

kenji21:
Alamofire hasn't yet support for it : https://github.com/Alamofire/Alamofire/issues/2990

I setted `BUILD_LIBRARY_FOR_DISTRIBUTION` to `YES` in this small framework: https://github.com/openium/SwiftiumKit

DavidBrunow:
> Alamofire hasn't yet support for it : [Alamofire/Alamofire#2990](https://github.com/Alamofire/Alamofire/issues/2990)
> 
> I setted `BUILD_LIBRARY_FOR_DISTRIBUTION` to `YES` in this small framework: https://github.com/openium/SwiftiumKit

@kenji21 I might be thinking about this wrong, but wouldn't there need to be a new release on that framework for me to test this functionality?

tmspzz:
@DavidBrunow I would checkout Almofire (or what @kenji21  has posted), change the project and produce the artifact. The create a `binary` dependency.

kenji21:
Just pushed a 1.6.0 tag with binary framework generated by the `tag-release.sh` script : https://github.com/openium/SwiftiumKit/releases/tag/v1.6.0

DavidBrunow:
Thanks to you both -- I'll test both scenarios.

DavidBrunow:
> Just pushed a 1.6.0 tag with binary framework generated by the `tag-release.sh` script : https://github.com/openium/SwiftiumKit/releases/tag/v1.6.0

It looks like that binary was built with Swift 4.2.1.

kenji21:
Sorry, I was re-archiving an in-house app... just  updated the zip with one compiled using:

```
$ xcodebuild -version
Xcode 11.2.1
Build version 11B53

$ carthage build --no-skip-current && carthage archive SwiftiumKit
*** xcodebuild output can be found in /var/folders/7v/dfg0p2vj1z58n65d99l2n1lh0000gp/T/carthage-xcodebuild.T2zD2j.log
Please update to the latest Carthage version: 0.34.0. You currently are on 0.33.0
*** Found Carthage/Build/iOS/SwiftiumKit.framework
*** Found Carthage/Build/iOS/SwiftiumKit.framework.dSYM
*** Found Carthage/Build/iOS/782D37DE-371F-3CC7-B1CC-F1EF9C87014C.bcsymbolmap
*** Found Carthage/Build/iOS/4AE0812B-5F19-3D66-B666-29E7284FB40D.bcsymbolmap
*** Found Carthage/Build/iOS/909D4621-BA13-3654-A185-9DE9453B0801.bcsymbolmap
*** Found Carthage/Build/iOS/E38D4DA2-8222-3C85-9557-E4836BC7F0F7.bcsymbolmap
*** Created SwiftiumKit.framework.zip
```

EmDee:
You can also try `Firebase`:

```
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseABTestingBinary.json" "6.9.0"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json" "6.9.0"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseRemoteConfigBinary.json" "6.9.0"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseMessagingBinary.json" "6.9.0"
```

DavidBrunow:
I was able to successfully test this change using the SwiftiumKit release in Xcode 11.1.

I did not see any issues with the Firebase binaries in any version of Xcode with the change removed so I don't think that is a good test scenario.

I was not able to get the Crashlytics example to work -- I got this error:

```shell
[34;1m***[0m Downloading binary-only framework [1mCrashlytics[0m at [32m"file:///Users/davidbrunow/Development/SmallSampleProject/Crashlytics.json"[0m
2019-11-07 10:20:35.903677-0600 carthage[30281:4932184] Task <EC8F7C5F-BF40-4187-AB44-04F8E0CE3687>.<3> finished with error [-1100] Error Domain=NSURLErrorDomain Code=-1100 "The requested URL was not found on this server." UserInfo={NSUnderlyingError=0x101021de0 {Error Domain=kCFErrorDomainCFNetwork Code=-1100 "(null)"}, NSErrorFailingURLStringKey=file:///Users/davidbrunow/Development/SmallSampleProject/Crashlytics.json, NSErrorFailingURLKey=file:///Users/davidbrunow/Development/SmallSampleProject/Crashlytics.json, NSLocalizedDescription=The requested URL was not found on this server.}
Failed to read file or folder at /Users/davidbrunow/Development/SmallSampleProject/Crashlytics.json: Error Domain=Result.AnyError Code=1 "The requested URL was not found on this server."
```

kenji21:
> I was able to successfully test this change using the SwiftiumKit release in Xcode 11.1.

nice to read, did you try both with the binary framework built using Xcode 11.2.1 ? and with the flag `--no-use-binaries` ?

DavidBrunow:
> > I was able to successfully test this change using the SwiftiumKit release in Xcode 11.1.
> 
> nice to read, did you try both with the binary framework built using Xcode 11.2.1 ? and with the flag `--no-use-binaries` ?

Yes I tested with the binary framework built using Xcode 11.2.1 and I just tested with the `--no-use-binaries` flag and everything builds with no errors.

MikeKasperlik:
I tested this branch locally and everything seems to work 👍 Thanks.

EmDee:
Same here. Works like a charm. Thanks, @DavidBrunow !

Also noticed that our own framework didn't have the `BUILD_LIBRARY_FOR_DISTRIBUTION` flag set to `true`, so yay 🎉

simonmitchell:
@ikesyo any idea when this may get merged and distributed in a release? 😊

EmDee:
Maybe @tmspzz or @mdiep can say something regarding a possible merge/release.

telip007:
Any updates on this? :)

hujunfeng:
Thanks, @DavidBrunow. The patch works well.

However, I find that [version files](https://github.com/Carthage/Carthage/blob/master/Documentation/VersionFile.md) needs to be checked against Swift 5.1 too. For Swift frameworks, version files save `swiftToolchainVersion` inside. Because of this, currently, Carthage will consider a mismatch of versions when resolving Cartfile, if a binary framework is built in 5.1 and the current Swift toolchain is 5.1.3, for example. As a result, Carthage will then download that binary framework again. Although Carthage will not throw the error of incompatible Swift version (because of this patch), it is annoying that it will download such frameworks every time I run `carthage bootstrap`.

dlewanda:
Thanks @DavidBrunow for your work here. Heads up that you will need to rebase on the latest `master` at this point, but otherwise it works for us. We'd love to see this PR get merged ASAP!

DavidBrunow:
I'm struggling with updating the tests -- any tips on only running specific tests or running the tests in Xcode?

DavidBrunow:
> Thanks, @DavidBrunow. The patch works well.
> 
> However, I find that [version files](https://github.com/Carthage/Carthage/blob/master/Documentation/VersionFile.md) needs to be checked against Swift 5.1 too. For Swift frameworks, version files save `swiftToolchainVersion` inside. Because of this, currently, Carthage will consider a mismatch of versions when resolving Cartfile, if a binary framework is built in 5.1 and the current Swift toolchain is 5.1.3, for example. As a result, Carthage will then download that binary framework again. Although Carthage will not throw the error of incompatible Swift version (because of this patch), it is annoying that it will download such frameworks every time I run `carthage bootstrap`.

I am seeing the proper behavior when running `carthage bootstrap`.

DavidBrunow:
I've added some tests but I'm concerned that they are too dependent upon the environment in which they are running to add value. Would appreciate input from folks familiar with writing tests.

hujunfeng:
> I'm not able to replicate that specific scenario but I'd lean towards making that fix a separate effort from this one.

I'm able to reproduce the scenario with `carthage bootstrap ----cache-builds`. 

Here are a test Cartfile and a binary spec for a framework built with Swift 5.1:
[ReproduceInvalidCache.zip](https://github.com/Carthage/Carthage/files/3883266/ReproduceInvalidCache.zip)

Used `carthage` built with commit [eb7d572](https://github.com/Carthage/Carthage/pull/2902/commits/eb7d572d0bec6992ff0119861e5954fcdd7cb513). This is the output:

```
~/Downloads/ReproduceInvalidCache
> carthage bootstrap --cache-builds
Please update to the latest Carthage version: 0.34.0. You currently are on 0.33.0
*** Downloading binary-only framework MyFramework at "file:///Users/jh/Downloads/ReproduceInvalidCache/MyFramework.json"
*** No cache found for MyFramework, building with all downstream dependencies
*** Downloading binary-only framework MyFramework at "file:///Users/jh/Downloads/ReproduceInvalidCache/MyFramework.json"
*** xcodebuild output can be found in /var/folders/p7/fx3v92297j95yqw4whx22b040000gn/T/carthage-xcodebuild.M4DuSb.log
~/Downloads/ReproduceInvalidCache
> carthage bootstrap --cache-builds
Please update to the latest Carthage version: 0.34.0. You currently are on 0.33.0
*** xcodebuild output can be found in /var/folders/p7/fx3v92297j95yqw4whx22b040000gn/T/carthage-xcodebuild.u0qIEt.log
*** Invalid cache found for MyFramework, rebuilding with all downstream dependencies
*** Downloading binary-only framework MyFramework at "file:///Users/jh/Downloads/ReproduceInvalidCache/MyFramework.json"
```

Note that when `carthage bootstrap --cache-builds` was run for the second time, it considered the cache was invalid, thus went to download the framework again.

EmDee:
@hujunfeng Are you seeing that cache mismatch happening with the current release build as well? I had the same happening to me with another framework (`Kingfisher` – https://github.com/onevcat/Kingfisher/issues/1307), but there it was due to missing header files in the framework itself.

I agree with @DavidBrunow though. If that's really a bug, it should be dealt with separately as it doesn't seem related to this one.

hujunfeng:
@EmDee Yes, I'm seeing the mismatch with the current release build. 

But that's the correct behavior for the release build because the Swift compiler version I was using was 5.1.3 and the binary framework was built with 5.1. There was indeed a mismatch. (I realized my previous post wasn't clear about that, so have just updated it with more context.) 

I also agree that the special scenario related to cache builds can be tackled in a separate PR.

patburns:
We're having problems trying to include Zendesk binaries (which just included module stability with their [4.0.0 release](https://developer.zendesk.com/embeddables/docs/ios_support_sdk/release_notes))

Excited to see that someone already is tackling issue in Carthage. Thanks for your work on this. Looking forward to the next release.

rainypixels:
@patburns Were you able to figure out a workaround for the Zendesk binaries?

simonmitchell:
@patburns @rainypixels we spent a lot of time in contact with Zendesk, because the release prior to that was built with Xcode 11.2 which obviously crashed our app when submitting feedback. They were absolutely useless and insisted on us updating to 4.0.0. We couldn't find a workaround, and I very much doubt they'll do anything to resolve it their end. Even though building a 4.0.1 with the only difference being using Xcode 11.2.1 would take them all of 10 minutes!

rainypixels:
@simonmitchell Thanks for circling back. And ugh, that sucks. But I can't say I'm surprised given my own experiences with their engineering process. 🤕I am on the verge of the manual install path as well; at least that will give me some control over managing my Zendesk dependency.

patburns:
@simonmitchell @rainypixels. To be clear at this point it's not a Zendesk issue. With 4.0.0 they have finally included the necessary changes to allow module stability. So if you manually install as @simonmitchell suggest you should be good to go. The next issue is that Carthage is making a check that doesn't account for module stability. Zendesk appears to have built 4.0.0 with Swift 5.1. So if you are running the latest xcode Carthage believes you'll be in conflict with your xcode version of swift which I believe is currently 5.1.2.

I believe this PR is attempting to fix the Carthage issue. @DavidBrunow does that sound correct to you?

DavidBrunow:
@patburns That is correct.

rzulkoski:
Any movement on this? I'm excited to see this merged and released!

anuragajwani:
Also keen to see this fix merged and released. It's blocking our customers from integrating with our framework when using Xcode 11.2.x. Our framework is also built with Xcode 11.1.

https://github.com/onfido/onfido-ios-sdk/issues/124

simonmitchell:
@anuragajwani don't be a Zendesk... just build a release using Xcode 11.2.1 😝 Is it really that hard to update it to Swift 5.2.1?

kenji21:
> @anuragajwani don't be a Zendesk... just build a release using Xcode 11.2.1 😝 Is it really that hard to update it to Swift 5.2.1?

I think you inverted minor/bugfix, it is 5.1.2 (and 5.1.3 for Xcode 11.3)

anuragajwani:
@simonmitchell our framework compiled with Xcode 11.2 was generating crashes which were not being encountered when using Xcode 11.1. And given module compatibility our integrators shouldn't have any problems using the our framework with Xcode 11.2.

kenji21:
> @simonmitchell our framework compiled with Xcode 11.2 was generating crashes which were not being encountered when using Xcode 11.1. And given module compatibility our integrators shouldn't have any problems using the our framework with Xcode 11.2.

Are you having this crash : https://stackoverflow.com/questions/58657087/after-upgrading-to-xcode-11-2-from-xcode-11-1-app-crashes-due-to-uitextlayoutv
what's why apple shipped Xcode 11.2.1...

anuragajwani:
@kenji21 

I can't recall the exact crash right now. Will have to check with the team. However regardless we deliver a compiled framework. That means that if we update to Xcode 11.2.1 integrators using Xcode 11.1 and using Carthage won't be able to use our latest release.

kenji21:
> @kenji21
> 
> I can't recall the exact crash right now. Will have to check with the team. However regardless we deliver a compiled framework. That means that if we update to Xcode 11.2.1 integrators using Xcode 11.1 and using Carthage won't be able to use our latest release.

No, but they can fix version in their Cartifle to point to your "last Xcode 11.1 build" version, latest version using latest Xcode...

choli:
> Any movement on this? I'm excited to see this merged and released!

We are interested in this as well. We have "workarounds" to make it work, these are not the answers we are interested in :)

So what I would like to know: Whats the status of this? What is preventing you from merging this? (Besides failing tests probably...)

anuragajwani:
> > @kenji21
> > I can't recall the exact crash right now. Will have to check with the team. However regardless we deliver a compiled framework. That means that if we update to Xcode 11.2.1 integrators using Xcode 11.1 and using Carthage won't be able to use our latest release.
> 
> No, but they can fix version in their Cartifle to point to your "last Xcode 11.1 build" version, latest version using latest Xcode...

We want to enable integrators to consume the latest version. Without going into much detail that is good for our integrator as well as the organisation I work for.

DavidBrunow:
I'm pushing on this to get it moved forward.

rzulkoski:
@DavidBrunow  If I'm reading this correctly, I think you made some of these tests too specific. It's probably sufficient to check for `Incompatible Swift version` rather than for the specific versions involved, as this test will have different output depending on the machine it is running on.

```
Test Case '-[CarthageKitTests.XcodeSpec determineSwiftInformation___should_determine_when_a_module_stable_Swift_framework_is_incompatible]' started.
/Users/travis/build/Carthage/Carthage/Tests/CarthageKitTests/XcodeSpec.swift:108: error: -[CarthageKitTests.XcodeSpec determineSwiftInformation___should_determine_when_a_module_stable_Swift_framework_is_incompatible] : failed - expected to equal <Incompatible Swift version - framework was built with 5.1.2 (swiftlang-1100.0.278 clang-1100.0.33.9) and the local version is 5.0 (swiftlang-1001.0.69.5 clang-1001.0.46.3).>, got <Incompatible Swift version - framework was built with 5.1.2 (swiftlang-1100.0.278 clang-1100.0.33.9) and the local version is 4.2.1 (swiftlang-1000.11.42 clang-1000.11.45.1).>
Test Case '-[CarthageKitTests.XcodeSpec determineSwiftInformation___should_determine_when_a_module_stable_Swift_framework_is_incompatible]' failed (0.726 seconds).
```

DavidBrunow:
> @DavidBrunow If I'm reading this correctly, I think you made some of these tests too specific. It's probably sufficient to check for `Incompatible Swift version` rather than for the specific versions involved, as this test will have different output depending on the version of Xcode installed on the machine it is running on.
> 
> ```
> Test Case '-[CarthageKitTests.XcodeSpec determineSwiftInformation___should_determine_when_a_module_stable_Swift_framework_is_incompatible]' started.
> /Users/travis/build/Carthage/Carthage/Tests/CarthageKitTests/XcodeSpec.swift:108: error: -[CarthageKitTests.XcodeSpec determineSwiftInformation___should_determine_when_a_module_stable_Swift_framework_is_incompatible] : failed - expected to equal <Incompatible Swift version - framework was built with 5.1.2 (swiftlang-1100.0.278 clang-1100.0.33.9) and the local version is 5.0 (swiftlang-1001.0.69.5 clang-1001.0.46.3).>, got <Incompatible Swift version - framework was built with 5.1.2 (swiftlang-1100.0.278 clang-1100.0.33.9) and the local version is 4.2.1 (swiftlang-1000.11.42 clang-1000.11.45.1).>
> Test Case '-[CarthageKitTests.XcodeSpec determineSwiftInformation___should_determine_when_a_module_stable_Swift_framework_is_incompatible]' failed (0.726 seconds).
> ```

@rzulkoski I think that might move us forward a little but still wouldn't be testing everything we need to test. I'm not sure these tests are useful at all without being able to specify a known local Swift version for  these tests. Without that, I don't think we can test these scenarios:

- Local Swift version >= 5.1 but framework version below 5.1
- Local Swift version >= 5.1 and framework version >= 5.1 but framework not built for distribution

rzulkoski:
@DavidBrunow I'm speaking from complete ignorance here as far as how this project is set up and what its testing capabilities are, but the scenarios you have outlined above seem like a job for mocked unit tests rather than integration tests. What we need to test is Carthage's decision-making in those scenarios given different inputs, not how specific Swift installations behave on a given machine.

So if you could mock out determining the local Swift version, determining a framework's version and determining whether the framework was built for distribution, you could write a few unit tests around these scenarios:

1. Swift >= 5.1, framework < 5.1, expect failure
2. Swift >= 5.1, framework >= 5.1 but not built for distribution, expect failure
3. Swift >= 5.1, framework >= 5.1 and built for distribution, expect success

You may also need to test determining if it's built for distribution, basically mocking out the results when looking for the .swiftmodule directory:

4. Directory not found, expect false
5. Directory empty or missing architectures (not sure if that's a valid check?), expect false
6. Directory found with expected contents, expect true

Does that make sense or am I completely off base here?

rzulkoski:
@DavidBrunow Another thought: If you do want to do integration testing then you should exclude your new tests when Local Swift < 5.1 since your changes aren't targeted at older versions. The Travis builds would have succeeded in that case.

DavidBrunow:
I think you are right about mocking and I think the simplest way to do so would be to create a way to specify a local version separate from the functionality to detect the local version. I'm not sure how I'd do that with the current way that things are built but I'll dig into it and see if I can figure something out.

Source/CarthageKit/Xcode.swift
```diff
@@ -134,12 +134,28 @@ internal func isSwiftFramework(_ frameworkURL: URL) -> Bool {
...
 	return SignalProducer.combineLatest(swiftVersion(usingToolchain: toolchain), frameworkSwiftVersion(frameworkURL))
 		.attemptMap { localSwiftVersion, frameworkSwiftVersion in
-			return localSwiftVersion == frameworkSwiftVersion
+			return localSwiftVersion == frameworkSwiftVersion || isModuleStableAPI(localSwiftVersion, frameworkSwiftVersion, frameworkURL)
 				? .success(frameworkURL)
 				: .failure(.incompatibleFrameworkSwiftVersions(local: localSwiftVersion, framework: frameworkSwiftVersion))
 		}
 }
 
+private func isModuleStableAPI(_ localSwiftVersion: String,
```
rzulkoski: Would you consider marking this method as `internal` instead of `private` and writing your tests against it instead of relying on resolving the toolchain?

mozeryansky:
@DavidBrunow Thanks for taking this on! Our team has really been struggling with this issue.

Source/CarthageKit/Xcode.swift
DavidBrunow: I think that is what I’ll have to do. 

Not exactly sure why but I’d prefer an integration test.

txaiwieser:
Any updates on this?

jackmichalak:
Thanks to all working to get this issue addressed. I wanted to check in on this since the holidays are now over. 

While in most ways Carthage is better than Cocoapods, I'm now spending lots of extra time building and testing releases only because of Carthage not understanding the module compatibility.

When might this be merged? What's preventing the merge? Is there anything I can do to help the process along?

rzulkoski:
@DavidBrunow I've done a little more digging and it looks like the `swiftVersion(usingToolchain:)` method that `checkSwiftFrameworkCompatibility(frameworkUrl:usingToolchain:)` depends on will attempt to retrieve the toolchain specified if it is installed, but will ultimately fall back to whatever xcode-select is currently pointed at. This means it is not suitable for our testing purposes. I have a proposed solution that I will reference here shortly.

DavidBrunow:
> Thanks to all working to get this issue addressed. I wanted to check in on this since the holidays are now over.
> 
> While in most ways Carthage is better than Cocoapods, I'm now spending lots of extra time building and testing releases only because of Carthage not understanding the module compatibility.
> 
> When might this be merged? What's preventing the merge? Is there anything I can do to help the process along?
> 
> Actually, why is this check performed at all? Which version of Xcode is selected at the time has no bearing on whether it will be usable in the specific project.

This check is performed because versions of Swift prior to Swift 5.1 were not module-stable.

I cannot speak to the timelines for when this will be merged since I'm not part of the core Carthage team. I will remove the failing integration tests and replace them with unit tests (or possibly the approach that @rzulkoski mentioned) over the next few days so hopefully it will be ready to be merged and a release can be cut.

In the meantime, you can also build Carthage from source and use that binary to avoid the extra time building and testing releases.

rzulkoski:
I've added a [PR](https://github.com/Shepherd-Dog/Carthage/pull/1) for your branch which includes my suggested changes.

jackmichalak:
@DavidBrunow Thank you. I don't mean to rush, and I know I wouldn't be able to make these changes myself. :)

I know that Swift versions weren't module stable previously, but I've always had multiple versions of Xcode on my computer and which one is selected is basically random. I don't understand why Carthage wants to prevent downloading based on which version of Xcode is selected; we'll get the same error when we try to build if it actually matters, and if it doesn't matter then that's great.

Unfortunately this is for distribution to clients, so I have to support every version of Xcode (particularly since CI isn't always very flexible). If it were internal I'd just be telling people to download my preferred version of Xcode.

DavidBrunow:
> I know that Swift versions weren't module stable previously, but I've always had multiple versions of Xcode on my computer and which one is selected is basically random. I don't understand why Carthage wants to prevent downloading based on which version of Xcode is selected; we'll get the same error when we try to build if it actually matters, and if it doesn't matter then that's great.

I wasn't part of the folks that decided that Carthage shouldn't download frameworks if they are not compatible with the system's Swift version but the decision makes sense to me. I'd rather know sooner rather than later that a framework won't work for my environment and have that framework built for me if I cannot download a compatible version.

DavidBrunow:
@tmspzz @jdhealy Any chance y'all could take a look at this in the near future?

jacobmitchinson:
@DavidBrunow Thank you for doing this! It's been a source of much headache in our team! I'd love to see this merged.

txaiwieser:
I know everyone is trying really hard to make this happen, but is there any blockers to this getting merged?

rzulkoski:
> @tmspzz @jdhealy Any chance y'all could take a look at this in the near future?

Any movement on this?

Sohil:
> @tmspzz @jdhealy Any chance y'all could take a look at this in the near future?

@tmspzz @jdhealy any chance we get a new release with this fix ?

mozeryansky:
@ikesyo  @tmspzz @jdhealy Just posting again on this. Could we get this merged into a new release soon?

michaeleisel:
Will there be a release soon that includes this?

michaeleisel:
Actually, I've tried building Carthage from source with the latest master (c7d6b8dbe), and I still see this problem. When I clone https://github.com/mozilla-mobile/firefox-ios and build with `carthage boostrap --platform ios --color auto --cache-builds`, I get:
```
*** Downloading application-services.framework binary at "v0.48.0"
***  Skipped installing application-services.framework binary due to the error:
	"Incompatible Swift version - framework was built with 5.1.3 (swiftlang-1100.0.282.1 clang-1100.0.33.15) and the local version is 5.1.2 (swiftlang-1100.0.278 clang-1100.0.33.9)."
```

DavidBrunow:
@michaeleisel When looking at the `common.xcconfig` used in `application-services.framework` I don't see the key `BUILD_LIBRARY_FOR_DISTRIBUTION` which is necessary for module stability. I'm not familiar with that project so please let me know if I am missing something.

michaeleisel:
Good catch, that's probably it

dchohfi:
hey, any idea when a new version of carthage is going to be released?

johnngoi:
@rzulkoski @tmspzz do we have a plan to add this enhancement in a release soon?

PR #2907, opened 2019-11-11
Prevent directory traversal

manicmaniac:
## Problem

When Cartfile contains a local dependency of which path ends with `..`, Carthage clones the repo into the wrong place.

I created the repository to reproduce the behavior: https://github.com/manicmaniac/CarthageDirectoryTraversal

## Cause

`GitURL.name` doesn't check if the own path is relative or absolute but just returns the last path component.

So the name could be `..` and later it will be used as the destination path where the dependency should be copied.

## Solution

I added some code to `GitURL` to convert a relative path to the absolute path before getting the name.

### DIFF
--- a/Source/CarthageKit/GitURL.swift
+++ b/Source/CarthageKit/GitURL.swift
@@ -47,7 +47,16 @@ public struct GitURL {
 
 	/// The name of the repository, if it can be inferred from the URL.
 	public var name: String? {
-		let components = urlString.split(omittingEmptySubsequences: true) { $0 == "/" }
+        let absoluteURLString: String
+        if urlString.hasPrefix(".") {
+            absoluteURLString = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
+                .appendingPathComponent(urlString)
+                .standardizedFileURL
+                .absoluteString
+        } else {
+            absoluteURLString = urlString
+        }
+		let components = absoluteURLString.split(omittingEmptySubsequences: true) { $0 == "/" }
 
 		return components
 			.last
--- a/Source/CarthageKit/GitURL.swift
+++ b/Source/CarthageKit/GitURL.swift
@@ -47,15 +47,15 @@ public struct GitURL {
 
 	/// The name of the repository, if it can be inferred from the URL.
 	public var name: String? {
-        let absoluteURLString: String
-        if urlString.hasPrefix(".") {
-            absoluteURLString = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
-                .appendingPathComponent(urlString)
-                .standardizedFileURL
-                .absoluteString
-        } else {
-            absoluteURLString = urlString
-        }
+		let absoluteURLString: String
+		if urlString.hasPrefix(".") {
+			absoluteURLString = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
+				.appendingPathComponent(urlString)
+				.standardizedFileURL
+				.absoluteString
+		} else {
+			absoluteURLString = urlString
+		}
 		let components = absoluteURLString.split(omittingEmptySubsequences: true) { $0 == "/" }
 
 		return components
--- a/Tests/CarthageKitTests/DependencySpec.swift
+++ b/Tests/CarthageKitTests/DependencySpec.swift
@@ -85,15 +85,44 @@ class DependencySpec: QuickSpec {
 					expect(dependency.name) == "whatisthisurleven"
 				}
 
-				it("should be the directory name if the given URL is relative local path") {
+				context("when a relative local path with dots is given") {
 					let fileManager = FileManager.default
-					let startingDirectory = fileManager.currentDirectoryPath
-					defer { fileManager.changeCurrentDirectoryPath(startingDirectory) }
-					let temporaryDirectoryURL = URL(fileURLWithPath: NSTemporaryDirectory())
-					fileManager.changeCurrentDirectoryPath(temporaryDirectoryURL.path)
-					let dependency = Dependency.git(GitURL(".."))
+					var startingDirectory: String!
+					var temporaryDirectoryURL: URL!
 
-					expect(dependency.name) == temporaryDirectoryURL.deletingLastPathComponent().lastPathComponent
+					beforeEach {
+						startingDirectory = fileManager.currentDirectoryPath
+						temporaryDirectoryURL = URL(fileURLWithPath: NSTemporaryDirectory())
+						fileManager.changeCurrentDirectoryPath(temporaryDirectoryURL.path)
+					}
+
+					afterEach {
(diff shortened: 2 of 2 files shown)


Source/CarthageKit/GitURL.swift
```diff
@@ -47,7 +47,16 @@ public struct GitURL {
 
 	/// The name of the repository, if it can be inferred from the URL.
 	public var name: String? {
-		let components = urlString.split(omittingEmptySubsequences: true) { $0 == "/" }
+        let absoluteURLString: String
```
DavidBrunow: It looks like the new indentation is using spaces instead of tabs. Could you please reformat with tabs to match the existing indentation?

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -84,6 +84,17 @@ class DependencySpec: QuickSpec {
 
 					expect(dependency.name) == "whatisthisurleven"
 				}
+
+                it("should be the directory name if the given URL is relative local path") {
```
DavidBrunow: Same comment here about indentation.

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -84,6 +84,17 @@ class DependencySpec: QuickSpec {
...
+                    let fileManager = FileManager.default
+                    let previousDirectory = fileManager.currentDirectoryPath
```
DavidBrunow: I'm not sure that `previousDirectory` accurate represents what this is -- would `currentDirectory` make more sense?

DavidBrunow: Looking again I understand why it is named that way a little better -- what about `startingDirectory` instead of `previousDirectory`?

Tests/CarthageKitTests/DependencySpec.swift
```diff
@@ -84,6 +84,17 @@ class DependencySpec: QuickSpec {
...
+                    defer { fileManager.changeCurrentDirectoryPath(previousDirectory) }
+                    let temporaryDirectoryURL = URL(fileURLWithPath: NSTemporaryDirectory())
```
DavidBrunow: I'm a tiny bit concerned about relying on the location of `NSTemporaryDirectory()`. In specific, this code is assuming that there will always be a directory below the temporary directory, right? Is that a safe assumption?

DavidBrunow:
After looking at this I am curious -- how does Carthage handle a single `.` in the path? Did you happen to run across that while creating this pull request?

Source/CarthageKit/GitURL.swift
manicmaniac: Fixed in https://github.com/Carthage/Carthage/pull/2907/commits/e561595a6fdfe2bc47ca79866104c396575d8183.

Tests/CarthageKitTests/DependencySpec.swift
manicmaniac: Fixed in https://github.com/Carthage/Carthage/pull/2907/commits/e561595a6fdfe2bc47ca79866104c396575d8183.

Tests/CarthageKitTests/DependencySpec.swift
manicmaniac: Agree. Fixed in https://github.com/Carthage/Carthage/pull/2907/commits/49b3e975c130cadcdde1a49722e9fd35770b5017.

Tests/CarthageKitTests/DependencySpec.swift
manicmaniac: I'm afraid I don't understand your concern precisely.

IMO it's enough because [the document of `NSTemporaryDirectory`](https://developer.apple.com/documentation/foundation/1409211-nstemporarydirectory) says
> Returns the path of the temporary directory for the current user.

In Darwin, `NSTemporaryDirectory` just returns the value defined in `confstr(_CS_DARWIN_USER_TEMP_DIR)` and fallbacks to `$TMPDIR` environment variable or `/tmp` (see [the implementation in Swift corelibs](https://github.com/apple/swift-corelibs-foundation/blob/3b471931ee535e8138bdada2f05ebb42095a7ed2/Foundation/NSPathUtilities.swift#L28-L57)).

So technically speaking, you can cheat `NSTemporaryDirectory` by setting `$TMPDIR` to the wrong path like a path which doesn't exist or a path to a regular file.
However that kind of settings violates to [the POSIX specification](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html) so I guess we don't have to consider the situation.

I don't know much about Windows but I suppose it may be the similar situation to Darwin.

manicmaniac:
> After looking at this I am curious -- how does Carthage handle a single . in the path? Did you happen to run across that while creating this pull request?

To be honest, I have never tested with a single `.` in the path.
Should I test it in this PR?

Tests/CarthageKitTests/DependencySpec.swift
DavidBrunow: It probably isn't something we need to worry about but the concern is that in a different environment the temporary directory might be at the root of the directory structure. If that were the case I think the test would break?

But really it is a tiny concern and not likely to happen. Thanks for taking the time to respond.

DavidBrunow:
> > After looking at this I am curious -- how does Carthage handle a single . in the path? Did you happen to run across that while creating this pull request?
> 
> To be honest, I have never tested with a single `.` in the path.
> Should I test it in this PR?

Since this line is testing for the existence of a prefix with `.` I think so:

```swift

Tests/CarthageKitTests/DependencySpec.swift
manicmaniac: > the temporary directory might be at the root of the directory structure.

Honestly I don't know whether if the environment like that exists or not.
But if it exists and it complies to POSIX, which means the root directory is the temporary directory where any users can move it, so I guess it won't break the test.
But if it exists and it is non-POSIX OS, I don't know what happens (again, I don't have a good example for this situation).

BTW I found Carthage already uses `NSTemporaryDirectory` in another place.

https://github.com/Carthage/Carthage/blob/7cb0e6ab571f0ce32e9f94bf9a24cb80371be226/Source/carthage/Build.swift#L184

So I think we don't have to care about the use of it.

manicmaniac:
> Since this line is testing for the existence of a prefix with `.` I think so.

Fixed in https://github.com/Carthage/Carthage/pull/2907/commits/bad4fb0a5b1a6bd4d2e0b4f3b1558447d0c8df14.

Please review it again 👍

DavidBrunow:
@manicmaniac Thank you for this effort. We need to get one of the maintainers to have a look at it and then hopefully we can get this merged in.

manicmaniac:
@tmspzz @DavidBrunow Could you please merge this one or let me now if there's any problem?

Issue #2760, opened 2019-04-06
Carthage crashes with Segmentation Fault: 11


JetForMe:
Carthage crashes with Segmentation Fault: 11, but running it again usually succeeds. It's not clear to me what changes to cause it to happen again.

* carthage install method: [ ] .pkg, [X] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.33.0
* `xcodebuild -version`: Xcode 10.1
Build version 10B61, but also happens with 10.2
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? no
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? no
* Are you using `--new-resolver`? no

**Cartfile**
```
github "agilebits/onepassword-app-extension" "add-framework-support"
github "Alamofire/Alamofire" ~> 4.0
github "dzenbot/DZNEmptyDataSet" ~> 1.8.1
github "m1entus/MZFormSheetPresentationController" "master"
github "mxcl/PromiseKit" ~> 6.0
github "PromiseKit/PMKAlamofire" ~> 3.2
github "segmentio/analytics-ios" ~> 3.6
github "square/Valet"
github "latencyzero/Marshal" "escapePeriods"
```

**Carthage Output**
```
$ carthage update --platform ios
*** Fetching Valet
*** Fetching Marshal
*** Fetching MZFormSheetPresentationController
*** Fetching onepassword-app-extension
*** Fetching DZNEmptyDataSet
*** Fetching analytics-ios
*** Fetching PromiseKit
*** Fetching Alamofire
*** Fetching PMKAlamofire
*** Fetching MZAppearance
*** Checking out MZAppearance at "de07b5fdd79465e4193095cd33a66843106e71de"
*** Checking out onepassword-app-extension at "bcc4cc97fed9a6e73fa204f2e61138e353cb3ef7"
*** Checking out MZFormSheetPresentationController at "6f4b7c8875baaa1a881c889934c7b9901155d393"
*** Checking out analytics-ios at "3.6.10"
*** Checking out DZNEmptyDataSet at "v1.8.1"
*** Checking out PromiseKit at "6.8.4"
*** Checking out Valet at "3.2.3"
*** Checking out Alamofire at "4.8.2"
*** Checking out Marshal at "cd203eb665d1ec0e1c04f26537316365bbecbb63"
*** Checking out PMKAlamofire at "3.2.3"
Segmentation fault: 11
```

[...]

PR #2908, opened 2019-11-20
Fix Segmentation Fault 11 while calling FetchCache.updateLastFetchTime

taisukeh:
## Problem

https://github.com/Carthage/Carthage/issues/2760#issuecomment-555964945

## Solution

Use NSLock to coordinates the operation of multiple threads of execution within the FetchCache.

### DIFF
--- a/Source/CarthageKit/Git.swift
+++ b/Source/CarthageKit/Git.swift
@@ -18,11 +18,16 @@ public struct FetchCache {
 
 	private static var lastFetchTimes: [GitURL: TimeInterval] = [:]
 
+	private static let lock = NSLock()
+
 	internal static func clearFetchTimes() {
 		lastFetchTimes.removeAll()
 	}
 
 	internal static func needsFetch(forURL url: GitURL) -> Bool {
+		lock.lock()
+		defer { lock.unlock() }
+
 		guard let lastFetch = lastFetchTimes[url] else {
 			return true
 		}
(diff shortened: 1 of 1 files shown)


peacemoon:
Would love to have this to be merged. My github action failed 100% now because of this `Segmentation Fault 11` issue

Source/CarthageKit/Git.swift
```diff
@@ -18,11 +18,16 @@ public struct FetchCache {
 
 	private static var lastFetchTimes: [GitURL: TimeInterval] = [:]
 
+	private static let lock = NSLock()
+
 	internal static func clearFetchTimes() {
 		lastFetchTimes.removeAll()
```
jonbrooks: If the lock is guarding usage of `lastFetchTimes`, do we need the lock here as well?

taisukeh: As you pointed out, it needed to be added!

daveluong:
Hi @taisukeh do you plan to merge this soon? Our builds are failing too frequently, would like to try out this fix. thank you!

ayachama:
Our bitrise builds are plagued with above issue. Can we get a release with the above fix?

joeydong:
tbh, i added this workaround a while ago and it's been fine 😄 

`carthage bootstrap --platform iOS || carthage bootstrap --platform iOS # Retry once`

PR #2917, opened 2019-12-12
Extend error messages when reading binaries

hlineholm:
If for instance your bundle doesn't contain a plist, the current error message says that Carthage cannot read the file or folder. The same goes for if the filename of the dSYM isn't in the correct format. So I've extended these messages.

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -1383,27 +1383,27 @@ public func binaryURL(_ packageURL: URL) -> Result<URL, CarthageError> {
 		let binaryName = packageURL.deletingPathExtension().deletingPathExtension().lastPathComponent
 		if binaryName.isEmpty {
 			return .failure(.readFailed(packageURL, NSError(
-                domain: NSCocoaErrorDomain,
-                code: CocoaError.fileReadInvalidFileName.rawValue,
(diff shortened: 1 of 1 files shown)


Source/CarthageKit/Xcode.swift
```diff
@@ -1381,13 +1381,29 @@ public func binaryURL(_ packageURL: URL) -> Result<URL, CarthageError> {
...
+		if binaryName.isEmpty {
+			return .failure(.readFailed(packageURL, NSError(
+                domain: NSCocoaErrorDomain,
+                code: CocoaError.fileReadInvalidFileName.rawValue,
+                userInfo: [
+                    NSLocalizedDescriptionKey: "Your dSYM has an invalid filename",
+                    NSLocalizedRecoverySuggestionErrorKey: "Make sure your dSYM has a an extension with the 'name.framework.dSYM' format"
+                ]
+            )))
+        } else {
```
DavidBrunow: Indentation looks a little off here and in the `return` below -- Carthage uses tabs instead of spaces, could you please reformat?

DavidBrunow:
Looks good to me other than the formatting I mentioned.

hlineholm:
Super @tmspzz 👍

PR #2924, opened 2020-01-16
Adds an override for Swift's "treat warnings as errors" build setting

1oo7:
Some dependency authors have "treat warnings as errors" turned on in their published frameworks. This is fine until Apple releases a new compiler version with a more thorough analyzer. Now your project won't build because of a single line of unreachable code, which was a fatalError() anyway, in a dependency that may not even be under active development anymore.

Or perhaps a build server has 15 jobs in the Jenkins queue, leading to type-checking exceeding 400 ms, triggering a Swift warning. Now your build cancels and to meet a deadline you're forced to disable "treat warnings as errors" in your git repo, since there was no way to simply override it from the command line.

Maybe your engineers can't update XCode to the latest version because of some new warnings about "let" vs. "var" etc. making carthage not `bootstrap`—but since they can't update XCode, they can't see all the new warnings at once, meaning they can't fix them. It's a sort of vicious cycle.

This PR fixes that by adding a simple new option to Carthage: `--warnings-as-errors X` where `X` is `yes`, `no`, `1`, `0`, `true`, `false`, `y`, `n` (not case sensitive).

Tests & documentation are included.

Note to reviewers: I don't know if this should be version 0.35.0 or 0.34.1. Updating to this version will not break anyone, but downgrading to 0.34.0 from this version might break a build script here and there if they might have added the new build setting. So I went with 0.35.0 to be on the safe side.


1oo7:
Thanks @tmspzz. It will be a big help to some of us. I built and tested it locally and it allowed us to successfully build our app with Xcode 11, whereas before, we weren't able to due to warnings being treated as errors in ReactiveSwift 5.

1oo7:
I added a few more reviewers since I'm guessing the first few were busy other than @tmspzz. Would really be nice since this will really help us go to Xcode 11.

1oo7:
If there are no objections, is there any way we can please get this merged and released? 

I can't imagine how anyone could argue Carthage should lack this capability, considering that Carthage's own SPM build configuration also sets `SWIFT_TREAT_WARNINGS_AS_ERRORS` to NO (otherwise ReactiveSwift 5 would not build). We just want Carthage to allow the same option to be overridden for projects that are still built using Carthage as opposed to SPM.

Thanks!!

Source/carthage/Build.swift
```diff
@@ -23,6 +23,23 @@ extension BuildOptions: OptionsProtocol {
...
+	}
+}
+
+/// Required to allow boolean argument to be passed.
+extension Bool: ArgumentProtocol {
+	public static let name = "bool"
+
+	public static func from(string: String) -> Bool? {
+		switch string.lowercased() {
+		case "yes", "1", "si", "y", "true", "ja":
```
NachoSoto: Is it standard to accept these strings in other languages?

Tests/XCDBLDTests/BuildArgumentsSpec.swift
```diff
@@ -61,10 +61,18 @@ class BuildArgumentsSpec: QuickSpec {
 				subject.derivedDataPath = ""
 			}
 
-			itCreatesBuildArguments("includes the the toolchain", arguments: ["-toolchain", "org.swift.3020160509a"]) { subject in
+			itCreatesBuildArguments("includes the toolchain", arguments: ["-toolchain", "org.swift.3020160509a"]) { subject in
 				subject.toolchain = "org.swift.3020160509a"
 			}
 
+			itCreatesBuildArguments("includes warnings as errors OFF", arguments: ["SWIFT_TREAT_WARNINGS_AS_ERRORS=NO"]) { subject in
```
NachoSoto: Maybe add another test to verify that the setting is not changed if left as `nil`?

Source/carthage/Build.swift
1oo7: Not sure, I can trim it to just English if that's preferred.

Tests/XCDBLDTests/BuildArgumentsSpec.swift
1oo7: OK, I added the nil test case, and it passes.

Source/carthage/Build.swift
1oo7: Well, to be on the safe side, I went ahead and removed the Spanish & German.

1oo7:
Hey so, if there's no objections, can we **pretty please** get this merged in? I realize there are quite a few PRs that haven't merged, but given this is a very minor non-breaking patch, and a kind of critical one to let folks build on Xcode 11, it would be super awesome if we could move forwards! You guys are great!

marcelofabri:
Can we revive this?

1oo7:
Why was this closed? I see that someone was merging README updates but not useful PRs like this... why?

PR #2945, opened 2020-03-03
Avoid unnecessary rebuilds when using --no-skip-current with --cache-builds

elliottwilliams:
When building the local project with `carthage build --no-skip-current --cache-builds`, Carthage writes a version file (since #2636). However, this file is never checked, and invoking Carthage again will rebuild the project even though the version file is valid.

I've added logic to Build.swift that checks for a valid version file before sending the `currentProducers`, analogous to what Carthage does for dependencies in `buildCheckedOutDependenciesWithOptions`.

Some changes made to support this:
- The logic for determining a project name based on its git origins is extracted to its own function in Git.swift.
- Version file APIs and `ProjectEvent`s printed to the console take a `String` dependency name rather than a `Dependency`. This allow them to support the local project, which does not have a corresponding `Dependency`.
- `GitURL.name` correctly handles repos that are at the "top level" of their server (i.e. "git@example.com:MyRepo.git"). Repos like this won't exist on GitHub but are possible on other servers, and a test case I wrote exposed this.

As always, happy to tweak this or break it into multiple PRs if necessary :)

### DIFF
--- a/IntegrationTests/Utilities/TestFramework.bash
+++ b/IntegrationTests/Utilities/TestFramework.bash
@@ -6,6 +6,10 @@ extract-test-frameworks-one-and-two() {
 	unzip ${BATS_TEST_DIRNAME:?}/../Tests/CarthageKitTests/fixtures/DependencyTest.zip 'DependencyTest/SourceRepos/TestFramework[12]/*' -d "${BATS_TMPDIR:?}"
 }
 
+extract-workspace-with-dependency() {
+	unzip -o ${BATS_TEST_DIRNAME:?}/../Tests/CarthageKitTests/fixtures/WorkspaceWithDependency.zip -d "${BATS_TMPDIR:?}"
+}
+
 branch-test-frameworks-one-and-two() {
 	directory_to_return_into="${PWD:?}"
 	branch="${1:?}" # parameter 1: branch name used in git repositories for both `TestFramework`s.
--- a/IntegrationTests/build.bats
+++ b/IntegrationTests/build.bats
@@ -24,12 +24,13 @@ EOF
     extract-workspace-with-dependency
     cd "${BATS_TMPDIR:?}/WorkspaceWithDependency"
     git init && git-commit 'Initialize project.'
+    git remote add origin "git@example.com:TestFramework1.git"
 
     run carthage build --no-skip-current --platform mac --cache-builds
     [ "$status" -eq 0 ]
-    [ "${lines[1]}" = "*** Invalid cache found for _Current, rebuilding with all downstream dependencies" ]
+    [ "${lines[1]}" = "*** Invalid cache found for TestFramework1, rebuilding with all downstream dependencies" ]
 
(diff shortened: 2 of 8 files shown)


tmspzz:
You're instructing carthage explicitly to build and cache the results. I don't agree with the change.

elliottwilliams:
Got it, no worries. One thing I hadn't thought about was how this would mean you can't make local changes to a project and then rebuild, which would cause false positives similar to #1785.

In your mind, how do the semantics of `build` change between the local project and the dependencies of the project? It sounds like the difference is that Carthage assumes that checked-out dependency code won't change after it builds, so skipping rebuilds makes sense. But the current project is liable to change, so --no-skip-current should always cause a rebuild.

tmspzz:
> It sounds like the difference is that Carthage assumes that checked-out dependency code won't change after it builds, so skipping rebuilds makes sense.

Why do you think this is so? You can still make local changes to projects in the checkout folder. Works even better if they checkout with `--use-submodules`

elliottwilliams:
> > It sounds like the difference is that Carthage assumes that checked-out dependency code won't change after it builds, so skipping rebuilds makes sense.
> 
> Why do you think this is so? You can still make local changes to projects in the checkout folder. Works even better if they checkout with `--use-submodules`

Right. There's nothing stopping you from making local changes to dependencies once they're in Checkouts, but if you do, `carthage build --cache-builds` won't honor those changes if it already has a cached binary for that version.

In https://github.com/Carthage/Carthage/issues/1785#issuecomment-282292563, @mdiep concluded that this caching behavior is a known limitation, and that Carthage should either document it or work around it (though this was never implemented).

This PR would give Carthage the same caching behavior for the local project as it already has for dependencies. But if the discrepancy is intentional, then I agree that we should drop this 🙂

tmspzz:
I'm on the fence here.  

On one side this is behavior changing but carthage will also notify that the version file is being used. Maybe @ikesyo or @jdhealy  can give their opinion too.

elliottwilliams:
Drawing from #1785, one thing that might make the ergonomics of this change safer is if we shipped this along with the implementation @BobElDevil [came up with](https://github.com/Carthage/Carthage/issues/1785#issuecomment-282052232) that [@mdiep approved of](https://github.com/Carthage/Carthage/issues/1785#issuecomment-282292563): 

> I think it would make sense to compare against the current 'subproject commit' value. That way we could also detect 'dirty' working trees as well and rebuild. (though with 'dirty' trees we would have to make sure that commit-dirty != commit-dirty).

I could change `versionFileMatches` so it returns false when the dependency's repo is dirty. This would mean that:
*  building with `--cache-builds --use-submodules` would invalidate cached versions when the submodule is dirty, fixing #1785, and that
* building with `--cache-builds --no-skip-current` would invalidate the cached current framework if the repo is dirty.

That said, I understand not wanting to change behavior here. Would be interested to hear more opinions 🙂

Issue #2941, opened 2020-02-24
Carthage "make install" does not handle /usr/local/bin not existing


aeberbach:
On a freshly installed Mac where Carthage is one of the first packages installed this directory may not exist. Result of make install is then a binary named '/usr/local/bin' rather than a binary named carthage _in_ /usr/local/bin.

Manual installation is necessary because of https://github.com/Carthage/Carthage/issues/2887

* carthage install method: [ ] .pkg, [ ] homebrew, [ X ] source
* `which carthage`: N/A
* `carthage version`: 0.33 (master)
* `xcodebuild -version`: N/A
* Are you using `--no-build`? N/A
* Are you using `--no-use-binaries`? N/A
* Are you using `--use-submodules`? N/A
* Are you using `--cache-builds`? N/A
* Are you using `--new-resolver`? N/A

**Cartfile**
N/A

**Carthage Output**
N/A

**Actual outcome**
Carthage makefile creates a binary named 'bin' in the /usr/local directory.

**Expected outcome**
Makefile should check for the presence of the directory, create it if it does not exist, and then create the Carthage binary in the right place. Or better, look for likely options that are already in the path and allow the install to choose one or too choose to create /usr/local/bin if that is what they want.

tmspzz:
Thanks for reporting. It would be great if you could make a PR to fix the issue.

jonnybach:
The referenced PR should resolve this issue.

PR #2952, opened 2020-03-17
Makefile - Check for and create binaries directory in install target

jonnybach:
Addresses Carthage/Carthage#2941

Modified the `install` make target so that the `BINARIES` directory is check for existence and create if not.


tmspzz:
@jonnybach thanks! 🎉

PR #2965, opened 2020-04-10
Don't filter workspaces that have valid schemes

MikeSilvis:
Presently with Carthage if you scheme is located inside the workspace, Carthage is unable to find a valid scheme and errors out.

See the following example of Alamofire where I moved the scheme into the workspace container, and out of the project:

https://github.com/Alamofire/Alamofire/commit/c32797bbaeec605c8ce647a6443718aa1794498d

Now when I try to build, I get the following:

```
2.4.4 ExampleCarthagePull
 ➲  carthage update
*** Fetching Alamofire
*** Checking out Alamofire at "93cc83453ffee1c6e9aee7bde60bb3732b15b858"
*** xcodebuild output can be found in /var/folders/td/0lcyjgbx5cz4tj2tgmckz8mc0000gn/T/carthage-xcodebuild.LXYoZ1.log
*** Skipped building Alamofire due to the error:
Dependency "Alamofire" has no shared framework schemes

If you believe this to be an error, please file an issue with the maintainers at https://github.com/MikeSilvis/Alamofire/issues/new
```

### DIFF
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -276,8 +276,8 @@ public func schemesInProjects(_ projects: [(ProjectLocator, [Scheme])]) -> Signa
 			switch project {
 			case .projectFile where !schemes.isEmpty:
 				return true
-            case .workspace where !schemes.isEmpty:
-                return true
+ 			case .workspace where !schemes.isEmpty:
+				return true
 			default:
 				return false
 			}
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -276,7 +276,8 @@ public func schemesInProjects(_ projects: [(ProjectLocator, [Scheme])]) -> Signa
 			switch project {
 			case .projectFile where !schemes.isEmpty:
 				return true
-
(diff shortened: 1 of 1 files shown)


tmspzz:
Can you please provide more context on this change? I understand the effect but I would like to know more.

MikeSilvis:
@tmspzz i updated the description to provide additional context. Thanks!

Source/CarthageKit/Xcode.swift
```diff
@@ -276,7 +276,8 @@ public func schemesInProjects(_ projects: [(ProjectLocator, [Scheme])]) -> Signa
 			switch project {
 			case .projectFile where !schemes.isEmpty:
 				return true
-
+            case .workspace where !schemes.isEmpty:
```
DavidBrunow, line 279: The indentation is a little off here -- Carthage uses tabs rather than spaces so I think that is the issue.

MikeSilvis:
Any update on this? Looks like another framework is having an issue

MikeSilvis:
@DavidBrunow any update?

philips77:
I manage to fix this issue on my side. Check out the commit that fixed it. I tried and Carthage import works now.

PR #2966, opened 2020-04-10
Update Carthage version to match production

DavidBrunow:


DavidBrunow:
This also fixes an integration test failure on `master` in `archive.bats`.

jdhealy:
Thanks @DavidBrunow! ✨

PR #2967, opened 2020-04-10
Fix integration test that is failing on newer versions of Xcode

DavidBrunow:
Newer versions of Xcode don't have an iPhone 6s by default so this should be a good compromise that works on older and newer versions of Xcode.


tmspzz:
@DavidBrunow Thanks 👍

PR #2971, opened 2020-04-13
Publish version functions

olejnjak:
Hi,

for a tool I am using, I would love to use version producers that also _CarthageKit_ uses, unluckily they are `internal`, is there an option to make them `public`? 🙂 

Thanks

### DIFF
--- a/Source/CarthageKit/SwiftVersionError.swift
+++ b/Source/CarthageKit/SwiftVersionError.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-internal enum SwiftVersionError: Error, Equatable {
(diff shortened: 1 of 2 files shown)


elliottwilliams:
I think there's an argument for keeping CarthageKit's API surface as small as possible, so that methods like these can freely change in the future without impacting tools that consume them.

Are there other tools you know of that need these methods / currently duplicate them? If multiple tools would benefit from having these be public, maybe that's a sign that they ought to be part of CarthageKit's API.

tmspzz:
It think that's fine given that it's understood that there's no commitment to keeping the interfaces stable.

PR #2972, opened 2020-04-14
Swift 5.2 migration

olejnjak:
This PR demonstrates what changes would be necessary (or useful) to migrate Carthage to use Swift 5.2. TL;DR not much, just removing custom `Result` dependency would be fine. And for further development it would be very useful as in Swift 5 custom `Result` can be easily mistaken with the built-in type.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/Carthage/Commandant.git",
         "state": {
           "branch": null,
-          "revision": "2cd0210f897fe46c6ce42f52ccfa72b3bbb621a0",
-          "version": "0.16.0"
+          "revision": "ab68611013dec67413628ac87c1f29e8427bc8e4",
+          "version": "0.17.0"
         }
       },
       {
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/Quick/Nimble.git",
         "state": {
           "branch": null,
-          "revision": "f8657642dfdec9973efc79cc68bcef43a653a2bc",
-          "version": "8.0.2"
+          "revision": "7fd118ec8795888bcbbebc1a41f6984454c4cd6f",
+          "version": "8.0.7"
         }
       },
       {
@@ -42,8 +42,8 @@
         "repositoryURL": "https://github.com/Quick/Quick.git",
         "state": {
           "branch": null,
-          "revision": "94df9b449508344667e5afc7e80f8bcbff1e4c37",
-          "version": "2.1.0"
+          "revision": "33682c2f6230c60614861dfc61df267e11a1602f",
+          "version": "2.2.0"
         }
       },
       {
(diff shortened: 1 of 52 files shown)


elliottwilliams:
In order for a Homebrew formula to support an OS, it has to be able to build on it. In other words, even if you keep the deployment target at macOS 10.13, upgrading to Swift 5 would mean that [Carthage's brew formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/carthage.rb) has to drop support for 10.13, because it can no longer build on a version of Xcode that runs on 10.13.

The last Xcode that runs on macOS 10.13 is Xcode 10.1, which uses Swift 4.2.1, so that's where things are stuck, unfortunately :(

@jdhealy [mentioned last fall](https://github.com/Carthage/Carthage/pull/2863#issuecomment-529561539) that it might be worth dropping High Sierra support once Catalina gets some adoption. Perhaps now's a good time to look into that?
> With Homebrew's criteria necessitating us to build with Swift 4.2.X in order to bottle binary releases for High Sierra — and High Sierra being just one behind the latest macOS release — I'm not comfortable dropping support for High Sierra until we see some significant numbers from Catalina.

thedavidharris:
The deadline for submitting this with the iOS 13 SDK is June 30th (see https://developer.apple.com/news/?id=03262020b), so I think that would make supporting Xcode 10.x unnecessary? That would be a better barometer to shift things.

jdhealy:
I imagine for people migrating (as it’s a process) untouched-for-periods-of-time apps that don‘t ’just compile‘ under Xcode 11.X — having Carthage available on the still-using-older-tools portion of that path would be important.

That said, there’s probably a middle ground to be had… in a number of ways. :sparkles: One great contribution — if someone were willing to explore it — would be examining a solid way for High-Sierra–based Homebrew-consuming Carthage users to be informed via Homebrew formula methods of ways to install Carthage (maybe older Carthage versions) outside the Homebrew ways.

thedavidharris:
@jdhealy Agreed, but given Carthage uploads binaries as part of releases I think it would be fair to expect them to manually install a working version and have the info appropriately available. 

https://github.com/Homebrew/homebrew-core/blob/master/Formula/carthage.rb just calls `prefix_install` so you could put some OS check in there and direct a STDOUT link to the last compatible version, if that's acceptable. Would need a roadmap/timing for when the last High Sierra supported release is though to merge it at the appropriate time.

eliperkins:
[I just released a v0.14.0 of Tentacle with Swift 5.2 support](https://github.com/mdiep/Tentacle/releases/tag/0.14.0), btw.

olejnjak:
Yep, I already removed my fork of Tentacle and will update this PR 🙂

jdhealy:
@eliperkins @olejnjak Appreciate that, great stuff! ☀️🚀 

(For those following along sporadically — that Tentacle release is **not** really the thing we're waiting for to move this Carthage migration the finish line; [other factors at play](https://github.com/Carthage/Carthage/pull/2972#issuecomment-613597401)).

PR #2980, opened 2020-04-28
Prevent directory traversal (in more piecemeal way) allowing «project-directory» flag to function.

jdhealy:
Also, sanitize [nul characters](https://en.wikipedia.org/wiki/Null_character) in `GitURL`s (transposing them to «␀».)

The cases in the modified test are — given a URL is specified and provided is a string without URL scheme — pathological, and as such the result for them from `GitURL.name` are maybe unintuitive, but decidedly safe.

The flag «--project-directory» clashed with the formerly placed (but never shipped in a tagged Carthage release) invocation of `FileManager.currentDirectoryPath`.

To avoid the clash, we settle on a less intuitive, but most importantly safe change — _to be clear_ a *breaking change* for some pathological URL-scheme–less or nul-character–containing Cartfile lines and `GitURL`s.

### DIFF
--- a/Source/CarthageKit/GitURL.swift
+++ b/Source/CarthageKit/GitURL.swift
@@ -47,21 +47,27 @@ public struct GitURL {
 
 	/// The name of the repository, if it can be inferred from the URL.
 	public var name: String? {
-		let absoluteURLString: String
-		if urlString.hasPrefix(".") {
-			absoluteURLString = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
-				.appendingPathComponent(urlString)
-				.standardizedFileURL
-				.absoluteString
-		} else {
-			absoluteURLString = urlString
-		}
-		let components = absoluteURLString.split(omittingEmptySubsequences: true) { $0 == "/" }
-
-		return components
+		let lastComponent = urlString
+			.replacingOccurrences(of: "\u{0000}", with: "\u{2400}") // can’t have those
+			.split(omittingEmptySubsequences: true) { $0 == "/" }
 			.last
 			.map(String.init)
-			.map(strippingGitSuffix)
+
+		precondition(lastComponent != .some("")) /* because of `omittingEmptySubsequences` */
+		
+		/// Potentially used to prevent backwards or noop directory traversal via «FULL STOP» characters…
+		/// …by deploying the «FULLWIDTH FULL STOP» character.
+		var replacementForEntirelyCharactersOfFullStop: [Character] = []
+		for char in (lastComponent ?? "") {
+			guard char == "." else { replacementForEntirelyCharactersOfFullStop = []; break }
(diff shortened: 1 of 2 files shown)


Source/CarthageKit/GitURL.swift
```diff
@@ -47,21 +47,27 @@ public struct GitURL {
...
-
-		return components
+		let lastComponent = urlString
+			.replacingOccurrences(of: "\u{0000}", with: "\u{2400}") // can’t have those
+			.split(omittingEmptySubsequences: true) { $0 == "/" }
 			.last
 			.map(String.init)
-			.map(strippingGitSuffix)
+
+		precondition(lastComponent != .some("")) /* because of `omittingEmptySubsequences` */
```
tmspzz, line 56: sure we want this in release? `name` is optional

Source/CarthageKit/GitURL.swift
```diff
@@ -47,21 +47,27 @@ public struct GitURL {
...
+		
+		/// Potentially used to prevent backwards or noop directory traversal via «FULL STOP» characters…
+		/// …by deploying the «FULLWIDTH FULL STOP» character.
+		var replacementForEntirelyCharactersOfFullStop: [Character] = []
+		for char in (lastComponent ?? "") {
+			guard char == "." else { replacementForEntirelyCharactersOfFullStop = []; break }
+			replacementForEntirelyCharactersOfFullStop.append("\u{FF0E}")
```
tmspzz, line 63: imho this is confusing as result, maybe just replace with `_` ?

Source/CarthageKit/GitURL.swift
jdhealy: I'll address the optionality first — (then, your deeper point about when `urlString` is empty):

If there’s no last path component (e.g. `urlString` is `//` · ⧼still a pathological case⧽), then `lastComponent` is `nil` and the precondition doesn't get hit.

For when `urlString` is empty (or similar whitespace), then in regular Carthage-the-tool usage, our Cartfile-from-NSScanner usage should preclude that from being hit by error-ing first with something like `Parse error: empty or unterminated string after dependency type in line: git "	" "master"`.

So, all good 👍

Source/CarthageKit/GitURL.swift
tmspzz: clarified over slack

jdhealy: So, that character, as an alternative to the «FULLWIDTH FULL STOP» character:

The «FULLWIDTH FULL STOP» character has a width to it (in macOS Terminal’s default state, in Finder, in others) that makes it noticeably different.

IMO that ‘different, but able to correspond it to what’s listed in a Cartfile’ nature give it utility. Since you mentioned you're good with the PR in overall approval, let’s get it merged 🚀

Issue #2925, opened 2020-01-23
is this project alive?


jafara:
no releases from last year, is it supported or should i move to or spm?

codeslubber:
Never mind releases, our build now takes an indeterminate amount of time because this thing hangs. Every single time.. seriously one of the worst tool breakdowns ever. Do NOT start using this unless the only other alternative is a year on an aleutian island scrounging for rat meat.

1oo7:
Well, the PRs have gotten some code reviews, but no merges in awhile. Not clear why.

evandcoleman:
Any updates here? Are you in need of new maintainers? I'd be happy to volunteer my time.

tmspzz:
@evandcoleman help is always appreciated, especially in triaging new issue

mozeryansky:
@tmspzz I'm waiting on the the issues that have been fixed to be released. When could we expect that? We're stuck using an old Xcode because Carthage isn't supporting prebuilt binaries.

michallaskowski:
I checked out the code and I am trying to add at least support for binary xcframeworks, based on open PRs from @tmspzz (thanks a lot for that work), cause it seems in the full scope of xcframework support they will not be merged soon. And I could use some xcframeworks for Unity and Kotlin Mutliplatform generated binaries. 

I always like to test my code, so I wanted to check it out, and apply some unit tests. It seems Project.swift could use some breakdown, into smaller classes with clear set of responsibilities (DependencyInstaller class could be one). This is what I am struggling most with any change here, and probably something that increases the level of effort for anyone willing to contribute.

Anyway, I got the feeling over the years of using Carthage, and reading through issues here, that changes are not always really welcome. Maybe this is just a feeling, maybe things changed, but if I am going to work on it, refactor, improve a bit in terms of SOLID, commit my personal time, are those changes going to be approved?

tmspzz:
@michallaskowski changes are welcome, however since I'm "in charge" of that PR if you want to continue I can give you some pointers on the direction to take. You can contact me [on twitter](https://twitter.com/tmpz) and we'll figure out how to sync.

PR #2981, opened 2020-04-29
Feature/xcframeworks binary download preparing refactor

michallaskowski:
As discussed in #2925
The refactor in itself gives this project and me nothing, but should enable easier addition of features (in my case I need binary xcframeworks) - by modularization and clear(er) set of responsibilities => allowing to unit tests particular classes => avoiding bugs and helping to understand the code faster. (In any particular order really, 3am for me now)

From Project.swift I carved out ~450 lines of code into separate classes
- BinaryInstaller
- BinaryFrameworkDownloader - download binary frameworks. Could be further split into GitHub downloader (or extension on Tentacle.Client) and the rest.
- FrameworkInformationProvider - provides info about frameworks already on the disk.


Source/CarthageKit/Project.swift
```diff
@@ -106,23 +106,38 @@ public final class Project { // swiftlint:disable:this type_body_length
...
+        set {
+            binaryInstaller.useNetrc = newValue
+        }
+    }
 
 	/// Sends each event that occurs to a project underneath the receiver (or
 	/// the receiver itself).
-	public let projectEvents: Signal<ProjectEvent, NoError>
-	private let _projectEventsObserver: Signal<ProjectEvent, NoError>.Observer
+    public var projectEvents: Signal<ProjectEvent, NoError>
```
tmspzz, line 120: private (set) ?

tmspzz:
Thanks for the PR and the refactoring work! This will certainly help going forward. I left one comment.

Source/CarthageKit/Project.swift
michallaskowski: hm, I think it is some leftover, it can be `let` as it was.

DavidBrunow:
I think all the new indentation uses spaces while the rest of Carthage uses tabs.

michallaskowski:
Yeah, I also noticed after the opening a PR. It is the only project so far I have seen with tabs, I am searching how to change Xcode to use tabs, per project, not for all of them.

michallaskowski:
@tmspzz thanks for the approval. What happens now? Is there a required set of approvers? I see some other PRs that are approved but sitting in open state, so best to be sure.
If this is merged I will prepare a draft with Readme updated, and then probably would start picking commits from your work and writing tests for binary xcframeworks. wdyt?

tmspzz:
@michallaskowski it would actually be better to merge this into my fork to minimize risk for the next release. For the steps after that it's best if we continue discussion on Twitter

jdhealy:
Appreciate you crafting on this @michallaskowski ☀️👍

tmspzz:
@michallaskowski changed destination branch from master to feature branch

PR #2987, opened 2020-05-14
Attempt to remove invalid archs by intersecting VALID_ARCHS and ARCHS

### LINKED ISSUE #2956: Why does Carthage use VALID_ARCHS and not ARCHS in copy-frameworks?
I skipped the first few questions because I'm not a direct user of Carthage.

We created an SDK for Android and iOS. Our iOS  SDK supports both CocoaPods and Carthage, our SDK is delivered both ways as a fat framework. 

Today one of our customers tried to integrate using Carthage, everything worked fine until he uploaded the app to TestFlight. The app was rejected with the following error:

```
ITMS-90381: Too many symbol files - These symbols have no corresponding slice in any binary
```

After some researching we found out that the resulting binary of our SDK in the app had architectures that the app does not use. I looked into the code of Carthage, and noticed that the `copy-frameworks` step does do stripping of the binary, but it bases the stripping on the `VALID_ARCHS` environment variable. Our customer had more archs in `VALID_ARCHS` than he had in `ARCHS`, which resulted in Carthage keeping the archs in our SDK binary that were in `VALID_ARCHS`, but the app itself was not build for those archs, because `ARCHS` contains less archs than `VALID_ARCHS`, which causes TestFlight to reject it.

Is there a good reason for Carthage to only strip archs that are not in `VALID_ARCHS`, and not the archs that are not in  `ARCHS`?

**Actual outcome**
Carthage keeps architectures in frameworks that the app is not compiled for, because it looks at the `VALID_ARCHS` variable.

**Expected outcome**
Carthage should strip the architectures based on `ARCHS` because those are the architectures the app is compiled for.

jerbob92:
This PR should fix #2956 
I'm not a Swift developer so I'm not sure this is the proper way to write this code. 

This PR makes sure that we intersect VALID_ARCHS and ARCHS so we only keep the ARCHS that we are actually building.

### DIFF
--- a/Source/carthage/CopyFrameworks.swift
+++ b/Source/carthage/CopyFrameworks.swift
@@ -165,9 +165,23 @@ private func frameworksFolder() -> Result<URL, CarthageError> {
 }
 
 private func validArchitectures() -> Result<[String], CarthageError> {
-	return getEnvironmentVariable("VALID_ARCHS").map { architectures -> [String] in
-		return architectures.components(separatedBy: " ")
-	}
+    let validArchs = getEnvironmentVariable("VALID_ARCHS").map { architectures -> [String] in
+        return architectures.components(separatedBy: " ")
+    }
+
+    if validArchs.error != nil {
+        return validArchs
+    }
+
+    let archs = getEnvironmentVariable("ARCHS").map { architectures -> [String] in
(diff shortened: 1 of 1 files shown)


jerbob92:
Notstale

tmspzz:
Thanks!

PR #2991, opened 2020-05-25
Support no_proxy/NO_PROXY environment variables

okaverin:
* As support for http_proxy/https_proxy environment already exists, there was also a need to support the no_proxy/NO_PROXY override.
* `kCFNetworkProxiesExceptionsList` is available only on macOS, hence the conditional statements.
* Made sure that downloaded binaries also use proxied environment, which was not the case.
* Moved `URLSession.proxiedSession` into Proxy.swift where it seems to belong a bit more.

### DIFF
--- a/Source/CarthageKit/GitHub.swift
+++ b/Source/CarthageKit/GitHub.swift
@@ -161,12 +161,3 @@ extension Client {
 		}
 	}
 }
-
-extension URLSession {
-	public static var proxiedSession: URLSession {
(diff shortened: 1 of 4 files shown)


jdhealy:
So, not sure I have the most context on this — how closely/loosely does this correspond with with `cURL`’s NO_PROXY environment variable handling?

tmspzz (approved):

tmspzz:
Thanks!

PR #2999, opened 2020-06-14
Use updated dependencies when building with Swift 5, but keep backwards-compatibility with Swift 4

elliottwilliams:
This is a complement to #2972, but doesn't make any breaking changes to CI / homebrew. It starts the work of upgrading to Swift 5.2 by using updated dependencies when Carthage is built with a 5.x compiler. These changes are helpful for users who depend on Carthage as a SwiftPM package [1], and should make it a little easier to drop support for Swift 4.x in the future.

[1]: Carthage doesn't build with Commandant >= 0.17.0, because that version drops support for antitypical/Result. You can manually pin Commandant to an older version, but then you can't depend on tools that have upgraded to the newer version of Commandant, like [SwiftLint](https://github.com/realm/SwiftLint/blob/bda1cd56e0bbcd4a68828394137e32d17c92963b/Package.swift#L17). So this PR helps keep users out of dependency hell when they're [using SwiftPM to version their build tools](https://artsy.github.io/blog/2019/01/05/its-time-to-use-spm/).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,23 @@
 // swift-tools-version:4.2
 import PackageDescription
 
+let compilerSpecificDependencies: [Package.Dependency]
+#if compiler(>=5)
+compilerSpecificDependencies = [
+    .package(url: "https://github.com/antitypical/Result.git", from: "5.0.0"),
+    .package(url: "https://github.com/Carthage/ReactiveTask.git", .branch("release-0.17.0")),
+    .package(url: "https://github.com/Carthage/Commandant.git", from: "0.17.0"),
+    .package(url: "https://github.com/ReactiveCocoa/ReactiveSwift.git", from: "6.0.0"),
+]
+#else
+compilerSpecificDependencies = [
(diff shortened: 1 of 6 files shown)


Package.swift
```diff
@@ -1,6 +1,23 @@
 // swift-tools-version:4.2
 import PackageDescription
 
+let compilerSpecificDependencies: [Package.Dependency]
+#if compiler(>=5)
+compilerSpecificDependencies = [
+    .package(url: "https://github.com/antitypical/Result.git", from: "5.0.0"),
+    .package(url: "https://github.com/Carthage/ReactiveTask.git", .branch("release-0.17.0")),
```
elliottwilliams, line 8: `release-0.17.0` is the branch of https://github.com/Carthage/ReactiveTask/pull/120, which has been open for a year :)

AFAICT, nothing is preventing that PR from being merged, there just hasn't been a reason to yet. It needs to get merged before this PR, since [versioned packages can't have branch dependencies](https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescription.md#package-dependency-requirement).

PR #3006, opened 2020-06-30
Failed to build on Xcode 12 and macOS 11.0

eofs:
The project failed to build on Xcode 12 and macOS 11.0 due to following error:

```
/Users/eofs/Projects/Xcode/Carthage/Source/XCDBLD/SDK.swift:80:3: error: call can throw, but errors cannot be thrown out of a global variable initializer
                KeyValuePairs.reduce([
                ^
```

This prevented installing latest Carthage release using `brew`.

Calling `reduce()` on the collection itself instead allowed the project to build successfully.


jdhealy:
Actually committed to `master` as https://github.com/Carthage/Carthage/commit/a33d3483b31e28e0488ea51e3efb6b44025f1b5f (via force push) 

###### _because I typed non-existent `KeyValuePair` instead of `KeyValuedPairs` (that'll teach me to not to forgo tests typing haphazardly at four in the morning)_  😓

eofs:
Thanks! :)

Issue #3013, opened 2020-07-16
Make dependencies sorted in the version file


peteranny:
* carthage install method: [ ] .pkg, [v] homebrew, [ ] source
* `which carthage`: /usr/local/bin/carthage
* `carthage version`: 0.34.0
* `xcodebuild -version`:
Xcode 10.1
Build version 10B61
* Are you using `--no-build`? NO
* Are you using `--no-use-binaries`?  NO
* Are you using `--use-submodules`?  NO
* Are you using `--cache-builds`?  NO
* Are you using `--new-resolver`?  NO

**Cartfile**
```
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json" == 6.27.0
```

**Carthage Output**
Run the bootstrapping for two times
```
# The first time
carthage bootstrap # The first bootstrap
cat Carthage/Build/.FirebaseAnalyticsBinary.version

# The second time
carthage bootstrap # The first bootstrap
cat Carthage/Build/.FirebaseAnalyticsBinary.version
```

Compare the version file **.FirebaseAnalyticsBinary.version** before and after.
```
{
  "Mac" : [

  ],
  "watchOS" : [

  ],
  "tvOS" : [

  ],
  "commitish" : "6.28.1",
  "iOS" : [
    {
      "name" : "nanopb",
      "hash" : "bd968d91c019c9f06c723ce8252f45ef15363b592d61bba2f56a3e1f943bbd6a",
      "linking" : "dynamic"
    },
    {
      "name" : "FirebaseCore",
      "hash" : "e9f35f56b1b2ffe905c1cb0cc716591507a8963e838488967fe5edcf9f16c32f",
      "linking" : "dynamic"
    },
    {
      "name" : "Firebase",
      "hash" : "a8669248da9519e9955e00448e518e1ca6becb32689eedddde237e19ad1136ae",
      "linking" : "dynamic"
    },
    {
      "name" : "FirebaseCoreDiagnostics",
      "hash" : "b9325775b24af655755a4cd25e7c74c7e69096900d48735b9e6633e52cff2796",
      "linking" : "dynamic"
    },
    {
      "name" : "FIRAnalyticsConnector",
      "hash" : "10dc1b7749e8dba04866da992c916b8a58c3254865f39e7b373896a287334c90",
      "linking" : "dynamic"
    },
    {
      "name" : "FirebaseAnalytics",
[...]

PR #3015, opened 2020-07-20
Sort frameworks in version file

peteranny:
This is enhancement as mentioned in #3013

Sort the frameworks by name in the version file
So we can easily identify which frameworks are altered when situations like running `--update`


tmspzz:
@peteranny  thanks for the patch! 🎉

PR #3024, opened 2020-07-30
Fix makefile

cfelder:
This patch fixes the usage of ``PREFIX`` and ``BINARIES_FOLDER`` environments in the Makefile.
Furthermore it removes ``sudo`` from the Makefile to allow unprivileged installs and avoid hidden
usage of sudo. For privileged installs simply call:

```bash
$ sudo make install
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,10 @@ MKDIR=mkdir -p
 SUDO=sudo
 CP=cp
 
+ifdef DISABLE_SUDO
+override SUDO:=
+endif
+
 .PHONY: all clean install package test uninstall xcconfig xcodeproj
 
 all: installables
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ INTERNAL_PACKAGE=CarthageApp.pkg
 OUTPUT_PACKAGE=Carthage.pkg
 
 CARTHAGE_EXECUTABLE=./.build/release/carthage
-BINARIES_FOLDER=/usr/local/bin
+BINARIES_FOLDER=$(PREFIX)/bin
 
 SWIFT_BUILD_FLAGS=--configuration release -Xswiftc -suppress-warnings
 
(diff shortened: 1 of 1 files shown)


tmspzz:
Thanks for opening a PR, but now you're giving privileged to the entire `install` target. Not just the `cp`

cfelder:
I see `installables` will be run with sudo as well.

```
install: installables
        if [ ! -d "$(BINARIES_FOLDER)" ]; then $(MKDIR) "$(BINARIES_FOLDER)"; fi
        $(CP) -f "$(CARTHAGE_EXECUTABLE)" "$(BINARIES_FOLDER)"
```

but shouldn't

```bash
$ make
$ sudo make install
```

just create the directory and copy the file using sudo? This is convenient for building several software.

imho I prefer to build and install unprivileged. I would like to have an option to install without using `sudo` though.

tmspzz:
This is the problematic line: https://github.com/Carthage/Carthage/pull/3024/files#diff-354f30a63fb0907d4ad57269548329e3R25

cfelder:
This PR does not remove sudo from the Makefile anymore but it allows disabling the usage of sudo using

    $ make install SUDO=false

nedmiz-6habsu-zyJcog (approved):

tmspzz:
The solution is still not viable as now, anyone can inject another command instead of `sudo`

cfelder:
> The solution is still not viable unfortunately. Now anyone can inject another command instead of `sudo`

May I kindly ask you to elaborate the disadvantage of that?

If one wants intentionally to do something different at this point, this is possible anyway.

tmspzz:
> May I kindly ask you to elaborate the disadvantage of that?

Sure, sorry for the short answer.

> If one wants intentionally to do something different at this point, this is possible anyway.

This is not the concern. The concern is when this script is used my automated system on CI or things like brew running on the user's machine.

cfelder:
Would you agree on using a different flag, e.g:

    make install DISABLE_SUDO=1

which simply overwrites SUDO with empty string and not using ``SUDO?=``

This way SUDO is just overwritten internally but not externally.

tmspzz:
> Would you agree on using a different flag, e.g:
> 
> ```
> make install DISABLE_SUDO=1
> ```
> 
> which simply overwrites SUDO with empty string and not using `SUDO?=`
> 
> This way SUDO is just overwritten internally but not externally.

This sounds reasonable to me

tmspzz (approved):

REGNANTGODDESSLISSA (commented):
- [ ]

PR #3035, opened 2020-08-24
Remove the cache if unarchiving was failed

tattn:
I found a cache-related bug.

I downloaded a zip file of the binary framework using Carthage, but it was failed because the zip file was broken.
So I corrected the zip file, and then I tried to download the framework.
But it occurred the same error.

The error was following:
```
*** xcodebuild output can be found in /var/folders/z3/lyh75tws0hg4nsp2wrg6gxnw0000gq/T/carthage-xcodebuild.OgWAAr.log
*** Downloading binary-only framework test at "file:///Users/<username>/carthage_binary_bug/test.json"
A shell task (/usr/bin/env unzip -uo -qq -d /var/folders/z3/lyh75tws0hg4nsp2wrg6gxnw0000gq/T/carthage-archive.559JOD /Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip) failed with exit code 9:
[/Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip or
        /Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip.zip, and cannot find /Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip.ZIP, period.
```

Reproduce:
```
$ git clone https://github.com/tattn/carthage_binary_bug
$ cd carthage_binary_bug
$ vi test.json # modify the path
$ carthage bootstrap # occur the error, because the test.zip is broken.
$ mv test.zip test3.zip
$ mv test2.zip test.zip # replace the broken file with the corrected file.
$ carthage bootstrap # occur the error, but the test.zip is corrected...
```

I changed to remove the broken cache if unarchiving was failed.
If there's a better way to fix it, I'll correct it.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -1011,7 +1011,12 @@ public final class Project { // swiftlint:disable:this type_body_length
 			.flatMap(.concat) { semanticVersion, frameworkURL in
 				return self.downloadBinary(dependency: Dependency.binary(binary), version: semanticVersion, url: frameworkURL)
 			}
-			.flatMap(.concat) { self.unarchiveAndCopyBinaryFrameworks(zipFile: $0, projectName: projectName, pinnedVersion: pinnedVersion, toolchain: toolchain) }
+			.flatMap(.concat) { zipFile in
+				self.unarchiveAndCopyBinaryFrameworks(zipFile: zipFile, projectName: projectName, pinnedVersion: pinnedVersion, toolchain: toolchain)
+ 					.on(failed: { _ in
+						try? FileManager.default.removeItem(at: zipFile)
+					})
+			}
 			.flatMap(.concat) { self.removeItem(at: $0) }
 	}
 


tmspzz:
Doesn't it make more sense to delete before unarchiving instead? 

Would the current change still propagate the error or hide it?

tattn:
@tmspzz 
> Doesn't it make more sense to delete before unarchiving instead?

Normally, the file specified in the URL will not change.
So it's more convenient to use the cache for the second time.

If the result changes for the same URL, as in this irregular case, I think it's better to delete the cache when it fails, and make it work the next time.

This irregular case often occurs when downloading frameworks that require authentication, such as .netrc.
Entering the wrong password will download an error message instead of the zipped framework.
That will treat the text as a zip file.

This change results in the following:
```
$ carthage bootstrap # occur the error
$ # <fix the the broken file>
$ carthage bootstrap # successful
```

tmspzz:
Sounds good, however carthage should still fail with an error or at least print a notice that XYZ failed unarchiving

tattn:
@tmspzz 
After this change, carthage still outputs the following error:

```
$ carthage bootstrap
*** xcodebuild output can be found in /var/folders/z3/lyh75tws0hg4nsp2wrg6gxnw0000gq/T/carthage-xcodebuild.OgWAAr.log
*** Downloading binary-only framework test at "file:///Users/<username>/carthage_binary_bug/test.json"
A shell task (/usr/bin/env unzip -uo -qq -d /var/folders/z3/lyh75tws0hg4nsp2wrg6gxnw0000gq/T/carthage-archive.559JOD /Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip) failed with exit code 9:
[/Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip or
        /Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip.zip, and cannot find /Users/<username>/Library/Caches/org.carthage.CarthageKit/binaries/test/1.0.0/test.zip.ZIP, period.

$ # <fix the problem such as a wrong password>
$ carthage bootstrap # successful
```

This error tells us which framework has problems, so I’m sure that we solve the problem.
Do you think we need a special error log for this error?

tmspzz:
@tattn No, I think this is enough. Thanks for the contribution!

REGNANTGODDESSLISSA (commented):
- [ ]

PR #3045, opened 2020-09-12
Update README to reflect the build requirements

chrstphrchvz:
I notice that Carthage 0.35.0 does not build with Xcode 9.4.1:
```
error: package at '/opt/local/var/macports/build/_Users_travis_build_macports_macports-ports_devel_carthage/carthage/work/Carthage-0.35.0' requires a minimum Swift tools version of 4.2.0 (currently 4.1.0)
```

My guess is the requirement was raised by #2797, and that the README should reflect this new requirement. Not sure if similar changes should be made elsewhere.


tmspzz:
Thanks for spotting this 👍

PR #3025, opened 2020-07-31
Don't use replaceItem if source and dest are on different filesystems. Fixes #3003

rudedogdhc:
`FileManager.replaceItem(_:withItemAt:backupItemName:options:)` does not work if the source and destination are on different volumes. This fix checks that the source and dest are on the different volumes and falls back to using `FileManager.copyItem(at:to:)` if that is the case.

PR #3047, opened 2020-09-17
Update createTemporaryDirectoryWithTemplate to create temp dirs on th…

rudedogdhc:
Pull request #3025 fixed an issue with installing frameworks while avoiding cross-device link errors. 

It seems that in complex projects, this is insufficient, as a copy-frameworks command with many input frameworks will sometimes still fail with errors such as 

```
copyItem failed: Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"
file:///var/folders/lq/yvbxlp_j64xfh9km1md802x00000gn/T/carthage-lipo-MIM7DX/SimpleKeychain.framework
file:///projects/build/dave/dmpremier-gvncjwnlubknpkfmnxmiipekxcnu/Build/Products/Debug-iphonesimulator/dha.app/Frameworks/SimpleKeychain.framework/
```

In this case, the source file in the temporary directory under `/var/folders` seems to disappear before the `copyItem` method gets called. It isn't always the same file that disappears; re-running the copy-frameworks command sometimes works and sometimes fails with a different file, which suggests that there is some kind of race condition if the install has to fall back to `copyItem` instead of `replaceItem`.

This patch creates a temporary directory that is on the same filesystem as the destination URL, thus ensuring that there is never an issue with cross-device links.

### DIFF
--- a/Source/CarthageKit/Archive.swift
+++ b/Source/CarthageKit/Archive.swift
@@ -56,7 +56,7 @@ private let archiveTemplate = "carthage-archive.XXXXXX"
 /// Unzips the archive at the given file URL into a temporary directory, then
 /// sends the file URL to that directory.
 private func unzip(archive fileURL: URL) -> SignalProducer<URL, CarthageError> {
-	return FileManager.default.reactive.createTemporaryDirectoryWithTemplate(archiveTemplate)
+	return FileManager.default.reactive.createTemporaryDirectoryWithTemplate(archiveTemplate, destinationURL: fileURL)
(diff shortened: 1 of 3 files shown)


tmspzz (approved):

tmspzz:
Thanks 🥳

vincentisambart:
Unfortunately this change seems to have caused issue #3056.
The temporary directories created here with `.itemReplacementDirectory` are never deleted, pile up and when you end up with more than 1000 of them on your file system, `.itemReplacementDirectory` fails.

PR #3052, opened 2020-09-19
 Archive framework where Mach-O Type == staticlib

### LINKED ISSUE #2623: Can't archive static framework
It seems because the output of building a static framework goes to `Carthage/Build/iOS/Static/XYZ.framework` (an additional `Static` directory), the archive script doesn't find it, outputting:
```
Could not find any copies of EngageSDK.framework. Make sure you're in the project's root and that the frameworks have already been built using 'carthage build --no-skip-current'.
```
And that's happening when both invoking `carthage build --archive` or `carthage build --no-skip-current && carthage archive`

3a4oT:
should fix #2623

### DIFF
--- a/Source/carthage/Archive.swift
+++ b/Source/carthage/Archive.swift
@@ -34,7 +34,15 @@ public struct ArchiveCommand: CommandProtocol {
 		}
 	}
 
-	public let verb = "archive"
+    private struct FrameworkMeta: Comparable, Hashable {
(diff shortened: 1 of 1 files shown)


tmspzz:
Please add tests

3a4oT:
Would love to do it if you can suggest where to start? I found that the Carthage tests environment is pretty complex and unfortunately there are no references on how to run tests (including integration tests) locally. Thanks!

gabrielmaldi:
This is still relevant.

PR #3055, opened 2020-09-21
Adding the ability to skip explicit schemes in a build or update step.

jeffctown:
### Summary

Some Carthage repositories have shared schemes that are never meant to be consumed by client developers.  [It's been mentioned before in this repo](https://github.com/Carthage/Carthage/issues/1227#issuecomment-308619580) that Carthage maintainers are sympathetic to these use cases.

#### Description 

This change provides a way to integrate a Carthage dependency and skip building explicit schemes while running the `build` or `update` commands.

I'm happy to add tests and update documentation for these changes, but wanted to see if this community would be open to such a change before going that far.

#### Future Direction:

In the future this should be declared in a `Cartfile`.  That way, a developer doesn't have to add the same schemes to be skipped every time they update Carthage dependencies.  

I am willing to draft a proposal for a `Cartfile` change to support that if there is still a desire for it.  I expect that I would need most of this work in this pull request to be able to have that change, so this pull request is a stepping stone on the path to providing an explicit list of schemes in a `Cartfile` that will be skipped.

### DIFF
--- a/Source/CarthageKit/Project.swift
+++ b/Source/CarthageKit/Project.swift
@@ -1126,6 +1126,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 	/// Returns a producer-of-producers representing each scheme being built.
 	public func buildCheckedOutDependenciesWithOptions( // swiftlint:disable:this cyclomatic_complexity function_body_length
 		_ options: BuildOptions,
+		skipSchemes: [String],
 		dependenciesToBuild: [String]? = nil,
 		sdkFilter: @escaping SDKFilterCallback = { sdks, _, _, _ in .success(sdks) }
 	) -> BuildSchemeProducer {
@@ -1212,7 +1213,7 @@ public final class Project { // swiftlint:disable:this type_body_length
 				options.derivedDataPath = derivedDataVersioned.resolvingSymlinksInPath().path
 
 				return self.symlinkBuildPathIfNeeded(for: dependency, version: version)
-					.then(build(dependency: dependency, version: version, self.directoryURL, withOptions: options, sdkFilter: sdkFilter))
+					.then(build(dependency: dependency, version: version, self.directoryURL, withOptions: options, skipSchemes: skipSchemes, sdkFilter: sdkFilter))
 					.flatMapError { error -> BuildSchemeProducer in
 						switch error {
 						case .noSharedFrameworkSchemes:
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -190,7 +190,8 @@ public func xcodebuildTask(_ task: String, _ buildArguments: BuildArguments, env
 public func buildableSchemesInDirectory( // swiftlint:disable:this function_body_length
 	_ directoryURL: URL,
 	withConfiguration configuration: String,
-	forPlatforms platformAllowList: Set<SDK>? = nil
+	forPlatforms platformAllowList: Set<SDK>? = nil,
+	skipSchemes: [String] = []
 ) -> SignalProducer<(Scheme, ProjectLocator), CarthageError> {
 	precondition(directoryURL.isFileURL)
 	let locator = ProjectLocator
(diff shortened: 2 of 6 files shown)


tmspzz:
@jeffctown I'm still open of it but no changes to the Cartfile, Cartfile.resolved or the Cartfile format. 

Also this would have to come after and/or be compatible with XCFrameworks support.

jeffctown:
Thanks for the feedback @tmspzz .  I'm okay skipping the Cartfile changes I mentioned in the `Future Direction` section.  

I'll add some tests to this soon.

jeffctown:
Added a few tests.

sandruhh:
Hiii why dont you go andsuck a nigger dick sandra? dont you love nigger dickk ?
att: devil u hoe

jeffctown:
This is still valid, just waiting on XCFrameworks to be merged.

REGNANTGODDESSLISSA (approved):

REGNANTGODDESSLISSA (approved):

REGNANTGODDESSLISSA (approved):

REGNANTGODDESSLISSA (approved):
Approved

elliottwilliams (changes requested):
I appreciate how simple this approach is and how it doesn't interfere with Carthage's conceptual model, where shared schemes are always buildable and one dependency can't interfere with build process of another. Seconding the consensus _not_ to go forward with the future direction right now (and to not relitigate #1227 #1990 or #1616).

Out of curiosity, what's your intended use case for this flag? I can imagine its utility in a complex dependency graph, but I worry a bit about projects "encouraging" users to use this flag to avoid build issues, when we'd really prefer that projects hide schemes which aren't meant to be consumed by Carthage.

jeffctown:
Hey @elliottwilliams .  

> Out of curiosity, what's your intended use case for this flag?

I have a dependency with a few shared schemes for targets used for only running tests.  The shared schemes are needed so they can use them during CI, but they do not need to be compiled for the library target I am integrating with.  The linked issues you mention seem to have the same root problem as me - there are projects out there with shared schemes and Carthage is wasting time and effort to compile them when they aren't needed.  My dependency is [pretty opinionated](https://github.com/groue/GRDB.swift/issues/433) about not fixing this on their side, so I'm just looking for a simple solution that allows shared schemes to be skipped by Carthage. 

> I worry a bit about projects "encouraging" users to use this flag

I hear your concern here - this would open up that possibility.  My hope is that asking integrators to take on the additional overhead of adding extra build flags on every Carthage invocation is enough of a deterrent that most library developers would not encourage it, but there's definitely no guarantee of that.  

Let me know if you'd still consider this.  TBH - I don't work on the codebase where I needed this anymore, but I think others (my previous co-workers for sure) could benefit from it.  I'd be happy to move the flag into `BuildOptions` like you suggested if you're still up for accepting this.

elliottwilliams:
@jeffctown yeah, thanks for this explanation!

> The shared schemes are needed so they can use them during CI, but they do not need to be compiled for the library target I am integrating with. The linked issues you mention seem to have the same root problem as me - there are projects out there with shared schemes and Carthage is wasting time and effort to compile them when they aren't needed

This is the use case I'm comfortable with – providing users an _optional_ performance optimization if they know they don't need some parts of a library. To me, this is spiritually similar to the `--platform` argument: if you know you only need to build for iOS, for instance, you can omit the other platforms your dependencies build for.

> Let me know if you'd still consider this. TBH - I don't work on the codebase where I needed this anymore, but I think others (my previous co-workers for sure) could benefit from it. I'd be happy to move the flag into BuildOptions like you suggested if you're still up for accepting this.

Go for it!

elliottwilliams:
Hey @jeffctown, I'd be interesting to hear what you think about this edge case: Currently, if you build with `--skip-schemes X`, scheme `X` won't be built but we'll still create a version file for the dependency. If you build again with `--cache-builds` _but omit --skip-schemes_, the version file will validate successfully and we won't rebuild anything.

This means that the combination of --cache-builds and --skip-schemes introduces some implicit state into your workspace. Is that something we should worry about?

1oo7:
Can you also add the ability to skip certain whole folders or xcode projects, not just schemes? 

There is a bug right now (https://github.com/Carthage/Carthage/issues/3148) where Carthage builds all .xcodeproj's that it can recursively find anywhere in the source repo's directory structure. 

There should at least be an option to skip projects in folders called "Demo" or whose project name contains "Demo" etc. Not just scheme, but project, or folder too.

jeffctown:
It’s been a while since I’ve touched this. I changed jobs and I haven’t figured out the open source contribution rules yet. If anyone is waiting on this change, feel free to take it over.

PR #3060, opened 2020-09-29
README: Minor correction re: "unintrusive"

jblz:
I'm not sure why, but "unintrusive" has always stuck out in my mind as something that people (including me) say, but isn't present in most reference sources as an antonym to the word "intrusive."

This suggested change replaces it in the project README with the word ["unobtrusive"](https://www.lexico.com/definition/unobtrusive). I think this conveys the intended meaning. Another option would be to say ["non-intrusive,"](https://www.merriam-webster.com/dictionary/nonintrusive) (I see it with and without the hyphen), though I'm in favor of the word choice submitted in this PR, myself.

Thanks for the lovely software.  I stumbled across it in the build instructions for https://devutils.app/ & couldn't pass up the opportunity to be pedantic /s

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,7 +409,7 @@ If you’re interested in using Carthage as part of another tool, or perhaps ext
 
 [CocoaPods](http://cocoapods.org/) is a long-standing dependency manager for Cocoa. So why was Carthage created?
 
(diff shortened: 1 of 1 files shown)


tmspzz:
Thanks 🎉

tmspzz:
Thanks for the suggestion! However we're fine having it slightly incorrect but more accessible non-native speakers.

Glad you like the project 👍

PR #3062, opened 2020-10-01
Use -c protocol.version=2 for submodule command

jonmagic:
Fixes https://github.com/realm/realm-cocoa/issues/6807

I've been trying to debug what caused https://github.com/realm/realm-cocoa/issues/6807 and https://github.com/realm/realm-cocoa/issues/6806 and figured out another way to get around the failure (besides switching to ssh) was to just tell the git client to use protocol version 2.


tmspzz:
I don't think this is a carthage issue.

PR #2881, opened 2019-09-26
[WIP] XCFramework support

tmspzz:

PR #3071, opened 2020-10-10
Create XCFrameworks and retain project compatibility

### LINKED ISSUE #3019: Carthage builds fail at xcrun lipo on Xcode 12 beta (3,4,5...)
* carthage install method: [ ] .pkg, [x] homebrew, [x] source
* `carthage version`: 0.35.0
* `xcodebuild -version`: 12A8169g

**Cartfile**
```
github "LoungeBuddy/Auth0.swift" "feature/decouple-keychain-manager"
binary "https://raw.githubusercontent.com/Instabug/Instabug-iOS/master/Instabug.json"
```

**Carthage Output**
```sh
> carthage build --platform iOS

*** xcodebuild output can be found in /var/folders/jm/pyy6p95j177_btm91tj0mkdm0000gn/T/carthage-xcodebuild.WPi9jL.log
*** Downloading binary-only framework Instabug at "https://raw.githubusercontent.com/Instabug/Instabug-iOS/master/Instabug.json"
*** Building scheme "JWTDecode-iOS" in JWTDecode.xcodeproj
Build Failed
	Task failed with exit code 1:
	/usr/bin/xcrun lipo -create /Users/swarn1/Library/Caches/org.carthage.CarthageKit/DerivedData/12.0_12A8169g/JWTDecode.swift/2.4.1/Build/Intermediates.noindex/ArchiveIntermediates/JWTDecode-iOS/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/JWTDecode.framework/JWTDecode /Users/swarn1/Library/Caches/org.carthage.CarthageKit/DerivedData/12.0_12A8169g/JWTDecode.swift/2.4.1/Build/Products/Release-iphonesimulator/JWTDecode.framework/JWTDecode -output /Volumes/Shared/LoungeBuddy/consumer-ios-app/Carthage/Build/iOS/JWTDecode.framework/JWTDecode

This usually indicates that project itself failed to compile. Please check the xcodebuild log for more details: /var/folders/jm/pyy6p95j177_btm91tj0mkdm0000gn/T/carthage-xcodebuild.WPi9jL.log
```

Running the problem command:
```sh
> /usr/bin/xcrun lipo -create /Users/swarn1/Library/Caches/org.carthage.CarthageKit/DerivedData/12.0_12A8169g/JWTDecode.swift/2.4.1/Build/Intermediates.noindex/ArchiveIntermediates/JWTDecode-iOS/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/JWTDecode.framework/JWTDecode /Users/swarn1/Library/Caches/org.carthage.CarthageKit/DerivedData/12.0_12A8169g/JWTDecode.swift/2.4.1/Build/Products/Release-iphonesimulator/JWTDecode.framework/JWTDecode -output /Volumes/Shared/LoungeBuddy/consumer-ios-app/Carthage/Build/iOS/JWTDecode.framework/JWTDecode

fatal error: /Applications/Xcode-12-beta-3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: /Users/swarn1/Library/Caches/org.carthage.CarthageKit/DerivedData/12.0_12A8169g/JWTDecode.swift/2.4.1/Build/Intermediates.noindex/ArchiveIntermediates/JWTDecode-iOS/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/JWTDecode.framework/JWTDecode and /Users/swarn1/Library/Caches/org.carthage.CarthageKit/DerivedData/12.0_12A8169g/JWTDecode.swift/2.4.1/Build/Products/Release-iphonesimulator/JWTDecode.framework/JWTDecode have the same architectures (arm64) and can't be in the same fat output file
```

**Actual outcome**
- Carthage fails to build any dependencies.
- Carthage advises to look at a build log which ends with `** BUILD SUCCEEDED **`

**Expected outcome**
- Carthage should not fail to build.
- On failures outside of `xcodebuild`, Carthage should not direct you to the build log.

**Workaround**
I guess the architecture for `iphonesimulator` has changed such that it now collides with `iphoneos`? Not sure how this stuff works. Either way, I've been able to work around the problem by changing Sources/CarthageKit/Xcode.swift:742 to `argsForBuilding.onlyActiveArchitecture = true`. Obviously, this is not a great solution.

elliottwilliams:
Hey! I want to propose an alternate xcframeworks implementation to #2801 and #2881. This PR fixes #3019 by combining built frameworks into an xcframework when supported. 

I don't mean to steal @tmspzz's thunder here, just wanted to give us another implementation to consider. I'm trying to keep the PR small and retain full compatibility with existing projects.

I've tested this in a large iOS+watchOS project (the Yelp app) and on some smaller multiplatform projects. I'm sure there are edge cases I'm missing -- would love for others to try this branch out and let me know what you find 🙂

### Summary of changes
  
- By default, **all frameworks built for a dependency will be merged into an xcframework in Carthage/Build**.
  - Building the same dependency multiple times with different `--platform` is supported -- new build products will be merged into an existing xcframework.
  - Similarly, rebuilding a dependency replaces previous builds for that platform in the xcframework.
  - xcframework creation can be disabled with `--no-create-xcframework`, and is disabled by default on systems without Xcode 12.

- **No configuration changes required for existing dependencies to support xcframeworks**.
	* The xcframework is created using  `-allow-internal-distribution` , which prevents it from requiring the usual `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` / `SKIP_INSTALL=NO`.
	* When building, Carthage extracts platform-specific frameworks to a temporary directory and passes that directory to Xcode. This allows frameworks to link against nested dependencies without adding xcframeworks to their project.

- Xcode 12 is required when creating an xcframework
	- This allows Carthage to embed dSYMs and bcsymbolmaps into the xcframework bundle, so we don't have to do additional work to manage multiple dSYMs.

- Version file (`--cache-builds`) support is complete.
	- Each entry in a version file still corresponds to one framework bundle, but frameworks contained in xcframeworks have additional `container` and `identifier` fields that describe their location within the .xcframework directory.
	- Cache tools like Rome can support xcframeworks by caching the `container` directory when present.

- No copy-frameworks support. Users should embed xcframeworks directly in Xcode instead of using `copy-frameworks`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,15 +6,15 @@ Carthage is intended to be the simplest way to add frameworks to your Cocoa appl
 
 Carthage builds your dependencies and provides you with binary frameworks, but you retain full control over your project structure and setup. Carthage does not automatically modify your project files or your build settings.
 
-# [Xcode >= 12 Workaround Guide - Follow this link](Documentation/Xcode12Workaround.md)
-
 - [Quick Start](#quick-start)
 - [Installing Carthage](#installing-carthage)
 - [Adding frameworks to an application](#adding-frameworks-to-an-application)
 	- [Getting started](#getting-started)
-		- [If you're building for macOS](#if-youre-building-for-macos)
-		- [If you're building for iOS, tvOS, or watchOS](#if-youre-building-for-ios-tvos-or-watchos)
-		- [For both platforms](#for-both-platforms)
+		- [Building platform-independent xcframeworks](#building-platform-independent-xcframeworks-xcode-12-and-above)
+		- [Building platform-specific framework bundles](#building-platform-specific-framework-bundles-default-for-xcode-11-and-below)
+			- [If you're building for macOS](#if-youre-building-for-macos)
+			- [If you're building for iOS, tvOS, or watchOS](#if-youre-building-for-ios-tvos-or-watchos)
+		- [For all platforms](#for-all-platforms)
 		- [(Optionally) Add build phase to warn about outdated dependencies](#optionally-add-build-phase-to-warn-about-outdated-dependencies)
 		- [Swift binary framework download compatibility](#swift-binary-framework-download-compatibility)
 	- [Running a project that uses Carthage](#running-a-project-that-uses-carthage)
@@ -52,26 +52,8 @@ Carthage builds your dependencies and provides you with binary frameworks, but y
 
 1. Run `carthage update`
 1. A `Cartfile.resolved` file and a `Carthage` directory will appear in the same directory where your `.xcodeproj` or `.xcworkspace` is
-1. Drag the built `.framework` binaries from `Carthage/Build/<platform>` into your application’s Xcode project.
-1. If you are using Carthage for an application, follow the remaining steps, otherwise stop here.
-1. On your application targets’ _Build Phases_ settings tab, click the _+_ icon and choose _New Run Script Phase_. Create a Run Script in which you specify your shell (ex: `/bin/sh`), add the following contents to the script area below the shell:
-
-    ```sh
-    /usr/local/bin/carthage copy-frameworks
-    ```
-
-- Add the paths to the frameworks you want to use under “Input Files". For example:
-
-    ```
-    $(SRCROOT)/Carthage/Build/iOS/Alamofire.framework
-    ```
-
-- Add the paths to the copied frameworks to the “Output Files”. For example:
-
-    ```
-    $(BUILT_PRODUCTS_DIR)/$(FRAMEWORKS_FOLDER_PATH)/Alamofire.framework
-    ```
-Another approach when having multiple dependencies is to use `.xcfilelist`s. This is covered in [If you're building for iOS, tvOS or watchOS](#if-youre-building-for-ios-tvos-or-watchos)
+1. Drag the built `.xcframework` bundles from `Carthage/Build` into the "Frameworks and Libraries" section of your application’s Xcode project.
+1. If you are using Carthage for an application, select "Embed & Sign", otherwise "Do Not Embed".
 
 For an in depth guide, read on from [Adding frameworks to an application](#adding-frameworks-to-an-application)
 
@@ -93,10 +75,23 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
 
 ### Getting started
 
-##### If you're building for macOS
+#### Building platform-independent xcframeworks (Xcode 12 and above)
 
 1. Create a [Cartfile][] that lists the frameworks you’d like to use in your project.
-1. Run `carthage update --platform macOS`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled framework.
+1. Run `carthage update`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled xcframework.
+1. On your application targets’ _General_ settings tab, in the _Embedded Binaries_ section, drag and drop each xcframework you want to use from the [Carthage/Build][] folder on disk.
+
+#### Building platform-specific framework bundles (default for Xcode 11 and below)
+
+**Xcode 12+ incompatibility**: Multi-architecture platforms are not supported when building framework bundles in Xcode 12 and above. Prefer [building with xcframeworks](#building-platform-independent-xcframeworks-xcode-12-and-above). If you need to build discrete framework bundles, [use a workaround xcconfig file](Documentation/Xcode12Workaround.md).
+
+##### If you're building for macOS
+
+<details>
+	<summary>macOS-specific instructions</summary>
+
+1. Create a [Cartfile][] that lists the frameworks you’d like to use in your project.
+1. Run `carthage update --platform macOS --no-create-xcframework`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled framework.
 1. On your application targets’ _General_ settings tab, in the _Embedded Binaries_ section, drag and drop each framework you want to use from the [Carthage/Build][] folder on disk.
 
 Additionally, you'll need to copy debug symbols for debugging and crash reporting on OS X.
@@ -105,10 +100,15 @@ Additionally, you'll need to copy debug symbols for debugging and crash reportin
 1. Click the _Destination_ drop-down menu and select _Products Directory_.
 1. For each framework you’re using, drag and drop its corresponding dSYM file.
 
+</details>
+
 ##### If you're building for iOS, tvOS, or watchOS
 
+<details>
+	<summary>Platform-specific instructions</summary>
+
 1. Create a [Cartfile][] that lists the frameworks you’d like to use in your project.
-1. Run `carthage update`. This will fetch dependencies into a [Carthage/Checkouts][] folder, then build each one or download a pre-compiled framework.
+1. Run `carthage update --no-create-xcframework`. This will fetch dependencies into a [Carthage/Checkouts][] folder, then build each one or download a pre-compiled framework.
 1. Open your application targets’ _General_ settings tab. For Xcode 11.0 and higher, in the "Frameworks, Libraries, and Embedded Content" section, drag and drop each framework you want to use from the [Carthage/Build][] folder on disk. Then, in the "Embed"  section, select "Do Not Embed" from the pulldown menu for each item added. For Xcode 10.x and lower, in the "Linked Frameworks and Libraries" section, drag and drop each framework you want to use from the [Carthage/Build][] folder on disk.
 1. On your application targets’ _Build Phases_ settings tab, click the _+_ icon and choose _New Run Script Phase_. Create a Run Script in which you specify your shell (ex: `/bin/sh`), add the following contents to the script area below the shell:
 
@@ -143,12 +143,13 @@ This script works around an [App Store submission bug](http://www.openradar.me/r
 With the debug information copied into the built products directory, Xcode will be able to symbolicate the stack trace whenever you stop at a breakpoint. This will also enable you to step through third-party code in the debugger.
 
 When archiving your application for submission to the App Store or TestFlight, Xcode will also copy these files into the dSYMs subdirectory of your application’s `.xcarchive` bundle.
+</details>
 
-##### For both platforms
+#### For all platforms
 
 Along the way, Carthage will have created some [build artifacts][Artifacts]. The most important of these is the [Cartfile.resolved][] file, which lists the versions that were actually built for each framework. **Make sure to commit your [Cartfile.resolved][]**, because anyone else using the project will need that file to build the same framework versions.
 
-##### (Optionally) Add build phase to warn about outdated dependencies
+#### (Optionally) Add build phase to warn about outdated dependencies
 
 You can add a Run Script phase to automatically warn you when one of your dependencies is out of date.
 
@@ -158,7 +159,7 @@ You can add a Run Script phase to automatically warn you when one of your depend
 /usr/local/bin/carthage outdated --xcode-warnings 2>/dev/null
 ```
 
-##### Swift binary framework download compatibility
+#### Swift binary framework download compatibility
 
 Carthage will check to make sure that downloaded Swift (and mixed Objective-C/Swift) frameworks were built with the same version of Swift that is in use locally. If there is a version mismatch, Carthage will proceed to build the framework from source. If the framework cannot be built from source, Carthage will fail.
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Carthage builds your dependencies and provides you with binary frameworks, but y
 	github "Alamofire/Alamofire" ~> 4.7.2
 	```
 
-1. Run `carthage update --create-xcframework`
+1. Run `carthage update --use-xcframeworks`
 1. A `Cartfile.resolved` file and a `Carthage` directory will appear in the same directory where your `.xcodeproj` or `.xcworkspace` is
 1. Drag the built `.xcframework` bundles from `Carthage/Build` into the "Frameworks and Libraries" section of your application’s Xcode project.
 1. If you are using Carthage for an application, select "Embed & Sign", otherwise "Do Not Embed".
@@ -78,7 +78,7 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
 #### Building platform-independent xcframeworks (Xcode 12 and above)
 
 1. Create a [Cartfile][] that lists the frameworks you’d like to use in your project.
-1. Run `carthage update --create-xcframework`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled xcframework.
+1. Run `carthage update --use-xcframeworks`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled xcframework.
 1. On your application targets’ _General_ settings tab, in the _Embedded Binaries_ section, drag and drop each xcframework you want to use from the [Carthage/Build][] folder on disk.
 
 #### Building platform-specific framework bundles (default for Xcode 11 and below)
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Carthage builds your dependencies and provides you with binary frameworks, but y
 	github "Alamofire/Alamofire" ~> 4.7.2
 	```
 
-1. Run `carthage update`
+1. Run `carthage update --create-xcframework`
 1. A `Cartfile.resolved` file and a `Carthage` directory will appear in the same directory where your `.xcodeproj` or `.xcworkspace` is
 1. Drag the built `.xcframework` bundles from `Carthage/Build` into the "Frameworks and Libraries" section of your application’s Xcode project.
 1. If you are using Carthage for an application, select "Embed & Sign", otherwise "Do Not Embed".
@@ -78,7 +78,7 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
 #### Building platform-independent xcframeworks (Xcode 12 and above)
 
 1. Create a [Cartfile][] that lists the frameworks you’d like to use in your project.
-1. Run `carthage update`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled xcframework.
+1. Run `carthage update --create-xcframework`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled xcframework.
 1. On your application targets’ _General_ settings tab, in the _Embedded Binaries_ section, drag and drop each xcframework you want to use from the [Carthage/Build][] folder on disk.
 
 #### Building platform-specific framework bundles (default for Xcode 11 and below)
@@ -91,7 +91,7 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
 	<summary>macOS-specific instructions</summary>
 
 1. Create a [Cartfile][] that lists the frameworks you’d like to use in your project.
-1. Run `carthage update --platform macOS --no-create-xcframework`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled framework.
+1. Run `carthage update --platform macOS`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled framework.
 1. On your application targets’ _General_ settings tab, in the _Embedded Binaries_ section, drag and drop each framework you want to use from the [Carthage/Build][] folder on disk.
 
 Additionally, you'll need to copy debug symbols for debugging and crash reporting on OS X.
@@ -108,7 +108,7 @@ Additionally, you'll need to copy debug symbols for debugging and crash reportin
 	<summary>Platform-specific instructions</summary>
 
 1. Create a [Cartfile][] that lists the frameworks you’d like to use in your project.
-1. Run `carthage update --no-create-xcframework`. This will fetch dependencies into a [Carthage/Checkouts][] folder, then build each one or download a pre-compiled framework.
+1. Run `carthage update`. This will fetch dependencies into a [Carthage/Checkouts][] folder, then build each one or download a pre-compiled framework.
 1. Open your application targets’ _General_ settings tab. For Xcode 11.0 and higher, in the "Frameworks, Libraries, and Embedded Content" section, drag and drop each framework you want to use from the [Carthage/Build][] folder on disk. Then, in the "Embed"  section, select "Do Not Embed" from the pulldown menu for each item added. For Xcode 10.x and lower, in the "Linked Frameworks and Libraries" section, drag and drop each framework you want to use from the [Carthage/Build][] folder on disk.
 1. On your application targets’ _Build Phases_ settings tab, click the _+_ icon and choose _New Run Script Phase_. Create a Run Script in which you specify your shell (ex: `/bin/sh`), add the following contents to the script area below the shell:
 
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ Carthage is intended to be the simplest way to add frameworks to your Cocoa appl
 
 Carthage builds your dependencies and provides you with binary frameworks, but you retain full control over your project structure and setup. Carthage does not automatically modify your project files or your build settings.
 
+# [Xcode >= 12 Workaround Guide - Follow this link](Documentation/Xcode12Workaround.md)
+
 - [Quick Start](#quick-start)
 - [Installing Carthage](#installing-carthage)
 - [Adding frameworks to an application](#adding-frameworks-to-an-application)
--- a/Source/CarthageKit/BuildOptions.swift
+++ b/Source/CarthageKit/BuildOptions.swift
@@ -17,7 +17,7 @@ public struct BuildOptions {
 	/// Whether to use downloaded binaries if possible.
 	public var useBinaries: Bool
 	/// Whether to create an XCFramework instead of lipoing built products.
-	public var createXCFramework: Bool
+	public var useXCFrameworks: Bool
 
 	public init(
 		configuration: String,
@@ -26,7 +26,7 @@ public struct BuildOptions {
 		derivedDataPath: String? = nil,
 		cacheBuilds: Bool = true,
 		useBinaries: Bool = true,
-		createXCFramework: Bool = false
+		useXCFrameworks: Bool = false
 	) {
 		self.configuration = configuration
 		self.platforms = platforms
@@ -34,6 +34,6 @@ public struct BuildOptions {
 		self.derivedDataPath = derivedDataPath
 		self.cacheBuilds = cacheBuilds
 		self.useBinaries = useBinaries
-		self.createXCFramework = createXCFramework
+		self.useXCFrameworks = useXCFrameworks
 	}
 }
--- a/Source/CarthageKit/BuildOptions.swift
+++ b/Source/CarthageKit/BuildOptions.swift
@@ -16,6 +16,8 @@ public struct BuildOptions {
 	public var cacheBuilds: Bool
 	/// Whether to use downloaded binaries if possible.
 	public var useBinaries: Bool
+	/// Whether to create an XCFramework instead of lipoing built products.
+	public var createXCFramework: Bool
 
 	public init(
 		configuration: String,
@@ -23,7 +25,8 @@ public struct BuildOptions {
 		toolchain: String? = nil,
 		derivedDataPath: String? = nil,
 		cacheBuilds: Bool = true,
-		useBinaries: Bool = true
+		useBinaries: Bool = true,
+		createXCFramework: Bool = true
 	) {
 		self.configuration = configuration
 		self.platforms = platforms
@@ -31,5 +34,6 @@ public struct BuildOptions {
 		self.derivedDataPath = derivedDataPath
 		self.cacheBuilds = cacheBuilds
 		self.useBinaries = useBinaries
+		self.createXCFramework = createXCFramework
 	}
 }
--- a/Source/CarthageKit/BuildOptions.swift
+++ b/Source/CarthageKit/BuildOptions.swift
@@ -26,7 +26,7 @@ public struct BuildOptions {
 		derivedDataPath: String? = nil,
 		cacheBuilds: Bool = true,
 		useBinaries: Bool = true,
-		createXCFramework: Bool = true
+		createXCFramework: Bool = false
 	) {
 		self.configuration = configuration
 		self.platforms = platforms
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -306,7 +306,7 @@ public struct BuildSettings {
 
 	/// The "OPERATING_SYSTEM" component of the target triple. Used in XCFrameworks to denote the supported platform.
 	public var platformTripleOS: Result<String, CarthageError> {
-		self["LLVM_TARGET_TRIPLE_OS_VERSION"].map { osVersion in
+		return self["LLVM_TARGET_TRIPLE_OS_VERSION"].map { osVersion in
 			// osVersion is a string like "ios8.0". Remove any trailing version number.
 			// This should match the OS component of an "unversionedTriple" printed by `swift -print-target-info`.
 			osVersion.replacingOccurrences(of: "([0-9]\\.?)*$", with: "", options: .regularExpression)
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -154,6 +154,10 @@ public struct BuildSettings {
 		}
 	}
 
+	public var archs: Result<Set<String>, CarthageError> {
+		return self["ARCHS"].map { Set($0.components(separatedBy: " ")) }
+	}
+
 	/// Attempts to determine the ProductType specified in these build settings.
 	public var productType: Result<ProductType, CarthageError> {
 		return self["PRODUCT_TYPE"].flatMap(ProductType.from(string:))
@@ -242,6 +246,11 @@ public struct BuildSettings {
 	/// Attempts to determine the name of the built product's wrapper bundle.
 	public var wrapperName: Result<String, CarthageError> {
 		return self["WRAPPER_NAME"]
+	}
+
+	/// Attempts to determine the name of the built product.
+	public var productName: Result<String, CarthageError> {
+		return self["PRODUCT_NAME"]
 	}
 
 	/// Attempts to determine the URL to the built product's wrapper, corresponding
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -304,6 +304,26 @@ public struct BuildSettings {
 		return self["TARGET_BUILD_DIR"]
 	}
 
+	/// The "OPERATING_SYSTEM" component of the target triple. Used in XCFrameworks to denote the supported platform.
+	public var platformTripleOS: Result<String, CarthageError> {
+		self["LLVM_TARGET_TRIPLE_OS_VERSION"].map { osVersion in
+			// osVersion is a string like "ios8.0". Remove any trailing version number.
+			// This should match the OS component of an "unversionedTriple" printed by `swift -print-target-info`.
+			osVersion.replacingOccurrences(of: "([0-9]\\.?)*$", with: "", options: .regularExpression)
+		}.flatMapError { _ in
+			// LLVM_TARGET_TRIPLE_OS_VERSION may be unavailable if `USE_LLVM_TARGET_TRIPLES = NO`.
+			// SWIFT_PLATFORM_TARGET_PREFIX anecdotally appears to contain the unversioned OS component, even in
+			// non-swift projects.
+			self["SWIFT_PLATFORM_TARGET_PREFIX"]
+		}
+	}
+
+	// The "ENVIRONMENT" component of the target triple, which is "simulator" when building for a simulator target
+	// and missing otherwise.
+	public var platformTripleVariant: Result<String, CarthageError> {
+		return self["LLVM_TARGET_TRIPLE_SUFFIX"].map { $0.stripping(prefix: "-") }
+	}
+
 	/// Add subdirectory path if it's not possible to paste product to destination path
 	public func productDestinationPath(in destinationURL: URL) -> URL {
 		let directoryURL: URL
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -173,6 +173,12 @@ public struct BuildSettings {
 		return productType.fanout(machOType).map(FrameworkType.init)
 	}
 
+	internal var frameworkSearchPaths: Result<[URL], CarthageError> {
+		return self["FRAMEWORK_SEARCH_PATHS"].map { paths in
+			paths.split(separator: " ").map { URL(fileURLWithPath: String($0), isDirectory: true) }
+		}
+	}
+
 	/// Attempts to determine the URL to the built products directory.
 	public var builtProductsDirectoryURL: Result<URL, CarthageError> {
 		return self["BUILT_PRODUCTS_DIR"].map { productsDir in
--- a/Source/CarthageKit/Xcode.swift
+++ b/Source/CarthageKit/Xcode.swift
@@ -586,7 +586,10 @@ private func mergeIntoXCFramework(in directoryURL: URL, settings: BuildSettings)
 		directoryURL.appendingPathComponent(productName).appendingPathExtension("xcframework")
 	}
 	let framework = SignalProducer(result: settings.wrapperURL.map({ $0.resolvingSymlinksInPath() }))
-	let buildDSYMs = SignalProducer(result: settings.wrapperURL).flatMap(.concat, createDebugInformation)
+
+	let buildDSYMs = SignalProducer(result: settings.wrapperURL)
+		.filter { _ in settings.machOType.value != .staticlib }
+		.flatMap(.concat, createDebugInformation)
 		.ignoreTaskData()
 		.map({ $0 })
 	let buildSymbolMaps = SignalProducer(result: settings.wrapperURL)
@@ -774,7 +777,7 @@ public func buildScheme( // swiftlint:disable:this function_body_length cyclomat
 		}
 		.flatMapTaskEvents(.concat) { builtProductURL -> SignalProducer<URL, CarthageError> in
 			guard !options.createXCFramework else {
-				// XCFrameworks already have debug information embedded in them.
+				// XCFrameworks have debug information embedded in them after being merged.
 				return SignalProducer(value: builtProductURL)
 			}
 			return UUIDsForFramework(builtProductURL)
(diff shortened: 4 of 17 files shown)


tmspzz:
Hey @elliottwilliams thanks for the great effort! 

I'm not sure however now what to do with this.  If I understand correctly this build one .xcframework per platform instead of a combined version (when needed). In the other PR the `--create-xcframework` flag was intended to support both combined and per platform version.

Can you think of a way of combining both PRs to get to finish line? I'm a bit wary of supporting both combined and singe xcframework packaging because it add a lot of complexity to an already complex project.

elliottwilliams:
> Can you think of a way of combining both PRs to get to finish line? I'm a bit wary of supporting both combined and singe xcframework packaging because it add a lot of complexity to an already complex project.

Yeah – I can see the value of building a combined xcframework, and agree with you there's no reason to support combined xcframeworks _and_ one-per-platform xcframeworks. Let me see if I can refactor to make that work!

elliottwilliams:
@tmspzz updated this to create a combined xcframework. I had to give up the behavior where it would automatically create an xcframework if build settings indicated `lipo` would fail, so I resorted to adding a flag and creating xcframeworks by default on Xcode 12+. I still agree with you that this is the right way to go – it makes the PR itself more complex but is a lot closer to how I'd expect this to work as a user.

tmspzz:
@elliottwilliams thanks! I'll review asap

Source/CarthageKit/Xcode.swift
```diff
@@ -188,6 +188,23 @@ public func xcodebuildTask(_ task: String, _ buildArguments: BuildArguments, env
 	return xcodebuildTask([task], buildArguments)
 }
 
+func shouldBuildXCFramework(deviceBuildSettings: BuildSettings, simulatorBuildSettings: BuildSettings) -> Result<Bool, CarthageError> {
```
jeffctown, line 191: This PR has me pretty excited, so I pulled this branch to try it out.  In my limited testing, everything is working well, but I ran into strange errors with Xcode 11.7.  

This method sounds like what is needed, but I don't think this is called from anywhere.

elliottwilliams: That fix makes total sense. I'll incorporate it when I get a chance to!

re: Xcode < 12, it sounds `--create-xcframework` should fail with an error like "you need to be on Xcode 12 or greater", right? I added an `.xcframeworkRequried` error in this PR, but didn't think to  incorporate it in the argument parser. Should be pretty easy to change though.

`shouldBuildXCFramework(deviceBuildSettings:simulatorBuildSettings:)` stuck around from the initial implementation of this PR, where Carthage would automatically make an xcframework if it determined that lipo would fail. Now that we're producing multi-platform xcframeworks, I'm not sure that it has any utility.

elliottwilliams:
We can't easily write unit tests for this, because Carthage's version of Nimble [needs to be upgraded to compile on Swift 5.2+](https://github.com/Quick/Nimble/releases/tag/v9.0.0), and _that_ would introduce breaking changes to the tests. So I'd prefer to tackle that later if possible :)

We _can_ write integration tests, though. I'd like to get the current suite of bats tests passing with --create-xcframework before merging.

chrisballinger:
Just want to say that I'm really excited about this, and I like the strategy of making things just work by default out of the box for existing projects.

However, I'm not sure if `-allow-internal-distribution` is the right path for all scenarios. This flag doesn't seem well documented, all I could find was "Specifies that the created xcframework contains information not suitable for public distribution." from the CLI help, but [this post](https://forums.swift.org/t/testable-imports-not-working-with-xcframeworks/35403/2) makes it seem like it may make xcframeworks produced this way not compatible between Xcode versions.

I don't think it should be a blocker for merging this PR, but I think that longer term that frameworks produced via `carthage archive` should be automatically built with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` and not use the `-allow-internal-distribution` flag. Otherwise stable binary framework releases published by others might not be compatible between Xcode / Swift compiler versions.

elliottwilliams:
@chrisballinger good points! Definitely agree that there's value in looking at BUILD_LIBRARY_FOR_DISTRIBUTION in the future, but that it's an intentional non-goal for this PR.

My understanding is that from Apple's perspective, "distribution" means "a binary framework that you provide to users without source code". If you _do_ ship source code, Apple wants you to use a Swift package, which is much more analogous to Carthage's model. So while BUILD_LIBRARY_FOR_DISTRIBUTION builds might be valuable for us in the future, the status quo is building an xcframework with only the local machine in mind.

mman:
@elliottwilliams BUILD_LIBRARY_FOR_DISTRIBUTION generates a stable swift interface and binary compatible object files that guarantee that the framework can be linked into the app with potentially newer versions of swift. So theoretically speaking, if you ``carthage build`` your dependencies with old Xcode and then update to new Xcode and try to rebuild a project, it may fail until you also rebuild the .xcframework.

elliottwilliams:
@mman 
> So theoretically speaking, if you `carthage build` your dependencies with old Xcode and then update to new Xcode and try to rebuild a project, it may fail until you also rebuild the .xcframework.

Exactly, and that's the status quo behavior of Carthage 🙂. Using --cache-builds is the current workflow to handle scenarios like these, and this PR doesn't change that. You can run

    carthage build --cache-builds

before you start building in Xcode to rebuild any out-of-date frameworks. (Depending on how your project is structured, you could even run this as a build phase!)

The [version file logic](https://github.com/elliottwilliams/Carthage/blob/9970edced88f7962522f07bb68ee6bf87f0dc4a9/Source/CarthageKit/VersionFile.swift#L217) will invalidate frameworks inside an xcframework unless:

* they were built for module stability, _or_
* they were built using the same compiler version

So if you care a lot about portability of built frameworks, you can set `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` yourself, and Carthage will still do the right thing.

Source/CarthageKit/Xcode.swift
elliottwilliams: @jeffctown https://github.com/Carthage/Carthage/pull/3071/commits/18991ec8348ec63be19318f967297e063257f200 has both of these incorporated, lmk if you run into any problems or have additional feedback!

I took a slightly different approach on static frameworks (using build settings instead of dwarfdump to save time), but I was able to reproduce and confirm locally. Thanks so much for reporting this :)

jeffctown: @elliottwilliams Awesome!  Thanks so much for bringing these changes in.  Your approach makes complete sense.  🎉

I'll give it a test run this week and let you know how it goes.

jeffctown: Just reporting back that the static frameworks fix is working like a charm for my project.  Thank you for this contribution @elliottwilliams !

README.md
```diff
@@ -93,10 +75,23 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
...
+1. Run `carthage update`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled xcframework.
+1. On your application targets’ _General_ settings tab, in the _Embedded Binaries_ section, drag and drop each xcframework you want to use from the [Carthage/Build][] folder on disk.
+
+#### Building platform-specific framework bundles (default for Xcode 11 and below)
+
+**Xcode 12+ incompatibility**: Multi-architecture platforms are not supported when building framework bundles in Xcode 12 and above. Prefer [building with xcframeworks](#building-platform-independent-xcframeworks-xcode-12-and-above). If you need to build discrete framework bundles, [use a workaround xcconfig file](Documentation/Xcode12Workaround.md).
+
 ##### If you're building for macOS
 
+<details>
```
jdhealy, line 90: This is a great idea! 👍

Source/CarthageKit/BuildSettings.swift
```diff
@@ -295,6 +304,26 @@ public struct BuildSettings {
...
+	public var platformTripleOS: Result<String, CarthageError> {
+		return self["LLVM_TARGET_TRIPLE_OS_VERSION"].map { osVersion in
+			// osVersion is a string like "ios8.0". Remove any trailing version number.
+			// This should match the OS component of an "unversionedTriple" printed by `swift -print-target-info`.
+			osVersion.replacingOccurrences(of: "([0-9]\\.?)*$", with: "", options: .regularExpression)
+		}.flatMapError { _ in
+			// LLVM_TARGET_TRIPLE_OS_VERSION may be unavailable if `USE_LLVM_TARGET_TRIPLES = NO`.
+			// SWIFT_PLATFORM_TARGET_PREFIX anecdotally appears to contain the unversioned OS component, even in
+			// non-swift projects.
+			self["SWIFT_PLATFORM_TARGET_PREFIX"]
```
jdhealy, line 317: Great sleuthing on this! 🎉

gjeck:
Thank you @elliottwilliams and all maintainers for providing this! Apologies if this is a distraction on the pull request, but I thought it might be worth mentioning that my team and I successfully utilized these changes to migrate a couple relatively large iOS/watchOS/tvOS applications from Xcode 11 to 12. Each app has a decent amount of internal and external (mixed dynamic and static framework) dependencies. Happy to provide more details if it's useful or zip it if it is not 🤐. Just wanted to chime in that the migration was relatively painless and addressed our needs. Hopefully that's a somewhat useful data point. 

I also read over the changes and think everything looks great for whatever that is worth 😅. Thanks again!

Source/CarthageKit/Xcode.swift
```diff
@@ -735,14 +811,64 @@ private func resolveSameTargetName(for settings: BuildSettings) -> SignalProduce
...
+	.flatten()
+	.filter { $0.pathExtension == "xcframework" }
+	.flatMap(.merge) { url -> SignalProducer<URL, CarthageError> in
+		frameworkBundlesInURL(url, compatibleWith: platformTripleOS, variant: settings.platformTripleVariant.value)
+			.mapError { .readFailed(url, $0 as NSError) }
+			.map { $0.bundleURL }
+	}
+
+	let makeTemporaryDirectory = SignalProducer<URL, CarthageError> { () -> Result<URL, CarthageError> in
+		var templatePath = (NSTemporaryDirectory() as NSString).appendingPathComponent("carthage-xcframework-XXXX").utf8CString
```
tmspzz, line 840: is this directory ever deleted?

tmspzz:
@elliottwilliams did I miss it or handling of binary dependencies with .xcframeworks is not handled here?

README.md
```diff
@@ -93,10 +75,23 @@ Once you have Carthage [installed](#installing-carthage), you can begin adding f
...
+	<summary>macOS-specific instructions</summary>
+
 1. Create a [Cartfile][] that lists the frameworks you’d like to use in your project.
-1. Run `carthage update --platform macOS`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled framework.
+1. Run `carthage update --platform macOS --no-create-xcframework`. This will fetch dependencies into a [Carthage/Checkouts][] folder and build each one or download a pre-compiled framework.
```
tmspzz, line 94: isn't the build argument false by default?

Source/CarthageKit/BuildOptions.swift
```diff
@@ -16,20 +16,24 @@ public struct BuildOptions {
...
 
 	public init(
 		configuration: String,
 		platforms: Set<SDK>? = nil,
 		toolchain: String? = nil,
 		derivedDataPath: String? = nil,
 		cacheBuilds: Bool = true,
-		useBinaries: Bool = true
+		useBinaries: Bool = true,
+		createXCFramework: Bool = false
```
tmspzz, line 29: see here

tmspzz:
I don't see the .version file being generated when using 

```
carthage update --create-xcframework --cache-builds --new-resolver --no-use-binaries
```

Cartfile:
```
github "Alamofire/Alamofire"
```

Source/CarthageKit/Xcode.swift
```diff
@@ -832,6 +966,15 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
 							]
 						}
 
+						if let extractedXCFrameworksDir = extractedXCFrameworksDir {
+							// If the project's working directory contains xcframeworks in Carthage/Build, target-specific
+							// frameworks will have been extracted to a temporary directory. Provide these frameworks as a fallback
+							// in case the project is not configured to build using xcframeworks.
+							result += [
```
tmspzz, line 973: I'm not sure we should be using this. If the project is not set to build with XCFrameworks then why are thy used in the first place?

tmspzz:
If I understood correctly, this works by re-creating the XCFramework each time a target finishes building? That's one way I guess 😅

README.md
elliottwilliams: It depends on the current Xcode version. If you're on 12, `--create-xcframework` is default, if you're on 11 or below, `--no-create-xcframework` is default: https://github.com/Carthage/Carthage/pull/3071/files#diff-1ef6915d12e67971af0a706b0f50605779bf84b0bb3593e90b0862a4f01b64af

So this section of the readme is written to be correct regardless of your Xcode.

tbh, I don't like how dynamic this is, but it feels like the right compromise. If the argument was false by default, this PR breaks backwards-compatibility for users on older Xcodes, _or_ it requires devs on Xcode 12 to always pass `--create-xcframework` to get a working build.

Source/CarthageKit/BuildOptions.swift
elliottwilliams: I think the only place this default value is actually used is in unit tests (see my comment above)

README.md
tmspzz: I wouldn't expect the behaviour of the default flags to change depending on external conditions (i.e. the installed version of Xcode). Imho it should always be `--no-create-xcframeworks` unless otherwise specified.

tmspzz: this makes it very hard to predict what carthage will do just by looking at the CLI. Think of scrips and other tooling (bug reports) that would also have to independently check for the installed version of Xcode.

elliottwilliams: Totally understand the concern. I'm willing to make --no-create-xcframeworks the default. I should be able to add an error when Carthage can anticipate that lipo will fail, so that users know they need to re-reun with --create-xcframeworks.

chrisballinger: Why keep a default that is broken on the latest version of Xcode? It will only lead to increased support burden as confused users don't understand why Carthage is still broken on Xcode 12 and requires looking at the documentation to discover the new special flag `--create-xcframeworks`. 

There are a lot of developers incorrectly removing ARM iOS Simulator slices to work around the Carthage Xcode 12 issues. There is no future workflow where the old behavior of `--no-create-xcframeworks` is correct: users of Apple Silicon Macs need xcframeworks for the ARM iOS Simulator, and as well as those who are shipping Mac Catalyst apps.

I would strongly urge to use defaults that "just work".

kohenkatz: To further support the above, Xcode 12 will be required by Apple for all iOS and iPadOS apps in just a few months, and I suspect the vast majority of developers will upgrade (even if they only develop macOS apps), so the default working only in old versions doesn't seem very useful.

elliottwilliams:
> @elliottwilliams did I miss it or handling of binary dependencies with .xcframeworks is not handled here?

@tmspzz That's correct. Binary dependencies and archiving are the two things I know still need work. I was thinking about doing that in a separate PR, since this one has been pretty stable for a couple weeks and since it handles most common workflows. Thoughts?

> If I understood correctly, this works by re-creating the XCFramework each time a target scheme finishes building? That's one way I guess 😅

Yeah 🙈 There's no way to "update" an xcframework using `xcodebuild -create-xcframework`, and since Carthage lets you build repeatedly with different `--platform` settings, I wanted to ensure that you couldn't accidentally overwrite a previously-built platform in your xcframework. Do you think that's a use case we should handle?

Source/CarthageKit/Xcode.swift
elliottwilliams: This logic is key to giving users control over whether they want to use xcframeworks, instead of requiring library maintainers to make that decision for everyone at once.

If we don't do something like this, then every project with Carthage dependencies would have to update itself to link xcframeworks. This would be a breaking change in the library's project, and they'd lose support for Xcode 11 / `carthage build --no-create-xcframeworks` in the process.

That said, this code only runs if you have xcframeworks in Carthage/Build, so we'd only be doing this if the user builds with --create-xcframework.

elliottwilliams: Oh and if it helps, this logic is similar to what Xcode does when it's told to link an xcframework 🙂 

It runs a pre-compile step which extracts the platform-specific framework bundle into $BUILT_PRODUCTS_DIR. These extracted frameworks get discovered by the compiler, since $BUILT_PRODUCTS_DIR is always implicitly part of the FRAMEWORKS_SEARCH_PATH.

README.md
elliottwilliams: Yeah @chrisballinger, I think that's the strongest argument to defaulting to true. It's also what I had in mind when I implemented this, and I agree that it's a "bad default" from the perspective of an Xcode 12 user.

If we take @tmspzz's suggestion and do the conservative thing, I think we _must_ give users a meaningful error directing them to the argument.

chrisballinger: @elliottwilliams Why not just always build xcframeworks on both Xcode 11 and Xcode 12? That would get rid of the ambiguity between versions, and will always do the correct thing going forward. Users are going to have to deal with a breaking change at some point, and it might as well be now. I guess Xcode 11 doesn't support the `-debug-symbols` flag for dSYMs/bcsymbolmaps in xcframeworks, but perhaps that could be guarded against.

elliottwilliams:
> I don't see the .version file being generated when using
> 
> ```
> carthage update --create-xcframework --cache-builds --new-resolver --no-use-binaries
> ```
> 
> Cartfile:
> 
> ```
> github "Alamofire/Alamofire"
> ```

Great catch! Looks like it wasn't able to get the swift version from dSYMs inside xcframeworks. Fixed in https://github.com/Carthage/Carthage/pull/3071/commits/8f8bbc66a756da944518f4ac843da2949d21f5b5

README.md
tmspzz: > There is no future workflow where the old behavior of --no-create-xcframeworks is correct: users of Apple Silicon Macs need xcframeworks for the ARM iOS Simulator, and as well as those who are shipping Mac Catalyst apps.
> I would strongly urge to use defaults that "just work".

Carthage is compatible all the way down to High Sierra.

A new flag should not be enabled by default.

tmspzz: A middle ground would be ending support for all systems where Xcode 12 is not supported before cutting the release with support for XCFrameworks.

The defaulting to `--create-xcframeworks` would make sense

jdhealy: > A new flag should not be enabled by default.

Also to consider, certain CI systems that upgrade Carthage without user confirmation. If avoidable, we aim to provide least-possible disruption to those users (continuing to compile their projects is ideal).

elliottwilliams: > A middle ground would be ending support for all systems where Xcode 12 is not supported before cutting the release with support for XCFrameworks.
>
> Then defaulting to `--create-xcframeworks` would make sense.

If you all are open to this, I think it's a good direction to move in. It would also give us an opportunity to drop source compatibility with Xcode 10.1 / Swift 4 and modernize some of Carthage's dependencies.

But I don't want this PR to hinge on the larger task of dropping support for older OSes and Carthage versions. And I'd like to avoid bikeshedding on the default value of this argument, as much as possible 😆

So how about: we do the safe thing and default to --no-create-xcframeworks, _with the understanding that --create-xcframeworks will become the default in the future_. Once this is reviewed and merged, dropping support for older Xcodes is one of the things I / others can start to work on. We can safely make a release with these changes before then.

afallon02:
Hey this patch looks great. I've been involved in looking at removing Carthage from a project due to this issue, but the fix would have been to generate XCFrameworks - So if Carthage can do this instead that would be great! Is it always expected that the end user will need to drag the xcframework into the project when using the functionality? What are the complexities around scripting that part?

elliottwilliams:
> Is it always expected that the end user will need to drag the xcframework into the project when using the functionality? What are the complexities around scripting that part?

@afallon02 That's correct! They will need to embed the xcframework using Xcode and _not_ use `carthage copy-frameworks`. Once it's embedded, Xcode will extract the correct bundle from the xcframework to the target's BUILT_PRODUCTS_DIR, which is how it actually gets picked up by the compiler.

Here are the new integration instructions from this PR. LMK if there's anything I can clarify: https://github.com/Carthage/Carthage/blob/8f8bbc66a756da944518f4ac843da2949d21f5b5/README.md#building-platform-independent-xcframeworks-xcode-12-and-above

ferologics:
Can't wait for this to get shipped so we can remove the workaround script from our codebase! 👍

superharoun:
Hi,
I am quite new to all of this, so I try to get how this works overall. 
Is there a way to install this "fork" (is that the right term) until it is release in master ? If so, how ? Can it be done through Homebrew ? Or how ?
Otherwise, I understand I should try out the .sh workaround until this is merged to Carthage:master ? Do we have an idea when that would happen, eventually ?
Sorry for the noob questions, I hope I'll catch up soon ! Regards

elliottwilliams:
> Hi,
> I am quite new to all of this, so I try to get how this works overall.
> Is there a way to install this "fork" (is that the right term) until it is release in master ? If so, how ? Can it be done through Homebrew ? Or how ?
> Otherwise, I understand I should try out the .sh workaround until this is merged to Carthage:master ? Do we have an idea when that would happen, eventually ?
> Sorry for the noob questions, I hope I'll catch up soon ! Regards

Hey @superharoun! I recommend sticking with the workaround script in master for the time being. This branch is in development and subject to change or break – you shouldn't depend on it until we've decided to merge :)

superharoun:
@elliottwilliams many thanks, will give it a try, i need to learn from all of this 👍🏼

Source/CarthageKit/Xcode.swift
elliottwilliams: It is now: https://github.com/Carthage/Carthage/pull/3071/commits/f151e9a14e6b2361ef78f905eec15b1898f631ee

README.md
elliottwilliams: This is done: https://github.com/Carthage/Carthage/pull/3071/commits/9b61e11133810e0ce0ab4454187389cc9457285d

One thing I was thinking about is that we _could_ make --create-xcframeworks the default _if and only if_ Carthage knows that it's going to encounter a lipo failure. This would preserve the behavior of any currently-successful carthage invocation, including when the user is using the EXCLUDED_ARCHS workaround. But for iOS devs building on Xcode 12+ (which I imagine is 90%+ of our users), they would always get xcframeworks instead of build errors.

Not sure I want to do this now, because it's additive and it would require refactoring the point in the build where we decide whether to make an xcframework or a universal framework. Currently we do this immediately after building a scheme, so we'd instead have to wait until all schemes are built to decide how to merge their products.

elliottwilliams:
> Edit : I have been trying around, managed to launch the .sh but somehow this still fails.

@superharoun this thread is not the right place to discuss build errors :) If you need help with the workaround script, I'd ask in #3019. Would you mind pinging me there? I wanna keep this thread as on-topic as possible!

nighthawk:
I gave this a shot, but it's stuck after the first archive succeeds. It just sits there and nothing else happens.

`carthage-xcf update --create-xcframework --cache-builds --new-resolver --no-use-binaries`

Is stuck at the first build:

```
*** xcodebuild output can be found in /var/folders/90/y0c1zrxj1kb8t7m29h8h03n00000gn/T/carthage-xcodebuild.l2ZUS1.log
*** Building scheme "ASWeekSelectorView" in ASWeekSelectorView.xcodeproj
```

Checking that log shows it's done and succeeded with:

```
** ARCHIVE SUCCEEDED ** [2.265 sec]
```

I removed that framework and tried again. Same issue, but with the next framework. The same thing happens if I also specify a platform or when using `bootstrap` or `build`. I'm unfamiliar with the Carthage internals, so I'm not sure how to dig deeper, but if this is an issue worth investigating, I'm happy to help someone who knows more about the internals. Otherwise, I'll be patient and wait for this to be merged.

elliottwilliams:
> I gave this a shot, but it's stuck after the first archive succeeds. It just sits there and nothing else happens.

Yep, I broke something and you caught this PR at a bad time, hang tight. The tests failed in the same manner.

renep:
I'm currently trying to get everything working on my new M1 MacMini and try to migrate to xcframework using this pull request in the hope that this pull request will get merged in the future.

My problem is now that I cannot create the xcframework with one of my depencencies. I get the following error: 
```
/Users/rene/workspace/github/SwiftHamcrest/Package.swift:4:8: error: no such module 'PackageDescription'
import PackageDescription
```
It does not matter if I try to build [SwiftHamcrest](https://github.com/nschum/SwiftHamcrest) as project dependency with `carthage update --create-xcframework` or try to create an archive build in the project itself with `carthage build --archive --create-xcframework`

My question is now if there is something not correct in the project setup of the SwiftHamcrest project, or is it an issue with this pull request.
For my test I have used the latest version of the pull request.

renep:
The `carthage archive` command does not work with xcframeworks. Here an error message is displayed: `Could not find any copies of <FrameworkName>.framework`.

I have investigated the code and found out the the `archive` searches for `*.framework` in the `Carthage/Build` folder and then add the directories to the zip. For xcframeworks the `.framework` is not located directly in `Carthage/Build` but in `Carthage/Build/<FrameworkName>.xcframework/<arch>/<FrameworkName>.framework`. So the framework is not found an the zip is not created. 
Also it is possible to specify the framework names with the archive command and only frameworks that are found are included in the zip.

The question is now how the solution to support xcframework in archives should look. Maybe the simplest solution would be to add also the option `--create-xcframework`. If this option is present all `*.xcframework` directories are added to the zip. 
If the framework names are specified, only `*.xcframework` are added that has this name specified. But it is not performed a deep check if the xcframework really does contain the a `<FrameworkName>.framework`

If that's fine I can try to implement this and make a pull request.

elliottwilliams:
> The `carthage archive` command does not work with xcframeworks. Here an error message is displayed: `Could not find any copies of <FrameworkName>.framework`.

@renep That's right. I'm not trying to handle archiving in the PR (at least for now), in the hopes that I can keep this changeset as small as possible, and get this code into master before building on top of it.

I've started to play with how we'd support this, though. If you're curious: https://gist.github.com/elliottwilliams/13f39f68c35de220be61dd2ba709c24a

You're welcome to start messing around yourself, but I fear that opening PRs based off of this one will be a headache for reviewers. And I can't stress enough that all of the code here is provisional, we make no commitments to merge it as-is, etc etc 🙂 

If you're on an M1 Mac and blocked by not being able to make archives, I believe you can use the workaround xcconfig described in #3019, edited to exclude x86_64 instead of arm64. Feel free to ping me over there if I can be helpful!

renep:
@elliottwilliams If course the priority should be to get the PR merged as fast as possible. Therefore I will not mess around in this PR. 
Having binary archives is not a show stopper for me, but it speeds up build on my CI a lot, so this is a thing a want to have when this PR is merged and I have updated my projects to use xcframeworks.

shaps80:
> If you're on an M1 Mac and blocked by not being able to make archives, I believe you can use the workaround xcconfig described in #3019, edited to exclude x86_64 instead of arm64. Feel free to ping me over there if I can be helpful!

I'm having the following issue, I _think_ its related to this. 

- I'm running on an M1 Mac mini
- I'm using the workaround script since I was previously using Xcode 12 on an Intel Mac mini
- `carthage.sh update` compiles the frameworks
- Xcode compile throws an error when selecting iPhone 11/12 simulators
    - Building for iOS Simulator, but the embedded framework was built for iOS
- I tried adjusting the script (not sure I know what I'm doing here though) but then `carthage.sh update` now fails

Any advice or example of the changes I should be making would be appreciated.

renep:
@shaps80 With the `carthage archive` it was meant to creaste a zip file with a prebuild xcframework binary. Similar to https://github.com/Carthage/Carthage#archive-prebuilt-frameworks-into-one-zip-file

When using xcframework and an M1 Mac you do not need the workaround the Xcode 12. The xcframework is the proper solution to solve this issue.

You should also checkout the Readme of this pull request how to setup your project, because the old copy frameworks script is also not needed anymore: https://github.com/elliottwilliams/Carthage/tree/xcframeworks#building-platform-independent-xcframeworks-xcode-12-and-above

shaps80:
Oh perfect I'll try this. Would definitely love to use the correct approach!

3a4oT:
> @renep That's right. I'm not trying to handle archiving in the PR (at least for now), in the hopes that I can keep this changeset as small as possible, and get this code into master before building on top of it.
> 
> I've started to play with how we'd support this, though. If you're curious: https://gist.github.com/elliottwilliams/13f39f68c35de220be61dd2ba709c24a
> 

@elliottwilliams Thank you! Great work!

3a4oT:
So I tried it locally with my Cartfile and everything seems to be working but I have two questions:

1.  `carthage update --create-xcframework` still generates `fat frameworks` and put those to corresponding folders "iOS", "Mac". etc. Maybe it can be disabled by default or some subcommand to disable it like `--no-fat-frameworks`?
2. Is it possible to add a macCatalyst variant of `xcframework`?

elliottwilliams:
> carthage update --create-xcframework still generates fat frameworks and put those to corresponding folders "iOS", "Mac". etc. Maybe it can be disabled by default or some subcommand to disable it like --no-fat-frameworks?

@3a4oT hmm, could you share a cartfile and the commit from this PR you were using? That  shouldn't happen. Might also be worth confirming that those platform-specific fat frameworks weren't from previous invocations of Carthage: running with `--create-xcframeworks` won't delete old framework bundles, it'll only make new xcframeworks.

> Is it possible to add a macCatalyst variant of xcframework?

Not in this PR - but I think that's an obvious next step.

fr-machlin:
maybe related to https://github.com/firebase/firebase-ios-sdk/discussions/7129

Source/CarthageKit/BuildSettings.swift
```diff
@@ -173,6 +173,12 @@ public struct BuildSettings {
 		return productType.fanout(machOType).map(FrameworkType.init)
 	}
 
+	internal var frameworkSearchPaths: Result<[URL], CarthageError> {
+		return self["FRAMEWORK_SEARCH_PATHS"].map { paths in
+			paths.split(separator: " ").map { URL(fileURLWithPath: String($0), isDirectory: true) }
```
elliottwilliams, line 178: this likely needs to avoid splitting escaped spaces

1oo7:
This PR has been open for over two months now. Is there still some issue that is preventing its approval or what's the status? We kind of need a path forwards since Xcode 12.3 has broken the ability to use the workaround script on frameworks that need to be embedded in other frameworks.

elliottwilliams:
> This PR has been open for over two months now. Is there still some issue that is preventing its approval or what's the status? We kind of need a path forwards since Xcode 12.3 has broken the ability to use the workaround script on frameworks that need to be embedded in other frameworks.

@1oo7 as I understand it, there’s no major hold-up, it’s just that this is a huge change and there’s not a ton of bandwidth from maintainers to review. I don’t think anyone is unaware of the urgency.

1oo7:
> > This PR has been open for over two months now. Is there still some issue that is preventing its approval or what's the status? We kind of need a path forwards since Xcode 12.3 has broken the ability to use the workaround script on frameworks that need to be embedded in other frameworks.
> 
> @1oo7 as I understand it, there’s no major hold-up, it’s just that this is a huge change and there’s not a ton of bandwidth from maintainers to review. I don’t think anyone is unaware of the urgency.

I've tested it locally with my org's pretty demanding Carthage flow and it seems fine to us. And switching to XCFrameworks resolves a couple of other build issues we're seeing with Xcode 12.3. 

However I'm not a maintainer so I can't really approve it. Hopefully someone has some coffee at home

CorbinMontague:
+1 to @1oo7's comments. My company also hasn't seen any issues with Carthage so far as we've tested this PR, but we do have more testing to do. With Xcode 12.3 breaking the existing Carthage workaround script, getting XCFrameworks support into Carthage soon would sure be nice!

kcharwood:
Migrating a project right now to this work.

For an app where I have a main target, and an app extension that depend on the same library, what settings should I set for embedding/signing?

shaps80:
> You should also checkout the Readme of this pull request how to setup your project, because the old copy frameworks script is also not needed anymore: https://github.com/elliottwilliams/Carthage/tree/xcframeworks#building-platform-independent-xcframeworks-xcode-12-and-above

So when I remove the workout and just run `carthage update` I still get a `lipo` failure so, I can't get the framework to build.
Does the framework provider need to make any changes on their end?

I also pulled this repo and did a `make install` in case I needed this version of `carthage` but it made no difference :(

renep:
> So when I remove the workout and just run `carthage update` I still get a `lipo` failure so, I can't get the framework to build.
> Does the framework provider need to make any changes on their end?

Run `carthage update --create-xcframework`

shaps80:
> Run `carthage update --create-xcframework`

Ahhh! Do I need to be on this branch?

shaps80:
Hmmm... so when I run `carthage update —create-xcframework` it looks like it obviously just does the right thing for the checkouts it can.

For pre-built frameworks, I still need to do a full `carthage update` (I assume not with the workaround?).

Trying it out, hopefully this works.

shaps80:
@renep thank you so much for your support/help here, really needed it.

Ok, so I’ve managed to get all the frameworks to compile as above which is great. Xcode also compiles for device.

However when building for iPhone 12 mini Simulator for example:
```
Failed to build module 'Onfido' from its module interface; the compiler that produced it, 'Apple Swift version 5.1.3 effective-4.1.50 (swiftlang-1100.0.282.1 clang-1100.0.33.15)', may have used features that aren't supported by this compiler, 'Apple Swift version 5.3.2 effective-4.1.50 (swiftlang-1200.0.45 clang-1200.0.32.28)'
```

This is a pre-built library, so Carthage isn’t building this one. I’m guessing this is unrelated?

CorbinMontague:
@elliottwilliams A few of my Carthage dependencies do not get an xcframework built when I run `carthage bootstrap --create-xcframework --cache-builds --platform iOS`. The one I'm looking at right now is [Branch](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution). The logs show that carthage is downloading the framework directly (`Downloading ios-branch-deep-linking-attribution.framework binary at "Branch iOS SDK 0.36.0"`). When I compare this to other carthage dependencies I have that successfully have an xcframework built, the logs says: `Building scheme "Foo" in Foo.xcodeproj`. Is there something that Branch needs to do to support being built as an xcframework by Carthage?

elliottwilliams:
> @elliottwilliams A few of my Carthage dependencies do not get an xcframework built when I run `carthage bootstrap --create-xcframework --cache-builds --platform iOS`. The one I'm looking at right now is [Branch](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution). The logs show that carthage is downloading the framework directly (`Downloading ios-branch-deep-linking-attribution.framework binary at "Branch iOS SDK 0.36.0"`). 

Downloading binary dependencies is a separate codepath, and is [something I intend to address in a follow-up PR](https://github.com/Carthage/Carthage/pull/3071#issuecomment-733180109). The "Downloading <name> at <version>" is a giveaway that Carthage has found a binary version and is preferring it.

Branch's repo has source code available, so you should be able to pass `--no-use-binaries` and have Carthage build a fresh xcframework.

elliottwilliams:
> For an app where I have a main target, and an app extension that depend on the same library, what settings should I set for embedding/signing?

@kcharwood See instructions in README.md on this branch: https://github.com/elliottwilliams/Carthage/blob/xcframeworks/README.md#building-platform-independent-xcframeworks-xcode-12-and-above

Signing shouldn't be any different than it has been. The only change wrt embedding is that you'll use a Copy Files phase instead of `carthage copy-frameworks`.

Patrick-Kladek:
I'm using `github "MessageKit/MessageKit"` in my Cartfile and got the following error (running `carthage update --create-xcframework`):

```
*** Building scheme "MessageKit" in MessageKit.xcodeproj
Build Failed
	Task failed with exit code 65:
	/usr/bin/xcrun xcodebuild -project /Users/patrick/Developer/Medbee/Carthage/Checkouts/MessageKit/MessageKit.xcodeproj -scheme MessageKit -configuration Release -derivedDataPath /Users/patrick/Library/Caches/org.carthage.CarthageKit/DerivedData/12.3_12C33/MessageKit/3.4.2 -sdk iphoneos ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES archive -archivePath /var/folders/_h/n2d19nrj2dvbxljvt7jrrzmr0000gn/T/MessageKit SKIP_INSTALL=YES GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=NO CLANG_ENABLE_CODE_COVERAGE=NO STRIP_INSTALLED_PRODUCT=NO (launched in /Users/patrick/Developer/Medbee/Carthage/Checkouts/MessageKit)

This usually indicates that project itself failed to compile. Please check the xcodebuild log for more details: /var/folders/_h/n2d19nrj2dvbxljvt7jrrzmr0000gn/T/carthage-xcodebuild.5hCKLp.log
```

The last part of the log

```
SetOwnerAndGroup patrick:staff /Users/patrick/Library/Caches/org.carthage.CarthageKit/DerivedData/12.3_12C33/MessageKit/3.4.2/Build/Intermediates.noindex/ArchiveIntermediates/MessageKit/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/MessageKit.framework (in target 'MessageKit' from project 'MessageKit')
    cd /Users/patrick/Developer/Medbee/Carthage/Checkouts/MessageKit
    /usr/sbin/chown -RH patrick:staff /Users/patrick/Library/Caches/org.carthage.CarthageKit/DerivedData/12.3_12C33/MessageKit/3.4.2/Build/Intermediates.noindex/ArchiveIntermediates/MessageKit/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/MessageKit.framework

SetMode u+w,go-w,a+rX /Users/patrick/Library/Caches/org.carthage.CarthageKit/DerivedData/12.3_12C33/MessageKit/3.4.2/Build/Intermediates.noindex/ArchiveIntermediates/MessageKit/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/MessageKit.framework (in target 'MessageKit' from project 'MessageKit')
    cd /Users/patrick/Developer/Medbee/Carthage/Checkouts/MessageKit
    /bin/chmod -RH u+w,go-w,a+rX /Users/patrick/Library/Caches/org.carthage.CarthageKit/DerivedData/12.3_12C33/MessageKit/3.4.2/Build/Intermediates.noindex/ArchiveIntermediates/MessageKit/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/MessageKit.framework

RegisterExecutionPolicyException /Users/patrick/Library/Caches/org.carthage.CarthageKit/DerivedData/12.3_12C33/MessageKit/3.4.2/Build/Intermediates.noindex/ArchiveIntermediates/MessageKit/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/MessageKit.framework (in target 'MessageKit' from project 'MessageKit')
    cd /Users/patrick/Developer/Medbee/Carthage/Checkouts/MessageKit
    builtin-RegisterExecutionPolicyException /Users/patrick/Library/Caches/org.carthage.CarthageKit/DerivedData/12.3_12C33/MessageKit/3.4.2/Build/Intermediates.noindex/ArchiveIntermediates/MessageKit/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/MessageKit.framework

Touch /Users/patrick/Library/Caches/org.carthage.CarthageKit/DerivedData/12.3_12C33/MessageKit/3.4.2/Build/Intermediates.noindex/ArchiveIntermediates/MessageKit/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/MessageKit.framework (in target 'MessageKit' from project 'MessageKit')
    cd /Users/patrick/Developer/Medbee/Carthage/Checkouts/MessageKit
    /usr/bin/touch -c /Users/patrick/Library/Caches/org.carthage.CarthageKit/DerivedData/12.3_12C33/MessageKit/3.4.2/Build/Intermediates.noindex/ArchiveIntermediates/MessageKit/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/MessageKit.framework

** ARCHIVE FAILED **


The following build commands failed:
	CompileSwift normal arm64
	CompileSwiftSources normal arm64 com.apple.xcode.tools.swift.compiler
(2 failures)
```

Not sure if its relevant but MessageKit has a subdependency. Cartfile:
```
github "nathantannar4/InputBarAccessoryView" ~> 5.2.1
```

And Cartfile.private
```
github "Quick/Quick" ~> 2.0.0
github "Quick/Nimble" ~> 8.0.0
```

Has someone an idea what's wrong?

renep:
> Not sure if its relevant but MessageKit has a subdependency:

This could be the case. Try to checkout the MessageKit (and maybe the other dependencies) and see if can be build as xcframwork using: 'carthage build --archive --create-xcframework'

I myself had to update some framework that I use to support xcframeworks.

shaps80:
@elliottwilliams / @renep sorry to be a pain. I have a closed-source dependency. I assume this is not something you can resolve, so if they don't support Apple Silicon, I'm essentially screwed?

Patrick-Kladek:
@renep Thanks for your help.

I tried that. I had to add `--platform iOS` so the project compiles but Carthage still showed an error message:

```
cd Developer/MessageKit
carthage build --archive --create-xcframework --platform iOS
*** xcodebuild output can be found in /var/folders/_h/n2d19nrj2dvbxljvt7jrrzmr0000gn/T/carthage-xcodebuild.Hmq82Z.log
*** Building scheme "InputBarAccessoryView" in InputBarAccessoryView.xcworkspace
*** Building scheme "Nimble-iOS" in Nimble.xcodeproj
*** Building scheme "Quick-iOS" in Quick.xcworkspace
*** Building scheme "MessageKit" in MessageKit.xcodeproj
Could not find any copies of MessageKit.framework. Make sure you're in the project's root and that the frameworks have already been built using 'carthage build --no-skip-current'.
```

`Carthage/Build` does contain 4 xcframeworks (the 3 dependencies + MessageKit.xcframework). Not sure if the error message from Carthage is a big issue. For me it looks like the xcframeworks were build successfully.

Bildschirmfoto 2020-12-17 um 15 09 32

However adding MessageKit as a dependency to my project results in the error message in my previous comment.

The same issue also happens for `CTAssetsPicker` (I have a fork which I maintain as official development is suspended). `CTAssetsPicker` uses `PureLayout` as a sub dependency so it's not specific to MessageKit.

PS: I'm using Version 3.0.0 of MessageKit if it helps reproducing.

renep:
> Could not find any copies of MessageKit.framework. Make sure you're in the project's root and that the frameworks have already been built using 'carthage build --no-skip-current'.

You can ignore this error message. The reason is that this pull request cannot create a zip file for the created binaries yet. As @elliottwilliams already mentioned, this will be part of an additional pull request.

tmspzz:
Thanks @elliottwilliams. Off we go 🚀

EmDee:
@tmspzz Do you have a rough ETA for a new release?

0x5e:
Thanks everyone!

kientux:
When I use both `--cache-builds` and `--use-xcframeworks`, Carthage won't create xcframeworks even when xcframeworks doesn't exist. It checks for `.framework` only and if the framework is built before, it assumes build is cached and skips building altogether.

Can Carthage check if `.framework` is available, skip build step, and run only `xcframework` step?

elliottwilliams:
> When I use both `--cache-builds` and `--use-xcframeworks`, Carthage won't create xcframeworks even when xcframeworks doesn't exist. It checks for `.framework` only and if the framework is built before, it assumes build is cached and skips building altogether.
> 
> Can Carthage check if `.framework` is available, skip build step, and run only `xcframework` step?

@kientux not without additional code changes. My intention was that you'd need to clear out your Carthage/Build folder when first switching to xcframeworks. 

Though I think it would be reasonable to have the version file logic only consider xcframeworks when `--use-xcframeworks` is passed, and ignore them when it isn't.

y7aa:
J

PR #3095, opened 2020-12-18
Fix Xcode 12 refusing to link fat binaries

tmspzz:


Source/CarthageKit/Xcode.swift
```diff
@@ -962,6 +962,12 @@ private func build(sdk: SDK, with buildArgs: BuildArguments, in workingDirectory
 					let actions: [String] = {
 						var result: [String] = [xcodebuildAction.rawValue]
 
+						if settings.contains(where: { UInt64($0["XCODE_VERSION_ACTUAL"].recover("")) ?? UInt64.zero >= 1200 }) {
+							// Fixes Xcode 12 refusing to link against fat binaries
+							// "Building for iOS Simulator, but the linked and embedded framework 'REDACTED.framework' was built for iOS + iOS Simulator."
+							result += [ "VALIDATE_WORKSPACE=NO" ]
+						}
```
elliottwilliams, line 969: We can skip this when using xcframeworks, right?

I'm thinking: pass `options.useXCFrameworks` into `build(sdk:with:in:)` and check it here.

tmspzz:
@elliottwilliams done

tmspzz:
@elliottwilliams  as discussed it makes sense to permanently inject this in the case a fat binary out of carthage's control is linked

elliottwilliams (approved):
Yeah, I'm comfortable with using it unconditionally given our discussion. To summarize for GH: we don't want to break projects which might ship with an existing universal framework (i.e. _not_ frameworks produced by carthage at `carthage build`-time).

Issue #3112, opened 2021-01-19
Carthage linker error


arunkkumarr2020:
* `carthage version`:  0.36.1
* `xcodebuild -version`:  12C33
* Are you using `--no-build`?  No
* Are you using `--no-use-binaries`?  No
* Are you using `--use-submodules`?  No
* Are you using `--cache-builds`?  No

After upgrading to latest carthage version when I compile the project it fails. I have copy script in  `Run Scrip` 
`/usr/local/bin/carthage copy-frameworks` if I comment that out it compiles otherwise I get following error:

`/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: one of -create, -thin <arch_type>, -extract <arch_type>, -remove <arch_type>, -replace <arch_type> <file_name>, -verify_arch <arch_type> ... , -archs, -info, or -detailed_info must be specified
`

This work working fine in 0.36.0

billybooth:
A number of [Loop](https://github.com/LoopKit/Loop) users have been experiencing this as well, and it has been established that pinning Carthage to 0.36.0 does work around the issue:
https://github.com/LoopKit/Loop/issues/1468

The same carthage command:
https://github.com/LoopKit/Loop/blob/18af414396f5eaad1f06b8efff369b77143e5f94/Scripts/copy-frameworks.sh#L40-L41

Same error:
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: one of -create, -thin <arch_type>, -extract <arch_type>, -remove <arch_type>, -replace <arch_type> <file_name>, -verify_arch <arch_type> ... , -archs, -info, or -detailed_info must be specified
```

Would be nice to see this one addressed before 0.37.0.  Thanks!

bwhiteley:
Same problem here. We have to pin to 0.36.0.

danielrbrowne:
I'm experiencing this same issue since upgrading from `0.36.0` to `0.37.0`.

Given this bug was seemingly introduced in `0.36.1`, I went looking at [the diff](https://github.com/Carthage/Carthage/compare/0.36.0...0.36.1) and I think that https://github.com/Carthage/Carthage/pull/3066 is the PR that likely introduced the bug (given the various changes involving `xcrun lipo` contained within).

danielrbrowne:
@tmspzz Pinging you since you were involved in that PR - does this seem like the right theory?

[...]

Issue #3120, opened 2021-01-29
tvOS build fail saying lipo errors, but iOS and macOS pass, on Xcode 12


liuxuan30:
* carthage install method: [ ] .pkg, [X] homebrew, [ ] source
* `which carthage`:homebrew
* `carthage version`: 0.36.1
* `xcodebuild -version`: Xcode12.0.1 & Xcode 12.4
* Are you using `--no-build`? no
* Are you using `--no-use-binaries`? yes, via carthage.sh
* Are you using `--use-submodules`? no
* Are you using `--cache-builds`? no
* Are you using `--new-resolver`? no

Hi there, 

We have been using the cathrage.sh since Xcode 12 and it seems working great, but suddenly the tvOS failed starting in this month, but we changed nothing about the process.

**And furthermore, iOS and CI passed at the same time, with same code and setup**. It's really confusing for us how to fix it.

The full travis build record:https://travis-ci.org/github/danielgindi/Charts/builds/756550941

The error for tvOS build:
```
Processing Info.plist
▸ Running script 'Copy Carthage Frameworks'
❌  error: /Applications/Xcode-12.0.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: one of -create, -thin <arch_type>, -extract <arch_type>, -remove <arch_type>, -replace <arch_type> <file_name>, -verify_arch <arch_type> ... , -archs, -info, or -detailed_info must be specified
Testing failed:
	/Applications/Xcode-12.0.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: one of -create, -thin <arch_type>, -extract <arch_type>, -remove <arch_type>, -replace <arch_type> <file_name>, -verify_arch <arch_type> ... , -archs, -info, or -detailed_info must be specified
	Testing cancelled because the build failed.
** TEST FAILED **
The following build commands failed:
	PhaseScriptExecution Copy\ Carthage\ Frameworks /Users/travis/Library/Developer/Xcode/DerivedData/ChartsDemo-iOS-aeselevbqicktffmzvanvfzicpzm/Build/Intermediates.noindex/Charts.build/Debug-appletvsimulator/ChartsTests.build/Script-6BA68666BDA3FCF79C2A6801.sh
(1 failure)
rake aborted!
```

iOS passed: https://travis-ci.org/github/danielgindi/Charts/jobs/756550942
[...]

PR #3155, opened 2021-03-25
Fix invalid lipo command in stripArchitectures() when embedding dSYMs

chriseplettsonos:
If the architectures argument is empty (because there are no architectures to be stripped), this method will produce and execute an invalid lipo command because it will contain no -remove arguments. To fix this, I make the method a no-op if the input list of architectures is empty.


elliottwilliams (approved):
Looks great, thank you! We do something like this already [in `nonDestructivelyStripArchitectures`](https://github.com/Carthage/Carthage/blob/5347deeade26f7388ff9c536b98fb78eab2b532f/Source/CarthageKit/Xcode.swift#L1397-L1400) but this impacts dSYM embedding via  copyDebugSymbolsForFramework.

danielrbrowne:
@elliottwilliams / @chriseplettsonos I realise this is merged in already, but I've been active on an open issue for a little while RE: invalid `lipo` commands observed whilst executing the Copy Frameworks build phase, and this change seems kind of related. Would this change resolve #3112 (and #3120 which seems like the same `lipo` error), and if so will this fix land in `0.38.0`?

PR #3196, opened 2021-06-03
Fix alt media firebase url parse

akaffenberger:


Source/CarthageKit/BinaryProject.swift
```diff
@@ -67,4 +69,12 @@ public struct BinaryProject: Equatable {
...
 			}
 	}
+    
+    public static func isValidBinaryUrl(value: String) -> Bool {
+        let detector = try! NSDataDetector(types: NSTextCheckingResult.CheckingType.link.rawValue)
+        let matches = detector.matches(in: value, options: [], range: NSRange(location: 0, length: value.utf16.count))
+        let match = matches.first(where: { Range($0.range, in: value) != nil })
+        
+        return match != nil
+    }
```
elliottwilliams, line 79: I think I'd prefer we use the same URL and scheme check that this function uses elsewhere. As written, this will accept a parameter like `?alt=foo.co`, which will just be rejected later on as a `.nonHTTPSURL`.

What would you think about doing something like this?
```suggestion
	private static func validateBinaryURL(_ value: String) -> Result<URL, BinaryJSONError> {
		guard let binaryURL = URL(string: value) else {
			return .failure(BinaryJSONError.invalidURL(value))
		}
		guard binaryURL.scheme == "file" || binaryURL.scheme == "https" else {
			return .failure(BinaryJSONError.nonHTTPSURL(binaryURL))
		}
		return .success(binaryURL)
	}
```


and then checking further up, like
```swift
else if item.name == "alt", let value = item.value, 
	 case .success = validateBinaryURL(value) {
	// ...
}
```

We could also refactor `from(jsonData:)` to use this function instead of repeating the checks.

elliottwilliams (changes requested):
Thanks for submitting this! I need to think about this a bit more but overall I like the approach.

One formatting nit: you'll need to use tabs and not spaces for these code changes. It's annoying b/c there's no way to hard-code that setting in our Package.swift, but in Xcode you can set `Indent Using: Tabs` in the File Inspector, and then reindent these blocks with `Editor > Structure > Re-Indent`.

akaffenberger:
@elliottwilliams  Finally had some down time to add in the requested changes, let me know what you think!

akaffenberger:
In the spirit of hacktoberfest.. any chance this could get reviewed again soon?

PR #3200, opened 2021-06-09
Use modern Alamofire version in README

jshier:
Thanks for the mention, but Alamofire 4 is no longer recommended, so this PR update the README to use Alamofire 5 in the example.


tmspzz:
Thanks! 🥳

PR #3202, opened 2021-06-10
Add "--valid-simulator-archs" as a workaround for arm simulator issue

Interfere:
There is a known issue, that `lipo` is unable to build a fat binary when used with Xcode 12 or later. It does happen due to a fact, that the recent xcodebuild supports also modern MacBooks, which are having arm64 CPU. 

There are two known workarounds:
1. Use XCFramework
2. Exclude arm64 architecture when building for iOS simulator

Considering migration to XCFrameworks as a task that takes time, the latter options is widely used as a workaround — corresponding script was introduced as a wrapper for carthage binary.

In thi PR I would like to introduce a useful option for carthage to explicitly specify target architectures for simulator, which might also be used as a workaround for the mentioned issue. The flag allows to list the valid architectures for iOS Simulator target only.

Example:
```
~$ carthage bootstrap --valid-simulator-archs "i386 x86_64"
```

I have also added this option for some test fixtures, to be able to run all tests on my local machine, using Xcode 12.

### DIFF
--- a/Source/CarthageKit/BuildOptions.swift
+++ b/Source/CarthageKit/BuildOptions.swift
@@ -18,6 +18,8 @@ public struct BuildOptions {
 	public var useBinaries: Bool
 	/// Whether to create an XCFramework instead of lipoing built products.
 	public var useXCFrameworks: Bool
+	/// Explicitly metntion architectures for simulator
+	public var validSimulatorArchs: String?
 
 	public init(
 		configuration: String,
@@ -26,7 +28,8 @@ public struct BuildOptions {
 		derivedDataPath: String? = nil,
 		cacheBuilds: Bool = true,
 		useBinaries: Bool = true,
-		useXCFrameworks: Bool = false
+		useXCFrameworks: Bool = false,
+		validSimulatorArchs: String? = nil
 	) {
 		self.configuration = configuration
 		self.platforms = platforms
@@ -35,5 +38,6 @@ public struct BuildOptions {
 		self.cacheBuilds = cacheBuilds
 		self.useBinaries = useBinaries
 		self.useXCFrameworks = useXCFrameworks
+		self.validSimulatorArchs = validSimulatorArchs
 	}
 }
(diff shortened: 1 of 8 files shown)


elliottwilliams:
Thanks for the contribution! I can see the value in making it less painful to not adopt xcframeworks, but I have a few objections here:

* VALID_ARCHS is considered deprecated as of Xcode 12.0, so I don't think we should be making it easier for users to set it. From [Xcode 12.0 Release Notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-12-release-notes):
    > The Build Settings editor no longer includes the Valid Architectures build setting (VALID_ARCHS), and its use is discouraged. Instead, there is a new Excluded Architectures build setting (EXCLUDED_ARCHS). If a project includes VALID_ARCHS, the setting is displayed in the User-Defined section of the Build Settings editor. (15145028)

* Who is this for? Any project that uses Carthage without xcframeworks is using the workaround script, or building with their own XCODE_XCCONFIG_FILE set which disables the arm64 simulator. Any new project should use xcframeworks. So unless I'm missing something, it feels like merging this adds a convenience for users who don't really need it, and potentially leads new users away from what we want, which is `--use-xcframeworks`.

* I haven't dug too deep into it, but I imagine there's a security risk of passing a string directly to xcodebuild without any sanitization.

Source/CarthageKit/BuildOptions.swift
```diff
@@ -26,7 +28,8 @@ public struct BuildOptions {
 		derivedDataPath: String? = nil,
 		cacheBuilds: Bool = true,
 		useBinaries: Bool = true,
-		useXCFrameworks: Bool = false
+		useXCFrameworks: Bool = false,
```
Devkey19, line 31: []()

Tests/CarthageKitTests/ArchiveSpec.swift
```diff
@@ -48,13 +48,13 @@ class ArchiveSpec: QuickSpec {
 
 			it("should zip relative paths into an archive") {
 				let subdirPath = "subdir"
-				expect { try FileManager.default.createDirectory(atPath: subdirPath, withIntermediateDirectories: true) }.notTo(throwError())
```
Devkey19, line 51: []()

Devkey19 (approved):
[]()

PR #3204, opened 2021-06-16
Add: --use-xcframeworks check

ryu1sazae:
Add one more field to issue template:
- --use-xcframeworks:  which is added at 0.37.0. i think this flag can be the trigger some issues occur or not.


elliottwilliams:
Thanks for this!

PR #3213, opened 2021-06-29
Avoid loading xcframeworks as Bundles

elliottwilliams:
(NS)Bundle caches bundles as it loads them, so if a loaded bundle's metadata changes during program execution, it will not be reflected by the Bundle API. Carthage does this when it updates a preexisting XCFramework using [`mergeIntoXCFramework(in:settings:)`](https://github.com/Carthage/Carthage/blob/19a7f97112052394f3ecc33dac3c67e5384b7514/Source/XCDBLD/FrameworkBundle.swift#L50). 

This leads to situations where a variant framework inside an XCFramework is "missed" by Carthage and not copied by [`extractXCFrameworks(in:for:)`](https://github.com/Carthage/Carthage/blob/19a7f97112052394f3ecc33dac3c67e5384b7514/Source/CarthageKit/Xcode.swift#L845) until `carthage build` is run a second time. For example, consider the following dependencies:

```
A: builds for iOS, watchOS
B: depends on A, builds for iOS
```

When `carthage build --use-xcframeworks --platform iOS,watchOS` with an empty Carthage/Build directory:

1. A's watchOS variant builds, and is stored in a new xcframework, `Carthage/Build/A.xcframework`.
2. A's iOS variant builds. Carthage loads the Bundle for `A.xcframework` but does not extract anything (because `A.xcframework`'s Info.plist  only indicates watchOS variant frameworks).
3. B's iOS variant builds. Carthage has already loaded the Bundle for `A.xcframework`, and does not extract anything (because, as cached by the system, `A.xcframework`'s Info.plist only indicates watchOS variant frameworks). Building B fails.

If I run `carthage build --use-xcframeworks --platform iOS,watchOS` again, Carthage loads the Bundle for `A.xcframework` and sees both the watchOS and iOS variants, so building B succeeds.

While there's not an official way to flush the Bundle system's metadata cache, there's a [long-standing private CoreFoundation call that does it](https://michelf.ca/blog/2010/killer-private-eraser/). We can use that, and assertion-fail in Debug builds if it ever goes away.

### DIFF
--- a/Source/XCDBLD/FrameworkBundle.swift
+++ b/Source/XCDBLD/FrameworkBundle.swift
@@ -3,18 +3,6 @@ import ReactiveSwift
 import ReactiveTask
 import Result
 
-/// Private CoreFoundation call which flushes cached metadata for bundle objects.
-///
-/// https://michelf.ca/blog/2010/killer-private-eraser/
-/// https://opensource.apple.com/source/CF/CF-550.13/CFBundlePriv.h
-private let _CFBundleFlushBundleCaches: ((CFBundle) -> Void)? = {
-	let sym = dlsym(
-		UnsafeMutableRawPointer(bitPattern: -3), // RTLD_SELF
-		"_CFBundleFlushBundleCaches"
-	)
-	return unsafeBitCast(sym, to: (@convention(c) (CFBundle) -> Void)?.self)
-}()
-
 /// Loads a bundle directory from a given URL and sends Bundle objects for each framework in it.
 ///
 /// If `url` is an XCFramework, sends a Bundle for each embedded framework bundle.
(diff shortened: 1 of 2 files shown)


Source/XCDBLD/FrameworkBundle.swift
```diff
@@ -11,25 +11,24 @@ import Result
...
 /// - parameter variant: If given along with `platformName`, only sends bundles from an XCFramework with a matching `SupportedPlatformVariant`.
 public func frameworkBundlesInURL(_ url: URL, compatibleWith platformName: String? = nil, variant: String? = nil) -> SignalProducer<Bundle, DecodingError> {
-	guard let bundle = Bundle(url: url) else {
-		return .empty
-	}
-
-	switch bundle.object(forInfoDictionaryKey: "CFBundlePackageType") as? String {
-	case "XFWK":
+	switch url.pathExtension {
+	case "xcframework":
```
elliottwilliams, line 15: Checking for an "xcframework" path extension is new behavior, but I think this is safe. `xcodebuild -create-xcframework` forces you to output to a bundle directory with an `.xcframework` extension, and Xcode only extracts an xcframework at build-time if it ends in `.xcframework`. 

So unless there are some intermediate-and-badly-named XCFrameworks passing around inside Carthage, this should match them all.

PR #3254, opened 2022-02-01
Fix 'objdump' call not working with Xcode 13.3 command line tools

### LINKED ISSUE #3253: Binary frameworks are not copied to 'Build' folder when using Xcode 13.3
* carthage install method: [ ] .pkg, [x] homebrew, [x] source
* `carthage version`: 0.38.0
* `xcodebuild -version`: Xcode 13.3, Build version 13E5086k
* Are you using `--use-xcframeworks`? No

**Cartfile**
```
binary "https://dl.google.com/geosdk/GooglePlaces.json" == 6.0.0
```

**Carthage Input**
```
carthage bootstrap
```

**Carthage Output**
```
*** Downloading binary-only framework GooglePlaces at "https://dl.google.com/geosdk/GooglePlaces.json"
*** xcodebuild output can be found in /var/folders/pq/n433_zn50fxbr6qm48hg2m_80000gq/T/carthage-xcodebuild.vsX42A.log
*** Downloading binary-only framework GooglePlaces at "https://dl.google.com/geosdk/GooglePlaces.json"
```

**Actual outcome**
Carthage did not copy binary framework `GooglePlaces.framework` to `./Carthage/Build/iOS/` folder.

**Expected outcome**
Carthage should copy all binary frameworks to `./Carthage/Build/<Platform>/` folders.

mikchmie:
Fixes #3253


tmspzz:
@mikchmie thanks! Will this still work with previous versions of Xcode or macOS?

mikchmie:
I'm checking it right now. So far I've checked options' syntax in following versions of `objdump:
- Apple LLVM version 13.1.6 (clang-1316.0.19.2) - Xcode 13.3
- Apple LLVM version 13.0.0 (clang-1300.0.29.30) - Xcode 13.2.1
- Apple LLVM version 12.0.5 (clang-1205.0.22.9) - Command_Line_Tools_macOS_10.12_for_Xcode_8.3.2
and it looks the same (with two dashes `--`).

@tmspzz  What is the minimum version of Xcode that Carthage support?

mikchmie:
Looks like the syntax was changed in LLVM 10.
Last LLVM 9 docs capture: 
https://web.archive.org/web/20190616123225/http://www.llvm.org:80/docs/CommandGuide/llvm-objdump.html
First LLVM 10 docs capture: 
https://web.archive.org/web/20190812120449/http://llvm.org:80/docs/CommandGuide/llvm-objdump.html

mikchmie:
I've also found LLVM's historic documentation and looks like the change actually happened in LLVM 9 (released 19 Sep 2019).
I suppose this is a more reliable way to pinpoint the moment of change, than the WaybackMachnie.

LLVM 8.0.1 docs:
https://releases.llvm.org/8.0.1/docs/CommandGuide/llvm-objdump.html

LLVM 9.0.0 docs:
https://releases.llvm.org/9.0.0/docs/CommandGuide/llvm-objdump.html

mikchmie:
@tmspzz If that's necessary I can add a fallback to the chain, e.g.:
```
return task.launch(standardInput: nil)
        .flatMapError { _ -> SignalProducer<TaskEvent<Data>, TaskError> in
            // Fallback for 'objdump' in LLVM <9.0.0 (Xcode <11.4) which uses `-` as option prefix instead of `--`
            Task("/usr/bin/xcrun", arguments: [
                "objdump",
                "-macho",
                "-private-header",
                "-non-verbose",
                url.resolvingSymlinksInPath().path,
                ]
            ).launch(standardInput: nil)
        }
```

krzyzanowskim:
I tried this patch with Xcode 13.2.1 and binary xcframework didn't land in `./Carthage/Build/<Platform>/` as per #3253

mikchmie:
@krzyzanowskim As far as I can tell XCFrameworks (`*.xcframework`) will never land in `./Carthage/Build/<Platform>/` - instead they should be placed in `./Carthage/Build/` - as they usually contain binaries for multiple platforms. Only "basic" frameworks (`*.framework`) land in `./Carthage/Build/<Platform>/`.

krzyzanowskim:
@mikchmie you're right. I just noticed I face a different issue, where the Xcode project that supposes to use .xcframework, links to a non-existing framework from `./Carthage/Build/<Platform>/`

mikchmie:
Is the project itself "stale"? Are there any active maintainers who could make Carthage compatible with latest Xcode versions?

CraigSiemens:
We're also waiting for this fix.
- https://github.com/Carthage/Carthage/issues/3269

Source/CarthageKit/MachHeader.swift
```diff
@@ -70,14 +70,25 @@ extension MachHeader {
...
-			"-non-verbose",
+			"--macho",
+			"--private-header",
+			"--non-verbose",
 			url.resolvingSymlinksInPath().path,
 			]
 		)
 
 		return task.launch(standardInput: nil)
+            .flatMapError { _ -> SignalProducer<TaskEvent<Data>, TaskError> in
```
Ro-M, line 81: @mikchmie So to recap my understanding: You are
- launching the `Task` with the new `--` prefix option
- in case that fails assume that it failed because it still expects the old `-` prefix
- so you "restart" the `Task` but now with the old implementation

I think we should not just expect that the error is due to the new prefix. How about instead we check the currently installed version of the objdump via `objdump --version` which outputs something along the line of this for Command Line Tools bundled with Xcode 13.3.1:
```
Apple LLVM version 13.1.6 (clang-1316.0.21.2.3)
```
If the version is greater than or equal to 13.1 (or whatever version already works with it) we use the double-dash prefix, if not we use the single-dash prefix. Wdyt?

mikchmie: @Ro-M I've updated the implementation according to my proposal above. Please let me know what you think about it.

mikchmie: I've digged a bit deeper. The documentation seems to be incomplete for older versions.

LLVM `3.8.0` is the oldest version working with double-dash (which translates to Xcode `7.3`). 
In other words LLVM `3.7.0` (Xcode `7.2.1`) is the newest NOT working version. 

However LLVM `3.7.0` doesn't have the `-private-header` option - only `-private-headers` (with an 's'), which has a totally different output. LLVM `3.8.0` does have that option. So to me it seems like we don't need any fallback mechanism after all, because our command wouldn't work in `3.7.0` either way.

There are even 2 comments about Xcode `7.x` incompatibility in the `MachHeader.swift` file:
Line 64:
>	 Note: `objdump` invocation requires flags unimplemented in Xcode 7.X ⋯ as it stands,
	         only a moreso-alternative codepath invokes this; be cautious to not invert that.
	         
Line 68:
> TODO: Potentially, `mdfind` other non-version-7.X Xcodes that might contain valid `objdump`.

BTW I've realized that in Apple's builds `objdump --version` prints "Clang version" instead of "LLVM version" which used to be quite different in the past (e.g. check Xcode 11 in the [table](https://en.wikipedia.org/wiki/Xcode#Xcode_11.x_-_13.x_.28since_SwiftUI_framework.29_2) - LLVM `8.0.0` vs Clang `11.0.0`).

So to sum up - I think I can reset this PR to its initial commit.

Ro-M: > However LLVM 3.7.0 doesn't have the -private-header option - only -private-headers (with an 's'), which has a totally different output. LLVM 3.8.0 does have that option. So to me it seems like we don't need any fallback mechanism after all, because our command wouldn't work in 3.7.0 either way.

There is `--private-headers` (alias `-p` which is an option you can use for `objdumb` on its own (similar to `--macho`) and there is `--private-header` (NO alias) which is ONLY available as a sub-option for the `--macho` command (`MACH-O ONLY OPTIONS AND COMMANDS`)

mikchmie: But the current implementation doesn't use the `-private-headers` / `-p` option, so it doesn't matter, does it?

Running the current command (with single-dash syntax) using `objdump` `3.7.0`:
```
./objdump -macho -private-header -non-verbose ./GooglePlaces
```
returns an error:
```
objdump: Unknown command line argument '-private-header'.  Try: './objdump -help'
objdump: Did you mean '-private-headers'?
```

So it seems to me that changing single-dash to double-dash syntax will not introduce any regression (because this code doesn't currently  work with LLVM `< 3.8.0` either - the only difference will be a single character in the error message). Am I wrong?

Ro-M: I was only pointing out the differences in the two private-header options as that seemed to cause confusion on your side (it did initially for me 😄)

Ro-M:
@tmspzz could you please provide feedback here? Is there anything that you would like us to add / change / explain in order to move this forward?

mikchmie:
@Ro-M Thanks for the summary 👍 
Just 2 tiny corrections:
> * `objdump` is not accepting single-dash prefixed parameters starting with `LLVM 13.1.6` which is bundled with the Command Line Tools of `Xcode 13.3`

For the sake of clarity let's stick to one versioning system (the one described as `LLVM` in [table on Wikipedia](https://en.wikipedia.org/wiki/Xcode#Xcode_11.x_-_13.x_.28since_SwiftUI_framework.29_2)). `13.1.6` comes from the `Clang version string` which Apple's builds print as their version. The official `LLVM` version is `13.0.0`.

> * The new syntax is already supported since `LLVM 3.8.0` bundled with `Xcode 7.2.1`

Actually it's bundled since `Xcode 7.3`. `Xcode 7.2.1` used `LLVM 3.7.0`

Ro-M (approved):

irmakcan:
This change fixes my issue on Xcode 13.4 as well. Hope it gets merged and released soon.

mneuwert:
Hey dudes, did carthage bite the dust? Is somebody planning to merge it? Not usable anymore with Xcode 13.3 - 14. May be also just call `otool -h` as higher level abstraction?

Ro-M:
@mneuwert I'm not sure about the current plans from the maintainers. As for this PR I would like to move this forward but need direction as to what's missing here. We currently are using a fork of carthage that also includes this change. For the future we are looking into carthage alternatives that have a more active support 😞

mikchmie:
@giginet Could you review this PR?

PR #3334, opened 2023-04-17
avoid breaking lldb when copying the xcframeworks to other machines

Cyberbeni:
(debugging would also break when deleting the build intermediates from the cache, not just when transferring the xcframeworks to another machine)


Cyberbeni:
Here is the official source for the recommendation of using this build option: https://developer.apple.com/videos/play/wwdc2022/110370?time=1096

PR #3352, opened 2023-09-18
Fix tests — particularly «XcodeSpec» — and bump Quick to `v2.2.1` and Nimble to `v8.1.2`.

jdhealy:
Majority of changes are [in the tests file XcodeSpec.swift](https://github.com/jdhealy/Carthage/commits/b6ddf91cbfc1b8c87e796544cc57101e142665f6/Tests/CarthageKitTests/XcodeSpec.swift).

No changes to Carthage functionality.

No changes to which Xcode versions can build Carthage-the-tool and its unit tests.

-----

<ul>
<li>b6ddf91 · With Travis’ macOS specifics in the following link, our `.travis.yml` no longer launches any Travis runs: <a href=https://www.travis-ci.com/blog/2020-11-02-travis-ci-new-billing/>https://www.travis-ci.com/blog/2020-11-02-travis-ci-new-billing/</a>

For our future use on other CI platforms, I’ve included `for_use_during_unit_testing_for_carthage_on_intel_host_os.xcconfig` scripting.
</li><li>89251a1 · Tests: Recent Xcodes would fail to compile one or more of the projects of "carthage-fixtures-ReactiveCocoaLayout-master".

Remove comment "[t]he fixture is maintained at https://github.com/ikesyo/carthage-fixtures-ReactiveCocoaLayout" as we now use another (already-existing in github.com/carthage/carthage) fixture.

Three renamed tests now have names like "should build [solely] platform […] in total" instead of "should build for all platforms."

Two other tests are renamed: "should locate the project" → "should locate the workspace" and same for "from the parent directory."
〰️ A single new expectation is added to each of these that provides more elaboration due to limited debugDescription of `ProjectLocator`.
〰️ But expectations don’t short circuit, so the existing ‘last expectation’ will still always run.

Use `apple archive` (with CLI `aa`) to do file replacement in the `beforeEach` of `XcodeSpec`.

Copy `SampleMultipleSubprojects` an additional time in `copy-fixures` shell script (to `"$destination/SampleMultipleSubprojectsSameOldSameOld"`.)
</li><li>f80a77e - Tests: Remove (just for one commit) test which will be fully superseded by "should build for platform […] in total" case.

That test had a title of "should build all subprojects for all platforms by default".

See subsequent commit for the superseding test case.
</li><li> 2de64f1 · Tests: Comment on already existing coverage of most/some of `!SWIFT_PACKAGE` ones which are conditionalized as such because the fixtures for those don’t exist as `.frameworks` in their prior location when we made the years-ago switch to building with Swift Package Manager.
</li><li> fa47bca · Tests: Remove a test which provided limited value — it was titled `"should not locate the project from a directory not containing it"`.
</li><li> 890ddf1 · Tests: Prevent the XCTest agent from dying if xcodebuild dies during some ProjectSpec tasks.

Verified it compiles on Xcode 11.3.1 nicely. Untested at runtime. Best effort.
</li><li> bd25e31 · Tests: Use `example.com` in `BinaryProjectSpecs` and `ProjectSpec` test values.
</li><li> 52e3d71 · Tests: Change our URLComponents invalid url example — in one test variable — to use square brackets instead of an emoji.

This will make the test behave correctly on newer Foundation versions.
</li><li> b5c1bec - Improve Makefile.

Minor note: nothing about the 'MKDIR=mkdir -p' line was wrong — we’re just doing a more verbose thing everywhere that variable was previously used.

This is just based on a maintainer (me) spotting this and desiring the Makefile to be hardened. Nothing concrete led to this.
</li><li> f03c471 · Bump Quick to `v2.2.1` and Nimble to `8.1.2`.
</li><li> 1c0f79d · Use `Nimble.Predicate` with the module name spelled out in spots in «CarthageKitTests».
</li>
</ul>


Makefile
```diff
@@ -72,15 +81,15 @@ package: installables
...
 
 install: installables
-	if [ ! -d "$(BINARIES_FOLDER)" ]; then $(SUDO) $(MKDIR) "$(BINARIES_FOLDER)"; fi
-	$(SUDO) $(CP) -f "$(CARTHAGE_EXECUTABLE)" "$(BINARIES_FOLDER)"
+	if [ ! -d "$${BINARIES_FOLDER:?}" ]; then $(SUDO) mkdir -p "$${BINARIES_FOLDER:?}"; fi
+	$(SUDO) $(CP) -v -f "$(CARTHAGE_EXECUTABLE)" "$${BINARIES_FOLDER:?}"
 
 uninstall:
-	$(RM) "$(BINARIES_FOLDER)/carthage"
+	/bin/zsh --no-globalrcs --no-rcs -c 'set -x; rm -f -v "$${BINARIES_FOLDER:?}"'
```
jdhealy: Just fixed up (well before landing this PR) in https://github.com/Carthage/Carthage/pull/3352/commits/a9908175afa3ff0540cedb8e00d072d4c39933a9.

PR #3368, opened 2024-03-22
Fix endianness typo in MachHeader unused computed property internal to CarthageKit.

RoboSchmied:
Fix typo: endianess -> endianness


jdhealy:
Thanks for the contribution! 🎉

Issue #3375, opened 2024-06-11
[Xcode 16.0 Beta] Carthage fails to `xcodebuild archive -showBuildSettings`, mentions `no destinations`.


henrik-dmg:
* carthage install method: [ ] .pkg, [x] homebrew, [ ] source
* `which carthage`: `/opt/homebrew/bin/carthage`
* `carthage version`: 0.39.1
* `xcodebuild -version`: `Xcode 16.0 Build version 16A5171c`
* Are you using `--no-build`? No
* Are you using `--no-use-binaries`? No
* Are you using `--use-submodules`? No
* Are you using `--cache-builds`? No
* Are you using `--new-resolver`? No
* Are you using `--use-xcframeworks`? Yes

**Cartfile**
```
github "CocoaLumberjack/CocoaLumberjack"
```

**Carthage Output**
```
*** Fetching CocoaLumberjack
*** Checking out CocoaLumberjack at "3.8.5"
*** xcodebuild output can be found in /var/folders/4n/hzc7z9y54g3dmn6204bg0lfm0000gq/T/carthage-xcodebuild.6LK8Jt.log
*** Skipped downloading CocoaLumberjack binary due to the error:
	"Bad credentials"
A shell task (/usr/bin/xcrun xcodebuild -project /Users/hpanhans/Developer/CocoaLumberjackDebug/Carthage/Checkouts/CocoaLumberjack/Tests/Tests.xcodeproj -scheme iOS\ Tests -configuration Release CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES archive -showBuildSettings -skipUnavailableActions) failed with exit code 6:
2024-06-11 14:40:27.934 xcodebuild[99335:9919534] Writing error result bundle to /var/folders/4n/hzc7z9y54g3dmn6204bg0lfm0000gq/T/ResultBundle_2024-11-06_14-40-0027.xcresult
xcodebuild: error: Found no destinations for the scheme 'iOS Tests' and action archive.
2024-06-11 14:40:29.049 xcodebuild[99335:9919534] Writing error result bundle to /var/folders/4n/hzc7z9y54g3dmn6204bg0lfm0000gq/T/ResultBundle_2024-11-06_14-40-0029.xcresult
2024-06-11 14:40:29.049 xcodebuild[99335:9919534] *** -[__NSCFConstantString stringByAppendingPathExtension:]: cannot append extension 'xcactivitylog' to path ''
[...]

PR #3381, opened 2024-08-29
Update `-showBuildSettings` “Found no destinations for the scheme” pattern to permit more before “Found” and after “error” (due to Xcode 16 Beta 6 and beyond.)

prassyy:
The patch added for picking right targets in Xcode16 beta1 https://github.com/Carthage/Carthage/issues/3375 broke due to the change in the error message in Xcode 16 beta6. This modifies the pattern to match the message. Thanks to @apoltix for his suggestion [here](https://github.com/Carthage/Carthage/issues/3375#issuecomment-2317014794).

### DIFF
--- a/Source/CarthageKit/BuildSettings.swift
+++ b/Source/CarthageKit/BuildSettings.swift
@@ -65,7 +65,7 @@ public struct BuildSettings {
 			.flatMapError { error in
 				switch error {
 				case let .shellTaskFailed(_, _, standardError):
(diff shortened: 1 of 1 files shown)


jdhealy:
Thanks for the PR!

Could you make the `[^\n]+Found` into `[^\n]*Found` to have it accommodate the pre-beta-6 Xcode 16s?

jdhealy:
Awesome! Gonna test this out locally later today.

Ch070-ch (commented):
Esta bien hay que ver a detalle

PR #3395, opened 2025-04-02
FIX: Unable to determine framework Swift version: No architectures found in dSYM. when Swift header is not -Swift.h

erneestoc:
## Issue

When using Carthage to install [Google's Recaptcha xcframework 18.7.0](https://dl.google.com/recaptchaenterprise/v18.7.0/RecaptchaEnterprise_iOS_xcframework/recaptcha-xcframework.xcframework.zip), I'm getting the following error:
```
Unable to determine framework Swift version: No architectures found in dSYM
```

The issue seems to be carthage is not able to determine swift versions based on a header or dsym.
* dsym is not shipped with this vendored framework.
* swift header is used, but it doesn't uses the `-Swift.h` convention Carthage expects.

Error coming from here: https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/Xcode.swift#L89

Happens because it fails to find header URL here https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/Xcode.swift#L64C42-L64C56

Header finder defined here https://github.com/Carthage/Carthage/blob/master/Source/CarthageKit/FrameworkExtensions.swift#L297

For the last several versions of swift, `-Swift.h` has a `// Generated by Apple Swift version ` at the top. Instead of letting carthage fail because it's not finding the correct -Swift.h, carthage could fallback to check the `// Generated by Apple Swift` at the top of the header.


erneestoc:
@jdhealy if you have time to review, that'd be great. 🙏🏻

erneestoc:
https://github.com/GoogleCloudPlatform/recaptcha-enterprise-mobile-sdk/issues/129#issuecomment-2772713584

Source/CarthageKit/FrameworkExtensions.swift
```diff
@@ -297,7 +297,34 @@ extension URL {
...
 		let headersURL = self.appendingPathComponent("Headers", isDirectory: true).resolvingSymlinksInPath()
 		let dirContents = try? FileManager.default.contentsOfDirectory(at: headersURL, includingPropertiesForKeys: [], options: [])
-		return dirContents?.first { $0.absoluteString.contains("-Swift.h") }
+
+		// First try the standard naming pattern
+		if let standardHeader = dirContents?.first(where: { $0.absoluteString.contains("-Swift.h") }) {
+			return standardHeader
+		}
+
+		// Fallback: scan only the first line of header files
```
AshleyDeng, line 306: maybe it's okay to add the fallback here, given that you have tests for it.
I was wondering if we should add the fallback after we've checked the dSYM

erneestoc: think this approach should be fine, checking the first 1024 bytes out of several hundred header files (if it was the case for some) should be pretty quick - plus the fallback should be fairly rare, most people defaults to `-Swift.h` for swift generated headers cause that's what Xcode does... I'd defer to @jdhealy

AshleyDeng (approved):

erneestoc:
cc: @giginet @tmspzz @ikesyo @elliottwilliams

erneestoc:
@jdhealy if this is not what we want, I can fix it other way, I'm open to suggestions.

tmspzz (approved):
Looks good to me but I've been out of the game for quite a while.

AshleyDeng:
@tmspzz Thank you for approving this PR!
What is the release process like? Anything we can facilitate?

erneestoc:
@jdhealy @tmspzz @giginet gentle bump, in case this could make a patch 🙏🏻

Source/CarthageKit/FrameworkExtensions.swift
amir-glassboxdigital: I have 2 issues with the Pull request:
1. The assumption that a xcframework must contain a header file that contains the swift version is incorrect, for
    example an xcframework written in objective C for example does not have such a header file, and may not 
    even contain any swift in it at all.
   Add a check for arm64-apple-ios.swiftinterface located under 
   YourFrameworkName.xcframework/ios-arm64/  
   YourFrameworkName.framework/Modules/YourFrameworkName.swiftmodule/
    this file contains the following comment in it, for example:
    "// swift-compiler-version: Apple Swift version 5.7.1 (swiftlang-5.7.1.135.3 clang-1400.0.29.51)" 
2. Swift version 6.1.2 breaks the regular expression, here is the line in arm64-apple-ios.swiftinterface from 
    xcframework built with 6.1.2 see the "effective-5.10"
   "// swift-compiler-version: Apple Swift version 6.1.2 effective-5.10 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)"

amir-glassboxdigital (changes requested):
See my comments on the headers file check and on the regular expression that will break by swift 6.1.2

